PR #1, opened 2017-03-17
Feature/repo structure

kurtwheeler:
The goal of this PR is to establish the organization of the repo for this project. There will probably be at least two more top level directories:
- foreman: where the code for the surveyor/supervisor will live
- terraform: where the terraform configuration will live

As part of verifying that the repo structure will work I've proven that it can support:
- Having multiple Dockerfiles in the top level directory and still run code from sub-modules
- Have these containers import the database model code from a shared Django project that exists for that purpose. (Right now it only looks like one service is importing the model, but later the foreman and the aggregator will need to as well.)
- Have a celery worker run in a Docker container and save a row to a postgres server running on the container's host.

As you review this please consider:
- Whether there's any improvements that can be made to any of the directory structures (pay particular attention to the two Django projects, these are my first two I've made for real).
- If all the names are clear and self-explanatory.
- If there are any best practices I should be following but aren't.

The one outstanding question/issue with the repo structure that could come up is sharing celery tasks between the foreman and the workers. The foreman needs to be aware of what tasks it can call. This may end up being as simple as having the foreman rely on the workers as a dependency.


data_models/environments/dev
```diff
@@ -0,0 +1,8 @@
+DJANGO_SECRET_KEY=!f3p74+hi5omz+)m((+8597&&01h-&bf(0#d1mwvak&clmv*%g
```
mhuyck: This secret key will never be a secret again. What I have done for adage-server is to create a configuration file that contains all of the secrets and immediately added that file name to .gitignore. To have something to represent that file in source control, I created a template (just added `.template` to the end of the file name`) with all of the secrets blanked out by comments.

I guess this is "just" a developer key, but I hope your dev server isn't exposed to the internet.

Dockerfile.foreman
```diff
@@ -0,0 +1,18 @@
...
+RUN groupadd user && useradd --create-home --home-dir /home/user -g user user
+WORKDIR /home/user
+
+ENV CELERY_VERSION 4.0.2
+
+COPY workers/requirements.in .
+
+RUN pip install pip-tools && \
+    pip-compile requirements.in && \
+    pip install -r requirements.txt
```
rzelayafavila: I see that there is this ``requirements.txt`` file referenced here, but I do not see it added in the initial commit. Will it be added later? Also, I am assuming this file is what will hold the version numbers for the different dependencies (as opposed to the requirements.in file, which just holds the file names)?

Dockerfile.foreman
```diff
@@ -0,0 +1,18 @@
+FROM python:3.5-slim
```
rzelayafavila: Hooray for using Python 3!!!

Dockerfile.foreman
```diff
@@ -0,0 +1,18 @@
...
+
+RUN groupadd user && useradd --create-home --home-dir /home/user -g user user
```
rzelayafavila: Do we want this ``user`` user to have a more specific name down the road?

Dockerfile.worker
```diff
@@ -0,0 +1,18 @@
+FROM python:3.5-slim
```
rzelayafavila: Some of the same comments I had from Dockerfile.foreman apply here as well.

data_models/bioinformatics_mill_models/models.py
```diff
@@ -0,0 +1,42 @@
+from django.db import models
+from django.utils import timezone
+
+
+class TimeTrackedModel(models.Model):
+    created_at = models.DateTimeField(editable=False)
+    updated_at = models.DateTimeField()
+
+    def save(self, *args, **kwargs):
```
rzelayafavila: I think you won't need this code here if you add the ``auto_now__add`` and ``auto_now`` arguments to ``created_at`` and ``updated_at``, respectively, right? https://docs.djangoproject.com/en/1.10/ref/models/fields/#django.db.models.DateField  I think they will do that automatically for you.

data_models/bioinformatics_mill_models/models.py
```diff
@@ -0,0 +1,42 @@
...
+    raw_format = models.CharField(max_length=256)
+    processed_format = models.CharField(max_length=256)
+    processor_required = models.IntegerField()
+    accession_code = models.CharField(max_length=256)
+
+    # This field where denote where in our system the file can be found
+    internal_location = models.CharField(max_length=256)
+
+    # This will at some be a meaningful integer->organism lookup thing
+    organism = models.IntegerField()
```
rzelayafavila: So this will be the organism taxonomy ID from NCBI?

data_models/manage.py
```diff
@@ -0,0 +1,22 @@
...
+        try:
+            import django
+        except ImportError:
+            raise ImportError(
+                "Couldn't import Django. Are you sure it's installed and "
+                "available on your PYTHONPATH environment variable? Did you "
+                "forget to activate a virtual environment?"
+            )
+        raise
+    execute_from_command_line(sys.argv)
```
mhuyck: I'm assuming this is just the Django-generated file so I'm skipping review here. Please let me know if you've altered anything and I should take a closer look. (The adage-server version is slightly different, but we're using Django 1.8.)

workers/downloaders/tasks.py
```diff
@@ -0,0 +1,25 @@
+from __future__ import absolute_import, unicode_literals
+from celery import shared_task
+
+from bioinformatics_mill_models.models import Batch
+
+
+@shared_task
+def save_test_batch():
+    print("saving batch")
```
rzelayafavila: I think we should also put in some Python logging statements as we grow this so that we save some output of the processes that are going on as we run this: https://docs.python.org/3/library/logging.html

Dockerfile.foreman
```diff
@@ -0,0 +1,18 @@
...
+WORKDIR /home/user
+
+ENV CELERY_VERSION 4.0.2
+
+COPY workers/requirements.in .
+
+RUN pip install pip-tools && \
+    pip-compile requirements.in && \
```
mhuyck: I have not seen `pip-compile` before. Cool!

Dockerfile.worker
```diff
@@ -0,0 +1,18 @@
...
+
+RUN pip install pip-tools && \
+    pip-compile requirements.in && \
+    pip install -r requirements.txt
+
+COPY workers/ .
+
+USER user
+
+CMD ["celery", "-A", "bioinformatics_mill_workers", "-c", "1", "worker"]
```
mhuyck: I'm not familiar with the Docker foreman/worker pattern. Do you have a reference I can skim to see how that works?

Dockerfile.foreman
```diff
@@ -0,0 +1,18 @@
+FROM python:3.5-slim
+
+RUN groupadd user && useradd --create-home --home-dir /home/user -g user user
+WORKDIR /home/user
+
+ENV CELERY_VERSION 4.0.2
+
+COPY workers/requirements.in .
```
mhuyck: Is it okay that this is using `workers/requirements.in` even though this is `Dockerfile.foreman`?

Dockerfile.foreman
kurtwheeler: Hmm perhaps, but I don't really know how much it matters because there will only ever be one user per container so we don't have to worry about them clashing with anything.

Dockerfile.foreman
kurtwheeler: The `Dockerfile.foreman` name here is a bit confusing because it's not actually doing any `foreman` stuff yet, it's just sending a celery message from a container. I could probably rename this `Dockerfile.tester` to clear up confusion.

Dockerfile.foreman
kurtwheeler: The `requirements.txt` file is actually generated by `pip-compile`. I think I'm doing this wrong here. I should be running `pip-compile` as a pre-deployment step and then be keeping the resulting `requirements.txt` under source control so that as `pip-compile` changes version numbers the changes are tracked. Then in these dockerfiles I should just be installing from the `requirements.txt` without compiling as part of building the container.

Dockerfile.worker
kurtwheeler: It's not so much a pattern as a workaround I came up with to solve the issue I was having where I didn't yet have a foreman to actually send the workers messages, so I made a very simple foreman app (that's what `master.py` is) that just runs `save_test_batch.delay()` to make sure the containers are talking to each other.

data_models/bioinformatics_mill_models/models.py
kurtwheeler: I found a comment online mentioning that those fields are considered evil and should be avoided. The explanation linked [here](http://stackoverflow.com/questions/1737017/django-auto-now-and-auto-now-add/1737078#1737078) is where I got this method of doing this.

data_models/bioinformatics_mill_models/models.py
kurtwheeler: Do you that ID system will cover all of our needs? If so: yes :)

workers/downloaders/tasks.py
kurtwheeler: Yes. I agree wholeheartedly. I haven't gotten to handling logging properly yet but it's on my TODO list.

workers/bioinformatics_mill_workers/celery.py
```diff
@@ -0,0 +1,23 @@
+from __future__ import absolute_import, unicode_literals
+import os
+from celery import Celery
```
kurtwheeler: That's a good point. I am definitely not in favor of leaving this file named celery. It was named so in an example I was following and I forgot to change it.

data_models/bioinformatics_mill_models/models.py
```diff
@@ -0,0 +1,42 @@
...
+    class Meta:
+        abstract = True
+
+
+# This model is still has a prototypical status, but I needed something to
+# test with and it's at least in the right ballpark
+class Batch(TimeTrackedModel):
+    source_type = models.CharField(max_length=256)
+    size_in_bytes = models.IntegerField()
+    download_url = models.CharField(max_length=256)
```
kurtwheeler: Good call.

data_models/bioinformatics_mill_models/models.py
```diff
@@ -0,0 +1,42 @@
...
+    raw_format = models.CharField(max_length=256)
+    processed_format = models.CharField(max_length=256)
+    processor_required = models.IntegerField()
+    accession_code = models.CharField(max_length=256)
+
+    # This field where denote where in our system the file can be found
```
kurtwheeler: Yup.

data_models/bioinformatics_mill_models/models.py
```diff
@@ -0,0 +1,42 @@
...
+# This model is still has a prototypical status, but I needed something to
+# test with and it's at least in the right ballpark
+class Batch(TimeTrackedModel):
+    source_type = models.CharField(max_length=256)
+    size_in_bytes = models.IntegerField()
+    download_url = models.CharField(max_length=256)
+    raw_format = models.CharField(max_length=256)
+    processed_format = models.CharField(max_length=256)
+    processor_required = models.IntegerField()
+    accession_code = models.CharField(max_length=256)
```
kurtwheeler: The latter. Right now I think that 256 is excessively long but I figure we can run a migration to make this size more appropriate later.

data_models/install.sh
```diff
@@ -0,0 +1,20 @@
...
+# Run this script with 'sudo -u postgres install.sh'
+
+# This script defaults to postgres version 9.5.
+# To use a different version set the POSTGRES_VERSION env var
+# (i.e. 'sudo POSTGRES_VERSION=9.4 -u postgres install.sh')
+if [ -z $POSTGRES_VERSION ]; then POSTGRES_VERSION=9.5; fi
+
+createdb bioinformatics_mill
+
+psql bioinformatics_mill -c "CREATE ROLE bioinformatics_mill_user WITH LOGIN PASSWORD 'bioinformatics_password';"
```
kurtwheeler: This password is only for use during development. This script won't be used in production. This doesn't actually go through Docker either, it's to be run on the host machine because that's where the database needs to be hosted from.

data_models/install.sh
```diff
@@ -0,0 +1,20 @@
...
+psql bioinformatics_mill -c 'GRANT ALL PRIVILEGES ON DATABASE bioinformatics_mill TO bioinformatics_mill_user;'
+psql bioinformatics_mill -c 'ALTER USER bioinformatics_mill_user CREATEDB;'
+
+iptables -A INPUT -s 172.17.0.0/16 -j ACCEPT
+
+echo 'host    all             all             172.17.0.0/16           md5' >> /etc/postgresql/$POST_GRES_VERSION/main/pg_hba.conf
```
kurtwheeler: Yeah that's a good call, I'll find the sources that eventually led me to find these three lines. (It took me an afternoon of googling to finally find all three of these.)

data_models/environments/dev
kurtwheeler: I did intentionally exposed this secret because it will only be used for development. I didn't think there'd be any issues with having only the development key exposed. Thus far I am not even running a django server anywhere (and won't be for a while, only the Aggregator will actually need to do that).

Dockerfile.foreman
mhuyck: I've observed that a number of good practices from the world before containerization no longer apply or are turned on their heads by something like Docker. This is one: I think a generic account name is good here. The point is to simplify the infrastructure to the point of being as generic as possible.

Dockerfile.foreman
mhuyck: Nice catch. I think it's important to capture version numbers in source control.

data_models/bioinformatics_mill_models/models.py
```diff
@@ -0,0 +1,42 @@
...
+
+class TimeTrackedModel(models.Model):
+    created_at = models.DateTimeField(editable=False)
+    updated_at = models.DateTimeField()
+
+    def save(self, *args, **kwargs):
+        ''' On save, update timestamps '''
+        if not self.id:
+            self.created_at = timezone.now()
+        self.updated_at = timezone.now()
```
mhuyck: Interesting. The documentation at https://docs.djangoproject.com/en/1.10/topics/i18n/timezones/ is confusing. Near the top, it says

> Time zone support is disabled by default. To enable it, set USE_TZ = True in your settings file. 

Then, in the first note it says

> The default settings.py file created by django-admin startproject includes USE_TZ = True for convenience.

I guess that's actually consistent, but why should `startproject` second-guess the decision that "time zone support is disabled by default?"

Dockerfile.foreman
mhuyck: Gotcha. Thanks!

Dockerfile.worker
kurtwheeler: Although I've thought a bit about it and while it's not specific to Docker [this article](https://aws.amazon.com/ec2/spot/spot-and-science/) does describe the "Grid Architecture" pattern that the foreman/workers are going to be implementing. (Jump down to the section titled `"Grid" Architecture`, it refers to something called a "master node" which I've renamed Foreman for our purposes since we're a refinery.)

kurtwheeler:
@mhuyck and @rzelayafavila I've either responded to all your comments or made the changes you recommended. Please let me know what you think.

Dockerfile.worker
mhuyck: Makes sense. I think I assumed from the file names "Docker.foreman" and "Docker.worker" that there was some Docker-based support for the pattern. A Google search for "Docker foreman" even turned up some meaty-looking hits...

data_models/environments/dev
mhuyck: I agree that the actual risk for a dev server is fairly small. Still, we do much of our development "in the open" and it would be a bummer to find that someone trashed our dev instance database. I see that the same key is also used and exposed for `environments/test`, a server which will almost certainly be exposed to the internet at some point.

A bigger concern is having this key *accidentally* re-used in a more important context. I've tried to prevent that entirely by not having any secrets embedded in source control anywhere. Maybe that's too extreme. What are your thoughts, @dongbohu and @rzelayafavila?

data_models/install.sh
mhuyck: I think if it's really true that this script won't be used in production then the risk is relatively low, but my concern highlighted above about *accidentally* leaving this in place still applies. At the very least, I think we'd need a big noisy warning in the comments above and written to the output of this command to the effect that this is **NOT FOR PRODUCTION** to mitigate that risk.

data_models/environments/dev
dongbohu: Since this secret is only used for development right now, it is okay with me to have it exposed. Maybe you can add an issue to remind yourself this fact. 

On the other hand, since it is exposed to the rest of the world, can we simply use something as simple as "foobar"?

kurtwheeler: Hmm, maybe I'll just make it `THIS_IS_NOT_A_SECRET_DO_NOT_USE_IN_PROD` :)

dongbohu: Sure. I guess any string that will remind both users and developers immediately of its fakeness is good.

kurtwheeler:
Okay I've added a github issue about the secrets: #2 and have a card in Trello: https://trello.com/b/tM4P1lOU/refineries about it as well. I've also changed the values of the django secret keys to be "THIS_IS_NOT_A_SECRET_DO_NOT_USE_IN_PROD" to make it extra obvious.

@mhuyck @rzelayafavila am I missing anything else or is this looking good?

data_models/manage.py
rzelayafavila: I was wondering about that too @kurtwheeler, since the ones for other projects do look a little bit different (most likely due to the version difference, as @mhuyck pointed out). Is this the ``manage.py`` file exactly as it was created by ``django-admin``?

kurtwheeler: Yes, I haven't modified this one at all except while making name changes from `bioinformatics_mill` to `data_refinery`.

rzelayafavila: Ok, cool!

data_models/bioinformatics_mill_models/models.py
rzelayafavila: Ok, as we discussed in person, I think there are both benefits and drawbacks from both. Im ok with sticking with this option for now, and we can discuss some more about it after the PR has been merged.

rzelayafavila:
Ok, I went through everything and made some more comments but I think you got all the stuff I commented on on my end.

PR #3, opened 2017-03-22
Feature/database models

kurtwheeler:
These are the models I think I will need for this project. I don't yet have perfect knowledge of what the metadata that needs to be stored in the Batch table will look like so I've left some of those fields at just `max_length=256` for now. I'm open to changing any of them now if you have any recommendations, however they can always be changed later as well. The rest of the tables I feel more confident about.


data_models/data_refinery_models/models.py
```diff
@@ -17,26 +18,101 @@ class Meta:
...
-        ("NEW", "New"),
-        ("DOWNLOADED", "Downloaded"),
-        ("PROCESSED", "Proccessed"),
-    )
-    status = models.CharField(max_length=10, choices=STATUSES)
+    class Meta:
+        db_table = "batches"
+
+
+# This table is used for tacking fields onto a Batch record that would
```
dongbohu: You may want to change these comments into docstring of `BatchKeyValue` class. By the way, is `tacking` on this line supposed to be `tracking`?

data_models/data_refinery_models/models.py
```diff
@@ -17,26 +18,101 @@ class Meta:
...
+# be sparsly populated if it was its own column.
```
dongbohu: `sparsly` should be `sparsely`?

data_models/data_refinery_models/models.py
```diff
@@ -17,26 +18,101 @@ class Meta:
...
+    key = models.CharField(max_length=256)
+    value = models.CharField(max_length=256)
+
+    class Meta:
+        db_table = "batch_key_values"
+
+
+class ProcessorJob(TimeTrackedModel):
+    batch_id = models.ForeignKey(Batch, on_delete=models.CASCADE)
+    pipeline_applied = models.CharField(max_length=256)
```
dongbohu: Can you explain to me what this field will be? A description of pipeline?

data_models/data_refinery_models/models.py
```diff
@@ -17,26 +18,101 @@ class Meta:
...
+    # At some point there will probably be some code like:
+    # if job.num_retries >= 3:
+    #     # do a bunch of logging
+    # else:
+    #     # retry the job
+    num_retries = models.IntegerField(default=0)
+
+    # This point of this field is to identify what worker ran the job.
+    # A few fields may actually be required or something other than just an id.
+    worker_id = models.CharField(max_length=256)
```
dongbohu: Do you need a separate `worker` table and this `worker_id` will be the foreign key to the "worker" table?

data_models/data_refinery_models/models.py
```diff
@@ -17,26 +18,101 @@ class Meta:
...
+class DownloaderJob(TimeTrackedModel):
+    batch_id = models.ForeignKey(Batch, on_delete=models.CASCADE)
+    start_time = models.DateTimeField(null=True)
+    end_time = models.DateTimeField(null=True)
+    success = models.NullBooleanField(null=True)
+
+    # These two fields are analagous to the fields with the same names
+    # in ProcessorJob, see their descriptions for more information
+    num_retries = models.IntegerField(default=0)
+    worker_id = models.CharField(max_length=256)
```
dongbohu: Same question as the one on `worker_id` in `ProcessorJob` table.

data_models/data_refinery_models/models.py
```diff
@@ -17,26 +18,101 @@ class Meta:
...
+    # This field represents how many times this job has been retried. It starts
+    # at 0 and each time the job has to be retried it will be incremented.
+    # At some point there will probably be some code like:
+    # if job.num_retries >= 3:
+    #     # do a bunch of logging
+    # else:
+    #     # retry the job
+    num_retries = models.IntegerField(default=0)
+
+    # This point of this field is to identify what worker ran the job.
```
dongbohu: Does it sound better if this line is changed into:
```The point of this field is to identify which worker runs the job.```
(I am not a native speaker of English. If you prefer the original version, it is okay with me.)

File: Dockerfile.worker [hunk 7]
```diff
@@ -14,6 +14,11 @@ COPY data_models/dist/data-refinery-models-* data_models/
 # Get the latest version from the dist directory.
 RUN pip install data_models/$(ls data_models | sort -r | head -1)
 
+COPY data_models/dist/data-refinery-models-* data_models/
+
+# Get the latest version from the dist directory.
+RUN pip install data_models/$(ls data_models | sort -r | head -1)
```
dongbohu: Actually `ls` command already sorts the filenames alphabetically, so the command can be more compact: 
`ls -r1 | head -1` 
or 
`ls -1 | tail -1`

kurtwheeler: Hmm, I'm not sure why but `sort` was able to handle all the file names being on the same line somehow. However `ls -1 | tail -1` is certainly better so I'll use that.

data_models/data_refinery_models/models.py
kurtwheeler: This is a good idea. I'm not sure what we would want to store in the worker table though so for now I think I will leave it like this until I actually make a worker and have an idea of what the columns of a worker table would be. I will add this as a TODO though.

data_models/data_refinery_models/models.py
kurtwheeler: I'm going to change it to a hybrid of these two options `The point of this field is to identify which worker ran the job.` because it should be in past tense because we won't know what worker is going to run the job until one pulls it off the queue.

data_models/data_refinery_models/models.py
kurtwheeler: I had this as `processor_applied` at some point but I think that a single processor is rarely going to be enough so there will be pipelines of processors which will have identifiers. This field will store the identifier of the processor pipeline which was applied during the job.

As an example a processor pipeline could be described in code like this:
```
processor_pipelines = {
  "EXPRESSION_TO_PCL": ["PROCESSOR_A", "PROCESSOR_B", "PROCESSOR_C"],
  ...
}
```
I could potentially rename this field to `processor_pipeline_applied` if you think that's more clear.

data_models/data_refinery_models/models.py
```diff
@@ -17,26 +18,101 @@ class Meta:
...
     download_url = models.CharField(max_length=2048)
-    raw_format = models.CharField(max_length=256)
-    processed_format = models.CharField(max_length=256)
-    processor_required = models.IntegerField()
-    accession_code = models.CharField(max_length=256)
+    raw_format = models.CharField(max_length=256, null=True)
+    processed_format = models.CharField(max_length=256, null=True)
+    pipeline_required = models.CharField(max_length=256)
+    accession_code = models.CharField(max_length=32)
+    status = models.CharField(max_length=20)
```
kurtwheeler: Fair point. For the statuses I already have it is large enough (values should only be entered into this field by using `BatchStatuses.NEW.value`). I hope to not be adding statuses often so if I end up needing one large than 20 I can just run a migration to widen the column.

data_models/data_refinery_models/models.py
```diff
@@ -17,26 +18,101 @@ class Meta:
...
     size_in_bytes = models.IntegerField()
     download_url = models.CharField(max_length=2048)
-    raw_format = models.CharField(max_length=256)
-    processed_format = models.CharField(max_length=256)
-    processor_required = models.IntegerField()
-    accession_code = models.CharField(max_length=256)
+    raw_format = models.CharField(max_length=256, null=True)
+    processed_format = models.CharField(max_length=256, null=True)
+    pipeline_required = models.CharField(max_length=256)
+    accession_code = models.CharField(max_length=32)
```
kurtwheeler: I don't know. I've only seen accession codes as long as ~10 characters. Do you know if they get longer than this?

data_models/data_refinery_models/models.py
```diff
@@ -17,26 +18,101 @@ class Meta:
...
+    start_time = models.DateTimeField(null=True)
+
+    # The end time of the job
+    end_time = models.DateTimeField(null=True)
+
+    class Meta:
+        db_table = "survey_jobs"
+
+
+class BatchStatuses(Enum):
```
kurtwheeler: If the download fails it will be recorded in the `status` field of the downloader_job. There will be code to check the statuses of jobs and retry the ones that fail. I don't think the batch needs to worry about this.

kurtwheeler:
Ok @dongbohu I've either responded to all your comments, made the suggested change, or both.

data_models/data_refinery_models/models.py
dongbohu: You may want to check with Rene or Casey. I don't know how long the access code could be.

data_models/data_refinery_models/models.py
dongbohu: Got it. If you prefer keeping the original field name, can you add some comments to make it easier to understand?

data_models/data_refinery_models/models.py
kurtwheeler: @cgreene and @jaclyn-taroni do either of you know what the upper limit on accession codes is?

cgreene: Hmm, SRA looks to be ~10 (also appears to be incrementing):
https://www.ncbi.nlm.nih.gov/sra/?term=human

GEO also appears to increment so let's say roughly 10.

The most recent ones in arrayexpress seem to be ~10:
https://www.ebi.ac.uk/arrayexpress/search.html?query=human&sortby=releasedate&sortorder=descending

ENA also appears to be incrementing with ~750k samples:
http://www.ebi.ac.uk/ena/data/search?query=human

I guess 32 characters should be enough for anybody (at least for a while?). Such statements are never proven wrong by history ;)

kurtwheeler: Ok, I'll leave it as this then. If we ever need to we can always widen it later, but given that these are alphanumeric, 32 characters should be enough to reference 6.33e49 samples.

PR #4, opened 2017-03-24
Feature/surveyor init

kurtwheeler:
This pull request starts adding the surveyor framework. Included with it is the ArrayExpressSurveyor which is only meant as a proof of concept to demonstrate how the surveyor framework will work and to test it. There is still lots to be done on this framework including making sure that we are not creating Batches which represent data we have already downloaded/processed and scheduling SurveyJobs.

Additionally I have not done the following correctly, but I have trello cards to fix that for each of them:
- [Logging](https://trello.com/c/QKS1NKj4/54-set-up-logging)
- [Testing](https://trello.com/c/v6cMyFWB/53-set-up-testing-framework)
- [Docstrings](https://trello.com/c/zC45GOGm/56-format-docstrings-correctly-to-generate-pydocs)

I could do these in this PR but it is already large so I'd rather just create a separate PR to handle each of those. I am planning on tackling those three tickets next once I am able to return to working on the Data Refinery project after my detour in deploying Cognoma.


foreman/data_refinery_foreman/surveyor/surveyor.py
```diff
@@ -0,0 +1,64 @@
...
+        log_message = log_message + " because: " + str(e)
+        print(log_message)
+
+        _end_job(survey_job, False)
+        return survey_job
+
+    try:
+        job_success = surveyor.survey(survey_job)
+    except Exception as e:
+        print("Exception caught while running job # " + survey_job.id +
```
rzelayafavila: Ok, I know these will get changed to logging statements in a future PR from our conversation.

foreman/data_refinery_foreman/surveyor/array_express_surveyor.py
```diff
@@ -0,0 +1,65 @@
...
+    def source_type(self):
+        return "ARRAY_EXPRESS"
+
+    def determine_pipeline(self,
+                           batch: Batch,
+                           key_values: List[BatchKeyValue] = []):
+        return ProcessorPipeline.MICRO_ARRAY_TO_PCL
+
+    def survey(self, survey_job: SurveyJob):
+        # Obviously this shouldn't be hardcoded in but the Array Express
```
kurtwheeler: No actually I think it should be a field on the SurveyJob. However it probably won't be an entity for each source we'll have so I think it should probably be in a key value table. I had considered making a key value table for SurveyJobs like I have for the Batches table but I guess I decided not too. This seems like a good reason to change my mind.

Dockerfile.surveyor
```diff
@@ -0,0 +1,21 @@
...
+RUN groupadd user && useradd --create-home --home-dir /home/user -g user user
+WORKDIR /home/user
+
+ENV CELERY_VERSION 4.0.2
+
+COPY foreman/requirements.txt .
+
+RUN pip install -r requirements.txt
+
+COPY data_models/dist/data-refinery-models-* data_models/
```
kurtwheeler: > Would it make sense for them to be packages on PyPI that we can install?

Yes definitely. I actually had made a PyPI package for this and was installing it but that necessitated having to change the version each time I wanted to make a change even while developing. Once the data models are a bit more stable and I'm ready to release an initial version I will change this back to installing from PyPI.

foreman/environments/dev
```diff
@@ -0,0 +1,10 @@
+DJANGO_SECRET_KEY=#29)1-w_!5x40ibbl0me3p^!f_j@_gpvre1cf+pj3cijp2h_18
```
kurtwheeler: Good call.

kurtwheeler:
@rzelayafavila I've made changes based on your comments. Take a look and lemme know whatchya think.

Dockerfile.surveyor
rzelayafavila: Ok, sounds good

PR #6, opened 2017-04-10
Start testing

kurtwheeler:
This PR adds tests for models, the Surveyor, and the Array Express Surveyor. It also adds scripts and Dockerfiles for running tests.

It also utilizes logging instead of print commands.

I also added some instructions to a couple READMEs.

Finally in this PR I cleaned up the organization of the project a bit. I found a good pattern for having Dockerfiles live in the project directories instead of having to be in the root directory of the repo.


kurtwheeler:
In the interest of not having this triple reviewed I've assigned it to @mhuyck but feel free to glance over it @rzelayafavila @dongbohu @cgreene

data_models/install.sh
```diff
@@ -16,6 +16,8 @@ psql data_refinery -c 'ALTER USER data_refinery_user CREATEDB;'
 # See https://www.howtogeek.com/177621/the-beginners-guide-to-iptables-the-linux-firewall/
 # for more information.
 iptables -A INPUT -s 172.17.0.0/16 -j ACCEPT
+/sbin/iptables-save > /etc/iptables/rules
```
mhuyck: Do you need to manually write output to `/etc/iptables/rules`? The [reference](https://www.howtogeek.com/177621/the-beginners-guide-to-iptables-the-linux-firewall/) from the comment above says you just need `sudo /sbin/iptables-save` for Ubuntu.

data_models/run_tests.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+# script for executing Django Unittests within Docker container.
```
mhuyck: typo: Unittests => unit tests

data_models/run_tests.sh
```diff
@@ -0,0 +1,11 @@
...
+
+docker build -t models_tests -f Dockerfile.tests .
+
+HOST_IP=$(ifconfig eth0 | grep "inet " | awk -F'[: ]+' '{ print $4 }')
+docker run \
+       --add-host=database:$HOST_IP \
+       --env-file /home/kurt/Development/data_refinery/data_models/environments/test \
+       -i models_tests "$@"
```
mhuyck: This `env-file` path does not look like something that would work for someone else's build... unless they also had a username of `kurt`.

foreman/Dockerfile.tests
```diff
@@ -0,0 +1,21 @@
+# This Dockerfile is meant to be run from the directory above where it resides.
```
mhuyck: Is this true? It seems to be setting `WORKDIR` using a full path manually at line 6... maybe I'm missing something?

foreman/README.md
```diff
@@ -0,0 +1,21 @@
...
+To run the surveyor in a Docker container just run `./run_surveyor.sh`.
+
+You can also start an interactive python interpreter by running `../run_shell.sh`.
+This interpreter will actually be running within the surveyor Docker container
+with the environment configured appropriately.
+
+### Testing
+
+To run all tests for the project just run `./run_tests.sh`. This script will
+accept any arguments that are valid for `python -m unittest`.
```
mhuyck: This is nice and clear. Thank you!

foreman/data_refinery_foreman/settings.py
```diff
@@ -93,6 +93,10 @@ def get_env_variable(var_name):
         'PASSWORD': get_env_variable('DATABASE_PASSWORD'),
         'HOST': get_env_variable('DATABASE_HOST'),
         'PORT': get_env_variable('DATABASE_PORT'),
+
+        'OPTIONS': {
+            'connect_timeout': get_env_variable('DATABASE_TIMEOUT')
```
mhuyck: Out of curiosity: why did you need to adjust the timeout here? It's unusual to modify that.

data_models/run_tests.sh
kurtwheeler: Good catch, I fixed this for [`run_shell`](https://github.com/data-refinery/data_refinery/pull/6/commits/6c1e3cb11b3cb484306fd926cfe533d96f170261) but forgot I had done something similar here.

data_models/install.sh
kurtwheeler: Hmm that wasn't a good link to leave there. It was incomplete. I'll update it with a better one.

data_models/data_refinery_models/tests.py
```diff
@@ -0,0 +1,23 @@
...
+from django.utils import timezone
+from data_refinery_models.models import SurveyJob
+
+
+class TimeTrackedModelTestCase(TestCase):
+    def setUp(self):
+        SurveyJob.objects.create(source_type="ARRAY_EXPRESS")
+
+    def tearDown(self):
+        SurveyJob.objects.all().delete()
```
kurtwheeler: But it only removes it when all the tests have completed. By doing this I ensure that if I write any other tests they won't interfere.

foreman/Dockerfile.tests
kurtwheeler: That `WORKDIR` is referring to a directory within the docker container, not a location on the host machine. See https://docs.docker.com/engine/reference/builder/#workdir for more information.

foreman/data_refinery_foreman/settings.py
kurtwheeler: I was running into an issue with my database timing out during development. While debugging that I changed the timeout to 5 seconds and decided that there wasn't really any time during development I'd want to wait 300 seconds to find out my database is timing out. So I made it an environment variable so I could have the short timeout during development..

foreman/data_refinery_foreman/surveyor/external_source.py
```diff
@@ -58,13 +57,14 @@ def handle_batch(self, batch: Batch, key_values: BatchKeyValue = None):
...
-                "is of the type DiscoveryPipeline."
+            message = ("Batches must have the processed_format field set " +
+                       "unless the pipeline returned by determine_pipeline" +
+                       "is of the type DiscoveryPipeline.")
             # Also should be more specific
             raise Exception(message)
 
-        if(batch.save()):  # This is also where we will queue the downloader job
+        # This is also where we will queue the downloader job
+        if(batch.save()):
```
dongbohu: Can you add a space after "if"?

foreman/run_surveyor.sh
```diff
@@ -0,0 +1,20 @@
...
+# the directory it lives in.
+cd "$( dirname "${BASH_SOURCE[0]}" )"
+
+# However in order to give Docker access to all the code we have to
+# move up a level
+cd ..
+
+docker build -t dr_surveyor -f foreman/Dockerfile.surveyor .
+
+HOST_IP=$(ifconfig eth0 | grep "inet " | awk -F'[: ]+' '{ print $4 }')
```
dongbohu: See my previous comment on finding IP address.

foreman/run_surveyor.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+# Script for running the Data Refinery Surveyor container
+
+# This script should always run as if it were being called from
+# the directory it lives in.
+cd "$( dirname "${BASH_SOURCE[0]}" )"
```
dongbohu: I guess this line can be a little simpler:
```
cd `dirname "${BASH_SOURCE[0]}"`
```

foreman/run_tests.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+# Script for executing Django PyUnit tests within a Docker container.
+
+# This script should always run as if it were being called from
+# the directory it lives in.
+cd "$( dirname "${BASH_SOURCE[0]}" )"
```
dongbohu: See my previous comment.

foreman/run_tests.sh
```diff
@@ -0,0 +1,19 @@
...
+
+# However in order to give Docker access to all the code we have to
+# move up a level
+cd ..
+
+docker build -t foreman_tests -f foreman/Dockerfile.tests .
+
+HOST_IP=$(ifconfig eth0 | grep "inet " | awk -F'[: ]+' '{ print $4 }')
```
dongbohu: See my previous comment.

run_shell.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+# Running this script will start an interactive python shell running
+# within the context of a Docker container.
+# By default the Docker container will be for the foreman project.
+# This can be changed by modifying the --env-file command line arg
+# and by modifying the Dockerfile.shell file appropriately.
+
+HOST_IP=$(ifconfig eth0 | grep "inet " | awk -F'[: ]+' '{ print $4 }')
```
dongbohu: See my previous comment.

workers/run_tester.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+# Script for running a simple tester container to test the worker.
+
+# This script should always run as if it were being called from
+# the directory it lives in.
+cd "$( dirname "${BASH_SOURCE[0]}" )"
```
dongbohu: See my previous comment.

workers/run_tester.sh
```diff
@@ -0,0 +1,20 @@
...
+
+# However in order to give Docker access to all the code we have to
+# move up a level
+cd ..
+
+docker build -t test_master -f workers/Dockerfile.tester .
+
+HOST_IP=$(ifconfig eth0 | grep "inet " | awk -F'[: ]+' '{ print $4 }')
```
dongbohu: See my previous comment.

workers/run_worker.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+# Script for running a Data Refinery Worker container.
+
+# This script should always run as if it were being called from
+# the directory it lives in.
+cd "$( dirname "${BASH_SOURCE[0]}" )"
```
dongbohu: See my previous comment.

workers/run_worker.sh
```diff
@@ -0,0 +1,21 @@
...
+
+# However in order to give Docker access to all the code we have to
+# move up a level
+cd ..
+
+docker build -t dr_worker -f workers/Dockerfile.worker .
+
+HOST_IP=$(ifconfig eth0 | grep "inet " | awk -F'[: ]+' '{ print $4 }')
```
dongbohu: See my previous comment.

Dockerfile.shell
```diff
@@ -0,0 +1,21 @@
+FROM python:3.5
+
+RUN groupadd user && useradd --create-home --home-dir /home/user -g user user
```
kurtwheeler: This code is not going to be executed within the host machine, but rather within the docker container, which is always going to be a Linux OS whether it is run on Mac or Linux. (In some ways a container is like a virtual machine and so these commands are setting up the environment within it.)

Dockerfile.shell
```diff
@@ -0,0 +1,21 @@
...
+# To use this container to interact with a different project change
+# "foreman" to the project name you'd like to interact with.
+COPY foreman/requirements.txt .
+
+RUN pip install -r requirements.txt
+
+COPY data_models/dist/data-refinery-models-* data_models/
+
+# Get the latest version from the dist directory.
+RUN pip install data_models/$(ls data_models -1 | tail -1)
```
kurtwheeler: Yes, that should be true because the `data_models/dist` directory is populated by the `python setup.py sdist` command and it names the tarballs it creates in this pattern `<django project name>-<version>.tar.gz`. Since the project name will always be the same the sorting will essentially be on the version numbers. This will work for versions 0-9, but once I hit version 10 this sorting will break. Therefore I'll change this to be `ls -1 | sort --version-sort | tail -1` which will sort on the version numbers properly.

>What if the file that you just copied (data-refinery-models-*) is a directory?
It shouldn't be. The only thing in that directory should be tarballs created by `python setup.py sdist`. If someone sticks a directory in there called `data-refinery-models-*` then this will probably break and I will have no sympathy for them because they shouldn't go sticking directories into dist directories :P

data_models/run_tests.sh
```diff
@@ -0,0 +1,15 @@
...
+
+# script for executing Django PyUnit Tests within a Docker container.
+
+# This script should always run as if it were being called from
+# the directory it lives in.
+cd "$( dirname "${BASH_SOURCE[0]}" )"
+
+docker build -t models_tests -f Dockerfile.tests .
+
+HOST_IP=$(ifconfig eth0 | grep "inet " | awk -F'[: ]+' '{ print $4 }')
```
kurtwheeler: Ah cool, I'll change that. Sometimes I wish linux would follow `There should be one-- and preferably only one --obvious way to do it.` a bit better.

foreman/data_refinery_foreman/surveyor/array_express.py
```diff
@@ -0,0 +1,75 @@
...
+
+            # If there is only one file object in data_files,
+            # ArrayExpress does not put it in a list of size 1
+            if (type(data_files) != list):
+                data_files = [data_files]
+
+            for data_file in data_files:
+                if (data_file["kind"] == "raw"):
+                    url = data_file["url"].replace("\\", "")
+                    # This is another place where this is still a POC.
```
kurtwheeler: Proof of Concept

kurtwheeler:
@dongbohu I've made those changes. Is it lookin good?

PR #7, opened 2017-04-12
Array Express Downloader

kurtwheeler:
This PR adds the array_express downloader. This downloader is a Celery task which takes a DownloaderJob ID. It then looks up the batch for the job, determines the source URL and destination URL, and downloads the file to it. I've created a Django command to setup the required database records and to call the task to test it. I didn't do this in a unit test because the task doesn't do much other than interact with Array Express and I don't want to download a file from them every time the full test suite is run.

I have tested this by running `./run_worker.sh` and `./run_tester.sh` and verifying that the file was downloaded to `workers/volume/expression_data/array_express/A-AFFY-1/E-GEOD-96.raw.1.zip`.

I also tested this by putting `raise Exception("test")` within the try block of the task and made sure that the logging was properly formatted and that the job was marked as a failure.

Many of the files changed are not related to this functionality because I am improving the configuration of the project incrementally as I work on it. I also upgraded the python version to 3.6.1.


workers/data_refinery_workers/downloaders/management/commands/queue_task.py
```diff
@@ -0,0 +1,47 @@
...
+        batch = Batch(
+            survey_job=survey_job,
+            source_type="ARRAY_EXPRESS",
+            size_in_bytes=0,
+            download_url="http://www.ebi.ac.uk/arrayexpress/files/E-MTAB-3050/E-MTAB-3050.raw.1.zip",  # noqa
+            raw_format="MICRO_ARRAY",
+            processed_format="PCL",
+            pipeline_required="MICRO_ARRAY_TO_PCL",
+            accession_code="A-AFFY-1",
+            internal_location="expression_data/array_express/A-AFFY-1/",
```
rzelayafavila: Would it make sense to abstract some of these parameters as command-line arguments or is this command only going to ever be used to load this one batch? In this case should we call it a test command or queue_task_E-MTAB-3050 or something? I don't have particularly strong opinions on this, but I'm wondering what the purpose of this file will be in the future.

workers/data_refinery_workers/downloaders/utils.py
```diff
@@ -0,0 +1,20 @@
+from django.utils import timezone
+from data_refinery_models.models import Batch, BatchStatuses, DownloaderJob
+
+
+def start_job(job: DownloaderJob):
+    """Record in the database that this job is being started. """
+    job.worker_id = "For now there's only one. For now..."
```
rzelayafavila: Where will we get these worker_ids from in the future? Are they automatically assigned by Celery?

data_models/data_refinery_models/models.py
```diff
@@ -119,7 +117,7 @@ class ProcessorJob(TimeTrackedModel):
 
     # This point of this field is to identify which worker ran the job.
     # A few fields may actually be required or something other than just an id.
-    worker_id = models.CharField(max_length=256)
+    worker_id = models.CharField(max_length=256, null=True)
```
kurtwheeler: That would be bad. However I am going to force this to get set by requiring every job to call a provided `start_job` function which will set this.

The reason I don't want to enforce this at the database level though is because I want to be able to create the job record before the job is picked up. Until the job is picked up by a worker, we can't know what the `worker_id` is.

foreman/data_refinery_foreman/surveyor/management/commands/start.py
```diff
@@ -1,4 +1,4 @@
-from django.core.management.base import BaseCommand, CommandError
+from django.core.management.base import BaseCommand
```
kurtwheeler: Correct. Survey Jobs will actually be queued by a scheduler in production.

foreman/Dockerfile.tests
```diff
@@ -1,6 +1,6 @@
 # This Dockerfile is meant to be run from the directory above where it resides.
 
-FROM python:3.5
+FROM python:3.6.1-slim
```
kurtwheeler: Hmm good call, I made some fairly large changes based on that feedback and am very happy with the way it turned out.

workers/data_refinery_workers/downloaders/array_express.py
```diff
@@ -0,0 +1,70 @@
...
+# This path is within the Docker container.
+ROOT_URI = "/home/user/data_store/"
+
+
+@shared_task
+def download_array_express(job_id):
+    job = (DownloaderJob
+           .objects
+           .filter(id=job_id)
+           [:1]
```
kurtwheeler: I think you're very right. I think I found this way of doing it online without understanding what was going on well enough. I agree that this should use the `get` method.

workers/data_refinery_workers/downloaders/array_express.py
```diff
@@ -0,0 +1,70 @@
...
+           .get())
+
+    utils.start_job(job)
+
+    batch = (Batch
+             .objects
+             .filter(id=job.batch_id)
+             [:1]
+             .get())
```
kurtwheeler: Hmm, I should take a look at that and make sure it will fail the right way.

workers/data_refinery_workers/downloaders/management/commands/queue_task.py
kurtwheeler: I meant lines 18 and 19 to clarify that, but I agree it can be more clear. I'll change the name to queue_test_task.py. But yeah it's only for testing, not to be used in production ever.

workers/data_refinery_workers/downloaders/utils.py
kurtwheeler: TBH I haven't thought this all the way through yet. Celery does assign a worker id to each worker however I would think that every time it is restarted these might be lost. There may be an instance id we can get through amazon which might help out. I mention [here](https://github.com/data-refinery/data_refinery/blob/master/data_models/data_refinery_models/models.py#L120) that this part of the model is still up in there air. I have only already included this to both remind myself to change it at some point and to make these functions seem necessary.

kurtwheeler:
Ok @rzelayafavila I've made some changes. Whaddaya think?

data_models/data_refinery_models/models.py
rzelayafavila: Ah I see, that makes sense.

foreman/data_refinery_foreman/surveyor/management/commands/start.py
rzelayafavila: Ah ok - you might have mentioned this somewhere else, but any reason why we are not using Django's test suite to run the tests?

workers/data_refinery_workers/downloaders/array_express.py
```diff
@@ -0,0 +1,72 @@
...
+# chunk_size is in bytes
+CHUNK_SIZE = 4096
+# This path is within the Docker container.
+ROOT_URI = "/home/user/data_store/"
+
+
+@shared_task
+def download_array_express(job_id):
+    try:
+        job = DownloaderJob.objects.get(id=job_id)
```
rzelayafavila: :+1:

workers/data_refinery_workers/downloaders/array_express.py
```diff
@@ -0,0 +1,72 @@
...
+    except ObjectDoesNotExist:
+        logger.error("Cannot find downloader job record with ID %d.", job_id)
+        return
+
+    success = True
+    utils.start_job(job)
+
+    try:
+        batch = Batch.objects.get(id=job.batch_id)
```
rzelayafavila: Nice! :+1:

workers/data_refinery_workers/downloaders/management/commands/queue_test_task.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+# Import and set logger
+import logging
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+# Just a temporary way to queue a celery task
+# without running the surveyor.
```
rzelayafavila: This is one small nitpicky thing that I went back on forth on for a little bit - I think in general we want comments that describe "the operation of ... [a] function or class" in the form of a docstring (as these docs state: http://python-guide-pt-br.readthedocs.io/en/latest/writing/documentation/#docstrings-versus-block-comments), and reserve block comments for things that add extra information for a programmer reading the code (if I am understanding this correctly). However, since this is just a temporary way to do this, I do not feel strongly about it having to be in docstring versus block comments, so it is up to you if you would like to change it or not.

workers/data_refinery_workers/downloaders/utils.py
rzelayafavila: Ok, I see, so this will be defined more in the future.

foreman/Dockerfile
```diff
@@ -13,9 +13,10 @@ COPY data_models/dist/data-refinery-models-* data_models/
 RUN pip install \
     data_models/$(ls data_models -1 | sort --version-sort | tail -1)
 
-
 COPY foreman/ .
 
 USER user
 
-CMD ./manage.py start
+ENTRYPOINT ["python3.6", "manage.py"]
```
rzelayafavila: :+1:

foreman/data_refinery_foreman/surveyor/management/commands/start.py
kurtwheeler: The reason was because this is an end-to-end test and I wasn't sure how to do that yet. It seems like it wouldn't be too hard to do though: http://docs.celeryproject.org/en/latest/userguide/testing.html.

I would agree that I should add them in this PR however I remembered that the next PR I am going to start will be significantly changing this downloader. Therefore rather than writing a bunch of tests now that I know will become obsolete in the very near future.

workers/data_refinery_workers/downloaders/management/commands/queue_test_task.py
kurtwheeler: I went ahead and made that change. But thanks for linking that! I had read it at some point in the past but I guess I misremembered seeing it in pep8 so when I went looking for it I couldn't find it.

foreman/data_refinery_foreman/surveyor/management/commands/start.py
rzelayafavila: Ok gotcha, sounds good!

PR #8, opened 2017-04-12
Link surveyor and downloader

kurtwheeler:
This small PR links together the separate surveyor and downloader code I have already added to the project. The surveyor will now call the downloader task for Array Express for each batch it discovers. It also adds a bit of extra logging and instructions for installing the workers project into the foreman project.

I have tested this by modifying the array express surveyor code to stop after discovering one batch (since I don't yet want to run this for a large amount of files. I then ran `workers/run_worker.sh` and `foreman/run_surveyor.sh`. A Survey Job ran which created Batch and DownloaderJob records along with queuing a download_array_express task. The task then ran and downloaded the file to `workers/volume/ARRAY_EXPRESS/A-AFFY-1/E-MTAB-3050.raw.1.zip`.


mhuyck:
Getting my head around this finally, @kurtwheeler. Will have some comments today.

workers/data_refinery_workers/downloaders/array_express.py
```diff
@@ -33,20 +30,16 @@ def download_array_express(job_id):
...
                 batch.download_url,
-                target_file_name,
+                target_file_path,
                 batch.id,
                 job_id)
 
     try:
-        target_file = open(target_file_name, "wb")
+        target_file = open(target_file_path, "wb")
         request = requests.get(batch.download_url, stream=True)
```
mhuyck: It's good that you have a `target_file.close()` in the `finally` block below, but check to see if you also need to clean up the connection created by `requests.get(..., stream=True)`. See:

> http://docs.python-requests.org/en/master/user/advanced/#body-content-workflow

where it says "If you set stream to True when making a request, Requests cannot release the connection back to the pool unless you consume all the data or call Response.close."

workers/data_refinery_workers/downloaders/management/commands/queue_task.py
```diff
@@ -0,0 +1,47 @@
...
+            internal_location="expression_data/array_express/A-AFFY-1/",
+            organism=1,
+            status=BatchStatuses.NEW.value
+        )
+        batch.save()
+
+        downloader_job = DownloaderJob(batch=batch)
+        downloader_job.save()
+        logger.info("Queuing a task.")
+        download_array_express.delay(downloader_job.id)
```
mhuyck: The dummy data here could make a nice test case...

foreman/data_refinery_foreman/surveyor/external_source.py
```diff
@@ -51,7 +63,8 @@ def handle_batch(self, batch: Batch, key_values: BatchKeyValue = None):
         batch.survey_job = self.survey_job
         batch.source_type = self.source_type()
         batch.status = BatchStatuses.NEW.value
-        batch.internal_location = None
+        batch.internal_location = (self.source_type() + "/"
+                                   + batch.accession_code + "/")
```
kurtwheeler: Ah very nice.

foreman/data_refinery_foreman/surveyor/external_source.py
```diff
@@ -60,14 +73,12 @@ def handle_batch(self, batch: Batch, key_values: BatchKeyValue = None):
             message = ("Batches must have the processed_format field set " +
                        "unless the pipeline returned by determine_pipeline" +
                        "is of the type DiscoveryPipeline.")
```
kurtwheeler: Nifty.

workers/data_refinery_workers/downloaders/management/commands/queue_task.py
kurtwheeler: I would agree except that it'd be hitting the Array Express API for test purposes. I should mock out the API for the test, but in the current branch I'm working on I both significantly rewrote this downloader and also am making tests for it in that branch.

I'd rather not add tests in this PR just to rip them out in the branch I'm currently working on.

workers/data_refinery_workers/downloaders/array_express.py
```diff
@@ -33,20 +30,16 @@ def download_array_express(job_id):
...
 
         for chunk in request.iter_content(CHUNK_SIZE):
```
kurtwheeler: TBH I didn't change this download code too much from what it was when [Rene wrote it](https://bitbucket.org/greenelab/arrayprocess/src/a34824df6100d8bd999b99cf4cab03d502b985ee/download_raw_data_files_by_platform.py?at=default&fileviewer=file-view-default). Now that I've put some thought into this I'd agree this `CHUNK_SIZE` is rather small. However what we actually want it to be will probably depend on the amount of memory on the machines we run it on. I think a quarter of a GB might be a good size for now though. It's large enough to get some performance gains but even if four of them are running in sync the memory usage won't be much over a GB.

And I agree that could be a good future feature. However it seems likely that downloading is going to take much less time than processing so unless this becomes an issue where we have trouble keeping a connection open long enough to download large files I think that this optimization would have a fairly small relative effect. Definitely something to keep in mind for the future though.

workers/data_refinery_workers/downloaders/array_express.py
kurtwheeler: Oh wow yeah that's a good catch!

workers/data_refinery_workers/downloaders/array_express.py
mhuyck: Sounds good. Mostly wanted to be sure we were thinking about these things here, but also hadn't seen the `CHUNK_SIZE` code before and wondered why it was there.

workers/data_refinery_workers/downloaders/management/commands/queue_task.py
mhuyck: Not in this PR then. I'd agree that a mock API is required first.

foreman/data_refinery_foreman/surveyor/external_source.py
kurtwheeler: So I'm definitely all for using `os.path.join()` and have in fact already made the change locally, however how big of an issue do you think portability is given that this project is being build around being run in a docker container? For example [here](https://github.com/data-refinery/data_refinery/pull/8/files#diff-5185347fdc47e66538c13d8187765e49R6) I list the root path as a Linux path which is pointing to a specific place within the docker container. I could turn that into an environment variable in order to make it easier to change based on the environment, however I plan on keeping that location within the Docker container consistent across environments.

kurtwheeler:
@mhuyck I've made a bunch of changes based on your feedback. How's it look?

kurtwheeler:
Notice that #9 was approved and merged into master so I've pulled the changes into this PR. This was mostly seamless but there was some path manipulation that was added which I've changed to use `os.path` functions.

workers/data_refinery_workers/downloaders/array_express.py
```diff
@@ -55,6 +55,7 @@ def download_array_express(job_id):
                      e)
     finally:
         target_file.close()
+        request.close()
```
mhuyck: 👍

foreman/data_refinery_foreman/surveyor/external_source.py
mhuyck: Good question. If we're running a specific flavor of Linux within Docker, I am a bit less concerned but we should still be writing solid Pythonic code wherever possible. (What if we wanted to switch to a different OS in Docker, for instance? Loose coupling is good.) I guess the example of the root path is something I would suggest is a setting that should ideally come in via a settings file or, as you say, an environment variable.

kurtwheeler: I agree with all of that :) At some point I'll probably change that root path to an environment variable.

PR #9, opened 2017-04-20
First processor and the configuration to support it

kurtwheeler:
This PR adds our first processor. It's still in a nascent state, however what data a processor will and will not work for will require additional investigation. However it is capable of processing at least SOME data from Array Express, which is good enough to be a POC.

I think I have also come up with a good way of dealing with R dependencies in this PR. Basically it comes down to putting any R package we want to use into Conda and managing dependencies entirely through Conda (both R and python).

So in order to support all the R dependencies required for this processor I have opened these PRs:
  - https://github.com/bioconda/bioconda-recipes/pull/4468
  - https://github.com/greenelab/conda-channel/pull/1

Which resulted in this conda channel being created and populated: https://anaconda.org/greenelab


workers/r_processors/cel_to_pcl.r
```diff
@@ -0,0 +1,40 @@
+ProcessCelFiles <- function(process.dir, organism.code, output.file) {
+    library("affy")
```
jaclyn-taroni: Which of the functions below comes from these libraries? We typically like to use [double colon operators](https://stat.ethz.ch/R-manual/R-devel/library/base/html/ns-dblcolon.html). I think this is helpful from a readability standpoint (not necessarily always faster than using `library`).

workers/r_processors/cel_to_pcl.r
```diff
@@ -0,0 +1,40 @@
...
+    library("affyio")
+    ##files now holds a list of all celfiles in the directory to be processed
+    files = list.celfiles(process.dir)
```
jaclyn-taroni: So here you would do

```R
files <- affy::list.celfiles(process.dir)
```

Also note the use of `<-` rather than `=` for assignment

workers/r_processors/cel_to_pcl.r
```diff
@@ -0,0 +1,40 @@
...
+    ##mapping holds the three major human platforms, the ReadAffy function used below
+    ##will read the dataset using the CDF specified in here.  This should correspond
+    ##to the brainarray CDFs for Entrez.  I was only using three human platforms before
+    ##so they were hand defined but this loop should map them.
+    mapping <- list()
+    for (level in ptype.levels) {
+        strip.version = gsub('v2', '', level)
+        strip.hyphen = gsub('-', '', strip.version)
+        stripped = gsub('_', '', strip.hyphen)
+        mapping[[level]] <- paste(stripped, organism.code, 'ENTREZG', sep='_')
```
jaclyn-taroni: Spacing should be `sep = '_'`

workers/r_processors/cel_to_pcl.r
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+
+    for (level in ptype.levels) {
+        ##pfiles vector only contains files of this type
+        pfiles = paste(process.dir, subset(files, ptype==level), sep="/")
```
jaclyn-taroni:
```R
pfiles <- paste(process.dir, subset(files, ptype == level), sep = "/") 
```
I would also look into [file.path](https://stat.ethz.ch/R-manual/R-devel/library/base/html/file.path.html) because it should be faster than using paste in this manner.

jaclyn-taroni:
The R LGTM 👍

workers/data_refinery_workers/processors/utils.py
```diff
@@ -0,0 +1,72 @@
...
+        success = True
+
+    job.success = success
+    job.end_time = timezone.now()
+    job.save()
+
+    batch.status = BatchStatuses.PROCESSED.value
+    batch.save()
+
+    return {}
```
dongbohu: What is this return value for?

workers/r_processors/cel_to_pcl.r
```diff
@@ -0,0 +1,43 @@
+ProcessCelFiles <- function (process.dir, organism.code, output.file) {
```
dongbohu: Not an expert on R, so I skipped this file.

workers/tester.sh
```diff
@@ -4,7 +4,7 @@
 
 # This script should always run as if it were being called from
 # the directory it lives in.
-script_directory=`dirname "${BASH_SOURCE[0]}"`
+script_directory=`cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd`
```
dongbohu: Since you run `cd` here already, isn't line 8 redundant?

workers/worker.sh
```diff
@@ -4,7 +4,7 @@
 
 # This script should always run as if it were being called from
 # the directory it lives in.
-script_directory=`dirname "${BASH_SOURCE[0]}"`
+script_directory=`cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd`
```
dongbohu: See my previous comment.

workers/data_refinery_workers/processors/array_express.py
```diff
@@ -0,0 +1,49 @@
...
+    target_directory = utils.ROOT_URI + "processed/" + batch.internal_location
+    os.makedirs(temp_directory, exist_ok=True)
+    os.makedirs(target_directory, exist_ok=True)
+
+    raw_file_name = batch.download_url.split('/')[-1]
+    zip_location = (utils.ROOT_URI + "raw/" + batch.internal_location
+                    + raw_file_name)
+
+    # I need to figure out location tracking
+    zip_ref = zipfile.ZipFile(zip_location, 'r')
```
dongbohu: What if `zip_location` doesn't exist in the file system? Or in the typical use case this issue should never happen?

workers/data_refinery_workers/processors/array_express.py
```diff
@@ -0,0 +1,49 @@
...
+    zip_ref.extractall(temp_directory)
+    zip_ref.close()
+
+    # Experiment code should be added to the batches data model
+    experiment_code = raw_file_name.split('/')[0]
+    new_name = experiment_code + ".pcl"
+
+    ro.r('source("/home/user/r_processors/cel_to_pcl.r")')
```
dongbohu: So this `ro.r()` function will evaluate the R script, I suppose?

workers/data_refinery_workers/processors/management/commands/queue_processor.py
```diff
@@ -0,0 +1,47 @@
+from django.core.management.base import BaseCommand
```
dongbohu: It will be nice if you can write some docstring at the very beginning to describe the purpose and usage of this management command (and any prerequisite if it has).

workers/data_refinery_workers/downloaders/management/commands/queue_downloader.py
```diff
@@ -0,0 +1,47 @@
+from django.core.management.base import BaseCommand
```
dongbohu: It will be nice if you can write some docstring at the very beginning to describe the purpose and usage of this management command (and any prerequisite if it has).

workers/data_refinery_workers/processors/utils.py
kurtwheeler: Honestly not much. I've created a convention where each `processor` should accept a dictionary and return a dictionary so I was following it. However the `end_job` processor should always be the last thing run so it doesn't need to really return anything. Do you think I should remove the return statement entirely?

workers/tester.sh
kurtwheeler: No I don't believe it is. The first `cd` command is run in a subshell so once that subshell exits the current shell hasn't changed directories. To prove this:

```
kurt@moss-greenelab:~/Development/data_refinery/workers$ echo `cd .. && pwd` && pwd
/home/kurt/Development/data_refinery
/home/kurt/Development/data_refinery/workers
```

workers/data_refinery_workers/processors/array_express.py
kurtwheeler: `ro.r()` will execute whatever string is passed to it as R code. Since I want to call the function defined in `cel_to_pcl.r` I first need to load the file into the R environment with `source` and then the subsequent `ro.r()` call actually invokes the function with parameters being passed from python.

workers/data_refinery_workers/processors/utils.py
dongbohu: It is okay if you want to keep the convention. You can write a brief comment here so that the other readers (and maybe yourself in the future) will understand the reason.

workers/tester.sh
dongbohu: You are right. Then what is the advantage of this revision compared to the original one?

kurtwheeler: `dirname` returns a relative path so for just ensuring that the script is running in the correct location it is sufficient, but to get a full path to that directory this change is necessary.

dongbohu: This is a good point. An alternative that I can think of is:
```
script_directory=`realpath $(dirname ${BASH_SOURCE[0]})`
```
Please feel free to use the one that you like more.

kurtwheeler: That makes the intent of the command much more clear. I like it :)

workers/data_refinery_workers/downloaders/management/commands/queue_downloader.py
kurtwheeler: I put it inside the command class itself, do you think I should have put it at the top of this file instead?

kurtwheeler:
Ok @dongbohu I think I covered everything you asked about. However I did significantly alter the processors/utils.py file to handle errors well. Wanna take a look at my changes and see if there's anything else that needs to be fixed?

workers/data_refinery_workers/downloaders/management/commands/queue_downloader.py
dongbohu: Probably better to put it at the very top so that a reader can find it immediately. And the usage part will fit better at the top too because it is the usage of this management command (instead of the class `Command`).

workers/data_refinery_workers/processors/management/commands/queue_processor.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+# Import and set logger
+import logging
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    """This command is intended for development purposes.
```
dongbohu: See my comments on its location.

workers/data_refinery_workers/processors/utils.py
```diff
@@ -0,0 +1,112 @@
...
+    adds them to the dictionary passed in with the keys 'job'
+    and 'batch' respectively."""
+    job = kwargs["job"]
+    job.worker_id = "For now there's only one. For now..."
+    job.start_time = timezone.now()
+    job.save()
+
+    try:
+        batch = Batch.objects.get(id=job.batch_id)
+    except ObjectDoesNotExist:
```
dongbohu: By the way, you can also write:
```
except Batch.DoesNotExist:
```
then you don't need to import `ObjectDoesNotExist` on line 4. Just a shortcut that I think is good to know.

workers/tester.sh
```diff
@@ -4,7 +4,7 @@
 
 # This script should always run as if it were being called from
 # the directory it lives in.
-script_directory=`dirname "${BASH_SOURCE[0]}"`
+script_directory=`realpath $( dirname "${BASH_SOURCE[0]}" )`
```
dongbohu: Just realized another optionthat uses a pipe:
```
script_directory=`dirname "${BASH_SOURCE[0]}"  | xargs realpath`
```
Please pick up whatever way you feel makes more sense.

workers/worker.sh
```diff
@@ -4,7 +4,7 @@
 
 # This script should always run as if it were being called from
 # the directory it lives in.
-script_directory=`dirname "${BASH_SOURCE[0]}"`
+script_directory=`realpath $( dirname "${BASH_SOURCE[0]}" )`
```
dongbohu: See my previous comment.

workers/data_refinery_workers/processors/utils.py
```diff
@@ -0,0 +1,112 @@
...
+    processed.
+    It is required that the dictionary returned by each processor
+    function preserve the mappings for 'job' and 'batch' that were
+    passed into it.
+    """
+
+    job_id = start_value["job_id"]
+    try:
+        job = ProcessorJob.objects.get(id=job_id)
+    except ObjectDoesNotExist:
```
dongbohu: See my previous comment.

workers/r_processors/process_cel_to_pcl.R
```diff
@@ -0,0 +1,42 @@
...
+
+
+  for (level in ptype.levels) {
+    # pfiles vector only contains files of this type
+    pfiles <- file.path(process.dir, subset(files, ptype == level))
+    # ReadAffy loads the array data using the custom CDF
+    data <- affy::ReadAffy(cdfname = mapping[[level]], filenames = pfiles)
+    express <- affy::rma(data)
+
+    write.exprs(express, file = output.file)
```
jaclyn-taroni: It just occurred to me that this should probably be `Biobase::write.exprs(express, file = output.file)`

kurtwheeler:
Ok @dongbohu I think this should be good to go at this point. What do you think?

PR #10, opened 2017-05-04
Link up processor

kurtwheeler:
This is a fairly straightforward change which just queues up a processor job after finishing a downloader job. However to support that change there was some processor discovery code that was needed. Additionally this PR ran into the difficulty of trying to import the processor dependencies into the foreman project. This was an issue because the processors now have a lot of dependencies. To work around this I am now calling the downloader task from the surveyor via the task name, rather than importing the task and calling via it's `delay()` method.


foreman/data_refinery_foreman/surveyor/external_source.py
```diff
@@ -82,6 +80,9 @@ def handle_batch(self, batch: Batch, key_values: BatchKeyValue = None):
                        "is of the type DiscoveryPipeline.")
             raise InvalidProcessedFormatError(message)
 
+        batch.internal_location = os.path.join(batch.accession_code,
+                                               batch.pipeline_required)
+
```
kurtwheeler: I'm not sure why this needed to be changed. I swear this was an issue in 3 different merge conflicts I resolved and I put it afterwards anyway. This should have already been like this, but I guess maybe I messed up one of the merge conflicts?

foreman/data_refinery_foreman/surveyor/external_source.py
```diff
@@ -31,7 +31,7 @@ class PipelineEnums(Enum):
 
 class ProcessorPipeline(PipelineEnums):
     """Pipelines which perform some kind of processing on the data."""
-    MICRO_ARRAY_TO_PCL = "MICRO_ARRAY_TO_PCL"
+    AFFY_TO_PCL = "AFFY_TO_PCL"
```
rzelayafavila: So this just to narrow it down to just AFFY platforms instead of any type of microarray?

rzelayafavila: By the way, will more attributes be added here as we add more capabilities (such as ``AGILENT_TO_PCL`` or something like that)?

foreman/data_refinery_foreman/surveyor/external_source.py
rzelayafavila: Hmm interesting - yeah, it makes sense that it should be here

foreman/data_refinery_foreman/surveyor/message_queue.py
```diff
@@ -0,0 +1,24 @@
...
+os.environ.setdefault('DJANGO_SETTINGS_MODULE',
+                      'data_refinery_workers.settings')
+
+app = Celery('data_refinery_workers')
+
+# Using a string here means the worker don't have to serialize
+# the configuration object to child processes.
+# - namespace='CELERY' means all celery-related configuration keys
+#   should have a `CELERY_` prefix.
+app.config_from_object('django.conf:settings', namespace='CELERY')
```
rzelayafavila: I think this file looks pretty good and straightforward.

foreman/data_refinery_foreman/surveyor/array_express.py
```diff
@@ -27,12 +25,13 @@ def source_type(self):
         return "ARRAY_EXPRESS"
 
     def downloader_task(self):
-        return download_array_express
+        return ("data_refinery_workers.downloaders"
+                + ".array_express.download_array_express")
```
kurtwheeler: Hmm, I'm actually not sure. The way I see it is that this is essentially a constant right now. What do you think the benefits of defining it as a constant instead of as an attribute are? Explicitness about it's status as a constant?

foreman/data_refinery_foreman/surveyor/external_source.py
kurtwheeler: Yes to both questions. Once we actually start adding lots of processors the naming conventions may need to be even more specific (such as AFFY_HGU133A and AFFY_HGU133B).

rzelayafavila: Ok, gotcha

foreman/data_refinery_foreman/surveyor/array_express.py
rzelayafavila: Yeah - I think it is better to define it as a constant up top. That way we know it is currently set at whatever value it is. I think it would not be too far-fetched for the module name to change at some point.

PR #12, opened 2017-05-10
Improve metadata

kurtwheeler:
This PR got out of hand. The original scope of it was to improve the metadata we were collecting about batches of data from Array Express. However halfway through completing that it was decided that we should change the batch size for Array Express from an experiment to a single sample. Since the batches the metadata was about were changing, it didn't seem to make sense to change that metadata without also changing the batch size. However in the process of doing that I discovered that it's not possible to download a single batch worth of data (i.e. a single sample). Therefore I needed to also refactor the assumption that `one downloader job`==`one batch`. This was a large change that affected a lot of the existing code. I also added a lot of new test code to previously untested functionality, because up until these changes that functionality was changing so much that I was hesitant to write tests for it.

So that's the story of why I now have so many changes in a single PR. I'm not entirely sure how to proceed with this PR. I don't think that too much of it can be broken out into smaller PRs. The only large chunk of code that makes sense on its own is the organism model code. However breaking this out will make things somewhat more complicated because the addition of the organism model code prompted me to break the `models.py` file up into several files contained in a module. I then made changes to other models as I implemented the rest of the functionality in this branch.

Anyway... I'm happy to do whatever makes this PR more digestible for reviewers. If that means breaking out the organism model code into it's own PR and duplicating the split of `models.py` in two PRs I can do that. If it's easier to leave it in this PR but have a separate reviewer look at the organism model code I can do that.


kurtwheeler:
Oh yeah I forgot to mention that there's some new migrations that this introduces. I'm going to reset the migrations so that there's just one migration file after this is merged. However I am waiting on figuring out what to do about the organism model stuff before I do so.

kurtwheeler:
I've been asked to split up the files in this PR into 3 parts. I think the easiest way to do this (and one that is actually roughly equal while still keeping the parts logically grouped) is to have one person look at the files within `data_models/`, one person look at files within `foreman/` and one person to look at the rest (which is mostly within `workers/` with a few in the top level directory).

dongbohu:
Thank you @kurtwheeler. I will take care of the files in `data_models` (unless @rzelayafavila or @mhuyck has strong preferences on it).

mhuyck:
It probably makes sense for me to look at the `workers/`, etc., group since that includes the code that interacts with ArrayExpress, and I have some experience with that from the `adage-server` project.

data_models/data_refinery_models/models/test_organisms.py
```diff
@@ -0,0 +1,238 @@
...
+        taxonomy_id = Organism.get_id_for_name("Human")
+
+        self.assertEqual(taxonomy_id, 9606)
+        mock_get.assert_has_calls([
+            call(ESEARCH_URL,
+                 {"db": "taxonomy", "field": "scin", "term": "HUMAN"}),
+            call(ESEARCH_URL,
+                 {"db": "taxonomy", "term": "HUMAN"})])
+
+        # The first call should have stored it in the database so
```
dongbohu: I suppose "it" here refers the requested organism? Do you mind elaborating this line a little bit to make the comment more readable? Thanks.

data_models/data_refinery_models/models/test_organisms.py
```diff
@@ -0,0 +1,238 @@
...
+
+        organism_name = Organism.get_name_for_id(9606)
+
+        self.assertEqual(organism_name, "HOMO SAPIENS")
+        mock_get.assert_called_once_with(
+            EFETCH_URL,
+            {"db": "taxonomy", "id": "9606"}
+        )
+
+        # The first call should have stored it in the database so
```
dongbohu: See my previous comment on "it".

data_models/data_refinery_models/models/test_organisms.py
```diff
@@ -0,0 +1,238 @@
...
+            call(ESEARCH_URL,
+                 {"db": "taxonomy", "term": "BLAH"})])
+
+        # The first call should have stored it in the database so
+        # this call should not make a request.
+        mock_get.reset_mock()
+        new_id = Organism.get_id_for_name("BLAH")
+
+        self.assertEqual(new_id, 0)
+        mock_get.assert_not_called()
```
dongbohu: Great test cases overall! :100:

data_models/data_refinery_models/migrations/0001_initial.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Generated by Django 1.11 on 2017-05-02 18:50
+# Generated by Django 1.10.6 on 2017-05-23 15:17
```
dongbohu: I skipped this file, is it okay?

data_models/data_refinery_models/models/batches.py
```diff
@@ -0,0 +1,53 @@
...
+    NEW = "NEW"
+    DOWNLOADED = "DOWNLOADED"
+    PROCESSED = "PROCESSED"
+
+
+class Batch(TimeTrackedModel):
+    survey_job = models.ForeignKey(SurveyJob, on_delete=models.PROTECT)
+    source_type = models.CharField(max_length=256)
+    size_in_bytes = models.IntegerField()
+    download_url = models.CharField(max_length=2048)
```
kurtwheeler: Hmm I suppose it'd be better to have an unnecessarily large field here than have to make a migration later just to extend this.

data_models/data_refinery_models/models/organism.py
```diff
@@ -0,0 +1,121 @@
...
+from data_refinery_models.models.base_models import TimeTrackedModel
+
+# Import and set logger
+import logging
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+NCBI_ROOT_URL = "https://eutils.ncbi.nlm.nih.gov/entrez/eutils/"
+ESEARCH_URL = NCBI_ROOT_URL + "esearch.fcgi"
+EFETCH_URL = NCBI_ROOT_URL + "efetch.fcgi"
```
kurtwheeler: These are very stable. The docs haven't changed since 2013.

data_models/data_refinery_models/models/organism.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+class UnscientifcNameError(BaseException):
```
kurtwheeler: Definitely a typo. Good catch.

data_models/data_refinery_models/models/organism.py
```diff
@@ -0,0 +1,121 @@
...
+    pass
+
+
+class InvalidNCBITaxonomyId(BaseException):
+    pass
+
+
+def get_scientific_name(taxonomy_id: int):
```
kurtwheeler: You're right. I started off with 3.5 so I didn't realize these were new in 3.5, I thought they were just a Python 3 thing. I'll update the README. I'll also add in some return values.

data_models/data_refinery_models/models/organism.py
```diff
@@ -0,0 +1,121 @@
...
+    root = ElementTree.fromstring(response.text)
+    taxon_list = root.findall("Taxon")
+
+    if(len(taxon_list) == 0):
+        logger.error("No names returned by ncbi.nlm.nih.gov for organism "
+                     + "with taxonomy ID %d.",
+                     taxonomy_id)
+        raise InvalidNCBITaxonomyId
+
+    return taxon_list[0].find("ScientificName").text.upper()
```
kurtwheeler: I'm going to say yes. Of course it's always possible for the API to have a bug and fail here, but it's an old API which is widely used so I'm going to go ahead and trust them.

data_models/data_refinery_models/migrations/0001_initial.py
kurtwheeler: Yep, it was automatically generated.

data_models/data_refinery_models/models/organism.py
dongbohu: :+1:

data_models/data_refinery_models/models/base_models.py
```diff
@@ -0,0 +1,18 @@
+from django.db import models
+from django.utils import timezone
+
+
+class TimeTrackedModel(models.Model):
```
dongbohu: Just realized that this class needs docstring too. Thanks.

data_models/data_refinery_models/models/batches.py
```diff
@@ -0,0 +1,62 @@
+from enum import Enum
+from django.db import models
+from data_refinery_models.models.base_models import TimeTrackedModel
+from data_refinery_models.models.surveys import SurveyJob
+
+
+class BatchStatuses(Enum):
```
dongbohu: Docstring please.

rzelayafavila:
Ok cool, I am reviewing the files in the ``foreman/`` directory.

data_models/data_refinery_models/models/organism.py
rzelayafavila: I actually have a couple of questions here (even though I am reviewing the ``foreman/`` folder):
* So you decided to make a new Organism model instead of adding functionality to the one in ``django-organisms`` (https://bitbucket.org/greenelab/django-organisms)? I think that's totally fine, but any reason you decided to go this route?
* I do think that even if these URLs are relatively stable, it might be nice for the user to be able to override them in the ``settings.py`` file (but we can set these as defaults here in case the user does not specify them). I think this would be convenient in the future.

data_models/data_refinery_models/models/organism.py
```diff
@@ -0,0 +1,129 @@
...
+
+    name = models.CharField(max_length=256)
+    taxonomy_id = models.IntegerField()
+    is_scientific_name = models.BooleanField(default=False)
+
+    @classmethod
+    def get_name_for_id(cls, taxonomy_id: int) -> str:
+        try:
+            organism = (cls.objects
+                        .filter(taxonomy_id=taxonomy_id)
```
rzelayafavila: Is there any case when the taxonomy_id would not be unique?

data_models/data_refinery_models/models/organism.py
kurtwheeler: There was definitely a reason or two why I decided to go this route, however they happened in person. What I can think of now though is that the `django-organisms` model has two names, common and scientific. However the NCBI API has a list structure for names which can include more than two and I wanted to create a model which could support that. Ultimately I think that the API I was trying to create (`get_id_for_name` and `get_name_for_id`) was more important than the data model I was creating to support it.

As to your other point, I'm not sure I like the idea of allowing someone to override the URLs without modifying the source code because it seems highly unlikely that someone could just pop in a different URL and have it work. The parameters that used for each request would almost certainly need to change and if they didn't it could cause undefined behavior. What use case do you think might be supported by allowing the URLs to be via the `settings.py` file?

data_models/data_refinery_models/models/organism.py
kurtwheeler: Yes. If the same organism was listed with two different names in two different batches but they both corresponded to the same NCBI taxonomy_id (since NCBI has a list structure for names the same ID can match multiple names), then there would be one record for each of those names, however they would share an id.

rzelayafavila: Hmmm... I think it would be interesting to see if that just corresponded to different names for the same actual organism object vs. different names for different organism objects (but then why would they have the same taxonomy ID)? Are there any examples in the stuff that was returned from the NCBI API of this?

data_models/data_refinery_models/models/organism.py
rzelayafavila: I think that sounds reasonable for the organisms - if you think this is a better fit then go for it.

In terms of the URLs, I think that at some point we might want to get this information from another URL endpoint. Also, what if we have a saved XML file with the organism information we want and want to pass it to these methods and model? What parameters would we need to abstract to make this possible?

data_models/data_refinery_models/models/organism.py
kurtwheeler: I'm not quite sure I understand what you're asking. What do you mean by an organism object?

workers/data_refinery_workers/processors/test_utils.py
```diff
@@ -0,0 +1,179 @@
...
+        processor_job.save()
+        processor_jobs_to_batches = ProcessorJobsToBatches(batch=batch,
+                                                           processor_job=processor_job)
+        processor_jobs_to_batches.save()
+
+        mock_processor = MagicMock()
+        mock_dict = {"something_to_pass_along": True,
+                     "job": processor_job,
+                     "batches": [batch]}
+        mock_dict["something_to_pass_along"] = True
```
mhuyck: This line is redundant with the end of line 156. Did you mean to leave it here?

workers/data_refinery_workers/processors/utils.py
```diff
@@ -16,20 +13,21 @@
...
     job.worker_id = "For now there's only one. For now..."
     job.start_time = timezone.now()
     job.save()
 
-    try:
-        batch = Batch.objects.get(id=job.batch_id)
-    except Batch.DoesNotExist:
-        logger.error("Cannot find batch record with ID %d.", job.batch_id)
+    batch_relations = ProcessorJobsToBatches.objects.filter(processor_job_id=job.id)
+    batches = list(map(lambda x: x.batch, batch_relations))
```
mhuyck: This line can be simpler as described above:

`batches = [br.batch for br in batch_relations]`

foreman/data_refinery_foreman/surveyor/array_express.py
```diff
@@ -16,68 +17,111 @@
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
 
+EXPERIMENTS_URL = "https://www.ebi.ac.uk/arrayexpress/json/v3/experiments/"
```
kurtwheeler: I don't think it would because if that URL ever changes then I think there will be additional changes that need to happen to work with that URL. We can also just change this in the code instead of in the `settings.py` file if we need to.

data_models/data_refinery_models/migrations/0001_initial.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Generated by Django 1.11 on 2017-05-02 18:50
+# Generated by Django 1.10.6 on 2017-05-26 15:12
```
kurtwheeler: I did this a while ago so I'm not 100% certain what it was but I believe I found a bug that was introduced in 1.11. I wish I had left a comment in the `requirements.in` file.

workers/Dockerfile.tests
```diff
@@ -0,0 +1,23 @@
+FROM continuumio/miniconda3:4.3.11
+
+RUN groupadd user && useradd --create-home --home-dir /home/user -g user user
+WORKDIR /home/user
+
+COPY workers/environment.yml .
```
kurtwheeler: This [COPY command](https://docs.docker.com/engine/reference/builder/#copy) is not just copying the file from one directory to another, but rather from the directory that the Dockerfile is being run from into the docker container. Docker containers start off completely empty so

```
RUN conda env create -f workers/environment.yml
```

wouldn't work because the container can't access a file until it's copied into the container.

workers/Dockerfile.tests
```diff
@@ -0,0 +1,23 @@
...
+
+RUN conda env create -f environment.yml
+
+ENV PATH "/opt/conda/envs/workers/bin:$PATH"
+
+COPY data_models/dist/data-refinery-models-* data_models/
```
kurtwheeler: The answer to this question is similar to the answer to the one above it. Because the [RUN command](https://docs.docker.com/engine/reference/builder/#run) is always executed within the context of the container I cannot run `(ls data_models -1 | tail -1)` in the Dockerfile and have it interact with files/directories within the host machine without first copying the files into the container. If it were possible, I certainly would agree that figuring out the correct file and THEN copying it into the container would be greatly preferable.

workers/data_refinery_workers/downloaders/array_express.py
```diff
@@ -1,19 +1,71 @@
...
+        zip_ref = zipfile.ZipFile(file_path, 'r')
+        zip_ref.extractall(os.path.dirname(file_path))
+    except Exception:
+        logging.exception("Exception caught while extracting %s during Job #%d.",
+                          file_path,
+                          job_id)
+        raise
+    finally:
+        zip_ref.close()
+        os.remove(file_path)
```
kurtwheeler: I think this is correct because if extraction fails then the downloader job will fail and need to be retried. When that happens the zip file will be re-downloaded so it won't be an issue that it's been deleted. Additionally the worker that picks up the new downloader job won't necessarily be the same one so the zip file could end up sitting on the original worker's disk forever.

workers/data_refinery_workers/downloaders/array_express.py
```diff
@@ -23,44 +75,29 @@ def download_array_express(job_id):
...
     utils.start_job(job)
 
-    try:
-        batch = Batch.objects.get(id=job.batch_id)
-    except ObjectDoesNotExist:
-        logger.error("Cannot find batch record with ID %d.", job.batch_id)
-        utils.end_job(job, None, False)
-        return
+    batch_relations = DownloaderJobsToBatches.objects.filter(downloader_job_id=job_id)
+    batches = list(map(lambda x: x.batch, batch_relations))
```
kurtwheeler: Oh yes, that's very nice. List comprehensions are probably the feature of python I need to become comfortable with the most.

workers/data_refinery_workers/downloaders/array_express.py
```diff
@@ -23,44 +75,29 @@ def download_array_express(job_id):
...
-                target_file_path,
-                batch.id,
-                job_id)
+    if len(batches) > 0:
+        target_file_path = utils.prepare_destination(batches[0])
+        download_url = batches[0].download_url
+    else:
+        logger.error("No batches found for job #%d.",
+                     job_id)
+        success = False
```
kurtwheeler: Yes. Good catch. I should wrap that entire `try` block in a `if success:`.

workers/data_refinery_workers/downloaders/array_express.py
```diff
@@ -23,44 +75,29 @@ def download_array_express(job_id):
...
-        request = requests.get(batch.download_url, stream=True)
-
-        for chunk in request.iter_content(CHUNK_SIZE):
-            if chunk:
-                target_file.write(chunk)
-                target_file.flush()
-    except Exception as e:
+        _verify_batch_grouping(batches, job_id)
+        _download_file(download_url, target_file_path, job_id)
+        _extract_file(target_file_path, job_id)
```
kurtwheeler: Yep you've got it and I agree a comment would be helpful

workers/data_refinery_workers/downloaders/management/commands/queue_downloader.py
```diff
@@ -33,18 +34,27 @@ def handle(self, *args, **options):
...
             status=BatchStatuses.NEW.value
         )
         batch.save()
 
-        downloader_job = DownloaderJob(batch=batch)
+        downloader_job = DownloaderJob()
         downloader_job.save()
+        downloader_job_to_batch = DownloaderJobsToBatches(batch=batch,
+                                                          downloader_job=downloader_job)
+        downloader_job_to_batch.save()
```
kurtwheeler: That is a good though, I like it. I'll make an issue about it and add it once the git tree isn't quite so hairy.

workers/data_refinery_workers/downloaders/management/commands/queue_downloader.py
```diff
@@ -33,18 +34,27 @@ def handle(self, *args, **options):
...
         logger.info("Queuing a task.")
         download_array_express.delay(downloader_job.id)
```
kurtwheeler: That's exactly what's going on here. `@shared_task` turns the function into an instance of Celery's Task class.

workers/data_refinery_workers/downloaders/test_array_express.py
```diff
@@ -0,0 +1,117 @@
...
+                   Batch(download_url="https://example.com"),
+                   Batch(download_url="https://example.com")]
+        job_id = 1
+        self.assertIsNone(array_express._verify_batch_grouping(batches, job_id))
+
+    def test_bad_batch_grouping(self):
+        """Raises exception if all batches don't have the same download_url."""
+        batches = [Batch(download_url="https://example.com"),
+                   Batch(download_url="https://example.com"),
+                   Batch(download_url="https://wompwomp.com")]
```
kurtwheeler: Only for 1800 bucks lol. `sadtrombone.com` is great though, it's everything I'd ever want that site to be.

workers/data_refinery_workers/downloaders/utils.py
```diff
@@ -27,31 +29,37 @@ def start_job(job: DownloaderJob):
...
-    """Record in the database that this job has completed,
-    create a processor job, and queue a processor task."""
-    job.success = success
-    job.end_time = timezone.now()
-    job.save()
-
+def end_job(job: DownloaderJob, batches: Batch, success):
+    """Record in the database that this job has completed.
+    Create a processor job and queue a processor task for each batch
+    if the job was successful."""
```
kurtwheeler: I do have flycheck and autopep8 both running haha. Since I've written this I've read up more on doc string conventions though so I've already started doing it more correctly in my new code. I'll definitely fix this up though.

workers/data_refinery_workers/processors/test_utils.py
kurtwheeler: I don't think so, good catch.

foreman/data_refinery_foreman/surveyor/array_express.py
```diff
@@ -16,68 +17,111 @@
...
+                    organism_id=organism_id,
+                    organism_name=organism_name,
+                    experiment_title=experiment["name"],
+                    release_date=experiment["release_date"],
+                    last_uploaded_date=experiment["last_update_date"],
+                    name=sample_file["name"]
+                ))
+
+        # Group batches based on their download URL and handle each group.
+        download_urls = set(map(lambda x: x.download_url, batches))
```
rzelayafavila: I think @mhuyck had a similar suggestion in his review. I think you can do a set comprehension for this line to make it simpler:
``download_urls = {bt.download_url for bt in batches}``

foreman/data_refinery_foreman/surveyor/array_express.py
```diff
@@ -16,68 +17,111 @@
...
+        for url in download_urls:
+            batches_with_url = list(filter(lambda x: x.download_url == url, batches))
```
rzelayafavila: Likewise, I think you can do:
``batches_with_url = [bt.download_url for bt in batches if bt.download_url == url]``

foreman/data_refinery_foreman/surveyor/array_express.py
```diff
@@ -16,68 +17,111 @@
...
+
+            if organism_name == "UNKNOWN":
+                logger.error("Sample from experiment %s did not specify the organism name.",
+                             experiment_accession_code)
+                organism_id = 0
+            else:
+                organism_id = Organism.get_id_for_name(organism_name)
+
+            for sample_file in sample["file"]:
+                if sample_file["type"] != "data" and sample_file["name"] is not None:
```
rzelayafavila: Would there be any case where ``sample_file["name"]`` would be ``False`` or ``""`` instead of ``None``, but we would still not want the sample_file?

data_models/data_refinery_models/migrations/0001_initial.py
mhuyck: Sounds intentional then. 👍

foreman/data_refinery_foreman/surveyor/management/commands/survey.py
```diff
@@ -0,0 +1,30 @@
+"""
+This command will create and run survey jobs for each experiment in the
+experiment_list. experiment list should be a file containing one
+experiment accession code per line.
```
rzelayafavila: :+1:

workers/Dockerfile.tests
mhuyck: That was the piece I was missing. Thanks! Makes sense now.

foreman/data_refinery_foreman/surveyor/management/commands/survey.py
```diff
@@ -0,0 +1,30 @@
...
+            "experiment_list",
+            help=("A file containing a list of experiment accession codes to "
+                  "survey, download, and process. These should be listed one "
+                  "per line. Should be a path relative to the foreman "
+                  "directory."))
+
+    def handle(self, *args, **options):
+        if options["experiment_list"] is None:
+            logger.error("You must specify an experiment list.")
+            return 0
```
rzelayafavila: Doesn't a ``0`` return value mean that a function has succeeded?

workers/data_refinery_workers/downloaders/array_express.py
mhuyck: Good plan, then. Thanks for the clear explanation!

I was thinking that it would be easier to diagnose a failure if the `.zip` file were left behind, but automated re-trying makes sense. If the error is repeatable it could probably be diagnosed by a newly downloaded `.zip` file as well.

foreman/requirements.txt
```diff
@@ -7,11 +7,16 @@
 amqp==2.1.4               # via kombu
 billiard==3.5.0.2         # via celery
 celery==4.0.2
-django==1.10.6
+django==1.11
```
rzelayafavila: I see you are running different versions of django in different parts of this project. I'm just curious, any reason for that?

foreman/data_refinery_foreman/surveyor/test_surveyor.py
```diff
@@ -26,7 +26,7 @@ def test_calls_survey(self, survey_method):
         job.save()
         surveyor.run_job(job)
 
-        survey_method.assert_called_with(job)
+        survey_method.assert_called_with()
```
rzelayafavila: Hmmm, now that this method isn't taking this argument, I'm wondering what exactly it is doing. Should it be called ``assert_called_with_job`` or something of the sort? Also, where is it defined again?

foreman/data_refinery_foreman/surveyor/array_express.py
```diff
@@ -16,68 +17,111 @@
...
+            experiment["last_update_date"] = parsed_json["releasedate"]
+
+        return experiment
+
+    def survey(self):
+        experiment_accession_code = (
+            SurveyJobKeyValue
+            .objects
+            .get(survey_job_id=self.survey_job.id,
+                 key__exact="experiment_accession_code")
```
kurtwheeler: This `survey` function is actually called by `surveyor.run_job` within a `try` block. If there's no match here the exception will be caught, logged, and the survey_job will be marked as a failure. Given that an `experiment_accession_code` is required here I think that's the appropriate behavior.

foreman/data_refinery_foreman/surveyor/array_express.py
```diff
@@ -16,68 +17,111 @@
...
+            .value
+        )
+
+        logger.info("Surveying experiment with accession code: %s.", experiment_accession_code)
+
+        experiment = self.get_experiment_metadata(experiment_accession_code)
+
+        r = requests.get(SAMPLES_URL.format(experiment_accession_code))
+        samples = r.json()["experiment"]["sample"]
```
kurtwheeler: My answer here is basically the same as the one above except that I'll add that at some point surveying will be triggered by a foreman who will be responsible for starting and retrying survey_jobs. If the API is down and the job fails then once I get around to writing the foreman code it'll have mechanisms to retry the survey_jobs a few times so hopefully the API will come back up before the retry limit is reached.

foreman/data_refinery_foreman/surveyor/management/commands/survey.py
kurtwheeler: Oh yeah, good catch.

foreman/data_refinery_foreman/surveyor/test_surveyor.py
kurtwheeler: This should just be `assert_called`. The `survey_method` Mock is passed in by the `@patch` decorator, which is mocking `surveyor.ArrayExpressSurveyor`'s `survey` method.

The reason I no longer am passing job to this method is that the `ArrayExpressSurveyor` object has the job as an instance variable, so it doesn't need to have it passed in.

foreman/requirements.txt
kurtwheeler: Hmm, whoops. I don't think I have a reason to be doing that. I think I may have run into a bug that was introduced in `1.11` though so that's why most of the project is `1.10.6`. I'll change this to match that.

foreman/data_refinery_foreman/surveyor/array_express.py
rzelayafavila: Ok, that sounds good

foreman/data_refinery_foreman/surveyor/array_express.py
rzelayafavila: :+1:

foreman/requirements.txt
rzelayafavila: :+1:

foreman/data_refinery_foreman/surveyor/test_surveyor.py
rzelayafavila: Ok, if you change it to ``assert_called``, I think that it makes sense

run_rabbitmq.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+docker run -d --hostname rabbit-queue --name message-queue rabbitmq:3
```
mhuyck: 👍

workers/data_refinery_workers/downloaders/array_express.py
```diff
@@ -23,44 +75,34 @@ def download_array_express(job_id):
...
+        logger.error("No batches found for job #%d.",
+                     job_id)
+        success = False
 
-    logger.info("Downloading file from %s to %s. (Batch #%d, Job #%d)",
-                batch.download_url,
-                target_file_path,
-                batch.id,
-                job_id)
+    if success:
```
mhuyck: 👍

workers/data_refinery_workers/downloaders/array_express.py
```diff
@@ -23,44 +75,34 @@ def download_array_express(job_id):
...
-                     + "with message: %s",
-                     job_id,
-                     batch.id,
-                     e)
-    finally:
-        target_file.close()
-        request.close()
+            # The files for all of the batches in the grouping are
+            # contained within the same zip file. Therefore only
+            # download the one.
```
mhuyck: 👍

workers/data_refinery_workers/processors/test_utils.py
```diff
@@ -0,0 +1,180 @@
...
+
+        kwargs = utils.start_job({"job": processor_job})
+        # start_job preserves the "job" key
+        self.assertEqual(kwargs["job"], processor_job)
+
+        # start_job finds the batches and returns them
+        self.assertEqual(len(kwargs["batches"]), 2)
+
+    def test_failure(self):
+        """Fails because there are no batches for the job."""
```
mhuyck: 👍

workers/data_refinery_workers/downloaders/array_express.py
```diff
@@ -23,44 +75,34 @@ def download_array_express(job_id):
...
     utils.start_job(job)
 
-    try:
-        batch = Batch.objects.get(id=job.batch_id)
-    except ObjectDoesNotExist:
-        logger.error("Cannot find batch record with ID %d.", job.batch_id)
-        utils.end_job(job, None, False)
-        return
+    batch_relations = DownloaderJobsToBatches.objects.filter(downloader_job_id=job_id)
+    batches = [br.batch for br in batch_relations]
```
mhuyck: 👍

workers/data_refinery_workers/processors/utils.py
```diff
@@ -15,27 +12,34 @@
...
     job.worker_id = "For now there's only one. For now..."
     job.start_time = timezone.now()
     job.save()
 
-    try:
-        batch = Batch.objects.get(id=job.batch_id)
-    except Batch.DoesNotExist:
-        logger.error("Cannot find batch record with ID %d.", job.batch_id)
+    batch_relations = ProcessorJobsToBatches.objects.filter(processor_job_id=job.id)
+    batches = [br.batch for br in batch_relations]
```
mhuyck: 👍

mhuyck:
Except for a handful of `"""` at the end of multi-line docstrings that should be on a line by themselves, you got everything I noted. Take one more pass to find those before you merge, but everything else I reviewed looks good now. Nice work, @kurtwheeler!

foreman/data_refinery_foreman/surveyor/array_express.py
kurtwheeler: Ok this line of code confused me. I can't tell why we would ever want to continue if `sample_file["name"]` is `None` since we use that value later and call split on it. I think it's probably just a bug so I'm going to remove everything from `and` onward.

kurtwheeler:
OK @rzelayafavila I just pushed the last change you requested I think. Is it lookin good?

PR #15, opened 2017-06-05
Use rds instance

kurtwheeler:
This pull requests integrates AWS' RDS and SQS services into the data refinery. The configuration of these is in Terraform. The RDS instance is used via Django's ORM still, the needed credentials are encrypted via git crypt in the `*/environments/prod` files. In fact the integration with SQS is also done in the environment file via the `CELERY_BROKER_URL` variable. Keeping these changes in the environment files makes it easy to switch between developing on a personal machine and deploying within the cloud: just change the environment from dev to prod.

I tested these integrations by runnning some jobs with the prod environment set, looking at the SQS queue via AWS' interface, and by connecting to the RDS instance via `psql` and verifying that records were being created and used in it.


kurtwheeler:
I'm going to assign this one to @rzelayafavila since you've read up on Terraform.

foreman/requirements.txt
```diff
@@ -6,6 +6,7 @@
 #
 amqp==2.1.4               # via kombu
 billiard==3.5.0.2         # via celery
+boto==2.46.1
```
kurtwheeler: Celery is still using boto v2.x.x for it's SQS integration. I am using boto3 for the S3 integration though!

.gitattributes
```diff
@@ -0,0 +1,4 @@
+prod filter=git-crypt diff=git-crypt
+*.tfstate filter=git-crypt diff=git-crypt
+terraform.tfvars filter=git-crypt diff=git-crypt
+*.tfstate.backup filter=git-crypt diff=git-crypt
```
kurtwheeler: Yup, they contain the Database password.

requirements.txt
```diff
@@ -6,7 +6,9 @@
 #
 autopep8==1.3.1
 click==6.7                # via pip-tools
-django==1.10.6
+configparser==3.5.0       # via flake8
+django==1.11
+enum34==1.1.6             # via flake8
```
kurtwheeler: > Do you mean that these packages (configparser, enum34, mccabe, etc) get installed not because they are direct dependencies to your project, but rather because they are dependencies of flake8?

Correct.

It is only for development/linting purposes. Actually nothing in this requirements.txt file is need to run the project because each sub project takes care of its own environment within the docker containers. I think in the past some of these were needed and are no longer. I've made an issue to clean that up: https://github.com/data-refinery/data_refinery/issues/17

workers/environment.yml
```diff
@@ -11,9 +11,11 @@ dependencies:
   - greenelab::all-entrezg-brainarray-cdfs
   - python=3.6
   - r::rpy2=2.8.5
+  - pycurl=7.43.0
```
kurtwheeler: You know what, this is a good question. I think I wanted to try it out for something but ended up not using it. I should remove this.

workers/environment.yml
```diff
@@ -11,9 +11,11 @@ dependencies:
...
   - pip:
     - celery==4.0.2
     - django==1.10.6
     - requests==2.13.0
     - psycopg2==2.7.1
+    - boto==2.46.1
```
kurtwheeler: Same reply as before.

terraform/database.tf
```diff
@@ -0,0 +1,130 @@
+provider "aws" {
+  region = "us-east-1"
+}
+
+variable "database_password" {}
```
kurtwheeler: Exactly!

terraform/database.tf
```diff
@@ -0,0 +1,130 @@
...
+  vpc_id = "${aws_vpc.data_refinery_vpc.id}"
+  map_public_ip_on_launch = true
+
+  tags {
+    Name = "data-refinery-1a"
+  }
+}
+
+resource "aws_subnet" "data_refinery_1b" {
+  availability_zone = "us-east-1b"
```
kurtwheeler: No. We don't actually have to worry about being in multiple availability zones for the workers because they don't have to always be available. In fact because we're planning on using spot instances they really won't be. Once there is a webserver then this will be a concern. (If this question was about why we have 2, that's to make sure that they can coordinate with each other to not repeat work etc.)

kurtwheeler:
@rzelayafavila I've answered your questions and removed the pycurl dependency.

foreman/requirements.txt
rzelayafavila: Interesting, it looks like support for boto3 will be available pretty soon though (https://github.com/celery/kombu/issues/592). Maybe in the next celery release!

PR #16, opened 2017-06-05
Process single sample

kurtwheeler:
This switches the processor method from a modified R script to [SCAN.UPC::scanFast](https://bioconductor.org/packages/3.5/bioc/html/SCAN.UPC.html). There will need to be a lot more R dependencies like `bioconductor-pd.hugene.1.0.st.v1` for this to work on most experiments however this works for the first test case and I am working on getting those dependencies into Conda. I'll add them in a subsequent PR. This has been tested by running a survey job for the experiment `E-GEOD-50790` and by running the tests.


workers/data_refinery_workers/processors/array_express.py
```diff
@@ -23,16 +23,23 @@ def cel_to_pcl(kwargs: Dict):
...
+    target_directory = os.path.join(utils.ROOT_URI, "processed", batch.internal_location)
     os.makedirs(target_directory, exist_ok=True)
-    new_name = batch.name + "." + batch.processed_format
 
-    ro.r('source("/home/user/r_processors/process_cel_to_pcl.R")')
-    ro.r['ProcessCelFiles'](
-        from_directory,
-        "Hs",  # temporary until organism handling is more defined
-        target_directory + new_name)
+    processed_file = os.path.join(target_directory, batch.name.split(".")[0])
```
mhuyck: If there is meaningful structure in the `batch.name` property, do we want to rely on the code here in the `array_express` processor to understand that structure (and stay in sync with it as modified), or would it be better to define a method on `batch` for this such as `batch.get_filename()`?

workers/data_refinery_workers/processors/array_express.py
```diff
@@ -23,16 +23,23 @@ def cel_to_pcl(kwargs: Dict):
...
+    processed_file = processed_file + "." + batch.processed_format
+
+    # It's necessary to load the foreach library before calling SCANfast
+    # because it doesn't load the library before calling functions
+    # from it.
+    ro.r("library('foreach')")
```
mhuyck: LOL. Well, I guess this is still a net win if `SCAN.UPC` is well maintained and we don't have to write our own `process_cel_to_pcl.R` now.

kurtwheeler: Hahaha yup :)

workers/data_refinery_workers/processors/array_express.py
kurtwheeler: I have a branch that is dealing with these issues: https://github.com/data-refinery/data_refinery/blob/run-in-ec2/common/data_refinery_common/file_management.py

PR #18, opened 2017-06-28
Use s3

kurtwheeler:
This PR utilizes S3 for file storage instead of local disk. I have created a new namespace `file_management` to componentize all file related operations. Now any projects which will need to access the files stored in S3 can utilize this namespace to do so. In addition to providing a standard way of determining where files should be located both locally (for temporary storage during use) and remotely (currently in S3), this should make switching to a different block storage provider easy in the future as only the `file_management` namespace will need to be changed.

Additionally this has the added benefit of making development vs production environment changes very easy via the `USE_S3` environment variable.


common/data_refinery_common/file_management.py
```diff
@@ -0,0 +1,178 @@
...
+def get_processed_path(batch: Batch) -> str:
+    return os.path.join(get_processed_dir(batch),
+                        _get_processed_name(batch))
+
+
+def upload_raw_file(batch: Batch, dir_name: str = None) -> None:
+    """Moves the batch's raw file out of the temp directory.
+
+    Depending on the value of the USE_S3 environment variable this may
+    just be to the RAW_PREFIX directory or it may be to S3.
```
dongbohu: Do you mean "just be uploaded to ..."?

common/data_refinery_common/file_management.py
```diff
@@ -0,0 +1,178 @@
...
+            bucket.put_object(Key=raw_path, Body=temp_file)
+    else:
+        shutil.copyfile(temp_path, raw_path)
+
+
+def download_raw_file(batch: Batch, dir_name: str = None) -> None:
+    """Moves the batch's raw file to the temp directory.
+
+    Depending on the value of the USE_S3 environment variable this may
+    just be from the RAW_PREFIX directory or it may be from S3.
```
dongbohu: I guess it means "just be downloaded from ..."?

common/data_refinery_common/file_management.py
```diff
@@ -0,0 +1,178 @@
...
+    temp_path = get_temp_pre_path(batch, dir_name)
+    if USE_S3:
+        bucket = boto3.resource("s3").Bucket(S3_BUCKET_NAME)
+        with open(temp_path, 'wb') as temp_file:
+            bucket.download_fileobj(raw_path, temp_file)
+    else:
+        shutil.copyfile(raw_path, temp_path)
+
+
+def upload_processed_file(batch: Batch, dir_name: str = None) -> None:
```
dongbohu: This function seems similar to `upload_raw_file()` function. Do you think it is worth the time to merge them into one?

common/data_refinery_common/settings.py
```diff
@@ -0,0 +1,141 @@
...
+
+USE_L10N = True
+
+USE_TZ = True
+
+
+# Static files (CSS, JavaScript, Images)
+# https://docs.djangoproject.com/en/1.10/howto/static-files/
+
+STATIC_URL = '/static/'
```
dongbohu: 👍

common/manage.py
```diff
@@ -0,0 +1,23 @@
...
+
+if __name__ == "__main__":
+    os.environ.setdefault("DJANGO_SETTINGS_MODULE",
+                          "data_refinery_common.settings")
+    try:
+        from django.core.management import execute_from_command_line
+    except ImportError:
+        # The above import may fail for some other reason. Ensure that the
+        # issue is really that Django is missing to avoid masking other
+        # exceptions on Python 2.
```
dongbohu: 👍

common/setup.py
```diff
@@ -0,0 +1,30 @@
...
+    version="0.1.0",
+    packages=find_packages(),
+    install_requires=["django>=1.10.6",
+                      "data_refinery_models>=0.1.0",
+                      "boto3>=1.4.4"],
+    license="BSD License",
+    description="Common functionality to be shared between Data Refinery sub-projects.",
+    url="https://www.greenelab.com",
+    author="Kurt Wheeler",
+    author_email="kurt.wheeler91@gmail.com",
```
dongbohu: You may want to use the lab's email address instead. Ask Rene about it. I don't have it at hand now.

foreman/data_refinery_foreman/surveyor/test_array_express.py
```diff
@@ -264,25 +267,27 @@ def test_experiment_object(self, mock_get):
     @patch('data_refinery_foreman.surveyor.message_queue.app.send_task')
     def test_survey(self, mock_send_task, mock_get):
         """survey generates one Batch per sample with all possible fields populated.
```
dongbohu: Minor: Capitalize 's' in "survey"?

workers/data_refinery_workers/downloaders/array_express.py
```diff
@@ -32,7 +34,7 @@ def _verify_batch_grouping(batches: List[Batch], job_id):
             raise ValueError("A batch doesn't have the same download url as other batches.")
 
 
-def _download_file(download_url, file_path, job_id):
+def _download_file(download_url: str, file_path: str, job_id: int) -> None:
```
dongbohu: 👍

common/data_refinery_common/file_management.py
kurtwheeler: This sentence is supposed to be modifying the one line description, but I can make it more clear by changing this to `Depending on the value of the USE_S3 environment variable this will either move the batch's raw file to the RAW_PREFIX directory or to S3.` if that is more clear. I don't want to use the word `upload` because if `USE_S3` is false then there won't be any uploading going on.

common/data_refinery_common/file_management.py
kurtwheeler: I agree these two functions are similar, but honestly I like that every operation this namespace performs has its own function. I think it'll make it more clear to users of these functions which function they want and what they should pass it. If I tried to merge them then I'd need an additional parameter to differentiate whether to upload a raw or a processed file, but only with those two options. So that parameter could either be a string which I'd have to validate is only `"raw"` or `"processed"` or I could make it a boolean. I there's a bit of code duplication here, which I think I could solve by making them both call a helper function `upload_file` which would just take a from path and a to path. I think that actually makes sense, but I do think I like the readability/usability of: `upload_processed_file(batch)` over `upload_file(batch, raw=False)`.

foreman/data_refinery_foreman/surveyor/test_array_express.py
kurtwheeler: `survey` is the name of a function.

common/data_refinery_common/file_management.py
rzelayafavila: Nice, I do think it was a good idea to abstract the common functionality into the `_upload_file` function.

common/Dockerfile.tests
```diff
@@ -0,0 +1,19 @@
...
+WORKDIR /home/user
+
+COPY common/requirements.txt .
+
+RUN pip install -r requirements.txt
+
+COPY data_models/dist/data-refinery-models-* data_models/
+
+# Get the latest version from the dist directory.
+RUN pip install data_models/$(ls data_models -1 | tail -1)
```
rzelayafavila: Hmmm are you sure this is always going to return the latest version? I think it might not, since (for example) `ls` lists version `1.10` first, before `1.9`, so `1.9` would be your last version listed on the dist directory.

common/Dockerfile.tests
```diff
@@ -0,0 +1,19 @@
+FROM python:3.6.1-slim
+
+RUN groupadd user && useradd --create-home --home-dir /home/user -g user user
+WORKDIR /home/user
+
+COPY common/requirements.txt .
+
+RUN pip install -r requirements.txt
```
rzelayafavila: So I forget, did we decide not to use python virtualenvironments in these Docker containers at all?

common/data_refinery_common/file_management.py
```diff
@@ -0,0 +1,184 @@
...
+def get_raw_path(batch: Batch) -> str:
+    return os.path.join(get_raw_dir(batch), batch.name)
+
+
+# The following four functions use the ID of the batch in the
+# temporary paths so it can be removed after processing is complete
+# without interfering with other jobs. An optional dir_name can be
+# used instead.
+def get_temp_dir(batch: Batch, dir_name: str = None) -> str:
+    dir_name = dir_name if dir_name is not None else str(batch.id)
```
rzelayafavila: :+1:

workers/Dockerfile
```diff
@@ -16,6 +16,11 @@ COPY data_models/dist/data-refinery-models-* data_models/
 # Get the latest version from the dist directory.
 RUN pip install data_models/$(ls data_models -1 | tail -1)
 
+COPY common/dist/data-refinery-common-* common/
+
+# Get the latest version from the dist directory.
+RUN pip install common/$(ls common -1 | tail -1)
```
rzelayafavila: Same comment here and in `workers/Dockerfile.tests` as above (about getting the latest version).

common/requirements.txt
```diff
@@ -0,0 +1,21 @@
+#
+# This file is autogenerated by pip-compile
+# To update, run:
+#
+#    pip-compile --output-file requirements.txt requirements.in
+#
+boto3==1.4.4
```
rzelayafavila: So re: our conversation in the previous PR, we are using boto3 instead of boto 2 because this does not deal with Celery workers?

foreman/data_refinery_foreman/surveyor/test_array_express.py
rzelayafavila: In that case I would put it in quotes, or specifically say something like "The survey function" or "The survey() function", since it is kind of unclear that it is talking about the survey function there. Or maybe something like `ArrayExpressSurveryor.survey() generates...`?

kurtwheeler: Good call.

common/Dockerfile.tests
kurtwheeler: Correct. Docker containers already isolate their environments so there's not really any point.

common/Dockerfile.tests
kurtwheeler: You're definitely right about this. I fixed this once but I forgot to port it everywhere: https://github.com/data-refinery/data_refinery/blob/master/foreman/Dockerfile#L14

common/requirements.txt
kurtwheeler: Celery needs boto2 but any code I write that uses boto will use version 3. That way once Celery upgrades to boto3 we're already on it and can just remove the dependence on boto2 without changing anything except a celery version.

kurtwheeler:
@rzelayafavila @dongbohu is this good to go? I believe I have addressed both of your concerns.

PR #21, opened 2017-07-05
Circle ci

kurtwheeler:
This PR integrates CircleCI with the data-refinery and uses it to run tests automagically. To support that there's a few other things this PR changes. It introduces a way to do migrations via a Docker container so that the environment on the host machine doesn't need any development packages. It also removes the foreman's dependency on the workers since that was mostly taken out a while ago. It also uses a database timeout in a couple more places so that tests fail fast if there's a database issue.


kurtwheeler:
> Seems like there is inherently a good functional test in place for this PR - does everything work with CircleCI?

Haha yeah right?

PR #22, opened 2017-07-05
Run in ec2

kurtwheeler:
This PR allows the workers to be run in EC2 instances. It's mostly terraform code with only some minor changes to the worker project to record worker ids now that there's actually workers.


kurtwheeler:
@rzelayafavila We should get you set up with git-crypt so you can decrypt the task definition.

terraform/container-registry.tf
```diff
@@ -0,0 +1,64 @@
+resource "aws_ecs_cluster" "data_refinery" {
+  name = "data-refinery"
+}
+
+resource "aws_iam_role" "ecs_service_role" {
+  name = "data-refinery-ecs-service"
+
+  assume_role_policy = <<EOF
+{
```
rzelayafavila: It would be good to have some comments before this describing what this iam policy is doing exactly.

terraform/container-registry.tf
```diff
@@ -0,0 +1,64 @@
...
+  "Version": "2012-10-17",
```
rzelayafavila: Where is this "Version" date coming from? Why 2012-10-17? It might be good to put that in the docs as well.

terraform/container-registry.tf
```diff
@@ -0,0 +1,64 @@
...
+EOF
+}
+
+resource "aws_iam_role_policy" "ecs_service" {
+  name = "data-refinery-ecs-service-policy"
+  role = "${aws_iam_role.ecs_service_role.name}"
+
+  policy = <<EOF
+{
+    "Version": "2012-10-17",
```
rzelayafavila: Same comments here as in the iam policy above.

terraform/database.tf
```diff
@@ -1,75 +1,5 @@
-provider "aws" {
```
rzelayafavila: Good call sending all of this over to a main.tf file

terraform/main.tf
```diff
@@ -0,0 +1,234 @@
...
+  name  = "data-refinery-ecs-instance-profile"
+  roles = ["${aws_iam_role.ecs_instance.name}"]
+}
+
+resource "aws_iam_role" "ecs_instance" {
+  name = "data-refinery-ecs-instance"
+
+  assume_role_policy = <<EOF
+{
+  "Version": "2008-10-17",
```
rzelayafavila: Same questions for the iam policies here. Why does this one have a 2008 date?

terraform/main.tf
```diff
@@ -0,0 +1,234 @@
...
+  }
+}
+
+# Note: this is a insecure practice long term, however it's
+# necessary to access it from lab machines.
+resource "aws_route_table" "data_refinery" {
+  vpc_id = "${aws_vpc.data_refinery_vpc.id}"
+
+  route {
+    cidr_block = "0.0.0.0/0"
```
rzelayafavila: Is doing a cidr block of `0.0.0.0` here customary?

terraform/main.tf
```diff
@@ -0,0 +1,234 @@
...
+  name = "data-refinery-worker"
+  description = "data-refinery-worker"
+  vpc_id = "${aws_vpc.data_refinery_vpc.id}"
+
+  tags {
+    Name = "data-refinery-worker"
+  }
+}
+
+# resource "aws_security_group_rule" "cognoma-service-http" {
```
rzelayafavila: Hmm, do we wanna remove this altogether? It looks like some code borrowed from cognoma

common/Dockerfile.tests
```diff
@@ -12,6 +12,7 @@ COPY data_models/dist/data-refinery-models-* data_models/
 # Get the latest version from the dist directory.
 RUN pip install data_models/$(ls data_models -1 | sort --version-sort | tail -1)
 
+
```
kurtwheeler: There isn't one. Musta slipped through :(

terraform/main.tf
kurtwheeler: Yes. It's the way to denote all IPv4 traffic. (Since the /0 part means that none of the IP address is significant.)

terraform/main.tf
kurtwheeler: Yes.

terraform/container-registry.tf
kurtwheeler: The date is coming from AWS' documentation where it says "Copy this text". I bet it's not truly necessary, but the documentation has it and it works so I went with it. I'm adding links to the sources of the policies in comments.

kurtwheeler:
Ok @rzelayafavila I've tried to address your questions and concerns. Take a look and lemme know if anything is unclear!

PR #23, opened 2017-07-06
Fix worker ids

kurtwheeler:
This PR handles an issue where process ids were unique to the instance they were on, but not across the entire cluster.


dcgoss:
What are the different environment files for? Is there a reason they aren't python files?

kurtwheeler:
The different environment files allow me to easily switch the environment between dev, test, and prod. There is a reason they aren't python files, in that they allow me to use the `--env-file` option for the `docker run` command. For example I use the dev file [here](https://github.com/data-refinery/data_refinery/blob/master/foreman/run_surveyor.sh#L28) and the test file [here ](https://github.com/data-refinery/data_refinery/blob/master/foreman/run_tests.sh#L20).

PR #24, opened 2017-07-06
Use cloudwatch

kurtwheeler:
This uses `awslogs` to send logs from AWS instances into cloudwatch. I tested this by running the data_refinery on ec2 and looking at the logs going into cloudwatch.


dcgoss:
What is the rationale for having both a Dockerfile and Dockerfile.tests?

kurtwheeler:
Having a different `CMD` for each. However given the presence of a `run_tests.sh` in each project and the fact that workers don't need an `ENTRYPOINT` I think I could easily combine these. I'll do that now.

kurtwheeler:
@dcgoss I've removed a couple Dockerfile.tests!

PR #26, opened 2017-07-06
Foreman

kurtwheeler:
This PR adds the foreman, who's job it is to make sure that jobs that fail, hang, or get lost get retried a few times and then alerting if they still don't succeed.

Within this PR I have removed DownloaderJobsToBatches and ProcessorJobsToBatches and am instead using Django's built in many-to-many relationship. This cause changes in several additional places.


foreman/data_refinery_foreman/foreman/main.py
```diff
@@ -0,0 +1,260 @@
...
+
+@do_forever(MIN_LOOP_TIME)
+def retry_lost_downloader_jobs() -> None:
+    """Retry downloader jobs that went too long without being started.
+
+    Idea: at some point this function could integrate with the spot
+    instances to determine if jobs are hanging due to a lack of
+    instances. A naive time-based implementation like this could end
+    up retrying every single queued job if there were a long period of
+    spot instance bid price being very high.
```
dongbohu: I don't understand this comment line. Can you elaborate?

foreman/data_refinery_foreman/foreman/main.py
```diff
@@ -0,0 +1,260 @@
...
+
+    The new downloader job will have num_retries one greater than
+    last_job.num_retries.
+    """
+    num_retries = last_job.num_retries + 1
+
+    new_job = ProcessorJob.create_job_and_relationships(num_retries=num_retries,
+                                                        batches=list(last_job.batches.all()),
+                                                        pipeline_applied=last_job.pipeline_applied)
+    processor_task = PROCESSOR_PIPELINE_LOOKUP[last_job.pipeline_applied]
```
dongbohu: Do you need to take care of the possible error that `last_job.pipeline_applied` is not found in `PROCESSOR_PIPELINE_LOOKUP`, or this error will never happen?

foreman/data_refinery_foreman/foreman/main.py
```diff
@@ -0,0 +1,260 @@
...
+    app.send_task(processor_task, args=[new_job.id])
+
+    last_job.retried = True
+    last_job.success = False
+    last_job.save()
+
+
+def handle_processor_jobs(jobs: List[ProcessorJob]) -> None:
```
dongbohu: I wonder whether it is possible to merge these `downloader` and `processor` jobs methods into one set. Some of them seem very similar.

foreman/data_refinery_foreman/foreman/main.py
```diff
@@ -0,0 +1,260 @@
...
+        created_at__lt=minimum_creation_time
+    )
+
+    handle_processor_jobs(lost_jobs)
+
+
+def monitor_jobs():
+    """Runs a thread for each job monitoring loop."""
+    threads = []
+    thread = Thread(target=retry_failed_downloader_jobs,
```
dongbohu: Since the second argument of `Thread()` is always the function's name, do you think it will be better to create a list of functions first, then use a loop to iterate the loop and get each function's name by `f.func_name`? Something like this:
```
functions = [f1, f2, ...]
threads = []
for f in functions: 
    thread = Thread(target=f, name=f.func_name)
    thread.start()
    threads.append(thread)
```

foreman/data_refinery_foreman/foreman/test_main.py
```diff
@@ -0,0 +1,274 @@
...
+            status=BatchStatuses.NEW.value
+        )
+        batch.save()
+        return batch
+
+    @patch('data_refinery_foreman.surveyor.message_queue.app.send_task')
+    def test_requeuing_downloader_job(self, mock_send_task):
+        batch = self.insert_batch()
+        job = DownloaderJob.create_job_and_relationships(
+            num_retries=0, batches=[batch], downloader_task="dummy_task")
```
dongbohu: Lines 52-54 are needed by each test case of downloader job. I wonder whether it is a good idea to put them in a separate function.

foreman/data_refinery_foreman/foreman/test_main.py
```diff
@@ -0,0 +1,274 @@
...
+        self.assertFalse(original_job.success)
+
+        retried_job = jobs[1]
+        self.assertEqual(retried_job.num_retries, 1)
+
+    @patch('data_refinery_foreman.surveyor.message_queue.app.send_task')
+    def test_requeuing_processor_job(self, mock_send_task):
+        batch = self.insert_batch()
+        job = ProcessorJob.create_job_and_relationships(
+            num_retries=0, batches=[batch], pipeline_applied="AFFY_TO_PCL")
```
dongbohu: Lines 165-167 are available in each test case of processor job. I wonder whether it is a good idea to put them in a separate function too.

foreman/data_refinery_foreman/foreman/test_main.py
```diff
@@ -0,0 +1,274 @@
...
+        survey_job = SurveyJob(source_type="ARRAY_EXPRESS")
+        survey_job.save()
+        self.survey_job = survey_job
+
+    def tearDown(self):
+        Batch.objects.all().delete()
+        DownloaderJob.objects.all().delete()
+        ProcessorJob.objects.all().delete()
+
+    def insert_batch(self):
```
dongbohu: Optional: This function doesn't really need `self`, so it can be a static function.

foreman/data_refinery_foreman/surveyor/surveyor.py
```diff
@@ -13,7 +13,7 @@ class SourceNotSupportedError(BaseException):
     pass
 
 
-def _get_surveyor_for_source(survey_job: SurveyJob):
+def get_surveyor_for_source(survey_job: SurveyJob):
```
dongbohu: I kind of like the convention of having a prefix of `_` to indicate that a function is private. Any reason why you removed it?

workers/data_refinery_workers/downloaders/array_express.py
```diff
@@ -76,11 +73,9 @@ def _extract_file(batches: List[Batch], job_id: int) -> None:
 @shared_task
 def download_array_express(job_id: int) -> None:
     job = utils.start_job(job_id)
+    batches = list(job.batches.all())
```
dongbohu: You don't have to convert the QuerySet `job.batches.all()` into a list to get its size and the first element. You can use `count()` and `first` methods instead.

workers/data_refinery_workers/processors/utils.py
```diff
@@ -21,8 +21,7 @@ def start_job(kwargs: Dict):
     job.start_time = timezone.now()
     job.save()
 
-    batch_relations = ProcessorJobsToBatches.objects.filter(processor_job_id=job.id)
-    batches = [br.batch for br in batch_relations]
+    batches = list(job.batches.all())
```
dongbohu: Again, this conversion is not really necessary if you only need to get the number of elements in the QuerySet.

data_models/data_refinery_models/models/jobs.py
```diff
@@ -1,73 +1,64 @@
...
-        db_table = "processor_jobs"
+    # This point of this field is to identify which worker ran the
+    # job. A few fields may actually be required or something other
+    # than just an id.
+    worker_id = models.CharField(max_length=256, null=True)
 
+    @classmethod
+    @transaction.atomic
+    def create_job_and_relationships(cls, *args, **kwargs):
+        """Inits and saves a job and its relationships to its batches."""
```
kurtwheeler: I agree with you, but doing so pushes the docstring line over 72 characters which breaks PEP8. I couldn't figure out how to make this line any shorter while still describing the function, so I shortened that word. Any ideas of better ways to deal with that?

dongbohu: I see. I have to confess that I often forget the 72-char-limit of docstrings and comments. In that case  I would prefer multi-line docstrings than ambiguous comments. What do you think?

foreman/data_refinery_foreman/foreman/main.py
```diff
@@ -0,0 +1,260 @@
...
+MAX_NUM_RETRIES = 2
+
+# For now it seems like no jobs should take longer than a day to be
+# either picked up or run.
+MAX_DOWNLOADER_RUN_TIME = timedelta(days=1)
+MAX_PROCESSOR_RUN_TIME = timedelta(days=1)
+MAX_QUEUE_TIME = timedelta(days=1)
+
+# To prevent excessive spinning loop no more than once every 10
+# seconds.
```
kurtwheeler: Because if it's all on one line then the line length is exactly 72 characters long, where PEP8 specifies that all comments and docstrings should be shorter than 72...  It's not my favorite style rule but I've gotten in the habit of just hitting `M-q` and letting emacs decide how to space it out (since it follows the guide exactly).

dongbohu: I see. Never mind then.

data_models/data_refinery_models/models/jobs.py
kurtwheeler: I agree with you about that. However the convention for multi-line docstrings is that you still start with a one line summary, followed by a newline, followed by more explanation. I don't know what I would change the one line summary to (I'd still have to abbreviate Initialize to inits to make it fit). Do you think the docstring could use more detail?

data_models/data_refinery_models/models/test_jobs.py
```diff
@@ -0,0 +1,65 @@
...
+        status=BatchStatuses.NEW.value
+    )
+    batch.save()
+    return batch
+
+
+class DownloaderJobTestCase(TestCase):
+    def test_create_job_and_relationships(self):
+        """DownloaderJob, Batches, and relationships are created."""
+        batches = [get_batch(), get_batch()]
```
kurtwheeler: I think I might disagree. The point of having two batches here is to make sure the code handles more than one well, it doesn't really do anything with any of the values so if any of them were different I don't see what it would change. If you're worried about the batches being compared at some point and potentially that comparison returning that they're equal, I don' think that's an issue because `get_batch` saves the batches which will give them an id which won't be the same between the two.

data_models/data_refinery_models/models/jobs.py
dongbohu: Hmm... I noticed that this class method needs some arguments and inside the body it plays with the arguments. Since this method has been called a lot of times by other modules, it will be great to add some detailson how it should be called.

foreman/data_refinery_foreman/foreman/main.py
kurtwheeler: The only way this error could happen is if a processor was added to the workers project without also being added here, which would be a mistake. If that were to happen I would want this to fail loudly and spectacularly so it immediately became obvious. I don't think this is an error that would come up in a normal scenario.

data_models/data_refinery_models/models/test_jobs.py
dongbohu: I forgot that the batch IDs are still different. That is a good point.

foreman/data_refinery_foreman/foreman/main.py
kurtwheeler: I'll do it here since that comment is mostly a note to myself anyway. (Unless of course you'd like me to add it in the comment.)

Basically we're planning on having the Data Refinery use AWS spot instances (or the google/microsoft equivalent). The way those work is that you set a bid price, and when the price of the spot instances drops below your bid price you get however many machines you specified you wanted until the price rises above your bid price.

This comment is referring to a scenario where the spot instance price is higher than our bid price for long enough that a job got queued by the surveyor (which runs on a persistent instance) and then sat in the SQS queue long enough that the foreman thought it somehow got lost forever (since there were no worker instances to pick it up). There's probably a number of ways this scenario could be handled, this comment is mostly a reminder that once I actually use spot instances that there could be an issue with jobs getting "lost" when actually they just haven't had a chance to get picked up yet.

foreman/data_refinery_foreman/foreman/main.py
kurtwheeler: I had the same thought several times. I've also thought about it in a more general sense too, as in do we really need to make any distinction between processor jobs and downloader jobs? However there are some small differences which would require many functions dealing with jobs to have handle them differently based on what they are. I think for the time being the overlap between the two is better than having an `if` statement in each function saying "is this job a processor or a downloader? ok, so then I set `pipeline_applied` instead of that `downloader_task`" or something.

If I really wanted to I'm sure I could get everything working in such a way that downloader jobs are just an instance of processor pipelines, but I think there might be some finagling that'd have to be done to make that work and I'm worried it'd be a lot harder to split the two in the future if there ever was a reason they did have to be two separate entities than it would be to merge them.

dongbohu: I took a closer look at the code and saw the small differences between these two jobs. I agree that it is a little tricky to have a uniform API for the two jobs. It will be even worse if the code includes `if` statements everywhere.

Do you think you may have other different job types in the future? What I am worrying about now that if you add another job type later, you may have to repeat these requeue/retry functions again, and that will be ugly. If you only have `downloader` and `processor` jobs, then the code duplication is not so bad. 

Since  am leaving on Friday, I don't think this dilemma can be solved neatly in this PR. Can you add an issue to remind yourself?

foreman/data_refinery_foreman/foreman/main.py
dongbohu: Thank you for the explanation. What I was confused about was the term of `spot instance bid price`. If I understand correctly, the last sentence actually means:
```
... if there were a long period during which the price of spot instance is higher than the bid price.
```
right?

data_models/data_refinery_models/models/jobs.py
kurtwheeler: Yeah that's a good call, I'll describe that in the docstring.

foreman/data_refinery_foreman/foreman/main.py
kurtwheeler: Yes you're right, that's far more clear. I'll change it to that.

foreman/data_refinery_foreman/foreman/main.py
kurtwheeler: As far as I can tell at the moment I don't think I'll need another job type later. However it can be hard to make such predictions accurately. I think the best approach might be to leave it as it is now, and then if I get to a point where I need a new kind of job I'll come up with a more elegant way of dealing with them all at that point.

I don't think I want to open an issue about it though because unless a new job type becomes necessary it won't ever get resolved, and the necessity of the new job type should remind me that I need to rework this.

foreman/data_refinery_foreman/foreman/main.py
kurtwheeler: This is beautiful :heart_eyes:

foreman/data_refinery_foreman/foreman/test_main.py
kurtwheeler: Actually, it accesses the `self.survey_job` which is created in the setUp function. I can see how you missed it though.

dongbohu: Never mind. Thanks.

kurtwheeler:
@dongbohu I believe my last commits addresses the remainder of your concerns. Let me know if this isn't the case.

PR #27, opened 2017-07-14
Use custom brainarray cdfs

kurtwheeler:
This PR rebuilds the workers environment from scratch, completely cutting conda out of the equation. This is because conda wasn't building the environment properly. Sometimes it would say that a package wasn't found, once I resolved such an issue by moving that package's location within `environment.yml`. Additionally I had some issues with disk space while building the docker image with conda and this way seems to be no worse and at least a bit better.

This also properly utilizes `SCAN.UPC` because before I was not specifying the `probeSummaryPackage` parameter. Yes, I know that it looks like this is introducing `SCAN.UPC` for the first time but actually I resolved a merge conflict badly resulting in the call to `cel_to_pcl.R` script to accidentally reenter the code.


kurtwheeler:
@mhuyck What does your work-load look like early next week? This is one of 5 PRs I have opened simultaneously today and I'm hoping to get them all merged in ASAP so I can try to start processing a large set of data for @jaclyn-taroni in the cloud early next week. (I foresee a number of issues coming up while running in the cloud for real and we're trying to process or at least download all those files by August 1st for a grant proposal.)

Anyway, if you don't think you'll be able to get to this on Monday perhaps I can reassign it to someone who can.

mhuyck:
Looks like a lot going on in here, but I'm happy to dive in. It appears that CircleCI is not happy with this... do you want to check that out first and see if you can fix it?

workers/r_dependencies.R
```diff
@@ -0,0 +1,205 @@
+install.packages('ff', repos='https://cloud.r-project.org/')
+install.packages('XML', repos='https://cloud.r-project.org/')
+install.packages('RCurl', repos='https://cloud.r-project.org/')
+install.packages('RSQLite', repos='https://cloud.r-project.org/')
+
+source('https://bioconductor.org/biocLite.R')
+biocLite('oligo')
```
jaclyn-taroni: Is there a reason for the decision to not include the result of these packages in one vector (perhaps called `bioc.packages`) and then lapply over the whole thing?
```R
lapply(bioc.packages, biocLite)
```

workers/data_refinery_workers/processors/array_express.py
```diff
@@ -31,14 +37,32 @@ def cel_to_pcl(kwargs: Dict):
...
+
+    # It's necessary to load the foreach library before calling SCANfast
+    # because it doesn't load the library before calling functions
+    # from it.
+    ro.r("library('foreach')")
+
+    ro.r['::']('SCAN.UPC', 'SCANfast')(
+        input_file,
+        output_file,
+        probeSummaryPackage=brainarray_package
```
jaclyn-taroni: This implementation of `SCAN` LGTM

workers/data_refinery_workers/processors/array_express.py
```diff
@@ -31,14 +37,32 @@ def cel_to_pcl(kwargs: Dict):
...
+    header = ro.r['::']('affyio', 'read.celfile.header')(input_file)
+
+    # header is a list of vectors. [0][0] contains the package name.
+    table = str.maketrans(dict.fromkeys(string.punctuation))
+    package_name = header[0][0].translate(table).lower()
+    brainarray_package = package_name + "hsentrezgprobe"
+
+    # Some CEL headers have a v1 in the package name which is not part
+    # of the brainararry package name. We have a table which corrects
+    # for this.
```
jaclyn-taroni: We might need a better long term solution here -- it could work to strip anything after the `v` for version in most cases. Will think about it more.

workers/r_dependencies.R
```diff
@@ -0,0 +1,205 @@
...
+
+
+packages <- row.names(available.packages(contriburl="http://brainarray.mbni.med.umich.edu/bioc/src/contrib/"))
+
+## This narrows down from all the files gotten just having entrezg in the name
+entrez_packages <- packages[grep("entrezg", packages)]
+
+install.packages(entrez_packages, repos="http://brainarray.mbni.med.umich.edu/bioc/", type="source")
+
+platform_designs <- c(
```
jaclyn-taroni: Same comment as above

workers/r_dependencies.R
```diff
@@ -0,0 +1,205 @@
...
+biocLite('org.Gg.eg.db')
+biocLite('org.Rn.eg.db')
+biocLite('org.Ss.eg.db')
+biocLite('org.Dr.eg.db')
+
+
+packages <- row.names(available.packages(contriburl="http://brainarray.mbni.med.umich.edu/bioc/src/contrib/"))
+
+## This narrows down from all the files gotten just having entrezg in the name
+entrez_packages <- packages[grep("entrezg", packages)]
```
jaclyn-taroni: If we're sticking with [Google's R Style Guide convention for identifiers](https://google.github.io/styleguide/Rguide.xml#identifiers), this would be `entrez.packages`

workers/data_refinery_workers/processors/array_express.py
kurtwheeler: I agree this might not be the best long term solution. I'm hoping that as we add more to this dictionary we'll notice some patterns.

kurtwheeler:
@mhuyck CircleCI seems to be having trouble building the docker image with all of the dependencies, which is kind of a duh moment for me because the thing takes over an hour to build on my beefy lab machine. I'm going to change things such that all the dependencies are built into a base image which the `workers` image will build off of. Then I can push the base image to a public image repository and CircleCI can pull it from there rather than having to build it itself.

That being said it won't affect most of this PR, basically the docker image will just get split into two.

workers/data_refinery_workers/processors/array_express.py
```diff
@@ -31,14 +37,32 @@ def cel_to_pcl(kwargs: Dict):
...
-    ro.r('source("/home/user/r_processors/cel_to_pcl.R")')
-    ro.r['ProcessCelFiles'](
-        temp_dir,
-        "Hs",  # temporary until organism handling is more defined
-        output_file)
+    header = ro.r['::']('affyio', 'read.celfile.header')(input_file)
+
+    # header is a list of vectors. [0][0] contains the package name.
+    table = str.maketrans(dict.fromkeys(string.punctuation))
+    package_name = header[0][0].translate(table).lower()
```
mhuyck: This is a nice, compact way to remove punctuation, but labeling it with something as generic as `table` isn't particularly helpful. Maybe something more descriptive, such as `remove_punctuation`?

workers/data_refinery_workers/processors/array_express.py
```diff
@@ -31,14 +37,32 @@ def cel_to_pcl(kwargs: Dict):
...
+    brainarray_package = package_name + "hsentrezgprobe"
+
+    # Some CEL headers have a v1 in the package name which is not part
+    # of the brainararry package name. We have a table which corrects
+    # for this.
+    if brainarray_package in PACKAGE_NAME_CORRECTIONS:
+        brainarray_package = PACKAGE_NAME_CORRECTIONS[brainarray_package]
```
mhuyck: I agree with @jaclyn-taroni's comment above, but also want to say that this is a nice simple workaround for that issue.

workers/data_refinery_workers/processors/array_express.py
mhuyck: I'm glad someone else knows this part... I have not had occasion to use `SCAN` before.

workers/Dockerfile.base
```diff
@@ -0,0 +1,53 @@
...
+
+RUN \
+  apt-get update -qq && \
+  apt-get install -y \
+                     lsb-release && \
+  echo "deb http://archive.ubuntu.com/ubuntu $(lsb_release -sc) multiverse\n" \
+      >> /etc/apt/sources.list.d/added_repos.list && \
+  echo "deb http://cran.cnr.berkeley.edu/bin/linux/ubuntu $(lsb_release -sc)/" \
+      >> /etc/apt/sources.list.d/added_repos.list && \
+  apt-key adv --keyserver keyserver.ubuntu.com --recv-keys E084DAB9 && \
```
kurtwheeler: Seems to be the recommended way to install R: see the `Secure APT` section of: https://cran.r-project.org/bin/linux/ubuntu/

workers/Dockerfile.base
```diff
@@ -0,0 +1,53 @@
...
+                     libcurl4-gnutls-dev \
+                     libssl-dev \
+                     curl \
+		     wget &&\
+  rm -rf /var/lib/apt/lists/*
+
+RUN groupadd user && useradd --create-home --home-dir /home/user -g user user
+WORKDIR /home/user
+
+ENV R_LIBS "/usr/local/lib/R/site-library"
```
kurtwheeler: It does not. I think this setting is to specify a usr-specific R library.

workers/Dockerfile.base
```diff
@@ -0,0 +1,53 @@
...
+RUN Rscript r_dependencies.R
+
+COPY workers/requirements.txt .
+
+RUN \
+  pip3 --no-cache-dir install pip --upgrade && \
+  pip3 --no-cache-dir install setuptools --upgrade && \
+  pip3 --no-cache-dir install wheel --upgrade && \
+  pip3 --no-cache-dir install -r requirements.txt --upgrade && \
+  rm -rf /root/.cache
```
kurtwheeler: You know, I'm not actually sure. I think I got most of this code from somewhere. `--no-cache-dir` just says it `Disable the cache.` I guess I'd assume that would mean that it both wouldn't read from nor write to the cache, however I think I'd rather keep this line in here just to be sure as it will reduce image size if it does in fact do anything, and if it doesn't do anything it probably only adds a few milliseconds to build time.

workers/Dockerfile.base
```diff
@@ -0,0 +1,53 @@
...
+
+# Run dev version of rpy2
```
kurtwheeler: You know, I kinda wondered about that myself. I took this code from https://hub.docker.com/r/rpy2/rpy2/~/dockerfile/ and they installed the dev version. I'm gonna try pinning it to a version and see if that works.

workers/r_dependencies.R
```diff
@@ -0,0 +1,209 @@
...
+  "pd.vitis.vinifera",
+  "pd.wheat",
+  "pd.xenopus.laevis",
+  "pd.x.laevis.2",
+  "pd.x.tropicalis",
+  "pd.yeast.2",
+  "pd.yg.s98",
+  "pd.zebgene.1.0.st",
+  "pd.zebgene.1.1.st",
+  "pd.zebrafish"
```
kurtwheeler: I agree it is long and it is closer to data than code. However as far as I'm concerned this R file really is a configuration file that only runs during the build of the image. It just happens to be written in R. I'm not sure I see any benefit to moving this list of packages out of this file into another one, only to read it in as data.

workers/Dockerfile
```diff
@@ -1,25 +1,14 @@
...
-RUN apt-get install libcurl4-openssl-dev
-
-COPY workers/environment.yml .
-
-RUN conda env create -f environment.yml
-
-ENV PATH "/opt/conda/envs/workers/bin:$PATH"
+FROM ccdl/data_refinery_worker_base
 
 COPY data_models/dist/data-refinery-models-* data_models/
```
kurtwheeler: The developer runs `python setup.py sdist` themselves when they've made changes to the data models. I included this in the workers' README.md: https://github.com/data-refinery/data_refinery/blob/master/data_models/README.md#developing

workers/environment.yml
```diff
@@ -1,25 +0,0 @@
...
-  - python=3.6
-  - r::rpy2=2.8.5
-  - pycurl=7.43.0
-  - pip:
-    - celery==4.0.2
-    - django==1.11.1
-    - requests==2.13.0
-    - psycopg2==2.7.1
-    - boto3==1.4.4
-    - retrying==1.3.3
```
kurtwheeler: Exactly.

workers/environments/dev
```diff
@@ -10,7 +10,7 @@ DATABASE_TIMEOUT=5
 
 CELERY_BROKER_URL=amqp://guest@rabbit
 
-USE_S3=True
+USE_S3=False
```
kurtwheeler: Yup yup.

workers/data_refinery_workers/downloaders/test_array_express.py
```diff
@@ -90,15 +90,16 @@ def test_download(self,
         # Call the task we're testing:
         array_express.download_array_express.apply(args=(downloader_job.id,)).get()
 
-        target_file_path = ("/home/user/data_store/temp/A-AFFY-1/AFFY_TO_PCL/{}"
+        target_file_path = ("/home/user/data_store/temp/A-AFFY-1/AFFY_TO_PCL/downloader_job_{}"
                             "/E-GEOD-59071.raw.3.zip").format(str(downloader_job.id))
 
         # Verify that all expected functionality is run:
-        _verify_batch_grouping.assert_called_once()
+        # _verify_batch_grouping.assert_called_once()
```
kurtwheeler: I think I meant to delete it but missed it because of the legit comment above it. Good catch.

workers/data_refinery_workers/processors/array_express.py
kurtwheeler: I'm gonna go with `punctuation_table` because `remove_punctuation` reads like a function name to me. Good point though.

workers/Dockerfile
mhuyck: Gotcha. 👍  That's why I missed it... I was looking for a script.

workers/Dockerfile.base
kurtwheeler: I tried installing version 2.8.6 of `ryp2` via pip but it didn't work. Going back to this version made things work again. I'd like to get this out so I can start processing data for Jackie. I've made an issue about this though: https://github.com/data-refinery/data_refinery/issues/28

workers/Dockerfile.base
mhuyck: Ok. Maybe a comment here (with a link to that R documentation) would be helpful. I guess maybe the comment would have to go above the "one-line" `RUN` statement.

`<rant>`Why doesn't Ubuntu offer a cleaner way to update R? Should we really be training users to install mysterious keys so they can install common software packages?`</rant>`

`<rant>`Okay, I wasn't actually done. I looked at the R documentation link you included. It says:

> Ubuntu archives on CRAN are signed with the key of “Michael Rutter marutter@gmail.com” with key ID E084DAB9

Really, r-project? REALLY? You couldn't come up with a more official key for distributing the software than "marutter@gmail.com"? So, if his Gmail gets hacked then everyone who has added this key is potentially susceptible to compromise?

Ok, now I'm really done.`</rant>`

workers/Dockerfile.base
mhuyck: Gotcha. That makes sense. I was overlooking the `/usr/` part, I guess, and jumping straight to `site-library`.

workers/Dockerfile.base
mhuyck: Works for me. I was guessing you had done this iteratively: added the `rm` at line 49, then discovered the `--no-cache-dir` param but didn't remove line 49.

workers/Dockerfile.base
mhuyck: I've found many Docker tutorials and documentation to be chasing a moving target. I do think it makes sense if we're deploying to production to not tie ourselves to a bleeding edge branch.

mhuyck: Sounds good.

workers/r_dependencies.R
```diff
@@ -0,0 +1,209 @@
+install.packages('ff', repos='https://cloud.r-project.org/')
+install.packages('XML', repos='https://cloud.r-project.org/')
+install.packages('RCurl', repos='https://cloud.r-project.org/')
+install.packages('RSQLite', repos='https://cloud.r-project.org/')
+
+source('https://bioconductor.org/biocLite.R')
```
mhuyck: This opens another big security hole into our image. We are placing lots of trust in the folks at `bioconductor.org` to keep tight control (and preferably an audit log) over changes to that `biocLite.R` file... in the wrong hands, that script can turn the installed base of bioconductor into a giant botnet.

workers/r_dependencies.R
mhuyck: Fair enough. I guess the name `r_dependencies.R` makes that purpose pretty clear so I'm okay with that.

mhuyck:
Ah, well... looks as if CircleCI found another issue. Code looks good to me though, so if you figure out why CircleCI is unhappy I'd love to know why.

workers/Dockerfile.tests
```diff
@@ -0,0 +1,3 @@
+FROM dr_worker
```
mhuyck: Should this line now say `FROM dr_worker_tests`?

mhuyck: Might be related to the CircleCI issue, if I'm following the Docker trail correctly.

kurtwheeler: It is. I was trying to trim down lines with this change but it doesn't work at all. I basically wanted a new dockerfile with just a different entrypoint, but doing it like this won't force a rebuild of dr_worker if the code changes, so the code stays stale. My latest change should fix this (I just go back to how this file was before.)

mhuyck: Ah.. just saw your revert.

So were you trying to take advantage of Docker's layering to do less rebuilding? Sounds good in theory, but not if it doesn't work, obviously. Maybe Docker needs a dependency tree feature so developers can link together related repositories and automatically rebuild upstream as necessary.

workers/data_refinery_workers/processors/array_express.py
mhuyck: Objects gotta noun, methods gotta verb. I'm good with that.

workers/Dockerfile.tests
mhuyck: Hmm. GitHub is being slow with your comments so mine is out of order, but I follow you.

PR #36, opened 2017-08-09
Microarray differentiation

kurtwheeler:
Adds a lot of code to figure out what to do with different sample types. This code implements the following heuristic:
   - If there are no raw files, download the processed file and run a no-op processor on it, effectively just putting the processed file into the compendium
    - Check to see if the sample file is a .CEL, if so process it with SCAN.UPC.
    - If there are raw files that are not .CELs, just download them without running any processor so the grant application is true.

Also fixes a bunch of minor bugs that came up during the big processing job, makes some infrastructure adjustments, and greatly improves celery task discovery so the Foreman can requeue jobs.


foreman/data_refinery_foreman/surveyor/external_source.py
```diff
@@ -72,8 +43,24 @@ def determine_pipeline(self,
...
+        Returns the Celery Downloader Task name from the
+        data_refinery_workers project which should be queued to
+        download Batches discovered by this surveyor.
+        """
+        return DOWNLOADER_TASK_LOOKUP[self.source_type()]
+
     def handle_batches(self, batches: List[Batch]):
+        new_batches = []
         for batch in batches:
+            if not Batch.is_new_batch(batch):
```
rzelayafavila: Hmmm - I'm not sure how much is being added by this `is_new_batch` function. Could we just call here (and wherever else this function is called) something like the original `if  Batch.objects.filter(name=batch.name).count() > 0:` line that the `is_new_batch` function calls (albeit somewhat differently)?

data_models/data_refinery_models/migrations/0002_auto_20170727_2100.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
```
kurtwheeler: Because I wanted to clear out the production DB, so I took the opportunity to clean up the migration files since I could just reapply them all from scratch. At this point, there is a lot of data in the database so I think this is the last time I can do this.

data_models/data_refinery_models/models/batches.py
```diff
@@ -44,6 +44,11 @@ class Batch(TimeTrackedModel):
     # This is the organism name as it appeared in the experiment.
     organism_name = models.CharField(max_length=256)
 
+    @classmethod
+    def is_new_batch(cls, batch) -> bool:
+        matching_batches = cls.objects.filter(name=batch.name)
+        return True if matching_batches.count() == 0 else False
```
kurtwheeler: Oh it definitely is, good call :+1:

foreman/Dockerfile
```diff
@@ -13,6 +13,12 @@ COPY data_models/dist/data-refinery-models-* data_models/
 RUN pip install \
```
kurtwheeler: The latter, although if you think it'd be more clear and/or explicit if I used pip3/python3 in this file I can do that.

foreman/Dockerfile
```diff
@@ -13,6 +13,12 @@ COPY data_models/dist/data-refinery-models-* data_models/
...
     data_models/$(ls data_models -1 | sort --version-sort | tail -1)
 
+COPY common/dist/data-refinery-common-* common/
```
kurtwheeler: The only file in there currently is `data-refinery-common-0.1.0.tar.gz`, however once I actually get to a good point I'll start being diligent about using semantic versioning for every change so I think there'll pretty much be the same file but with different version numbers.

workers/Dockerfile
```diff
@@ -10,11 +10,6 @@ COPY common/dist/data-refinery-common-* common/
 # Get the latest version from the dist directory.
 RUN pip3 install common/$(ls common -1 | sort --version-sort | tail -1)
 
-COPY common/dist/data-refinery-common-* common/
-
-# Get the latest version from the dist directory.
-RUN pip install common/$(ls common -1 | tail -1)
```
kurtwheeler: Yeah they were, I somehow duplicated them before without noticing.

foreman/data_refinery_foreman/foreman/main.py
```diff
@@ -215,7 +218,8 @@ def retry_lost_processor_jobs() -> None:
         start_time=None,
         end_time=None,
         created_at__lt=minimum_creation_time
-    )
+        # TEMPORARY for Jackie's grant
+    ).exclude(pipeline_applied=ProcessorPipeline.NONE.value)
```
kurtwheeler: Yes it is, although you actually just provided one example of why this is useful. `pipeline_applied="None"` is actually wrong, it would have to be `pipeline_applied="NONE"`. By avoiding the use of magic strings I guarantee that they are all consistent everywhere. Basically its the same argument as ever for using Enumerated Types.

workers/Dockerfile.base
```diff
@@ -39,6 +39,11 @@ COPY workers/r_dependencies.R .
 
 RUN Rscript r_dependencies.R
 
+# Move these apt-gets up into the other one once I have time for rebuild
+# libpq-dev install pg_config which is needed for psycopg2
```
kurtwheeler: Yeah that's not a bad call, I'll make one for it.

foreman/data_refinery_foreman/surveyor/external_source.py
kurtwheeler: I think clarity is being added by the `is_new_batch` function. I recently read https://martinfowler.com/bliki/FunctionLength.html and I thought it made a lot of sense. I think it validates the existence of this function pretty well.

workers/Dockerfile.base
kurtwheeler: https://github.com/data-refinery/data_refinery/issues/37

kurtwheeler:
Ok @rzelayafavila I've answered questions and made a single change. Lemme know what you think.

Also about your point about the breadth of this PR, I agree and I apologize. I just was quickly iterating on all the issues I was encountering while running it in production and didn't want to bog down that iteration by splitting it across a dozen PRs.

foreman/data_refinery_foreman/surveyor/external_source.py
rzelayafavila: Mhmm, yeah, I see how that makes sense. I could also argue that `Batch.objects.filter(name=batch.name).count() > 0` is a standard Django way of doing things - people who use Django will use statements like that a lot to check if there is already an object in the table with those characteristics. However, what you say, in the context of Martin Fowler's blog post, makes sense, so I think this form also makes sense.

workers/Dockerfile.base
rzelayafavila: :+1:

foreman/Dockerfile
rzelayafavila: Hmmm - do you want to keep this as is and copy all of the files or do something like what you have in the `pip install` command above and only copy the newest version?

foreman/Dockerfile
rzelayafavila: Ok, that's fine then. Maybe put in a comment in the first `pip install` statement that says that all pip installs in this file will use pip3.

data_models/data_refinery_models/migrations/0002_auto_20170727_2100.py
rzelayafavila: I thought that might be the case - why are there 2 migration files then though? Wouldn't that have reduced your migration files to just one?

foreman/Dockerfile
kurtwheeler: I believe I did this the way I did because what you're describing isn't possible. The only Dockerfile command capable of interacting with the host's filesystem is `COPY`, so until I copy the directory into the image I cannot run `$(ls data_models -1 | sort --version-sort | tail -1)` on it.

data_models/data_refinery_models/migrations/0002_auto_20170727_2100.py
kurtwheeler: It did. And then I realized I needed to make another change midway through the run of all the processors. I didn't want to blow away the DB because at that point there was production data in it, so I just ran the single additional migration.

kurtwheeler:
@rzelayafavila I've added a comment about pip3 and responded to your other questions. Lemme know what you think.

foreman/Dockerfile
```diff
@@ -5,6 +5,8 @@ WORKDIR /home/user
 
 COPY foreman/requirements.txt .
 
+# The base image does not have Python 2.X on it at all, so all calls
```
rzelayafavila: :+1:

PR #38, opened 2017-08-18
Job failure reasons

kurtwheeler:
This PR tracks the failure of jobs much much better. It both improves the logging for job failures to be more clear, and also adds a `failure_reason` to the jobs tables. It also adds some long overdue tests along with refactoring the processing code to be more readable, reusable, and testable.


workers/run_tests.sh
```diff
@@ -11,11 +11,19 @@ cd $script_directory
 # move up a level
 cd ..
 
+# Set up the test data volume directory if it does not already exist
+volume_directory="$script_directory/test_volume"
+if [ ! -d "$volume_directory" ]; then
+    mkdir $volume_directory
```
dongbohu: Do you need to consider the case of `$script_directory/test_volume` exists but is not a directory? In that case `mkdir` command will fail.

common/data_refinery_common/file_management.py
```diff
@@ -35,7 +35,7 @@ def _get_downloaded_name(batch: Batch) -> str:
     return os.path.basename(path)
 
 
-def _get_processed_name(batch: Batch) -> str:
+def get_processed_name(batch: Batch) -> str:
```
kurtwheeler: Yes, it was previously only used within the `file_management` namespace but now it is used elsewhere as well.

workers/data_refinery_workers/downloaders/array_express.py
```diff
@@ -21,34 +21,36 @@
...
+                     job.id)
         target_file = open(file_path, "wb")
         with closing(urllib.request.urlopen(download_url)) as request:
             shutil.copyfileobj(request, target_file, CHUNK_SIZE)
     except Exception:
-        logging.exception("Exception caught while running Downloader Job #%d.",
-                          job_id)
+        logging.exception("Exception caught while downloading batch in Downloader Job #%d.",
+                          job.id)
+        job.failure_reason = "Exception caught while downloading batch"
```
kurtwheeler: I think you misread my code a bit, because in lines 28-31 `job.failure_reason` does not include the `job.id`, only the message that gets logged does. In line 31 `job.failure_reason` gets set to `failure_message`, which has been defined as `"A Batch doesn't have the same download URL as the other batches"`. Within the call to `logger.error` I append `" in Downloader Job #%d."` to that base string and format it with the `job.id`.

Basically my thought is that the `job.failure_reason` does not need the id logged with it since it is on the job object and therefore has the ID. However I do want to log the `job.id` in the log message because otherwise I'll have no way to know which job failed.

workers/data_refinery_workers/downloaders/array_express.py
```diff
@@ -87,7 +89,8 @@ def _extract_file(batches: List[Batch], job_id: int) -> None:
     except Exception:
         logging.exception("Exception caught while extracting %s during Downloader Job #%d.",
                           zip_path,
-                          job_id)
+                          job.id)
+        job.failure_reason = "Exception caught while extracting " + zip_path
```
kurtwheeler: See my response to your previous comment.

workers/data_refinery_workers/downloaders/utils.py
```diff
@@ -77,6 +82,11 @@ def queue_task(processor_job):
                 processor_job = save_batch_create_job(batch)
                 if batch.pipeline_required != ProcessorPipeline.NONE.value:
                     success = queue_task(processor_job)
+                    if success:
+                        logger.info("Downloader Job %d completed successfully.")
```
kurtwheeler: Good catch :+1:

workers/data_refinery_workers/processors/array_express.py
```diff
@@ -11,97 +12,114 @@
...
 
-
-def cel_to_pcl(kwargs: Dict) -> Dict:
-    """Process .CEL files to .PCL format using R.
-
-    Moves the .CEL file from the raw directory to the temp directory,
-    calls ProcessCelFiles, uploads the processed file, and cleans up
-    the raw/temp files. Because it uses the file_management module
-    this works seamlessly whether S3 is being used or not.
+    Also adds the keys input_file and output_file to job_context so
```
kurtwheeler: Yes :)

workers/data_refinery_workers/processors/array_express.py
```diff
@@ -11,97 +12,114 @@
...
+    try:
+        header = ro.r['::']('affyio', 'read.celfile.header')(input_file)
+    except RRuntimeError as e:
+        # Array Express processor jobs have only one batch per job.
+        file_management.remove_temp_directory(job_context["batches"][0])
+
+        base_error_template = "unable to read Affy header in input file {0} due to error: {1}"
+        base_error_message = base_error_template.format(input_file, str(e))
+        log_message = "Processor Job %d running AFFY_TO_PCL pipeline " + base_error_message
+        logger.error(log_message)
```
kurtwheeler: Yep, yep I do. (I wasn't as careful about testing these log statements as I should have been because the next branch I worked on was about fixing log statements.)

workers/run_tests.sh
kurtwheeler: Seems like an extreme edge case, I can handle it if you think it's important but I would think that `mkdir` would give an error message and the user could then remove or rename their `test_volume` file.

kurtwheeler:
@dongbohu I've responded to all your comments (or just made the suggested change). Let me know what you think.

workers/data_refinery_workers/downloaders/array_express.py
dongbohu: Ah, my fault. Thanks for the explanation.

workers/run_tests.sh
dongbohu: I agree that this is a very rare case. So it is okay if you don't want to handle it.

PR #44, opened 2017-09-06
End to end tests

kurtwheeler:
Adds end-to-end tests for the Data Refinery. These run separately from all the other tests, using Docker Compose. Docker Compose runs 4 containers:
 - Workers (in normal mode, but with an environment of `test`)
 - Foreman (run via the test-runner)
 - Rabbit
 - Nginx (used to download data from within the workers)

The Foreman runs a survey job, which discovers a batch, queues a downloader job, and then waits for it to finish so it can discover its processor job, then waits for the processor job to finish. The workers run the downloader and processor jobs. The Foreman tests then check that `SCAN.UPC` outputting a processed file. More in depth testing of the contents of that file are upcoming.

I tested this without an internet connection to be sure that no external requests were actually being made. This will keep the tests from being finicky.

### DIFF
--- a/foreman/data_refinery_foreman/surveyor/end_to_end_tests.py
+++ b/foreman/data_refinery_foreman/surveyor/end_to_end_tests.py
@@ -1,6 +1,9 @@
 import time
+import json
 from django.test import TransactionTestCase
+from unittest.mock import patch, Mock
 from data_refinery_models.models import (
+    Organism,
     SurveyJob,
     SurveyJobKeyValue,
     Batch,
@@ -8,6 +11,7 @@ from data_refinery_models.models import (
     ProcessorJob
 )
 from data_refinery_foreman.surveyor import surveyor
+from data_refinery_foreman.surveyor.array_express import EXPERIMENTS_URL
 
 # Import and set logger
 import logging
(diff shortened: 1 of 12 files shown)


workers/data_refinery_workers/processors/test_array_express.py
```diff
@@ -165,13 +165,13 @@ def test_success(self, mock_file_management: MagicMock):
 
         # If this file already exists for any reason then we aren't
```
dongbohu: You may want to revise these two lines of comments due to the code change. The new code seems to always remove this file if it already exists.

foreman/data_refinery_foreman/surveyor/end_to_end_tests.py
```diff
@@ -0,0 +1,295 @@
...
+    while job.success is None:
+        logger.info("Still polling the %s.",
+                    job_class.__name__)
+        time.sleep(LOOP_TIME)
+        job = job_class.objects.filter(id=job.id).get()
+
+    return job
+
+
+class ScanUpcEndToEndTestCase(TransactionTestCase):
```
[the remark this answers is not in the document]
kurtwheeler: SCAN.UPC stands for: [Single-channel array normalization (SCAN) and Universal exPression Codes (UPC)](https://bioconductor.org/packages/release/bioc/html/SCAN.UPC.html).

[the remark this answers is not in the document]
dongbohu: Good to know. Thanks!

workers/data_refinery_workers/processors/test_array_express.py
kurtwheeler: Wait but I meant for these lines of comments to be explaining why I'm removing the file if it already exists. Is that unclear? Is there a way I could make that clearer?

workers/data_refinery_workers/downloaders/utils.py
```diff
@@ -78,15 +79,14 @@ def queue_task(processor_job):
...
-            with transaction.atomic():
-                processor_job = save_batch_create_job(batch)
-                if batch.pipeline_required != ProcessorPipeline.NONE.value:
-                    success = queue_task(processor_job)
-                    if success:
-                        logger.info("Downloader Job %d completed successfully.", job.id)
-                    else:
-                        failure_template = "Could not find Processor Pipeline {} in the lookup."
-                        job.failure_reason = failure_template.format(batch.pipeline_required)
+            processor_job = save_batch_create_job(batch)
```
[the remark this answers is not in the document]
kurtwheeler: I moved the transaction to only be around the job-creation call because including the message in the transaction can cause some weird behavior. I've created an issue to just do this within that function instead of wrapping every invocation of it: https://github.com/data-refinery/data_refinery/issues/45

kurtwheeler:
Ok @dongbohu I've made changes based on your feedback. One change that wasn't explicitly related to your suggestions was that when queuing a processor job I catch exceptions caused by the message queue not being available, and then delete the job record since it will never get picked up.

Take a look and let me know if there's anything else you think I need to address.

workers/data_refinery_workers/processors/test_array_express.py
dongbohu: I think what confused me in the comment is "then we aren't actually testing that it is generated". The code actually removes `output_file` when it already exists. Do you think a comment like this:
```If output_file exists, remove it first.```
would make more sense?

kurtwheeler: Yup, I dig that. Nice and brief :)

PR #46, opened 2017-09-19
Daiquiri logging

kurtwheeler:
Improves logging across the entire project (except the organisms code actually, explained below). The `common/logging` namespace provides a function `get_and_configure_logger` which returns a logger with a format that is consistent across the project:
```
2017-09-19 17:25:06,462 local/MainProcess data_refinery_workers.downloaders.utils INFO [downloader_job: 2]: Downloader job completed successfully.
```
(see line 8 of `common/logging` for the format string).

The `[downloader_job: 2]` part of that string was generated because the logger function was called like:
```
logger.info("Downloader job completed successfully.", downloader_job=downloader_job.id)
```
Any additional keywords passed to any of the logging functions will be formatted in this manner.

The reason why the organisms code doesn't use this logger is because it doesn't actually need to support the additional context like this. It still would be good to do for consistency sake, but first I need to address https://github.com/data-refinery/data_refinery/issues/39.


workers/data_refinery_workers/downloaders/management/commands/queue_downloader.py
```diff
@@ -51,5 +49,5 @@ def handle(self, *args, **options):
         batch.save()
 
         downloader_job = DownloaderJob.create_job_and_relationships(batches=[batch])
-        logger.info("Queuing a task.")
         download_array_express.delay(downloader_job.id)
+        logger.info("DownloaderJob queued.")
```
rzelayafavila: :+1: Nice - I agree this is more informative.

workers/data_refinery_workers/downloaders/test_array_express.py
```diff
@@ -148,7 +148,7 @@ def test_verification_failure(self,
         self.assertIsNotNone(downloader_job.start_time)
         self.assertIsNotNone(downloader_job.end_time)
         self.assertEqual(downloader_job.failure_reason,
-                         "A Batch doesn't have the same download URL as the other batches")
+                         "A Batch doesn't have the same download URL as the other batches.")
```
rzelayafavila: Would it make sense to output which Batch doesn't have the same download URL in the failure_reason of the downloader_job?

common/data_refinery_common/test_utils.py
```diff
@@ -8,6 +8,7 @@ class UtilsTestCase(TestCase):
     @patch('data_refinery_common.utils.requests.get')
     def test_get_worker_id_cloud(self, mock_get, mock_get_env_variable):
         """Test that a request is made and the global value is stored"""
+        utils.INSTANCE_ID = None
```
rzelayafavila: What is this change for? It doesn't seem like this change is part of the logging changes, right?

common/data_refinery_common/logging.py
```diff
@@ -0,0 +1,37 @@
...
+import logging
+import sys
+from data_refinery_common.utils import get_worker_id
+
+
+# base_format_string = "%(asctime)s {0} %(name)s %(color)s%(levelname)s"
+FORMAT_STRING = (
+    "%(asctime)s {0} %(name)s %(color)s%(levelname)s%(extras)s"
+    ": %(message)s%(color_stop)s"
+).format(get_worker_id())
```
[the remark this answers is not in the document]
kurtwheeler: Ah yeah I can see how that would be confusing, but you did in fact figure it out. I'll add some comments to make that more clear.

common/data_refinery_common/logging.py
```diff
@@ -0,0 +1,37 @@
+import daiquiri
+import logging
+import sys
+from data_refinery_common.utils import get_worker_id
+
+
+# base_format_string = "%(asctime)s {0} %(name)s %(color)s%(levelname)s"
```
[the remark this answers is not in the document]
kurtwheeler: I suspect the latter, I'll rip it out.

foreman/requirements.txt
```diff
@@ -9,7 +9,7 @@ billiard==3.5.0.2         # via celery
 boto==2.46.1
 celery==4.0.2
 django==1.11.1
-GEOparse==0.1.10
+geoparse==0.1.10
```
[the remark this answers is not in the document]
kurtwheeler: Oh good point. Seems annoying that they can't maintain Python's convention of lowercase package names. However I really should just remove this because I actually am not using it any more.

workers/data_refinery_workers/downloaders/test_array_express.py
kurtwheeler: I thought about that, however the issue is that it's not clear which is the culprit Batch. They could all have different URLs which would then make calling one of them out as being wrong a red herring. I think the correct debugging steps when seeing that `failure_reason` are to look at all the `download_urls` for the Batches in the DownloaderJob and I think the ambiguity would perhaps lead the person debugging it to look at all of them.

common/data_refinery_common/test_utils.py
kurtwheeler: I think that I realized that if the test runner ever ran these two tests in the opposite order that it could break them. It is deterministic so they aren't likely to fail until more tests are added, however it's brittle so I decided to fix it while I was thinking about it.

kurtwheeler:
Ok @rzelayafavila I think I have answered or addressed your concerns. Let me know what you think.

common/data_refinery_common/logging.py
```diff
@@ -4,7 +4,9 @@
 from data_refinery_common.utils import get_worker_id
 
 
-# base_format_string = "%(asctime)s {0} %(name)s %(color)s%(levelname)s"
+# Most of the formatting in this string is for the logging system. All
+# that the call to format() does is replace the "{0}" in the string
+# with the worker id.
```
rzelayafavila: :+1:

common/data_refinery_common/test_utils.py
rzelayafavila: Ok - maybe it would be good to put a quick inline comment to explain this, since it is not immediately apparent.

foreman/requirements.txt
```diff
@@ -9,15 +9,10 @@ billiard==3.5.0.2         # via celery
 boto==2.46.1
 celery==4.0.2
 django==1.11.1
-geoparse==0.1.10
 kombu==4.0.2              # via celery
-numpy==1.12.1             # via geoparse, pandas
-pandas==0.19.2            # via geoparse
 psycopg2==2.7.1
-python-dateutil==2.6.0    # via pandas
-pytz==2017.2              # via celery, django, pandas
```
rzelayafavila: Nice! I'm glad you commented which dependencies were needed through other dependencies so that they could all be cleaned out in a situation like this one!

common/data_refinery_common/logging.py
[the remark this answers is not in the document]
rzelayafavila: It's true, but I think we could all *eventually* figure most of each others' code out. I think the point is to make code easy and relatively fast to read, especially for someone who is not familiar with the file or the project (and who might not even be in the lab, etc.)

workers/data_refinery_workers/processors/test_array_express.py
```diff
@@ -127,7 +127,7 @@ def test_failure(self):
                        "input_file": input_file}
 
         job_context = array_express._determine_brainarray_package(job_context)
-        failure_reason_template = """unable to read Affy header in input file /home/user/data_store/temp/TEST2/AFFY_TO_PCL/batch_{0}/dummy due to error: Error in (function (filename, info = c("basic", "full"), verbose = FALSE)  : \n  Is /home/user/data_store/temp/TEST2/AFFY_TO_PCL/batch_{1}/dummy really a CEL file? tried reading as text, gzipped text, binary, gzipped binary, command console and gzipped command console formats\n"""  # noqa
+        failure_reason_template = """Unable to read Affy header in input file /home/user/data_store/temp/TEST2/AFFY_TO_PCL/batch_{0}/dummy while running AFFY_TO_PCL due to error: Error in (function (filename, info = c("basic", "full"), verbose = FALSE)  : \n  Is /home/user/data_store/temp/TEST2/AFFY_TO_PCL/batch_{1}/dummy really a CEL file? tried reading as text, gzipped text, binary, gzipped binary, command console and gzipped command console formats\n"""  # noqa
```
rzelayafavila: Does this string, and the one on line 204 need to be all in the same line or can they be broken up to fit our 80-char limit?

common/data_refinery_common/logging.py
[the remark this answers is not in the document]
kurtwheeler: Oh no, that's not what I meant at all. I totally agree with you, I was just letting you know that you had in fact figured it out correctly as a way of answering your question.

foreman/requirements.txt
kurtwheeler: Well actually, I just reran `pip-compile` on the `requirements.in` file and it automagically regenerated this `requirements.txt` file.

rzelayafavila: Nice :+1:

kurtwheeler:
Aight @rzelayafavila, this is back to you now.

common/data_refinery_common/test_utils.py
```diff
@@ -8,6 +8,8 @@ class UtilsTestCase(TestCase):
     @patch('data_refinery_common.utils.requests.get')
     def test_get_worker_id_cloud(self, mock_get, mock_get_env_variable):
         """Test that a request is made and the global value is stored"""
+        # Ensure utils.INSTANCE_ID hasn't been set yet in case the
+        # order the tests are run in ever changes
```
rzelayafavila: :+1:

workers/data_refinery_workers/processors/test_array_express.py
```diff
@@ -127,7 +127,16 @@ def test_failure(self):
...
+        failure_reason_template = (
+            "Unable to read Affy header in input file "
+            "/home/user/data_store/temp/TEST2/AFFY_TO_PCL/batch_{0}/dummy "
+            "while running AFFY_TO_PCL due to error: Error in (function "
+            '(filename, info = c("basic", "full"), verbose = FALSE)  : \n  '
+            "Is /home/user/data_store/temp/TEST2/AFFY_TO_PCL/batch_{1}/dummy "
+            "really a CEL file? tried reading as text, gzipped text, binary, "
+            "gzipped binary, command console and gzipped command console "
+            "formats\n"
+        )
```
rzelayafavila: :+1:

workers/data_refinery_workers/processors/test_array_express.py
```diff
@@ -201,7 +210,16 @@ def test_failure(self):
...
+        failure_reason_template = (
+            'Encountered error in R code while running AFFY_TO_PCL pipeline '
+            'during processing of '
+            '/home/user/data_store/temp/TEST2/AFFY_TO_PCL/batch_{0}/dummy: '
+            'Error in {{ : \n  task 1 failed - '
+            '"Is /home/user/data_store/temp/TEST2/AFFY_TO_PCL/batch_{1}/dummy '
+            'really a CEL file? tried reading as text, gzipped text, binary, '
+            'gzipped binary, command console and gzipped command console '
+            'formats\n"\n'
+        )
```
rzelayafavila: :+1:

PR #65, opened 2017-10-09
Files table

kurtwheeler:
This is a big PR, but it's not as bad as it looks. The main goal of the PR is to factor file information out of the Batches table so that each Batch may have more than one File. In the process of doing this I realized that this was a perfect (and perhaps even necessary) time to address https://github.com/data-refinery/data_refinery/issues/39 and https://github.com/data-refinery/data_refinery/issues/30. Therefore I have deleted the `data_models` sub-project and moved all of the `data_models` code into the `common` sub-project.

So a lot of this code is code moving around, a good amount it is changing `data_refinery_models.models` to `data_refinery_common.models`, and then only some of it is actually meaningful code changes cause by refactoring. I also added tests for the `no_op` processor because I realized I didn't have any and wanted to be sure my refactor didn't break it.


common/data_refinery_common/models/batches.py
```diff
@@ -0,0 +1,277 @@
...
+            this_batch.save()
+            this_batch.files.add(*files)
+
+        return this_batch
+
+    class Meta:
+        db_table = "batches"
+
+
+class File(TimeTrackedModel):
```
dongbohu: So `File` is a new class that was refactored out of the original `Batch` class, am I right?

foreman/data_refinery_foreman/surveyor/array_express.py
```diff
@@ -25,19 +25,29 @@ def source_type(self):
...
         else:
             return ProcessorPipeline.NONE
 
+    def group_batches(self) -> List[List[Batch]]:
+        """Groups batches based on the download URL of their only File"""
+        groups = []
+        download_urls = {file.download_url for file in [batch.files[0] for batch in self.batches]}
+
+        for url in download_urls:
+            groups.append([batch for batch in self.batches if batch.files[0].download_url == url])
```
dongbohu: This approach to group the urls is compact and seems "pythonic", but the performance may not be the best. Suppose the number of elements in `self.batches` is **b**, and the number of elements in `download_urls` is **d**, the double loop means that the time will be O(b * d). But if we build a dict whose key is download_url and the value is an array of batches, it can be done in O(b + d) time. 

Of course if b and d are both small, then this big O discussion is just academic trivia. Do you know the typical values of b and d?

workers/data_refinery_workers/processors/test_utils.py
```diff
@@ -2,50 +2,63 @@
...
     BatchStatuses,
+    File,
     ProcessorJob,
 )
 from data_refinery_workers.processors import utils
-from data_refinery_common import file_management
 
 
-def init_batch():
+def init_objects():
```
dongbohu: I noticed that `init_objects()` function are available in all test modules in this PR and they look similar. How easy is it to merge them into one function?

common/data_refinery_common/models/batches.py
kurtwheeler: Yes, the reason for this is that Batches should be able to have more than one File, which they can't if the fields are part of the Batch.

workers/data_refinery_workers/processors/test_utils.py
kurtwheeler: They are indeed similar, but the differences between them are pretty specific to the test files they're located in. Merging them into a single function might reduce the total number of lines, but I think the amount of configuration the function would have to accept to be able to meet all the different needs would end up making it much less clear. For example the `init_objects()` function in `test_utils.py` inserts two batches, whereas it only inserts one in the other two. The other two have some key differences in values for the objects it creates that are tied to the locations that the files the tests operate upon. 

Also I think that it's somewhat nice to have the initialization function defined in the test file itself so that when you work on the tests you can see what's been set up for the test without going to another file. It's not a huge benefit on it's own, but combined with the finagling that would be involved to merge the functions I think it's clearer and simpler to do keep these functions separate.

kurtwheeler:
@dongbohu I fixed the runtime issues with the `group_batches()` function and responded to your other comments. Let me know what you think.

foreman/data_refinery_foreman/surveyor/array_express.py
kurtwheeler: I think b and d will both be small enough that it won't be an issue, but that is a much more elegant solution so I went ahead and changed it. (I managed to do it in a single pass too). I even added a test for it because I probably should have done so originally.

foreman/data_refinery_foreman/surveyor/array_express.py
```diff
@@ -40,13 +40,19 @@ def determine_pipeline(self,
...
-            groups.append([batch for batch in self.batches if batch.files[0].download_url == url])
+        # Builds a mapping of each unique download_url to a list of
+        # Batches whose first File's download_url matches.
+        download_url_mapping = {}
+        for batch in self.batches:
+            download_url = batch.files[0].download_url
+            if download_url in download_url_mapping:
+                download_url_mapping[download_url].append(batch)
+            else:
+                download_url_mapping[download_url] = [batch]
```
dongbohu: :+1:

foreman/data_refinery_foreman/surveyor/array_express.py
```diff
@@ -40,13 +40,19 @@ def determine_pipeline(self,
...
 
-        return groups
+        # The values of the mapping we built are the groups the
+        # batches should be grouped into.
+        return list(download_url_mapping.values())
```
dongbohu: Isn't `download_url_mapping.values()` already a list?

foreman/data_refinery_foreman/surveyor/test_array_express.py
```diff
@@ -306,3 +306,20 @@ def test_survey(self, mock_send_task, mock_get):
...
+        file2 = File(download_url="a")
+        file3 = File(download_url="b")
+        file4 = File(download_url="a")
+        batch1 = Batch(files=[file1])
+        batch2 = Batch(files=[file2])
+        batch3 = Batch(files=[file3, file4])
+
+        surveyor.batches = [batch1, batch2, batch3]
+        groups = surveyor.group_batches()
+        self.assertEqual(groups, [[batch1, batch2], [batch3]])
```
dongbohu: :+1:

foreman/data_refinery_foreman/surveyor/array_express.py
kurtwheeler: No, in fact they are [dictionary view objects](https://docs.python.org/3/library/stdtypes.html#dictionary-view-objects). I believe they are iterable, but cannot be accessed via indices:
```python
>>> d = {"a": 1}
>>> v = d.values()
>>> v[0]
  File "<stdin>", line 1, in <module>
TypeError: 'dict_values' object does not support indexing
```

kurtwheeler:
I actually had not heard of DAMP, but I like the idea a lot and it seems to validate my earlier notions. Thanks for linking it to me :)

I responded to your last question, back to you.

foreman/data_refinery_foreman/surveyor/array_express.py
dongbohu: You are right. This seems to be a new feature in Python 3. I forgot it.

common/setup.py
```diff
@@ -1,13 +1,25 @@
 import os
 from setuptools import find_packages, setup
+import re
 
 # allow setup.py to be run from any path
 os.chdir(os.path.normpath(os.path.join(os.path.abspath(__file__), os.pardir)))
 
+# code found at: https://stackoverflow.com/a/7071358/6095378
+VERSIONFILE = "data_refinery_common/_version.py"
+verstrline = open(VERSIONFILE, "rt").read()
```
dongbohu: Do you mind using underscore to make `verstrline`, `VSRE` and `verstr` more readable?

common/setup.py
```diff
@@ -1,13 +1,25 @@
...
+VSRE = r"^__version__ = ['\"]([^'\"]*)['\"]"
```
dongbohu: Not sure whether you should add more comments. I always found code with regular expression hard to follow.

foreman/setup.py
```diff
@@ -0,0 +1,42 @@
+import os
+from setuptools import find_packages, setup
+import re
+
+
+# allow setup.py to be run from any path
+os.chdir(os.path.normpath(os.path.join(os.path.abspath(__file__), os.pardir)))
+
+# code found at: https://stackoverflow.com/a/7071358/6095378
```
dongbohu: See comments above.

kurtwheeler:
@dongbohu good call on the variable names, those were awful and I hated them. I just lazily kept them the same as the code I found. Also good call on the regex comment. I had to think for a moment about what it was actually doing myself.

Anyway, I made those changes. Look good?

PR #69, opened 2017-10-19
Reduce r output

kurtwheeler:
This PR reduces the output from R code by a lot. This output is noisy, and mostly useless because we know what is going to run so we don't want to see the output about it. The only thing we really care about is whether it succeeds or fails, which we can tell without looking at the output.


workers/data_refinery_workers/processors/array_express.py
```diff
@@ -88,21 +89,25 @@ def _run_scan_upc(job_context: Dict) -> Dict:
     input_file = job_context["input_file_path"]
 
     try:
+        # It's necessary to load the foreach library before calling SCANfast
+        # because it doesn't load the library before calling functions
+        # from it.
+        ro.r("suppressMessages(library('foreach'))")
```
mhuyck: Cool! I am happy to know about `suppressMessages`.

workers/data_refinery_workers/processors/array_express.py
```diff
@@ -88,21 +89,25 @@ def _run_scan_upc(job_context: Dict) -> Dict:
...
-        # from it.
-        ro.r("library('foreach')")
+        # All R messages are turned into Python 'warnings' by rpy2. By
+        # filtering all of them we silence a lot of useless output
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")
+            scan_upc = ro.r['::']('SCAN.UPC', 'SCANfast')
+            scan_upc(input_file,
+                     job_context["output_file_path"],
+                     probeSummaryPackage=job_context["brainarray_package"])
```
mhuyck: 👌

PR #80, opened 2017-12-11
Salmon quant

kurtwheeler:
This PR adds the Salmon processor to the data refinery. Pull request #70 added the Surveyor and the Downloader for SRA. The Salmon processor is a bit unique because it has to find the correct Salmon index for the organism it is processing. This Salmon index is generated by the Index Refinery, which was added in PR #77.

The only other particularly unique thing about this PR is that some of the test files were too large to store in the repo so I used Git LFS, but even then the index was too big when compressed using `gzip`. Therefore I had to compress it using `xz` instead. I have added code to the `workers/run_tests.sh` file which will make sure that this file is downloaded and then in the correct format that the code will be looking for.


workers/data_refinery_workers/processors/salmon.py
```diff
@@ -0,0 +1,258 @@
...
+
+    job_context["input_file_path"] = files[0].get_temp_pre_path(job_context["job_dir_prefix"])
+    # Salmon outputs an entire directory of files, so create a temp
+    # directory to output it to until we can zip it to
+    # files[0].get_temp_post_path()
+    job_context["output_directory"] = os.path.join(
+        files[0].get_temp_dir(job_context["job_dir_prefix"]), "output")
+    os.makedirs(job_context["output_directory"], exist_ok=True)
+
+    if len(files) == 2:
```
rzelayafavila: Do we ever expect the the length of files to be something other than 0, 1, or 2? Should we handle these cases?

workers/data_refinery_workers/processors/salmon.py
```diff
@@ -0,0 +1,258 @@
...
+        with gzip.open(job_context["input_file_path_2"], "rt") as input_file:
+            for line in input_file:
+                if counter % 4 == 2:
+                    total_base_pairs += len(line.replace("\n", ""))
+                    number_of_reads += 1
+
+                counter += 1
+
+    if total_base_pairs / number_of_reads > 75:
+        job_context["kmer_size"] = "31"
```
rzelayafavila: I'm curious - why are these numbers strings here instead of ints?

workers/data_refinery_workers/processors/salmon.py
```diff
@@ -0,0 +1,258 @@
...
+            if counter % 4 == 2:
+                total_base_pairs += len(line.replace("\n", ""))
+                number_of_reads += 1
+
+            counter += 1
+
+    if "input_file_path_2" in job_context:
+        with gzip.open(job_context["input_file_path_2"], "rt") as input_file:
+            for line in input_file:
+                if counter % 4 == 2:
```
rzelayafavila: I think a comment explaining the logic of this line here would be helpful.

workers/data_refinery_workers/processors/salmon.py
```diff
@@ -0,0 +1,258 @@
...
+
+    if total_base_pairs / number_of_reads > 75:
+        job_context["kmer_size"] = "31"
+    else:
+        job_context["kmer_size"] = "23"
+
+    return job_context
+
+
+def _download_index(job_context: Dict) -> Dict:
```
rzelayafavila: A docstring for this method would be helpful

workers/data_refinery_workers/processors/salmon.py
```diff
@@ -0,0 +1,258 @@
...
+    batch = job_context["batches"][0]
+    try:
+        index_batches = Batch.objects.filter(source_type=Downloaders.TRANSCRIPTOME_INDEX.value,
+                                             organism_id=batch.organism_id,
+                                             status=BatchStatuses.PROCESSED.value).all()
+        batch_ids = [batch.id for batch in index_batches]
+
+        batch_key_value = BatchKeyValue.objects.filter(batch_id__in=batch_ids,
```
rzelayafavila: Hmm, I'm a bit confused - why are we filtering for Batches with ID in batch_ids but then just getting the first element of that list?

workers/data_refinery_workers/processors/salmon.py
```diff
@@ -0,0 +1,258 @@
...
+    finally:
+        # Remove temp dir for the index batch since we've extracted it to
+        # the temp dir for this job.
+        index_file.remove_temp_directory(job_context["job_dir_prefix"])
+
+    return job_context
+
+
+def _zip_and_upload(job_context: Dict) -> Dict:
+    # If there are paired reads... the file name will be based off of
```
rzelayafavila: It would be good to have these comments in the Python docstring format `""" """`, especially in case we ever want to export the documentation for all of the functions (or access it from `__doc__` or `help()`).

workers/data_refinery_workers/processors/salmon.py
```diff
@@ -0,0 +1,258 @@
...
+    # the first file's name, but not the second.
+    # Note that this is a workaround for not having separate File
+    # objects to represent processed files. Once that is fixed this
+    # should simply use the processed file instead of one of the input
+    # files.
```
rzelayafavila: Also, I think this should say what you are adding to the `job_context` (like you do well in the other docstrings) before you return it

common/data_refinery_common/models/batches.py
```diff
@@ -215,6 +215,7 @@ def download_raw_file(self, dir_name: str=None) -> None:
         Depending on the value of the USE_S3 environment variable this
         will either move the batch's raw file from the RAW_PREFIX directory
         or from S3.
+        Returns the path the file was downloaded to.
```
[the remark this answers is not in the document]
kurtwheeler: Good catch!

workers/data_refinery_workers/processors/management/commands/queue_processor.py
```diff
@@ -40,46 +41,37 @@ def run_sra_processor(self):
         batch = Batch(
             survey_job=survey_job,
             source_type="SRA",
-            size_in_bytes=2214725074,
-            raw_format="fastq",
-            processed_format="sf",
```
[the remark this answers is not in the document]
kurtwheeler: Correct!

workers/data_refinery_workers/processors/salmon.py
kurtwheeler: There should always be exactly one or two files for salmon jobs, but I will add a check to make sure this is the case.

workers/data_refinery_workers/processors/salmon.py
kurtwheeler: Good call.

workers/data_refinery_workers/processors/salmon.py
kurtwheeler: The BatchKeyValue database table always stores strings because it's `value` column is defined as a varchar field.

workers/data_refinery_workers/processors/salmon.py
kurtwheeler: Very true, good call.

workers/data_refinery_workers/processors/salmon.py
kurtwheeler: As discussed in person, it's because I should do this with a join instead of two separate queries. Thanks for pointing me to `select_related`!

workers/data_refinery_workers/processors/salmon.py
kurtwheeler: I agree I should add a docstring here, but these comments are not about the function, they're about the 4 lines of code they immediately precede.

kurtwheeler:
Ok @rzelayafavila, I believe I have addressed your questions and concerns. Take a look and lemme know what you think.

workers/data_refinery_workers/processors/salmon.py
```diff
@@ -0,0 +1,285 @@
...
+    """
+    total_base_pairs = 0
+    number_of_reads = 0
+    counter = 1
+    with gzip.open(job_context["input_file_path"], "rt") as input_file:
+        for line in input_file:
+            # In the FASTQ file format, there are 4 lines for each
+            # read. Three of these contain metadata about the
+            # read. The string representing the read itself is found
+            # on the second line of each quartet.
```
rzelayafavila: :+1:

workers/data_refinery_workers/processors/salmon.py
rzelayafavila: Ah, I see

workers/data_refinery_workers/processors/salmon.py
```diff
@@ -0,0 +1,285 @@
...
+        index_file.remove_temp_directory(job_context["job_dir_prefix"])
+
+    return job_context
+
+
+def _zip_and_upload(job_context: Dict) -> Dict:
+    """Zips the directory output by Salmon into a single file and uploads it.
+
+    Adds the 'success' key to job_context because this function is the
+    last in the job.
```
rzelayafavila: :+1:

workers/data_refinery_workers/processors/test_salmon.py
```diff
@@ -0,0 +1,276 @@
...
+                                       "batches": [batch],
+                                       "index_directory": "missing",
+                                       "input_file_path": input_file_path_1,
+                                       "input_file_path_2": input_file_path_2,
+                                       "output_directory": "blah"})
+
+        # The function being tested.
+        job_context = salmon._run_salmon(job_context)
+
+        self.maxDiff = None
```
rzelayafavila: Ah, what is this `self.maxDiff` being used for? I don't see it anywhere else.

workers/data_refinery_workers/processors/salmon.py
```diff
@@ -0,0 +1,285 @@
...
+from data_refinery_common.models import Batch, BatchStatuses, BatchKeyValue
+from data_refinery_common.logging import get_and_configure_logger
+import subprocess
+
+
+logger = get_and_configure_logger(__name__)
+JOB_DIR_PREFIX = "processor_job_"
+
+
+def _set_job_prefix(job_context: Dict) -> str:
```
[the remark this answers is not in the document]
kurtwheeler: Yes, good catch!

workers/data_refinery_workers/processors/test_salmon.py
kurtwheeler: Good catch, I think I was using it at some point and then forgot to remove it once I no longer was.

kurtwheeler:
@rzelayafavila I think we're all good here!

workers/data_refinery_workers/processors/salmon.py
```diff
@@ -15,7 +15,7 @@
 JOB_DIR_PREFIX = "processor_job_"
 
 
-def _set_job_prefix(job_context: Dict) -> str:
+def _set_job_prefix(job_context: Dict) -> Dict:
```
rzelayafavila: :+1:

workers/data_refinery_workers/processors/test_salmon.py
```diff
@@ -246,7 +246,6 @@ def test_run_salmon_failure(self):
         # The function being tested.
         job_context = salmon._run_salmon(job_context)
 
-        self.maxDiff = None
```
rzelayafavila: :+1:

Issue #52, opened 2017-09-19
Create a script that can be used to run a Nomad job.


kurtwheeler:

kurtwheeler:
This is a duplicate of #51

Issue #50, opened 2017-09-19
Rip out celery from project


kurtwheeler:

Miserlou:
This happened! 🐰

Issue #48, opened 2017-09-19
Utilize the interface in place of sending celery messages


kurtwheeler:

Miserlou:
We don't send Celery messages any more!

PR #81, opened 2018-01-12
Nomad interface

kurtwheeler:
This PR replaces Celery with Nomad for sending messages. Nomad jobs are run within a container which is only alive for the duration of that job. Therefore I have set it up so that the container is expecting the job parameters, which it then passes into a django management command which actually runs the job.

I have not yet figured out how to inject environment variables into Nomad jobs, so therefore there are some hardcoded IP values and some REDACTED values for secrets.


kurtwheeler:
This comment exists and is formatted the way it is to link it with issues
Closes #47
Closes #48
Closes #49
Closes #50
Closes  #52

workers/data_refinery_workers/processors/management/commands/queue_processor.py
```diff
@@ -79,9 +81,12 @@ def run_sra_processor(self):
         )
         batch2.save()
 
+        # This is the old way of doing things. Should be updated once
+        # I'm sure it's working.
```
Miserlou: Are we sure it's working? :)

workers/data_refinery_workers/processors/management/commands/queue_processor.py
```diff
@@ -79,9 +81,12 @@ def run_sra_processor(self):
...
         processor_job = ProcessorJob.create_job_and_relationships(batches=[batch, batch2])
-        processor_task = PROCESSOR_PIPELINE_LOOKUP[batch.pipeline_required]
-        app.send_task(processor_task, args=[processor_job.id])
+        # processor_task = PROCESSOR_PIPELINE_LOOKUP[batch.pipeline_required]
+        processor_task = "temp cause it'll break if I use PROCESSOR_PIPELINE_LOOKUP"
+        send_job(processor_task, processor_job.id)
         logger.info("Processor Job queued.", processor_job=processor_job.id)
 
     def run_trasnscriptome_processor(self):
```
Miserlou: Docstring.

workers/data_refinery_workers/processors/management/commands/run_processor_job.py
```diff
@@ -0,0 +1,48 @@
...
+            help=("The processor job's name. Must be enumerated in job_lookup."))
+        parser.add_argument(
+            "--job-id",
+            type=int,
+            help=("The processor job's ID."))
+
+    def handle(self, *args, **options):
+        if options["job_id"] is None:
+            logger.error("You must specify a job ID.")
+            return 1
```
Miserlou: Why 0/1 rather than T/F?

Also, it looks like this file is copy pasted a few times maybe, could it be made generic?

workers/data_refinery_workers/processors/transcriptome_index.py
```diff
@@ -235,8 +234,7 @@ def _zip_index(job_context: Dict) -> Dict:
     return job_context
 
 
-@shared_task
-def build_index(job_id: int) -> None:
+def build_transcriptome_index(job_id: int) -> None:
```
Miserlou: Docstring

workers/data_refinery_workers/processors/utils.py
```diff
@@ -49,13 +49,14 @@ def end_job(job_context: Dict):
     job.end_time = timezone.now()
     job.save()
 
+    batches = job_context["batches"] if "batches" in job_context else None
+
     # Clean up temp directory.
-    if len(job_context["batches"]) != 0:
+    if batches is not None and len(batches) > 0 and len(batches[0].files) > 0:
```
Miserlou: Not checking if batches[0] has a files key

processor.nomad
```diff
@@ -0,0 +1,75 @@
...
+        S3_BUCKET_NAME = "data-refinery"
+        LOCAL_ROOT_DIR = "/home/user/data_store"
+
+        RAW_PREFIX = "raw"
+        TEMP_PREFIX = "temp"
+        PROCESSED_PREFIX = "processed"
+      }
+
+      resources {
+        cpu = 500
```
dongbohu: What does "500" mean here?

processor.nomad
```diff
@@ -0,0 +1,75 @@
...
+        memory = 4048
```
dongbohu: I suppose the unit here is MB?

workers/data_refinery_workers/processors/utils.py
kurtwheeler: Do you think that's really necessary? This is a list of `Batch` objects which each have a `files` attribute, it's not looking up a key.

run_nomad.sh
```diff
@@ -0,0 +1,33 @@
...
+
+# Start the nomad in development locally
+nomad agent -bind $HOST_IP \
+      -data-dir /home/kurt/Development/data_refinery/nomad_dir \
+      -dev \
+      > nomad.logs &
+
+docker build -t wkurt/nomad-test:second -f workers/Dockerfile .
+
+check_nomad_status () {
```
dongbohu: I think some comments for this function will be helpful.

run_nomad.sh
```diff
@@ -0,0 +1,33 @@
...
+    echo $(curl --write-out %{http_code} \
+                  --silent \
+                  --output /dev/null \
+                  http://$HOST_IP:4646/v1/status/leader)
+}
+
+# Wait for Nomad to get started.
+nomad_status=$(check_nomad_status)
+while [ $nomad_status != "200" ]; do
+    sleep 0.5
```
dongbohu: Do you know how portable this `sleep` command is? I know earlier Unix/Linux implementations only accept an integer. If it is not portable, then `sleep 1` is probably better.

processor.nomad
Miserlou: Good catch, and leaves me wanting more overall documentation in this file.

workers/data_refinery_workers/downloaders/management/commands/run_downloader_job.py
```diff
@@ -0,0 +1,47 @@
+from django.core.management.base import BaseCommand
```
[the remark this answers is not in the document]
kurtwheeler: I wasn't using Django because of Celery, it can work without Django. I used Django primarily because it was the recommended ORM.

workers/data_refinery_workers/downloaders/management/commands/run_downloader_job.py
```diff
@@ -0,0 +1,47 @@
...
+            help=("The downloader job's name. Must be enumerated in job_lookup."))
+        parser.add_argument(
+            "--job-id",
+            type=int,
+            help=("The downloader job's ID."))
+
+    def handle(self, *args, **options):
+        if options["job_id"] is None:
+            logger.error("You must specify a job ID.")
+            return 1
```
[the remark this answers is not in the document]
kurtwheeler: Management commands will be executed from the shell, so I see these more as bash exit codes than I do as seeing this return a True/False value. In fact if a function is supposed to do something and returns True I would expect it to have succeeded, whereas in this case it indicates that it failed.

workers/data_refinery_workers/processors/management/commands/run_processor_job.py
kurtwheeler: Same as above for the 0/1 things. I'm not actually entirely sure what you mean by this file being copied a few times, but except for checking that a `job_id` parameter is passed in I don't think too much functionality is repeated.

Miserlou:
The OSX HOST_IP detector is broken and returns a mangled IPv6 address, preventing the tests from running on OSX.

It can be replaced with (via S/O):

```bash
ifconfig | grep "inet " | grep -v 127.0.0.1 | cut -d\  -f2
```

Miserlou:
Ah, it looks like this is used in about 10 different files copy and pasted, so I suggest we instead have a `common.sh` and then use `source` to import the get IP function.

kurtwheeler:
Ok @Miserlou and @dongbohu, I thinkl this should be good to go at this point!

PR #116, opened 2018-02-09
Ensure that permissions on docker volumes gets intialized correctly.

kurtwheeler:
Rich and I ran into an issue where nomad containers could not write to their volumes. This will prevent this from happening when the project is setup on future boxes, however it will not retroactively fix the permissions for existing volumes.

This will also redirect stderr from Nomad into the nomad.logs file so the terminal doesn't output stuff randomly.


kurtwheeler:
Note that to fix the issue on existing volumes `chmod -R a+rwX workers/volume` can be run.

PR #143, opened 2018-02-27
Add variables, and use the <resource>-<user>-<stage> pattern. Commented…

Miserlou:
…, but needs proper documentation

This changes hardcoded resource names to use the `<resource>-<user>-<stage>` pattern. This means that all of our resources are at-a-glance recognizable, and we won't have any collisions from multiple developers working on the `dev` stage simultaneously.

This can be used by CI/CI, etc., by setting the `TF_VAR_user` and `TF_VAR_stage` variables. Developers can also instead modify the `variables.tf` file.

## Types of changes

_Put an `x` in the boxes that apply_

- [x] New feature (non-breaking change which adds functionality)

## Functional tests

n/a

## Checklist

### DIFF
--- a/infrastructure/variables.tf
+++ b/infrastructure/variables.tf
@@ -9,3 +9,4 @@ variable "user" {
 variable "stage" {
   default = "dev"
 }
+
--- a/infrastructure/variables.tf
+++ b/infrastructure/variables.tf
@@ -1,0 +1,11 @@
+# Variables
+# Only lowercase alphanumeric characters and hyphens allowed!
+# In CI, this can be set with TF_VAR_user, TF_VAR_stage, etc.
+
+variable "user" {
+  default = "my-user-name"
+}
+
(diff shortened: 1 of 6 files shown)


infrastructure/variables.tf
```diff
@@ -0,0 +1,11 @@
...
+# Only lowercase alphanumeric characters and hyphens allowed!
+# In CI, this can be set with TF_VAR_user, TF_VAR_stage, etc.
+
+variable "user" {
+  default = "my-user-name"
+}
+
+variable "stage" {
+  default = "dev"
+}
```
dongbohu: You may want to add a new line char at the end of this line.

kurtwheeler:
This LGTM. Once someone has set the appropriate variables they will just be able to run `terraform init && terraform apply` and then they'll have their own state file managing their own resources.

Issue #146, opened 2018-02-28
Implement System-Wide Centralized Logging


Miserlou:
We have numerous processes which are spitting out logs. Currently, these are all just spitting out to `stdout` or to plain text log files which live on their hosts.

In our initial executions at scale in the cloud, I am expecting that we will encounter many unforseen errors. Without a proper logging system (with associated alerting), we will not be able to investigate and fix these errors.

This means that we need to:

- [x] Create CloudWatchLog resources in our `terraform` infrastructure
- [x] Install, configure and enable `awslogs` ([Amazon's version](https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/QuickStartEC2Instance.html)) on our Nomad machines
- [x] Install and configure [watchtower](https://github.com/kislyuk/watchtower) (or similar) in our `foreman` and `worker` Django applications.
- [x] Possibly install configure and enable `awslogs` inside our worker Docker machines if we want Salmon/SCAN outputs to be centrally collected as well. [Possibly related: [CloudWatch and Moby](https://github.com/moby/moby/blob/master/daemon/logger/awslogs/cloudwatchlogs.go)] 
- [x] Document the correct usage of log watching/tailing tools such as (the _other_) [awslogs](https://github.com/jorgebastida/awslogs) for our system.

PR #142, opened 2018-02-27
Terraform docs

kurtwheeler:
## Issue Number

This is progress towards https://github.com/AlexsLemonade/refinebio/issues/57, but does not quite close it yet.

## Purpose/Implementation Notes

This PR renames the terraform directory to infrastructure, reorganizes the resources within the files, and adds some configuration and resources for production Nomad instances.

## Types of changes

What types of changes does your code introduce?
_Put an `x` in the boxes that apply_

- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Functional tests

I have run `terraform apply` within the Greenelab account to test my changes. The nomad server instance starts up properly and even registers the existing Nomad jobs. However the client does not know the server's IP address yet so it doesn't work yet.

## Checklist

- [x] Lint and unit tests pass locally with my changes
- [x] I have added necessary documentation (if appropriate)

PR #147, opened 2018-02-28
Adds logging.tf, creates new log groups and log streams, also standar…

Miserlou:
…dizes using _ for TF resource names and - for AWS names

## Issue Number

Related to: https://github.com/AlexsLemonade/refinebio/issues/146

Creates CloudWatch Log Groups and Log Streams for our resources. 

Also makes sure that all `name`s use "-" rather than "_" in `data-refinery-` so this is standard.

## Question

This is also a `conf/awslogs.conf` file, but I don't see it being used anywhere. Why is it there? Should it be kept and improved and documented, or can it be removed?

## Types of changes

_Put an `x` in the boxes that apply_

- [x] New feature (non-breaking change which adds functionality)


kurtwheeler:
Ah that `conf/awslogs.conf` file was used for logging from ECS, but #142 removed the references to him but I forgot to remove him as well. For more context you can take a look at [this](http://docs.aws.amazon.com/AmazonECS/latest/developerguide/using_cloudwatch_logs.html).

Issue #63, opened 2017-09-19
Startup scripts for the Nomad clients


kurtwheeler:

Miserlou:
Was this taken care of by the recent PR? Or is this more of a devops problem?

kurtwheeler:
More of a devops problem. This'll require work in the terraform repository.

kurtwheeler:
Relates to #56

PR #149, opened 2018-02-28
Infrastructure code to spin up Nomad Server and Client in AWS.

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/63
https://github.com/AlexsLemonade/refinebio/issues/56

Note that this is only part of the way to #56 because it only has the configuration code for one Nomad Server. I will do the additional two servers in a subsequent PR.

This PR successfully spins up a Nomad Server on one AWS instance, registers our Nomad Jobs to it, and then starts a Nomad Client on another AWS instance which then registers itself with the Nomad cluster.

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Functional tests

I have run `terraform apply` and witnessed it successfully complete. I then ssh'd onto the Nomad Client instance and made sure there was a line in `/tmp/nomad_client.log` like this:
```
2018/02/28 23:05:57.488116 [INFO] client: node registration complete
```
which indicates that the client node was able to register itself with the Nomad Server.
I then ssh'd onto the Nomad Server instance and ran:
```
$ nomad node-status
ID        DC   Name            Class   Drain  Status
b48e36af  dc1  ip-10-0-191-57  <none>  false  ready
```
which shows that the server is aware of the client instance. Finally, I ran

```bash
$ nomad status
ID          Type                 Priority  Status   Submit Date
DOWNLOADER  batch/parameterized  50        running  02/28/18 23:18:37 UTC
PROCESSOR   batch/parameterized  50        running  02/28/18 23:18:37 UTC
```
which shows that the jobs were successfully registered.

## Checklist


infrastructure/instances.tf
```diff
@@ -16,21 +16,86 @@ data "aws_ami" "ubuntu" {
...
 }
 
-resource "aws_instance" "data_refinery_worker_1" {
+# We need to transfer the Nomad Job Specifications onto the instance
+# so we can register them. However we use these job specifications to
+# store environment variables, so they are on Github as templates. To
+# format them with the production environment variables, we need to
+# run the following command. By doing so with a `null_resource` we're
+# able to avoid having a prerequisite step to running `terraform
+# apply`.
```
Miserlou: Good comment, thank you

infrastructure/instances.tf
```diff
@@ -16,21 +16,86 @@ data "aws_ami" "ubuntu" {
...
+}
+
+# This instance will run the Nomad Server. We will want three of
+# these, but one server needs to start first so that the others can be
+# aware of its IP address to join it.
+# Those other servers do not yet exist in the current configuration.
+# TODO: create those other servers!
+resource "aws_instance" "nomad-server-1" {
   ami = "${data.aws_ami.ubuntu.id}"
   instance_type = "t2.xlarge"
```
Miserlou: Is this instance type overkill for what it is used for? Still not super expensive but I don't know what the requirements are for these machines, it shouldn't be too huge.

infrastructure/instances.tf
```diff
@@ -43,21 +108,48 @@ resource "aws_instance" "data_refinery_worker_1" {
   }
```
Miserlou: It might be useful to add the IP address of this as an `output`.

infrastructure/instances.tf
```diff
@@ -43,21 +108,48 @@ resource "aws_instance" "data_refinery_worker_1" {
...
+  user_data = "${data.template_file.nomad-client-script-smusher.rendered}"
   key_name = "${aws_key_pair.data_refinery.key_name}"
 
   tags = {
-    Name = "data-refinery-2-${var.user}-${var.stage}"
+    Name = "nomad-client-1-${var.user}-${var.stage}"
   }
 
+  # I think these are the defaults provided in terraform examples.
+  # They should be removed or revisited.
```
Miserlou: Agreed, this is a point where AWS gets you.

infrastructure/nomad-configuration/nomad-client-instance-user-data.tpl.sh
```diff
@@ -0,0 +1,46 @@
...
+${install_nomad_script}
+EOF
+
+# Create the Nomad Client configuration.
+cat <<"EOF" > client.hcl
+${nomad_client_config}
+EOF
+
+
+# Create a directory for docker to use as a volume.
```
Miserlou: Why is this /tmp? Does the volume need to persist?

infrastructure/nomad-configuration/nomad-server-instance-user-data.tpl.sh
```diff
@@ -0,0 +1,65 @@
...
+
+# Start the Nomad agent in server mode.
+nohup nomad agent -config server.hcl > /tmp/nomad_server.log &
+
+# Give the Nomad server time to start up.
+sleep 30
+
+# Determine the IP address of this machine so we know where the Nomad
+# server's HTTP API is addressed.
+IP_ADDRESS=$(ip route get 8.8.8.8 | awk '{print $NF; exit}')
```
Miserlou: If we used an `output`, we could replace this. It'll also be usefull if we want to set up a proxy route to this or the UI.

infrastructure/security.tf
```diff
@@ -25,6 +30,53 @@ resource "aws_security_group_rule" "data_refinery_worker_custom" {
...
+  from_port = 4648
+  to_port = 4648
+  protocol = "udp"
+  self = true
+  security_group_id = "${aws_security_group.data_refinery_worker.id}"
+}
+
+# Allow SSH connections to Nomad instances. This is pretty much the
+# only way to debug issues until we get the Nomad UI up and running.
+# THIS DEFINITELY NEEDS TO BE REMOVED LONG TERM!!!!!!!!!!
```
Miserlou: Add `# XXX` so we can `ag` for it before prod deploy!

infrastructure/nomad-configuration/nomad-client-instance-user-data.tpl.sh
kurtwheeler: Hmm, good point. Long term it won't matter because we'll be using EFS, but I may as well just create this directory in `/home/ubuntu` anyway.

infrastructure/instances.tf
kurtwheeler: Oh yeah that's a really good point. I think this will be an area where we'll need to do tuning. [Nomad says](https://www.nomadproject.io/guides/cluster/requirements.html):

> Nomad servers may need to be run on large machine instances. We suggest having between 4-8+ cores, 16-32 GB+ of memory, 40-80 GB+ of disk and significant network bandwidth. The core count and network recommendations are to ensure high throughput as Nomad heavily relies on network communication and as the Servers are managing all the nodes in the region and performing scheduling. The memory and disk requirements are due to the fact that Nomad stores all state in memory and will store two snapshots of this data onto disk. Thus disk should be at least 2 times the memory available to the server when deploying a high load cluster.

but that's probably for a rather large cluster. I bet we can get away with a small or medium.

infrastructure/instances.tf
kurtwheeler: :+1:

infrastructure/instances.tf
Miserlou: Oh damn, that's more than I expected! Okay, let's leave it as is and then see how they perform at relative scale.

infrastructure/nomad-configuration/nomad-server-instance-user-data.tpl.sh
kurtwheeler: That won't work because the `output` won't be generated until after this script is run. Still might be good to have it an an output though.

infrastructure/security.tf
kurtwheeler: :+1:

infrastructure/nomad-configuration/nomad-server-instance-user-data.tpl.sh
Miserlou: Roger

PR #151, opened 2018-03-02
Kurtwheeler/more nomad servers

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/56
https://github.com/AlexsLemonade/refinebio/issues/57
https://github.com/AlexsLemonade/refinebio/issues/63

We need three Nomad servers to [prevent data loss](https://www.nomadproject.io/docs/internals/consensus.html#deployment-table). In https://github.com/AlexsLemonade/refinebio/pull/149 I added terraform configuration for one server instance and one client instance, along with startup scripts for both of them.

In this PR I add terraform configuration for the remaining two Nomad Servers along with a special startup script for the lead server which needs to be a special case because the other servers need to know its IP address so they can talk to it.

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Functional tests

I ran `terraform apply` and witnessed a successful deploy. I then ssh'd onto the lead Nomad Server and ran:

```bash
ubuntu@ip-10-0-54-44:~$ nomad node-status
ID        DC   Name            Class   Drain  Status
738a4041  dc1  ip-10-0-145-46  <none>  false  ready
```
Which shows the client node was able to register itself to the cluster.

```bash
ubuntu@ip-10-0-54-44:~$ nomad server-members
Name                   Address      Port  Status  Leader  Protocol  Build  Datacenter  Region
ip-10-0-3-4.global     10.0.3.4     4648  alive   false   2         0.7.1  dc1         global
ip-10-0-54-44.global   10.0.54.44   4648  alive   true    2         0.7.1  dc1         global
ip-10-0-59-224.global  10.0.59.224  4648  alive   false   2         0.7.1  dc1         global
```

Which shows that there's three servers with one of them being the leader.

```bash
ubuntu@ip-10-0-54-44:~$ nomad status
ID          Type                 Priority  Status   Submit Date
DOWNLOADER  batch/parameterized  50        running  03/02/18 15:58:10 UTC
PROCESSOR   batch/parameterized  50        running  03/02/18 15:58:10 UTC
```

Which shows the jobs were registered successfully.

## Checklist


Miserlou:
I think this is all good, but unfortunately there's now a merge conflict with the logging stuff that just landed. I think it might make sense to actually create new log streams for each of these new instances as well to make it easier to monitor debug - but, this will mean one more variable passed to the templated script, so `log-stream-nomad-server-${user}-${stage}` becomes `log-stream-nomad-server-${user}-${stage}-${membership-number}` or something like that.

Miserlou:
An alternative would be that we blow away all of the named log streams entirely and let everything create its own stream with a semi-deterministic pattern, ex. `log-stream-nomad-server-${user}-${stage}-${hostname}`.

infrastructure/instances.tf
```diff
@@ -111,29 +110,137 @@ resource "aws_instance" "nomad-server-1" {
...
+    region = "${var.region}"
+  }
+}
+
+# This is another Nomad Server instance since it's recommended to have
+# at least three to avoid data loss.
+resource "aws_instance" "nomad_server_2" {
+  ami = "${data.aws_ami.ubuntu.id}"
+  instance_type = "t2.small"
+  availability_zone = "us-east-1a"
```
Miserlou: change to use ${var.region}

infrastructure/instances.tf
```diff
@@ -111,29 +110,137 @@ resource "aws_instance" "nomad-server-1" {
...
+output "nomad_server_2_ip" {
+  value = "${aws_instance.nomad_server_2.public_ip}"
+}
+
+# This is another Nomad Server instance since it's recommended to have
+# at least three to avoid data loss.
+resource "aws_instance" "nomad_server_3" {
+  ami = "${data.aws_ami.ubuntu.id}"
+  instance_type = "t2.small"
+  availability_zone = "us-east-1a"
```
Miserlou: change to use ${var.region}

infrastructure/instances.tf
```diff
@@ -186,7 +186,7 @@ output "nomad_server_2_ip" {
 resource "aws_instance" "nomad_server_3" {
   ami = "${data.aws_ami.ubuntu.id}"
   instance_type = "t2.small"
-  availability_zone = "us-east-1a"
+  availability_zone = "${var.region}a"
```
Miserlou: Should we put one in a and one in b?

PR #152, opened 2018-03-02
Add Log Consumption To Readme // Add EFS Mounting

Miserlou:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/146

This documents the usage of the `awslogs` tool to consume Refine.bio logs.

UPDATE: This also adds EFS creation and mounting.


infrastructure/nomad-configuration/client-instance-user-data.tpl.sh
```diff
@@ -13,6 +13,15 @@
...
 # but those run after this script runs.
 
+# Mount the EFS.
+export DEBIAN_FRONTEND=noninteractive
+apt-get update -y
+apt-get upgrade -y
+apt-get install --yes nfs-common
+mkdir -p /var/efs/
+chown ubuntu:ubuntu /var/efs/
+echo "${file_system_id}.efs.${region}.amazonaws.com:/ /var/efs/ nfs4 nfsvers=4.1,rsize=1048576,wsize=1048576,hard,timeo=600,retrans=2 0 0" >> /etc/fstab
```
kurtwheeler: Can you document what some of these options are doing?

README.md
```diff
@@ -1,8 +1,20 @@
...
+Refine.bio currently has four sub-projects contained within this repo:
+- [common](./common) Contains code needed by both `foreman` and `workers`.
+- [foreman](./foreman) Discovers data to download/process and manages jobs.
+- [workers](./workers) Runs Downloader and Processor jobs.
+- [infrasctructure](./infrastructure) Manages infrastructure for Refine.bio.
+
 <!-- START doctoc generated TOC please keep comment here to allow auto update -->
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
-**Table of Contents**
+## Table of Contents
```
kurtwheeler: Was this generated by doctoc?

infrastructure/instances.tf
```diff
@@ -285,17 +265,15 @@ output "nomad_client_ip" {
...
 
-variable "database_password" {}
-
 resource "aws_db_instance" "postgres_db" {
   identifier = "data-refinery-${var.user}-${var.stage}"
   allocated_storage = 100
   storage_type = "gp2"
   engine = "postgres"
-  engine_version = "9.5.4"
+  engine_version = "9.5.10"
```
kurtwheeler: I made this change because one `terraform apply` said that it was going to change the version from `9.5.10 -> 9.5.4` so I think RDS upgraded postgres for us, so we might as well explicitly get on the latest version anyway.

workers/processor.nomad.tpl
```diff
@@ -44,6 +44,8 @@ job "PROCESSOR" {
         RAW_PREFIX = "${{RAW_PREFIX}}"
         TEMP_PREFIX = "${{RAW_PREFIX}}"
         PROCESSED_PREFIX = "${{PROCESSED_PREFIX}}"
+
+        NOMAD_HOST = "${{NOMAD_HOST}}"
```
kurtwheeler: This is probably unnecessary right? The downloaders template file doesn't have it and it's able to communicate with Nomad.

kurtwheeler:
Ok @Miserlou, I've changed the final things I wanted to change about this. Take another look and lemme know if there's anything else you think needs to be changed before we merge this in.

The only thing I'm still not sure is solid is all those `XXX`'s around the database code. Do we want to move those database commands out of the client init script and into deploy.sh?

kurtwheeler:
This addresses https://github.com/AlexsLemonade/refinebio/issues/153

foreman/surveyor.nomad.tpl
```diff
@@ -0,0 +1,80 @@
...
+
+        DATABASE_NAME = "${{DATABASE_NAME}}"
+        DATABASE_USER = "${{DATABASE_USER}}"
+        DATABASE_PASSWORD = "${{DATABASE_PASSWORD}}"
+        DATABASE_HOST = "${{DATABASE_HOST}}"
+        DATABASE_PORT = "${{DATABASE_PORT}}"
+        DATABASE_TIMEOUT = "${{DATABASE_TIMEOUT}}"
+
+        RUNNING_IN_CLOUD = "False"
+        WATCHTOWER="False"
```
Miserlou: This can be removed, we aren't going to use Watchtower for now.

foreman/surveyor.nomad.tpl
```diff
@@ -0,0 +1,80 @@
...
+        PROCESSED_PREFIX = "${{PROCESSED_PREFIX}}"
+        NOMAD_HOST = "${{NOMAD_HOST}}"
+      }
+
+      # The resources the job will require.
+      resources {
+        # CPU is in AWS's CPU units.
+        cpu = 500
+        # Memory is in MB of RAM.
+        memory = 2048
```
Miserlou: I think this value is probably too high.

foreman/surveyor.nomad.tpl
```diff
@@ -0,0 +1,80 @@
...
+        ]
+        ${{EXTRA_HOSTS}}
+        volumes = ["${{VOLUME_DIR}}:/home/user/data_store"]
+
+        logging {
+          type = "awslogs"
+          config {
+            awslogs-region = "${{REGION}}",
+            awslogs-group = "data-refinery-log-group-${{USER}}-${{STAGE}}",
+            awslogs-stream = "log-stream-nomad-docker-surveyor-${{USER}}-${{STAGE}}"
```
Miserlou: We're going to want to figure out how to include the instance or machine id here.

format_nomad_with_env.sh
```diff
@@ -0,0 +1,111 @@
+#!/bin/bash
```
Miserlou: This feel needs documentation about what it actually does.

format_nomad_with_env.sh
```diff
@@ -0,0 +1,111 @@
...
+    export EXTRA_HOSTS="
+        extra_hosts = [\"database:$DB_HOST_IP\",
+                       \"nomad:$NOMAD_HOST_IP\"]
+"
+else
+    export EXTRA_HOSTS=""
+fi
+
+
+while read line; do
```
Miserlou: All bash should be explained.

format_nomad_with_env.sh
```diff
@@ -0,0 +1,111 @@
...
+# template environment variables into the job specifications. Until
+# this issue is resolved, we are using perl to accomplish this. The
+# issue can be found here:
+# https://github.com/hashicorp/nomad/issues/1185
+
+if [[ ! -z $output_dir && ! -d "$output_dir" ]]; then
+    mkdir $output_dir
+fi
+
+# Perl magic found here: https://stackoverflow.com/a/2916159/6095378
```
Miserlou: Explain what this "perl magic" actually does. Ideally, remove the need for perl entirely.

infrastructure/deploy.sh
```diff
@@ -0,0 +1,117 @@
+#!/bin/bash -e
```
Miserlou: Document what this script does.

infrastructure/deploy.sh
```diff
@@ -0,0 +1,117 @@
...
+
+# This script must be run from /infrastructure!
+
+# Make our IP address known to terraform.
+source ../common.sh
+export TF_VAR_host_ip=`wget -qO- http://ipecho.net/plain ; echo`
```
Miserlou: This is an SPoF if we're using this in prod. It might be possible to do with with system DNS tools rather than a third party web service.

infrastructure/deploy.sh
```diff
@@ -0,0 +1,117 @@
...
+                  --output /dev/null \
+                  $NOMAD_ADDR/v1/status/leader)
+}
+
+
+# Wait for Nomad to get started in case the server just went up for
+# the first time.
+echo "Confirming Nomad cluster.."
+start_time=$(date +%s)
+diff=0
```
Miserlou: We're checking for the existence of the cluster, but we should be checking for the health of the _client_ or there could be a race condition on first deploy.

infrastructure/deploy.sh
```diff
@@ -0,0 +1,117 @@
...
+fi
+
+# Make sure that prod_env is empty since we are only appending to it.
+# prod_env is a temporary file we use to pass environment variables to
+# `docker run` commands when running migrations.
+rm -f prod_env
+
+# We have terraform output environment variables via a single output
+# variable, which we then read in as json using the command line tool
+# `jq`, so that we can use them via bash.
```
Miserlou: Dependency on jq isn't documented.

infrastructure/deploy.sh
```diff
@@ -0,0 +1,117 @@
...
+    env_var_assignment=$(echo $row | jq -r ".name")=$(echo $row | jq -r ".value")
+    export $env_var_assignment
+    echo $env_var_assignment >> prod_env
+done
+
+# Get an image to run the migrations with.
+docker pull $FOREMAN_DOCKER_IMAGE
+
+# Migrate auth.
+docker run \
```
Miserlou: Is this still necessary?

infrastructure/deploy.sh
```diff
@@ -0,0 +1,117 @@
...
+echo "Registering new job specifications.."
+nomad_job_specs=nomad-job-specs/*
+for nomad_job_spec in $nomad_job_specs; do
+    echo "registering $nomad_job_spec"
+    nomad run $nomad_job_spec
+done
+
+# Remove the ingress config so the next `terraform apply` will remove
+# access for Circle.
+echo "Removing ingress.."
```
Miserlou: We may want ingress allocation/deallocation to be separate steps of the production deployment pipeline - I am worried about the ingress being left open by a mid-script error.

infrastructure/deploy.sh
```diff
@@ -0,0 +1,117 @@
...
+# specs.
+mkdir -p nomad-job-specs
+../format_nomad_with_env.sh -p workers -e prod -o $(pwd)/nomad-job-specs/
+../format_nomad_with_env.sh -p foreman -e prod -o $(pwd)/nomad-job-specs/
+
+# Re-register Nomad jobs.
+echo "Registering new job specifications.."
+nomad_job_specs=nomad-job-specs/*
+for nomad_job_spec in $nomad_job_specs; do
+    echo "registering $nomad_job_spec"
```
Miserlou: r -> R

infrastructure/deploy/ci_ingress.tf
```diff
@@ -0,0 +1,18 @@
+resource "aws_security_group_rule" "data_refinery_ci_nomad" {
+
+  type = "ingress"
+  from_port = 4646
+  to_port = 4646
+  protocol = "tcp"
+  cidr_blocks = ["${var.host_ip}/32"]
```
Miserlou: This the whole /32 - I think this can be a /1. Is there a /1?

infrastructure/deploy/ci_ingress.tf
```diff
@@ -0,0 +1,18 @@
...
+  security_group_id = "${aws_security_group.data_refinery_worker.id}"
+}
+
+resource "aws_security_group_rule" "data_refinery_ci_postgres" {
+  type = "ingress"
+  from_port = 5432
+  to_port = 5432
+  protocol = "tcp"
+  cidr_blocks = ["${var.host_ip}/32"]
```
Miserlou: Same as above.

infrastructure/instances.tf
```diff
@@ -186,10 +161,10 @@ output "nomad_server_2_ip" {
 resource "aws_instance" "nomad_server_3" {
   ami = "${data.aws_ami.ubuntu.id}"
   instance_type = "t2.small"
-  availability_zone = "${var.region}b"
+  availability_zone = "${var.region}a"
   vpc_security_group_ids = ["${aws_security_group.data_refinery_worker.id}"]
-  iam_instance_profile = "${aws_iam_instance_profile.ecs_instance_profile.name}"
-  subnet_id = "${aws_subnet.data_refinery_1b.id}"
+  iam_instance_profile = "${aws_iam_instance_profile.data_refinery_instance_profile.name}"
+  subnet_id = "${aws_subnet.data_refinery_1a.id}"
```
Miserlou: We are still only operating in a single subnet of a single region.

infrastructure/instances.tf
```diff
@@ -147,7 +122,7 @@ resource "aws_instance" "nomad_server_2" {
   instance_type = "t2.small"
   availability_zone = "${var.region}a"
   vpc_security_group_ids = ["${aws_security_group.data_refinery_worker.id}"]
```
Miserlou: Servers are in the workers SG, they should have their own SG without DB access.

infrastructure/instances.tf
```diff
@@ -253,10 +233,10 @@ data "template_file" "nomad_client_script_smusher" {
 resource "aws_instance" "nomad_client_1" {
   ami = "${data.aws_ami.ubuntu.id}"
   instance_type = "t2.xlarge"
-  availability_zone = "${var.region}b"
+  availability_zone = "${var.region}a"
   vpc_security_group_ids = ["${aws_security_group.data_refinery_worker.id}"]
```
Miserlou: Again, using a generic `worker` SG. Should really be per task type, and need a further dive into how Nomad allocates jobs on difference instance types and in different SGs.

infrastructure/instances.tf
```diff
@@ -285,17 +265,15 @@ output "nomad_client_ip" {
...
-  engine_version = "9.5.4"
+  engine_version = "9.6.6"
   instance_class = "db.t2.micro"
-  name = "data_refinery_${var.user}_${var.stage}"
-  username = "data_refinery_user"
+  name = "data_refinery"
+  username = "${var.database_user}"
   password = "${var.database_password}"
   db_subnet_group_name = "${aws_db_subnet_group.data_refinery.name}"
   skip_final_snapshot = true
```
Miserlou: May want to be false in prod.

infrastructure/nomad-configuration/client-instance-user-data.tpl.sh
```diff
@@ -13,6 +13,37 @@
...
+mkdir -p /var/efs/
+echo "${file_system_id}.efs.${region}.amazonaws.com:/ /var/efs/ nfs4 nfsvers=4.1,rsize=1048576,wsize=1048576,hard,timeo=600,retrans=2 0 0" >> /etc/fstab
+mount -a -t nfs4
+chown ubuntu:ubuntu /var/efs/
+
+# Set up the database
+# XXX - Failure here is catastrophic - even normally "safe" operations like apply permissions to a user that already has them
+# XXX - We need to have way to do this "manually" for staging and automatically for prod, all without
+# XXX - the catastrophic failure.
+apt-get install --yes postgresql-client-common postgresql-client
```
Miserlou: Not sure this is all even necessary any more.

We could use Packer avoid needing do a lot of this stuff every time but I don't want to open that portal to hell right now.

infrastructure/nomad-configuration/client-instance-user-data.tpl.sh
```diff
@@ -13,6 +13,37 @@
...
+PGPASSWORD=${database_password} psql -c "CREATE ROLE data_refinery_user WITH LOGIN PASSWORD 'data_refinery_password';" -h ${database_host} -p 5432 -U ${database_user} -d ${database_name}
+PGPASSWORD=${database_password} psql -c 'GRANT ALL PRIVILEGES ON DATABASE data_refinery TO data_refinery_user;' -h ${database_host} -p 5432 -U ${database_user} -d ${database_name}
+PGPASSWORD=${database_password} psql -c 'ALTER USER data_refinery_user CREATEDB;' -h ${database_host} -p 5432 -U ${database_user} -d ${database_name}
+# This won't work on RDS. AWS won't give you the superuser ability.
+# PGPASSWORD=${database_password} psql -c 'ALTER ROLE data_refinery_user superuser;' -h ${database_host} -p 5432 -U ${database_user} -d ${database_name}
+PGPASSWORD=${database_password} psql -c 'CREATE EXTENSION IF NOT EXISTS hstore;' -h ${database_host} -p 5432 -U ${database_user} -d ${database_name}
+
+PGPASSWORD=${database_password} psql -c 'GRANT ALL ON ALL TABLES IN SCHEMA public to drpostgresuser;' -h ${database_host} -p 5432 -U ${database_user} -d ${database_name}
```
Miserlou: May or may not be necessary at all.

infrastructure/nomad-configuration/client-instance-user-data.tpl.sh
```diff
@@ -55,15 +86,18 @@ cat <<"EOF" > client.hcl
...
 
-
 # Install Nomad
 chmod +x install_nomad.sh
 ./install_nomad.sh
 
 # Start the Nomad agent in client mode.
 nomad agent -config client.hcl > /var/log/nomad_client.log &
+
+# XXX: TODO: Delete the cloudinit and syslog!
```
Miserlou: This could maybe be made stage-configurable.

infrastructure/nomad-configuration/lead-server-instance-user-data.tpl.sh
```diff
@@ -74,24 +65,9 @@ cat <<"EOF" > server.hcl
 ${nomad_server_config}
 EOF
 
-
 # Install Nomad.
 chmod +x install_nomad.sh
 ./install_nomad.sh
 
 # Start the Nomad agent in server mode.
 nohup nomad agent -config server.hcl > /tmp/nomad_server.log &
```
Miserlou: We need to confirm that this log gets rotated when executed this way.

infrastructure/users.tf
```diff
@@ -0,0 +1,206 @@
...
+
+resource "aws_iam_user" "data_refinery_user_worker" {
+  name = "data-refinery-user-worker-${var.user}-${var.stage}"
+}
+
+resource "aws_iam_user" "data_refinery_user_foreman" {
+  name = "data-refinery-user-foreman-${var.user}-${var.stage}"
+}
+
+# XXX: TODO: Lock these down!!!!
```
Miserlou: This is literally just copied out of Zappa and should be locked down ASAP.

infrastructure/users.tf
```diff
@@ -0,0 +1,206 @@
...
+}
+EOF
+}
+
+resource "aws_iam_user_policy" "data_refinery_user_foreman_policy" {
+  name = "data-refinery-user-foreman-key-${var.user}-${var.stage}"
+  user = "${aws_iam_user.data_refinery_user_foreman.name}"
+
+  policy = <<EOF
+{
```
Miserlou: Same as above.

infrastructure/variables.tf
```diff
@@ -15,3 +16,117 @@ variable "stage" {
...
+}
+variable "temp_prefix" {
+  default = "temp"
+}
+variable "processed_prefix" {
+  default = "processed"
+}
+
+# Output our production environment variables. These should be in
+# parity with the env files such as workers/environments/dev.
```
Miserlou: Sort these maybe?

infrastructure/variables.tf
```diff
@@ -15,3 +16,117 @@ variable "stage" {
...
+  value = [
+    {name = "REGION"
+      value = "${var.region}"},
+    {name = "USER"
+      value = "${var.user}"},
+    {name = "STAGE"
+      value = "${var.stage}"},
+    {name = "AWS_ACCESS_KEY_ID_WORKER"
+      value = "${aws_iam_access_key.data_refinery_user_worker_key.id}"},
+    {name = "AWS_SECRET_ACCESS_KEY_WORKER"
```
Miserlou: Why was this changed? Is there also a secret_access_key_not_worker?

infrastructure/deploy.sh
kurtwheeler: Good question. I'm not sure why auth was broken out in the first place. However when all migrations are run that's one of the things that's listed as getting migrated.

infrastructure/deploy/ci_ingress.tf
kurtwheeler: 32 means the entire IP address is relevant.

infrastructure/instances.tf
kurtwheeler: Isn't this intentional for now to make debugging easier?

infrastructure/deploy/ci_ingress.tf
Miserlou: Derp.

infrastructure/instances.tf
kurtwheeler: I would agree with this, but there's a [terraform issue](https://github.com/hashicorp/terraform/issues/5417) where you can't use terraform to destroy RDS instances with this property set to true without doing a workaround.

We'll definitely want this eventually, but the longer we punt on changing it the less headaches we'll have for dev infrastructure.

format_nomad_with_env.sh
Miserlou: Also, I think this file is duplicated and I don't think it needs to be.

infrastructure/variables.tf
kurtwheeler: Yes, the box's key. When we run `export AWS_SECRET_ACCESS_KEY=XXX` so it can be formatted into the job specs, the rest of the script then has that env var set so the next time we run `terraform apply` (at the end) we get errors about how the data refinery worker user doesn't have enough permissions to do all things terraform wants to do (including just refreshing some resources).

Issue #141, opened 2018-02-26
Make gene annotation (BrainArray, possibly Ensembl genome version) less human-specific


jaclyn-taroni:
### New Issue Checklist
- [x] The title is short and descriptive
- [ ] The issue contains an:
  - [ ] Idea (new feature, user story, etc)
  - [x] Problem
- [ ] You have explained the:
  - [x] Context
  - [x] Problem or idea
  - [x] Solution or next step
  
### Context

While @Miserlou and I were taking a look at #137, we discovered that some species (e.g., human) information was hard-coded. 

#### Microarray
When we get Affymetrix data from ArrayExpress and process it AND it is BrainArray supported (see also #64), we'll need to use the organism information (e.g., the first letter of the genus and first letter of the species) to specify which BrainArray package to use. Referenced here: https://github.com/data-refinery/data-refinery/pull/137/files#diff-f935c7992de9ff4aedc8bccb3910ab18R66 We'll also need to change the other part of the string, `entrezgprobe` will need to be changed to something that reflects that ENSG  are being used (see #85; this is also noted in that diff).

#### RNA-seq
Here, for Ensembl builds
https://github.com/data-refinery/data-refinery/blob/988885e5dce3b79df04294a969ae1971a82d99df/foreman/data_refinery_foreman/surveyor/transcriptome_index.py#L50

assembly version 37 is specified. It is unclear to me what this corresponds to and if it is specific to human -- will require more investigation. If it was the version of _Ensembl_ I would expect it to be version 91 (the current version).

### Solution or next step

@kurtwheeler -- let's look into what that assembly version 37 is. We should also consider putting together a timeline for #85 because I will need to provide a new PCL file for testing and presumably some guidance on the package & string to use.

jaclyn-taroni:
It may be the case that there is a species that _is_ supported by BrainArray, but a particular chip is not, e.g., a human Affymetrix chip that SCAN will process successfully but there's no corresponding BrainArray. Not sure how often that will be true.

jaclyn-taroni:
Based on a conversation with @kurtwheeler yesterday -- we're thinking assembly version 37 may be specific to Ensembl Plants. @kurtwheeler suggested a strategy where we make a table that contains the current versions for each of the Ensembl divisions.

kurtwheeler:
[...]

PR #162, opened 2018-03-15
Kurtwheeler/document ensembl releases

kurtwheeler:
## Issue Number

#141

We fetch genome assemblies from ensembl.org and ensemblgenomes.org. These are versioned and we have to specify the version. This was done in code in such a way that it wasn't clear how the version was determined or even that it was correct.

## Types of changes

- [X] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

## Functional tests

I ran
```
./foreman/run_surveyor.sh survey_transcriptome Ensembl 1
```
which tests the main division and I ran 
```
./foreman/run_surveyor.sh survey_transcriptome EnsemblPlants 1
```
which tests the plants division and the other divisions are similar to plants.

## Checklist

### DIFF
--- a/foreman/data_refinery_foreman/surveyor/transcriptome_index.py
+++ b/foreman/data_refinery_foreman/surveyor/transcriptome_index.py
@@ -37,6 +37,19 @@ DIVISION_LOOKUP = {"EnsemblPlants": "plants",
                    "EnsemblMetazoa": "metazoa"}
 
 
+# Ensembl will periodically release updated versions of the
+# assemblies.  All divisions other than the main one have identical
+# release versions.  The latest assembly version for the main division
+# can be found by going to ftp://ftp.ensembl.org/pub/ and looking for
+# the latest version. All other divisions' latest assembly version can
+# be found by going to ftp://ftp.ensemblgenomes.org/pub/plants. These
+# versions are the latest version as of whenever it was last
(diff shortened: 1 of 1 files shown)


Miserlou:
Comment is good, I guess while we're on this topic - how do we preserve this information in the ComputationalResult? Is the system version enough, or should this information be surfaced in another way? @jaclyn-taroni

jaclyn-taroni:
By system version do you mean refine.bio versions or ensembl version? I’d probably want to know the _genome assembly_ version — so for _Arabidopsis thaliana_ it would be TAIR10, in humans GRCh38.p10

Miserlou:
Ah okay. Yes, by system version I mean the deployed version of our entire application/stack, from which they could deduce the other information if needed. You probably want something more specific though, which may require a new field.

jaclyn-taroni:
The genome assembly bit is something folks would likely want to know “at a glance”

kurtwheeler:
I think this is a good idea, however such a change would need to happen after the refactor PR is merged. I've made a new issue so that this PR can be closed and merged: https://github.com/AlexsLemonade/refinebio/issues/163

Miserlou:
If we're tracking the ability to track this value at a glance in another ticket, then this PR can be merged with no issues.

PR #192, opened 2018-04-09
Kurtwheeler/dev env improvements

kurtwheeler:

I wanted to test something in my development environment but a lot of things didn't work. I also improved documentation and scripts in a few places.

## Types of changes

## Functional tests

I ran: `./foreman/run_surveyor.sh survey_array_express --accession GSE45547`. It ran successfully, created Downloader jobs which ran successfully and created Processor jobs which failed for expected reasons. (That was what I was testing, will be fixed in another PR.)

## Checklist


foreman/surveyor.nomad.tpl
```diff
@@ -58,24 +57,14 @@ job "SURVEYOR" {
 
         # The args to pass to the Docker container's entrypoint.
         args = [
-          # Uncomment when image is updated!
-          # "python",
-          # "manage.py",
+          "python3",
+          "manage.py",
```
Miserlou: Was the image updated?

workers/downloader.nomad.tpl
```diff
@@ -59,24 +58,15 @@ job "DOWNLOADER" {
 
         # The args to pass to the Docker container's entrypoint.
         args = [
-          # Uncomment when image is updated!
-          # "python",
-          # "manage.py",
+          "python3",
```
Miserlou: Same Q as above

format_nomad_with_env.sh
```diff
@@ -68,19 +69,38 @@ fi
 # We need to specify the database and Nomad hosts for development, but
 # not for production because we just point directly at the RDS/Nomad
 # instances.
+# Conversely, in prod we need AWS credentials and a logging config but
+# not in development.
```
Miserlou: We definitely still need to be able to log and have AWS credentials in development!

If you don't want to have a awslogger, we should check for something like `local` rather than `prod`.

foreman/surveyor.nomad.tpl
kurtwheeler: Yes.

workers/downloader.nomad.tpl
kurtwheeler: Yes.

format_nomad_with_env.sh
```diff
@@ -39,6 +35,11 @@ while getopts ":p:e:o:h" opt; do
     esac
 done
 
+if [[ $project != "workers" && $project != "foreman" && $project != "api" ]]; then
+    echo 'Error: must either specify project as either "api", workers", or "foreman" with -p.'
```
dongbohu: The first `either` is redundant.

format_nomad_with_env.sh
```diff
@@ -68,19 +69,38 @@ fi
 # We need to specify the database and Nomad hosts for development, but
 # not for production because we just point directly at the RDS/Nomad
 # instances.
+# Conversely, in prod we need AWS credentials and a logging config but
+# not in development.
+# We do these with multi-line environment variables so that they can
+# be formatted into development job specs.
 if [ $env != "prod" ]; then
```
dongbohu: Since we are using `prod` for production, does that mean the deployment command run by circleci should be:
```
... TF_VAR__stage=prod ./deploy.sh
```
instead of
```
 ... TF_VAR__stage=produnction ./deploy.sh
```

infrastructure/variables.tf
```diff
@@ -53,10 +53,10 @@ variable "running_in_cloud" {
 
 # This is a placeholder until there is a production image ready.
 variable "workers_docker_image" {
-  default = "wkurt/dr_worker:3"
+  default = "wkurt/dr_worker:5"
```
dongbohu: Why can't we use `ccdl/data_refinery_workers` and `ccdl/data_refinery_foreman` images?

kurtwheeler: Are those images ready to go? I definitely can switch to them, I just wasn't sure what their status was and was trying to keep this PR's scope small.

format_nomad_with_env.sh
kurtwheeler: I don't think it will matter as far as environments go. That `TF_VAR__stage` variable is what ends up in the name of AWS resources such as `data-refinery-dev-dhu-vpc` or whatever. I don't think its actually used for anything outside of that.

However `prod` might be better than `production` for that same reason, as `data-refinery-prod-circleci-vpc` is already a fairly long name.

infrastructure/variables.tf
dongbohu: Yes, circleci deployment always updates these two images after `data_refinery_worker_base`.

format_nomad_with_env.sh
dongbohu: Yeah, I agree that `prod` is better. Let me rebuild the deployment within circleci with this name.

kurtwheeler:
Ok @Miserlou and @dongbohu I believe I have now addressed both of your questions/comments. We good for a mergey-merge?

kurtwheeler:
Also I should note that I have set the Docker images to be `ccdl/data_refinery_workers:v0.1.0` and `ccdl/data_refinery_foreman:v0.1.0` with the intention of tagging this commit as `v0.1.0`.

PR #195, opened 2018-04-09
Cleans up the state of the models and fixes some mismatches between m…

kurtwheeler:
…odels and serializers.

Stef ran into some bugs related to models/serializers on the dev branch. There were some mismatches between the two. This branch resolves those mismatches, removes some cruftiness of models such as the models file being `new_models.py` instead of `models.py`, makes every model update its `created_at` and `last_modified` fields, standardizes the name of `last_modified`, and removes the TimeTrackedModel and WorkerJob abstract classes. It also adds some more test code to the API to make sure that more models are working correctly so as to catch issues like #194 

## Types of changes

## Functional tests

`api/serve.sh` starts the server successfully. I am relying on the automated tests for regressions.

## Checklist


common/data_refinery_common/models/jobs.py
```diff
@@ -1,16 +1,78 @@
+from typing import Dict
+
 from django.db import transaction
 from django.db import models
-from data_refinery_common.models.base_models import TimeTrackedModel
+from django.utils import timezone
+
+from data_refinery_common.models.models import Sample, Experiment, OriginalFile
+
+
```
Miserlou: Extra newline

common/data_refinery_common/models/models.py
```diff
@@ -108,13 +124,21 @@ def __str__ (self):
     publication_doi = models.CharField(max_length=64, blank=True)
     pubmed_id = models.CharField(max_length=32, blank=True)
     source_first_published = models.CharField(max_length=256, blank=True)
-    source_last_updated = models.CharField(max_length=256, blank=True)
+    source_last_modified = models.CharField(max_length=256, blank=True)
```
Miserlou: Should we try to parse these into actual times?

With charfields we can't do queries like, give me all samples original made after 2013, etc..

kurtwheeler: I want to say yes. My fear is that at some point we're gonna get some bad data from the external sources which isn't nicely parsable... Do we trust them? Is a badly-formatted data a good reason to skip an experiment? Once we make this a date or datetime field we'll have less options once we do run into a bad string.

I'm pretty balanced on the fence about this, so nudge me either way and I'll make it happen.

common/data_refinery_common/migrations/0002_auto_20180409_1908.py
```diff
@@ -0,0 +1,95 @@
+# Generated by Django 2.0.2 on 2018-04-09 19:08
```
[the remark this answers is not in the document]
kurtwheeler: Rolling it into initial will break migrations for development. If we want to avoid going past the initial one then we should just not source control the migration files so that `./update_models.sh` just does the right thing for each dev.

foreman/data_refinery_foreman/surveyor/array_express.py
```diff
@@ -104,8 +105,8 @@ def create_experiment_from_api(self, experiment_accession_code: str) -> Dict:
             experiment_object.description = parsed_json["description"][0]["text"]
             experiment_object.platform_name = experiment["platform_accession_name"]
             experiment_object.platform_accession_code = experiment["platform_accession_code"]
-            experiment_object.source_first_published = experiment["release_date"]
-            experiment_object.source_last_updated = experiment["last_update_date"]
+            experiment_object.source_first_published = parse_datetime(experiment["release_date"])
```
Miserlou: Quick little testerino might be nice here..

If it doesn't work, there is a library called `arrow` that I like for situations like this

PR #215, opened 2018-04-24
Kurtwheeler/track migrations again

kurtwheeler:

This starts to track migration files again because we had multiple issues with the models and the database getting out of sync in development (including cloud development). When we finally get to a release point, we can blow them all away and revert to an initial migration file.

Also we had an extra `SurveyJob` class defined in the common subproject so I removed that.

## Types of changes

## Functional tests

I ran the unit tests, they're still passing and this is mostly just removing cruft and tracking stuff that should have been tracked so I don't think it can impact too much.

## Checklist


Miserlou:
I think tracking migrations across branches before production is going to cause more problems than it solves. I don't see why this is necessary.

Issue #238, opened 2018-05-01
API Endpoint: Dataset


Miserlou:
### Context

Users using the web interface will want to construct a "data set" of files they want to explore and download. To make this shareable, we will need to persist it to a database.

### Problem or idea

We should have a CRUDdy set of `/dataset` endpoints that allows the interface to track these datasets and ultimately dispatch them to the smasher/notifier queues. 

### Solution or next step

The data representation is probably pretty easy, it can be a list of experiment IDs with a sublist of associated Sample IDs, as well as processing/smashing progress information.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

cgreene:
This will help us for #189 too. We will have a record of what samples go into constructed datasets.

PR #247, opened 2018-05-07
Adds Dataset API Endpoint

Miserlou:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/238

Adds a 'Dataset' type and and API interface for creating and managing Datasets, including starting smasher jobs.

## Types of changes

## Functional tests

Yep

## Checklist


Miserlou:
I added email to the create method, but I don't think there should be more than one way to start a job.

Issue #244, opened 2018-05-02
Add a unit test for salmontools extract-unmapped


jaclyn-taroni:
### Context

Once we implement `salmontools extract-unmapped` (#83), we'll need a unit test.

### Problem or idea

The point of this tool is to essentially "filter" FASTQ files we give `salmon quant` to only those reads that Salmon could not map to the transcriptome index we gave it (e.g., unmapped reads). (We use the [`--writeUnmapped`](http://salmon.readthedocs.io/en/latest/salmon.html#writeunmappednames) option for `salmon quant` to get the read names.)

Some cursory thoughts:
* We could use the [sample data from SalmonTools](https://github.com/COMBINE-lab/SalmonTools/tree/474ba5947711074a015a268b47058ac8930cf566/sample_data).
* We could do an more extensive test (likely as part of a test of the entire RNA-seq pipeline) using simulated data. Specifically, if we use a tool like [Polyester](https://github.com/alyssafrazee/polyester) (as suggested by Rob Patro [here](https://github.com/AlexsLemonade/refinebio/issues/117#issue-296499941)) to simulate human reads and some _P. aeruginosa_ reads and combine them into a FASTQ file where we know the number of bacterial reads. We could check if the FASTA we get out of `salmontools extract-unmapped` is the set of bacterial reads we would expect. (This might be a future thought!)

### Solution or next step

To be revisited after #83 is accomplished.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

dongbohu:
Done by #250.

Issue #83, opened 2018-01-16
Implement extraction of unmapped read sequences (unmapped by Salmon)


kurtwheeler:
Use [salmontools](https://github.com/COMBINE-lab/SalmonTools) extract-unmapped command.

kurtwheeler:
![salmon-gene-name](https://user-images.githubusercontent.com/4026833/35344221-908a92b8-00fa-11e8-9e42-5a9ec4d94c43.jpg)

kurtwheeler:
Ok, so it turns out that the github repo I linked to above pretty much says what needs to be done. `salmontools extract-unmapped` will expect an `unmapped_reads.txt` file output from Salmon and the original FASTQ file that was passed into Salmon. Therefore it will make sense to do this within the same Processor as the current Salmon processor.

kurtwheeler:
#84 answers the why of using STAR.

jaclyn-taroni:
After chatting with @dongbohu, here's a more up to date sketch

![4_27_18 10_49 am office lens](https://user-images.githubusercontent.com/19534205/39369096-5d406628-4a09-11e8-8955-4ba044b391be.jpg)

jaclyn-taroni:
Chatted with @dongbohu this afternoon - the output of `salmontools extract-unmapped` appears to be a FASTA file. We think it make sense to gzip it after the fact.

dongbohu:
Implemented in #250.

PR #250, opened 2018-05-08
New feature: SalmonTools (and corresponding tests) in Salmon pipeline

dongbohu:
## Issue Number
#83 and #244

This PR adds salmontools data analysis and two test cases, one for double reads, the other for single reads.

## Types of changes

## Functional tests

Unit tests were added and run to confirm the results.

## Checklist

## Screenshots

### DIFF
--- a/workers/data_refinery_workers/processors/salmon.py
+++ b/workers/data_refinery_workers/processors/salmon.py
@@ -263,15 +263,13 @@ def _run_salmontools(job_context: Dict, skip_processed=SKIP_PROCESSED) -> Dict:
         logger.info("Skipping pre-processed SalmonTools run!")
         skip = True
 
-    job_context['time_start'] = timezone.now()
     if skip:  # If this procedure should be skipped, return immediately
-        job_context['time_end'] = timezone.now()
         return job_context
 
     command_str = "salmontools extract-unmapped -u {unmapped_file} -o {output} "
     output_prefix = job_context["output_directory"] + "unmapped_by_salmon"
     command_str = command_str.format(unmapped_file=unmapped_filename,
-                                         output=output_prefix)
+                                     output=output_prefix)
     if "input_file_path_2" in job_context:
         command_str += "-1 {input_1} -2 {input_2}"
         command_str = command_str.format(input_1=job_context["input_file_path"],
@@ -280,6 +278,7 @@ def _run_salmontools(job_context: Dict, skip_processed=SKIP_PROCESSED) -> Dict:
         command_str += "-r {input_1}"
         command_str= command_str.format(input_1=job_context["input_file_path"])
 
+    start_time = timezone.now()
     logger.info("Running the following SalmonTools command: %s",
                 command_str,
                 processor_job=job_context["job_id"])
@@ -287,7 +286,7 @@ def _run_salmontools(job_context: Dict, skip_processed=SKIP_PROCESSED) -> Dict:
     completed_command = subprocess.run(command_str.split(),
                                        stdout=subprocess.PIPE,
                                        stderr=subprocess.PIPE)
-    job_context['time_end'] = timezone.now()
+    end_time = timezone.now()
 
     # As of SalmonTools 0.1.0, completed_command.returncode is always 0,
     # (even if error happens).  completed_command.stderr is not totally
@@ -305,8 +304,8 @@ def _run_salmontools(job_context: Dict, skip_processed=SKIP_PROCESSED) -> Dict:
         result = ComputationalResult()
         result.command_executed = command_str
         result.system_version = __version__
-        result.time_start = job_context['time_start']
-        result.time_end = job_context['time_end']
+        result.time_start = start_time
+        result.time_end = end_time
         result.program_version = subprocess.run(['salmontools', '--version'],
                                                 stderr=subprocess.PIPE,
                                                 stdout=subprocess.PIPE).stderr.decode().strip()
@@ -318,7 +317,7 @@ def _run_salmontools(job_context: Dict, skip_processed=SKIP_PROCESSED) -> Dict:
         logger.error("Shell call to salmontools failed with error message: %s",
                      status_str,
                      processor_job=job_context["job_id"])
-        job_context["job"].failure_reason = ("Shell call to salmon failed because: "
+        job_context["job"].failure_reason = ("Shell call to salmontools failed because: "
                                              + status_str[0:256])
         job_context["success"] = False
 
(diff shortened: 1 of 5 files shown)


dongbohu:
A few things that may be worth mentioning:
1. All test data for salmontools in this personal repo:
https://github.com/dongbohu/salmontools_tests
They are organized in the same structure as the data that will be generated by `salmon quant` command.

2. Because `salmontools` is not installed in the workers_base Docker image, it is manually installed in `Dockerfile.tests` to pass the test cases. @kurtwheeler, I know you are working on individual docker images for each processor. You can copy these two lines in `Dockerfile` that builds `salmon` docker image:
```
RUN git clone https://github.com/COMBINE-lab/SalmonTools.git
RUN cd SalmonTools && cmake . -DCMAKE_INSTALL_PREFIX=/usr/local && make install
```
Note that because the executable will be installed in `/usr/local/bin/`, we don't need to add the path. As a matter of fact, this line in `workers/Dockerfile.base` can be removed too:

```ENV PATH /home/user/salmon-${SALMON_COMMIT_HASH}/bin:${PATH}```

because `salmon` is also installed in /usr/local/bin/ thanks to `-DCMAKE_INSTALL_PREFIX=/usr/local` option.

**Comment by dongbohu** [comment]
I also did a little cleanup on a few other files, for instance, I moved the function `_zip_and_upload` in `salmon.py` to the bottom of this file, because it is the last step in this pipeline.

File: workers/data_refinery_workers/processors/salmon.py [hunk 1]
```diff
@@ -285,33 +252,113 @@ def _run_salmon(job_context: Dict, skip_processed=SKIP_PROCESSED) -> Dict:
...
+    # the status of SalmonTools execution.  Any text in stderr that is
+    # not in the above format is treated as error message.
+    status_str = completed_command.stderr.decode().strip()
+    success_pattern = r'^There were \d+ unmapped reads$'
+    if re.match(success_pattern, status_str):
+        result = ComputationalResult()
+        result.command_executed = command_str
+        result.system_version = __version__
+        result.time_start = job_context['time_start']
+        result.time_end = job_context['time_end']
```
kurtwheeler: If you're just going to use the variables within the same function, you can just use local variables rather than putting these onto the `job_context`. That object is a means for passing data from one step to another.

workers/data_refinery_workers/processors/salmon.py
```diff
@@ -285,33 +252,113 @@ def _run_salmon(job_context: Dict, skip_processed=SKIP_PROCESSED) -> Dict:
...
+        skip = True
 
-    try:
-        tx_out = ro.r['::']('tximport', 'tximport')(job_context['output_directory'] + 'quant.sf', type="salmon", txOut=True)
-    except RRuntimeError as e:
-        # XXX/TODO: This doesn't work yet.
-        # But, eventually, it will look something like this. ^^
+    job_context['time_start'] = timezone.now()
+    if skip:  # If this procedure should be skipped, return immediately
+        job_context['time_end'] = timezone.now()
```
kurtwheeler: I don't think anything else is going to use this, so if you're not running salmon tools it probably doesn't need to be set.

workers/data_refinery_workers/processors/salmon.py
```diff
@@ -285,33 +252,113 @@ def _run_salmon(job_context: Dict, skip_processed=SKIP_PROCESSED) -> Dict:
...
+        result.command_executed = command_str
+        result.system_version = __version__
+        result.time_start = job_context['time_start']
+        result.time_end = job_context['time_end']
+        result.program_version = subprocess.run(['salmontools', '--version'],
+                                                stderr=subprocess.PIPE,
+                                                stdout=subprocess.PIPE).stderr.decode().strip()
+        result.is_ccdl = True
+        result.save()
+        job_context["result"] = result
```
kurtwheeler: What is `job_context["result"]` for?

workers/data_refinery_workers/processors/salmon.py
```diff
@@ -285,33 +252,113 @@ def _run_salmon(job_context: Dict, skip_processed=SKIP_PROCESSED) -> Dict:
...
+        job_context["success"] = True
+    else:   # error in salmontools
+        logger.error("Shell call to salmontools failed with error message: %s",
+                     status_str,
+                     processor_job=job_context["job_id"])
+        job_context["job"].failure_reason = ("Shell call to salmon failed because: "
```
kurtwheeler: This should say "Shell call to salmontools"...

workers/data_refinery_workers/processors/salmon.py
dongbohu: What about `job_context['time_start']` if the job is skipped? Should we delete that one too?

workers/data_refinery_workers/processors/salmon.py
dongbohu: Are you saying that I don't have to set `job_context['time_start']` and `job_context['time_end']` at all?

workers/data_refinery_workers/processors/salmon.py
dongbohu: I copied it from `_run_salmon` function. It seems to be a key whose value is the job result.

workers/data_refinery_workers/processors/salmon.py
kurtwheeler: Not exactly. You could just instead use variables named `start_time` and `end_time` rather than putting them in the `job_context`.

workers/data_refinery_workers/processors/salmon.py
dongbohu: Good sight. This is a copy/paste error.

workers/data_refinery_workers/processors/salmon.py
kurtwheeler: If we switch to using local variables it won't be an issue.

workers/data_refinery_workers/processors/salmon.py
kurtwheeler: Yeah I figured that might be what happened, that one doesn't seem to be used anywhere either.

workers/data_refinery_workers/processors/salmon.py
dongbohu: Good to know. I thought each job would always set the keys such as `time_start` and `time_end` in `job_context`.

kurtwheeler: I can see how it would look like that. The `start_time` and `end_time` fields for the ProcessorJob are set in the utils functions `start_job` and `end_job`.

Issue #165, opened 2018-03-19
Add and Harmonize SDRF Data


Miserlou:
Samples on ArrayExpress have meta-information in `sdrf` files. We may want to turn some of this information into structured/harmonized fields on our sample table, and to define some harmony with samples from other data sources.

I'm currently crunching the statistics about what these values are. So far, it looks like quite sparse/noisy data.

Miserlou:
The stats are finally done. Here are the values:

https://gist.githubusercontent.com/Miserlou/479348184735a084107288116d2da8b3/raw/ce9fb186c9f84ba3c270f47b9c1fd82e4274b946/array_express_sdrf_values.txt

It is an extremely long tail. It could maybe be cleaned up further, (ex `FactorValue[` vs `Factor Value[` vs `FactorValue [` vs `Factor Value [`, etc), but I still think that it'd be a very long tail.

Not really sure what to do about this.

I think we could at least give it a shot for all of the various conceptions of "Organism Part".

jaclyn-taroni:
I will go through the very top of this list and identify fields I think are equivalent & my rationale for inclusion

Miserlou:
I was imagining both for filtering/searching - in which case perfect harmonization is less important - but also for choosing which information we want to prepare for _display_.

jaclyn-taroni:
As far as display is concerned, resolving some of this stuff
> ex `FactorValue[` vs `Factor Value[` vs `FactorValue [` vs `Factor Value [`, etc

Is probably a worthwhile starting point. For displaying a given sample (think table), I think we'll probably want to include _all_ of the information in the `sdrf` -- perhaps prioritizing "harmonized" categories, i.e., putting them as the first columns

jaclyn-taroni:
Is there any documentation about the difference between a characteristic and a factor value? In my experience, they can often be redundant.

jaclyn-taroni:
From [this documentation](https://www.ebi.ac.uk/arrayexpress/help/creating_a_sdrf.html):
> These Factor Value[] columns reference the experimental factor names defined in the IDF, and should be placed at the end of the SDRF. The contents of these columns will usually duplicate those in a material Characteristics column or a protocol Parameter Value column.

For filtering purposes, we can likely treat them as the same.

jaclyn-taroni:
[...]

PR #252, opened 2018-05-10
Harmonize SDRF and Other Types of Data

Miserlou:
## Issue Number
https://github.com/AlexsLemonade/refinebio/issues/165

This adds extracted properties to a Sample object based on various patterns that happen in metadata that we get from our three sources. There is a new 'harmony' file which handles all of this new functionality.

The weird thing about this PR is that includes a cherry-picked commit from my FastQC branch, so it also adds very crude Fast/MultiQC support which has been further fleshed out with ComputationalResults in another branch that will accompany this PR with a smaller PR immediately after merge.

## Types of changes

## Functional tests

There are new tests for the harmonization added.

## Checklist


foreman/data_refinery_foreman/surveyor/harmony.py
```diff
@@ -0,0 +1,699 @@
...
+    ##
+    # Sex!
+    ##
+    sex_fields = [    
+                    'sex',
+                    'gender',
+                    'subject gender',
+                    'subjext sex',
+                    'characteristic [sex]',
+                    'characteristics [sex]',
```
kurtwheeler: For sex, age, and part you include the `characteristic [{item}]` in this list, but it looks like `add_variants` will take care of those variants for you.

foreman/data_refinery_foreman/surveyor/harmony.py
```diff
@@ -0,0 +1,699 @@
...
+                    'stop time', 
+                    'time point', 
+                    'sampling time point', 
+                    'sampling time',
+                    'time post infection'
+                ]
+    time_fields = add_variants(time_fields)
+    for sample in original_samples:
+        title = sample['title']
+        for key, value in sample.copy().items():
```
kurtwheeler: This looks like a great method for dealing with the messiness of this metadata! Good work. The only comment I have on the `harmonize` function is about the lines like `for key, value in sample.copy().items():`. Why are you `copy()`ing every sample? The [view object](https://docs.python.org/3.6/library/stdtypes.html#dict-views) returned by `items()` doesn't allow you to mutate the original dictionary.

foreman/data_refinery_foreman/surveyor/test_harmony.py
```diff
@@ -0,0 +1,299 @@
...
+        self.sample.save()
+
+        self.samples = [self.sample]
+
+    def test_sdrf_harmony(self):
+        """ Harmonize SDRF test"""
+        
+        metadata = parse_sdrf("https://www.ebi.ac.uk/arrayexpress/files/E-MTAB-3050/E-MTAB-3050.sdrf.txt")
+        harmonized = harmonize(metadata)
+        print(harmonized)
```
kurtwheeler: Dangling print

foreman/data_refinery_foreman/surveyor/test_harmony.py
```diff
@@ -0,0 +1,299 @@
...
+            'SRR1797277',
+            'SRR1533126'
+        ]
+        for accession in lots:
+            try:
+                metadata = SraSurveyor.gather_all_metadata(accession)
+                harmonized = harmonize([metadata])
+                print(harmonized)
+            except UnsupportedDataTypeError as udte:
+                print(udte)
```
kurtwheeler: It looks like these tests will be pretty noisy, did you mean to leave all these prints in here?

workers/data_refinery_workers/processors/salmon.py
```diff
@@ -368,9 +400,14 @@ def salmon(job_id: int) -> None:
...
                         _set_job_prefix,
                         _prepare_files,
+                        _run_fastqc,
+
                         _determine_index_length,
                         _download_index,
+
                         _run_salmon,
                         _run_salmontools,
+                        _run_fastqc,
```
kurtwheeler: Looks like `_run_fastqc` is called twice. Is that intentional?

cgreene:
I want to make sure of something. Though we are attempting to harmonize metadata, are we still storing exactly what the submitter supplied somewhere? We should be able to provide the metadata that was submitter supplied along with our own `CCDL standardized` metadata and `predicted` metadata (not yet extant, but when we have ML models predicting metadata this would go there).

Miserlou:
Yes, we'll have three forms of sample metadata:

* `SampleAnnotation` where `annotation.is_ccdl = False`, which are provided by original data source and will be displayed in the "full" view in the interface
* Properties directly on the Sample object, which are harmonized here and will be displayed in the web interface as primary/at-a-glance information
* `SampleAnnotation` where `annotation.is_ccdl = True`, which are created by one of our future classifiers and associated with a `ComputationalResult` object. We haven't talked about these too much yet and I don't think we've decided how we want to display these results in the web interface

jaclyn-taroni:
Can you write up a description of the methods you've employed here? For example, you mentioned some "smoothing" being applied to the values in the sex (& related) field(s) yesterday during design review. It would be good to capture this information here for posterity/future documentation/future scientific writing. Other "want to know" things that come to mind: stripping punctuation, handling of [characteristic vs. factor value](https://github.com/AlexsLemonade/refinebio/issues/165#issuecomment-376565852)

common/data_refinery_common/models/models.py
```diff
@@ -45,9 +45,21 @@ def __str__ (self):
...
+    part = models.CharField(max_length=255, blank=True)
+    genotype = models.CharField(max_length=255, blank=True)
+    disease = models.CharField(max_length=255, blank=True)
+    disease_stage = models.CharField(max_length=255, blank=True)
+    cell_line = models.CharField(max_length=255, blank=True)
+    treatment = models.CharField(max_length=255, blank=True)
+    race = models.CharField(max_length=255, blank=True)
+    subject = models.CharField(max_length=255, blank=True)
+    compound = models.CharField(max_length=255, blank=True)
+    time = models.CharField(max_length=255, blank=True)
```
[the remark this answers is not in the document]
Miserlou: The field names are sourced from this comment original ticket: https://github.com/AlexsLemonade/refinebio/issues/165#issuecomment-376684079

With _all_ of these fields, the problem is going to be that they're overloaded in how they're used by submitters, so the more general we can be the better, then at least everything will make some type of sense, even if it's vague. For part, I'd think maybe `organism_part`, but then you could have an `organism_part` that doesn't actually occur organism-wide (ex a tumor), and `subject_part` doesn't always work because cell lines aren't really "subjects" in the sense that we're using it everywhere else. IDK. Time is similarly overloaded to the context of the exposure, but it's often things like exposure-to-a-compound-or-type-of-radiation, but obviously not always.

foreman/data_refinery_foreman/surveyor/array_express.py
```diff
@@ -269,13 +272,39 @@ def create_samples_from_api(self,
...
+            # Figure out the Organism for this sample
+            organism_name = UNKNOWN
+            for characteristic in sample["characteristic"]:
+                if characteristic["category"].upper() == "ORGANISM":
+                    organism_name = characteristic["value"].upper()
+
+            if organism_name == UNKNOWN:
+                logger.error("Sample from experiment %s did not specify the organism name.",
+                             experiment.accession_code,
+                             survey_job=self.survey_job.id)
```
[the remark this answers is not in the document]
Miserlou: 👍

foreman/data_refinery_foreman/surveyor/harmony.py
```diff
@@ -0,0 +1,699 @@
...
+        copy.append("comments [" + item + "]")
+        copy.append("comments[" + item + "]")
+        copy.append("factorvalue[" + item + "]")
+        copy.append("factor value[" + item + "]")
+        copy.append("factorvalue [" + item + "]")
+        copy.append("factor value [" + item + "]")
+        copy.append("sample_" + item)
+        copy.append("sample_host" + item)
+        copy.append("sample_sample_" + item) # Yes, seriously.
+    return copy
```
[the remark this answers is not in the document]
Miserlou: 👍

foreman/data_refinery_foreman/surveyor/harmony.py
Miserlou: This looks reduntant, but there are some samples actually have a nested characteristic field. Added a comment about it.

foreman/data_refinery_foreman/surveyor/harmony.py
```diff
@@ -0,0 +1,699 @@
...
+        for key, value in sample.copy().items():
+            lower_key = key.lower().strip() 
+            if lower_key in part_fields:
+                harmonized_samples[title]['part'] = value.lower().strip()
+                break
+
+    ##
+    # Genotype!
+    ##
+    genotype_fields = [
```
jaclyn-taroni: I am a bit uncomfortable with lumping these together given the way we now intend to use the harmonized data.

From my original comment (https://github.com/AlexsLemonade/refinebio/issues/165#issuecomment-376684079): 

> For filtering only (i.e., we're mostly interested in the presence or absence of metadata attached to a sample), I think it's appropriate to lump together the following categories: genotype, genetic background, variation, strain or line, background strain, cultivar and ecotype. This basically would indicate that there is some information about the genetic background (in a broad sense) that is associated with an individual sample. These are not the exact same information (multiple cultivars could potentially have the same genetic background and I bet people use "genotype" and "genetic background" a wide range of different ways depending on their individual experiments)

cgreene: Genetic information is better than genotype (since genotype is likely to be represented by both a strain background + variations from that such as deletions).

kurtwheeler:
This code looks good to me. I do think that at least one of @cgreene or @jaclyn-taroni should also approve this before it is merged.

common/data_refinery_common/models/models.py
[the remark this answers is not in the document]
cgreene: I agree that `part` and `time` are ambiguous. One thing that we will need to do (as @jaclyn-taroni noted somewhere) is explain to people _how_ we harmonized. Is there a way to include some sort of metadata on these fields in a way that would let us make the description available? If we added something like help text with this information, can that be accessed at the API level if we want it to be?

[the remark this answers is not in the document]
Miserlou: @cgreene @jaclyn-taroni do you have a conceptual preference for part:

```python
    part_fields = [
                    # AE
                    'organism part', 
                    'cell type', 
                    'tissue', 
                    'tissue type', 
                    'tissue source', 
                    'tissue origin', 
                    'source tissue', 
                    'tissue subtype', 
                    'tissue/cell type', 
                    'tissue region', 
                    'tissue compartment', 
                    'tissues', 
                    'tissue of origin', 
                    'tissue-type', 
                    'tissue harvested', 
                    'cell/tissue type', 
                    'tissue subregion', 
                    'organ',
                    'characteristic [organism part]',
                    'characteristics [organism part]',

                    # SRA
                    'cell_type'
                    'organismpart',

                    # GEO
                    'isolation source',
                    'tissue sampled',
                    'cell description'

                ]
```
`tissue` maybe?

Time is a lot harder as it's entirely experiment dependent, so the provided and harmonized information is less useful for large-scale analysis, but may be useful while browsing the web interface for samples:

```python
    time_fields = [
                    'time', 
                    'initial time point', 
                    'start time', 
                    'stop time', 
                    'time point', 
                    'sampling time point', 
                    'sampling time',
                    'time post infection'
                ]
```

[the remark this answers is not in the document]
jaclyn-taroni: How about `specimen_part` for the first one?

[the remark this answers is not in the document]
jaclyn-taroni: I like `time_point` for the second one.

jaclyn-taroni:
Recording some methods notes based in part on a conversation I had with @Miserlou:

* No fuzzy matching is done at this time. That is to say that we are not currently handling misspellings.
* > Values are stripped and forced to lowercase.
* > [G]ender, which is forced to being 'male' or 'female', and type-casting age values to doubles.  

  Gender is currently being determined by the presence of `f`.

Here is a table that reflects my understanding of what's happening:

| Harmonized key | Keys from data sources |
|:----------------:|-------------------------|
| `part` (or `specimen part`?) | `organism part`, `cell type`, `tissue`, `tissue type`, `tissue source`, `tissue origin`, `source tissue`, `tissue subtype`, `tissue/cell type`, `tissue region`,  `tissue compartment`,  `tissues`, `tissue of origin`, `tissue-type`,  `tissue harvested`, `cell/tissue type`, `tissue subregion`, `organ`, `characteristic [organism part]`, `characteristics [organism part]`, `cell_type`, `organismpart`, `isolation source`, `tissue sampled`, `cell description`
| `genetic information` | `strain/background`, `strain`,  `strain or line`, `background strain`, `genotype`, `genetic background`, `genotype/variation`, `ecotype`, `cultivar`, `strain/genotype`|
| `disease` |  `disease `, `disease state `, `disease status `, `diagnosis `, `disease `, `infection with `, `sample type ` |
| `disease stage` | `disease state `, `disease staging `, `disease stage `, `grade `, `tumor grade `,  `who grade `, `histological grade `, `tumor grading `, `disease outcome `, `subject status ` |
| `treatment` | `treatment`, `treatment group`, `treatment protocol`,  `drug treatment`, `clinical treatment` |
| `race` | `race`, `ethnicity`, `race/ethnicity`, `phenotype` |
| `subject` |  `subject `, `subject id `, `subject/sample source id `, `subject identifier `, `human subject anonymized id `, `individual `, `individual identifier `,  `individual id `, `patient `, `patient id `, `patient identifier `,  `patient number `, `patient no `,  `donor id `, `donor `, `sample_source_name `|
| `development stage` | `developmental stage`,  `development stage`, `development stages` |
| `compound` | `compound`, `compound1`, `compound2`, `compound name`, `drug`, `drugs`, `immunosuppressive drugs` |
| `time` | `initial time point`, `start time`, `stop time`, `time point`, `sampling time point`, `sampling time`, `time post infection` |

Outstanding questions/comments I have:
* Why is `phenotype` in `race`? That should not be the case.
* I also think `compound` might be better termed `treatment` at this time.
* Can you comment further on handling of characteristic vs. factor value? If something says `FactorValue[age]` vs. `Characteristic [age]` what happens?

Miserlou:
`compound` and `treatment` are currently distinct fields, are you saying they should be combined? Similarly, should `phenotype` be its own extract field, or should that be folded into `genetic information`?

jaclyn-taroni:
> `compound` and `treatment` are currently distinct fields, are you saying they should be combined? 

Whoops, my bad. I actually think that might be the way to go given how broad these categories are right now. 

> Similarly, should `phenotype` be its own extract field, or should that be folded into genetic information?

`phenotype` is too broad a term to go anywhere really at this stage.

cgreene:
I'm curious about this too:
> Can you comment further on handling of characteristic vs. factor value? If something says FactorValue[age] vs. Characteristic [age] what happens?

jaclyn-taroni:
> Can you comment further on handling of characteristic vs. factor value? If something says FactorValue[age] vs. Characteristic [age] what happens?

As noted in the [ArrayExpress documentation](https://www.ebi.ac.uk/arrayexpress/help/creating_a_sdrf.html), Factor Values are often redundant with Characteristics or Parameters which is the motivation behind this question.

Chatted with @Miserlou - in it's current state the code will take whichever it encounters _first_.

Relevant bit of code starting here: https://github.com/AlexsLemonade/refinebio/blob/bae319aaa434838b68a5d198d9c1d0f82c05ed20/foreman/data_refinery_foreman/surveyor/harmony.py#L627

jaclyn-taroni:
Note that as of https://github.com/AlexsLemonade/refinebio/pull/252/commits/e68baee8cc15b204b3e7444578b9169935f0d9bc, gender is no longer determined by the presence of `f`.

jaclyn-taroni:
Okay, this LGTM -- it doesn't look like `compound` and `treatment` have been combined. This might be the kind of decision we make once we take a look at what _values_ there are/how people use the two for that + what users think.

PR #255, opened 2018-05-15
Ramenhog/filter samples by ids

ramenhog:
Modified the `samples` endpoint to allow filtering by a list of sample ids.

## Types of changes

## Functional tests
* passed in a list of ids via a query param for the samples endpoint and verified it returned the expected results

* endpoint still works when no `ids` query param is present

### DIFF
--- a/api/data_refinery_api/views.py
+++ b/api/data_refinery_api/views.py
@@ -11,37 +11,6 @@ from rest_framework.response import Response
 from rest_framework.settings import api_settings
 from rest_framework import status, filters, generics
 
-<<<<<<< Updated upstream
-from data_refinery_common.models import (
-    Experiment, 
-    Sample, 
-    Organism, 
-    ComputationalResult,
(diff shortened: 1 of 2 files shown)


api/data_refinery_api/views.py
```diff
@@ -192,7 +194,13 @@ def get(self, request, format=None):
...
         filter_dict.pop('limit', None)
         filter_dict.pop('offset', None)
-        samples = Sample.objects.filter(**filter_dict)
+        ids = self.request.query_params.get('ids', None)
+
+        if ids is not None:
+            ids = [ int(x) for x in ids.split(',')]
+            samples = Sample.objects.filter(pk__in=ids)
+        else:
+            samples = Sample.objects.filter(**filter_dict)
```
Miserlou: Ideally, you should be able to handle both types of filter at the same time here, rather than just one or the other..

PR #257, opened 2018-05-16
Fix AE Accession Detector

Miserlou:
## Issue Number

n/a

Analyzing the initial Zebrafish crunch results determined that the AE accession code detector is broken in some cases (failed to return a value). This PR adds a return value in all cases and adds a test case.

## Types of changes

## Functional tests

New test added for failing cases in ZF list.

## Checklist

### DIFF
--- a/foreman/data_refinery_foreman/surveyor/management/commands/survey_all.py
+++ b/foreman/data_refinery_foreman/surveyor/management/commands/survey_all.py
@@ -43,7 +43,7 @@ Note: One entry per line, GSE* entries survey GEO, E-GEO-* entries survey ArrayE
                 s3 = boto3.resource('s3')
                 try:
                     filepath = "/tmp/input_" + str(uuid.uuid4()) + ".txt"
-                    s3.Bucket(bucket).download_file(key, filepath)
+                    s3.Bucket(bucket).download_file(key, "/tmp/input_" + str(uuid.uuid4()) + ".txt")
                 except botocore.exceptions.ClientError as e:
                     if e.response['Error']['Code'] == "404":
                         logger.error("The remote file does not exist.")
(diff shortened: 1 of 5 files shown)


Miserlou:
Additionally, this adds a fix whereby the AE accession detector could generate non-unique values for a unique field, ex "Sample 1" from `E-MTAB-6739`, which now becomes `E-MTAB-6739-Sample 1`

foreman/data_refinery_foreman/surveyor/management/commands/survey_all.py
```diff
@@ -43,7 +43,7 @@ def handle(self, *args, **options):
                 s3 = boto3.resource('s3')
                 try:
                     filepath = "/tmp/input_" + str(uuid.uuid4()) + ".txt"
-                    s3.Bucket(bucket).download_file(key, "/tmp/input_" + str(uuid.uuid4()) + ".txt")
```
cgreene: I like this commit using a defined variable instead of having this written in two places, but I'm confused about how it relates to the topic of this PR.

Miserlou: The code pre-change is actually broken since it's actually generating two separate uuids. It's not related to the other AE changes other than it allows to use S3-hosted accession lists when kicking off jobs. I can file a separate PR, but that seemed like overkill for a trivial 1-line fix.

cgreene: Ahh! Two separate uuids sound bad! I think it would be better to make this a separate PR because then it's not held up by anything else here & seems like a very fast merge. The two separate uuids bit sounds particularly urgent.

foreman/data_refinery_foreman/surveyor/test_array_express.py
```diff
@@ -52,3 +54,67 @@ def test_survey(self, mock_send_task):
...
+        ex_accessions = [
+                            "E-MTAB-3050", 
+                            "E-MEXP-669", 
+                            "E-MEXP-2215", 
+                            "E-MEXP-2288", 
+                            "E-MEXP-2381",
+                            "E-MTAB-6739",
+                        ]
+
+        for ex_accesion in ex_accessions:
```
kurtwheeler: > ex_accesion

should be ex_accession

Issue #256, opened 2018-05-16
Speed up salmon speed and salmon test speed


kurtwheeler:
### Context

Relates to #234 

The `salmon` tests are very slow and the slowest part of our test pipeline. However the samples that we run `salmon` on were not chosen based on how long we thought they would take to process.

### Problem or idea

There are two potential ways to speed up the `salmon` tests. One of them is to choose which sample we're going to process based on what will make them run fast. Salmon's docs about its [biasSpeedSamp option](http://salmon.readthedocs.io/en/latest/salmon.html#biasspeedsamp) state:

> When evaluating the bias models (the GC-fragment model specifically), Salmon must consider the probability of generating a fragment of every possible length (with a non-trivial probability) from every position on every transcript. This results in a process that is quadratic in the length of the transcriptome — though each evaluation itself is efficient and the process is highly parallelized.

This indicates that Salmon has a quadratic run time relative to the size of the index transcriptome, which means that if we use an organism that has a smaller index transcriptome we should see large speedups.

The second possibility is to make salmon run faster in the general case. After discussing this problem with @jaclyn-taroni, she suggested that we use `salmon`'s [biasSpeedSamp option](http://salmon.readthedocs.io/en/latest/salmon.html#biasspeedsamp) with a value of ~5 which based on the docs:

> It is possible to speed this process up by a multiplicative factor by considering only every ith fragment length, and interploating the intermediate results. The --biasSpeedSamp option allows the user to set this sampling factor. Larger values speed up effective length correction, but may decrease the fidelity of bias modeling. However, reasonably small values (e.g. 10 or less) should have only a minor effect on the computed effective lengths, and can considerably speed up effective length correction on large transcriptomes.

should result in a 5x speedup.

### Solution or next step

Switch the salmon tests to use a organism with a smaller index transcriptome. @jaclyn-taroni can possibly point us in the direction of a good organism.

Also, enable the `biasSampSpeed` option when running salmon with a value of 5.

These two changes should be made together so the impact they have on the time the salmon test takes can be gauged. It seems like we can potentially achieve a >20x speedup, which will probably mean that this test's runtime is no longer problematic. However if its still the slowest part of our test suite then we should be able to reduce it by another 4x by using a `biasSampSpeed` value of 20 while running tests. However switching this parameter based on whether or not its a test adds complexity to the implementation so it should only be done if necessary.

### New Issue Checklist

- [x] The title is short and descriptive.
[...]

Issue #234, opened 2018-04-30
Speed up end-to-end tests


kurtwheeler:
### Context

With the current base image, end to end tests cannot be done quickly because pulling the image and pushing it to a local registry take forever. However with the completion of #55 we will not have such a large image so this can be done quicker.

### Problem or idea

We should become paying clients of CircleCI so that we can parallelize our builds. Once we do so we  should shoot to get the tests running in under a half hour.

### Solution or next step

My initial suggestion for how to parallelize is to have these test threads running simultaneously:
- [ ] One to run the general tests that are fairly lightweight.
- [ ] One to run individual processors.
- [ ] One to run end to end tests.

These may or may not all be necessary or the best way to do this. 

SCAN.UPC is the processor which requires the annotation packages which really blow up the size of our Docker images the most. Therefore if we just make sure we don't run it during end-to-end tests, we won't need to push the image to a local registry which will speed things up a lot. We can instead pull that image and run a processor to test it, which I believe will probably be the longest test we'll have.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

cgreene:
I just looked into this. As an open source project, we do have parallelism already (4 containers). However, it needs to be enabled in the config file:
https://circleci.com/docs/2.0/configuration-reference/#jobs

kurtwheeler:
Oh awesome then.

kurtwheeler:
I dropped [this comment](https://github.com/AlexsLemonade/refinebio/issues/233#issuecomment-385493027) on https://github.com/AlexsLemonade/refinebio/issues/233 which potentially could have sped things up but it doesn't work.

Issue #55, opened 2017-09-19
Split workers image into job-specific images and configure Nomad to use the correct one for each job.


kurtwheeler:
### Context

As we continue to create different `Processors` the `ccdl/dr_workers` Docker image will continue to grow to contain all the dependencies and (at least some of the data) the different jobs need. This makes for a very bulky image that takes a long time to build, upload, and download along with requiring a non-trivial amount of disk space to store.

### Problem or idea

We should have a Docker image for every type of job we have. We should also have different Nomad job specs for each type of job, rather than having a single `processor.nomad` job spec for all of them.

### Solution or next step

Split the workers Docker image into separate images for each job type and create Nomad job specifications for each job type. Note that some jobs may still be able to an image, for example the SCAN.UPC image should be usable by both Affymetrix and Illumina specific jobs types.

Finally, change `data_refinery_common.message_queue.send_job` function  to specify job types directly to Nomad rather than specifying the job type via the job parameters.

jaclyn-taroni:
We just talked about what an "RNA-seq" Docker image _might_ contain: `RSEM`, `Salmon`, `salmontools`, `tximport` and for QC purposes `FastQC`, `MultiQC`

Just recording this conversation here, which is to say this is not a substitute for looking at the code for the processor!

PR #262, opened 2018-05-17
Split docker images and parallelize CircleCI 

kurtwheeler:
## Issue Number

#55 #234 

I know we like PRs to be specific to a single issue, but splitting the Docker images ended up making the tests very unwieldy so I parallelized the circle jobs so that the jobs would run much quicker.

The worker Docker image was huge and getting bigger as we jammed more dependencies into it. This was making things more difficult operationally because any time you wanted to do anything with the Docker image it was slow to build, push, or pull.

We still have a big Docker image, because the R dependencies for Affymetrix are large. However I have added some special handling for it so that it doesn't bog down CircleCI tests any more than strictly necessary. Essentially, I have made it so that there's no scripts that will automatically build that image. If you want to build that image, you need to be explicit about doing so.

This is a big PR, however a lot of the code isn't new. Whereas before we had one big Dockerfile for the workers and we only needed two nomad job templates, now we have one Dockerfile and one nomad job template for each kind of job. A lot of the other code changes are related to handling these extra images.

The changes to CircleCI stuff ended up not being too involved. I utilized Circle's workflows feature to create four jobs which each have their own workflow so they can be run simultaneously. The latest workflow that finished had the following timings for the different jobs:
* affymetrix_tests: 14:34
* main_tests: 12:05
* salmon_tests: 29:01
* transcriptome_and_downloaders_tests:17:08

This means that the `salmon_tests` is the test that we'll have to wait on. However #256 has two ideas which should speed up salmon. I implemented one of them in this PR because it was easy and made the iteration times much faster.  The additional idea should result in another serious reduction in the runtime for the salmon test, I'm hoping for at least 2x. This should  mean that our slowest test becomes the transcriptome_and_downloaders_tests, which means that a full build should be possible in ~17 minutes.

## Types of changes

## Functional tests

I have spun up the infrastructure and run jobs in the cloud.
I have also run jobs locally all the way through for every kind of processor.

## Checklist


kurtwheeler:
This PR makes the #robots slack channel much more chatty because each commit now triggers four builds instead of one. Unfortunately there's not yet a way to get only one workflow notification, although there is a feature request for it: https://discuss.circleci.com/t/notifications-for-workflows/14452

kurtwheeler:
> Is there any reason why we don't have a common nomad job base? We have 6 nearly-identical job spec template files now. It'd be great if we could just template out the part that changes, else we have to update tons of files whenever we add a new envvar, etc..

This is a good point, I'll make that change. Should reduce the line count of this PR by quite a bit too.

workers/dockerfiles/Dockerfile.affymetrix
```diff
@@ -0,0 +1,64 @@
+FROM ubuntu:16.04
+
+# Much of this dockerfile is based off of the rpy2 docker image found:
+# https://hub.docker.com/r/rpy2/rpy2/~/dockerfile/
+
+RUN \
+  apt-get update -qq && \
+  apt-get install -y lsb-release && \
+  echo "deb http://cran.cnr.berkeley.edu/bin/linux/ubuntu $(lsb_release -sc)/" \
```
dongbohu: I have been wondering why we used the mirror site at UCB. Wouldn't a mirror site on east coast (such as CMU or Duke) faster?
https://cran.r-project.org/mirrors.html

workers/dockerfiles/Dockerfile.illumina
```diff
@@ -0,0 +1,56 @@
+FROM ubuntu:16.04
+
+RUN \
+  apt-get update -qq && \
+  apt-get install -y lsb-release && \
+  echo "deb http://cran.cnr.berkeley.edu/bin/linux/ubuntu $(lsb_release -sc)/" \
```
dongbohu: Same question on why we are using UCB mirror site.

workers/dockerfiles/Dockerfile.downloaders
```diff
@@ -0,0 +1,50 @@
+FROM ubuntu:16.04
+
+RUN \
+  apt-get update -qq && \
+  apt-key adv --keyserver keyserver.ubuntu.com --recv-keys E084DAB9 && \
```
dongbohu: This line is needed only for the latest R installation, I believe.

workers/dockerfiles/Dockerfile.no_op
```diff
@@ -0,0 +1,39 @@
+FROM ubuntu:16.04
+
+RUN \
+  apt-get update -qq && \
+  apt-key adv --keyserver keyserver.ubuntu.com --recv-keys E084DAB9 && \
```
dongbohu: For R installation only.

workers/dockerfiles/Dockerfile.transcriptome
```diff
@@ -1,48 +1,23 @@
...
 
-# Much of this dockerfile is based off of the rpy2 docker image found:
-# https://hub.docker.com/r/rpy2/rpy2/~/dockerfile/
-
 RUN \
   apt-get update -qq && \
-  apt-get install -y lsb-release && \
-  echo "deb http://cran.cnr.berkeley.edu/bin/linux/ubuntu $(lsb_release -sc)/" \
-      >> /etc/apt/sources.list.d/added_repos.list && \
   apt-key adv --keyserver keyserver.ubuntu.com --recv-keys E084DAB9 && \
```
dongbohu: For R installation only.

workers/dockerfiles/Dockerfile.transcriptome
```diff
@@ -75,9 +50,22 @@ RUN curl -k -L https://github.com/COMBINE-lab/salmon/archive/${SALMON_COMMIT_HAS
 ENV PATH /home/user/salmon-${SALMON_COMMIT_HASH}/bin:${PATH}
```
dongbohu: The installation steps for salmon seem a little out-dated. The latest version is 0.9.1:
https://github.com/COMBINE-lab/salmon/releases
Since `-DCMAKE_INSTALL_PREFIX=/usr/local` option is used in cmake, the binary will be installed in /usr/local/bin, so line 50 is not needed.

workers/dockerfiles/Dockerfile.salmon
```diff
@@ -0,0 +1,67 @@
...
+
+RUN curl -k -L https://github.com/COMBINE-lab/salmon/archive/${SALMON_COMMIT_HASH}.zip \
+         -o salmon-${SALMON_COMMIT_HASH}.zip && \
+    unzip salmon-${SALMON_COMMIT_HASH}.zip && \
+    cd salmon-${SALMON_COMMIT_HASH} && \
+    mkdir build && \
+    cd build && \
+    cmake .. -DCMAKE_INSTALL_PREFIX=/usr/local && make && make install
+
+ENV PATH /home/user/salmon-${SALMON_COMMIT_HASH}/bin:${PATH}
```
dongbohu: See my comments earlier about salmon installation.

workers/dockerfiles/Dockerfile.salmon
```diff
@@ -0,0 +1,67 @@
...
+
+# Starting here, code is based on:
+# https://github.com/COMBINE-lab/salmon/blob/master/docker/Dockerfile
+ENV PACKAGES gcc make g++ cmake libboost-all-dev liblzma-dev libbz2-dev \
+    ca-certificates zlib1g-dev unzip autoconf
+
+# Version 0.8.2 has an outdated sha hash which breaks its make command.
+# I fixed this in the lastest commit, but that hasn't been cut to a
+# new version yet.
+ENV SALMON_COMMIT_HASH 0b9f1eb8a0a321adc6320f490b015e75ebf3d3bb
```
dongbohu: See my comments on the latest version of salmon.

workers/dockerfiles/Dockerfile.transcriptome
jaclyn-taroni: I agree that we should update Salmon to the latest version.

kurtwheeler:
> Is there any reason why we don't have a common nomad job base? We have 6 nearly-identical job spec template files now. It'd be great if we could just template out the part that changes, else we have to update tons of files whenever we add a new envvar, etc..

Ok so I dug into using [levant](https://github.com/jrasell/levant) to make this scenario better and it would be a good thing to do. However I think it would be a non trivial amount of work so I suggest we make another issue for it and address it after we process a million samples.

Miserlou:
I review this with Kurt today and I think it looks pretty good after the few improvements we made together. Once it gets rebased on dev and the tests all pass on CI this can be merged.

PR #263, opened 2018-05-17
Adds RDS Parameter Groups and Instance Type Variable

Miserlou:
## Issue Number

n/a

Connecting to the API during heavy scrape can cause `django.db.utils.OperationalError: FATAL:  remaining connection slots are reserved for non-replication superuser connections`.

This adds a) an RDS Parameter Group so we can configure the timeout and b) variablizes a hard-coded RDS instance type string so that we can beef it up to something better that a t2.micro in production.

This actually doesn't solve the underlying problem, but it does make it rarer. The correct solution in the future will be to have separate database users for API and Foreman/Worker instances, and then to reserve a minimum number of connections just for the API user (which should also have PgBouncer installed).

## Types of changes

## Functional tests

Tested, and got this result:

Before:
screen shot 2018-05-17 at 3 58 22 pm

After:
screen shot 2018-05-17 at 4 01 11 pm

## Checklist

### DIFF
--- a/infrastructure/instances.tf
+++ b/infrastructure/instances.tf
@@ -348,12 +348,12 @@ resource "aws_db_parameter_group" "postgres_parameters" {
 
   parameter {
     name = "deadlock_timeout"
-    value = "60000" # 60000ms = 60m
+    value = "60000" # 60000ms = 60s
   }
 
   parameter {
     name = "statement_timeout"
-    value = "60000" # 60000ms = 60m
+    value = "60000" # 60000ms = 60s
   }
 }
 
--- a/infrastructure/instances.tf
+++ b/infrastructure/instances.tf
@@ -341,23 +341,43 @@ resource "aws_autoscaling_policy" "clients_scale_down" {
 # Database
 ##
 
+resource "aws_db_parameter_group" "postgres_parameters" {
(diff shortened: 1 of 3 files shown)


infrastructure/instances.tf
```diff
@@ -341,23 +341,43 @@ resource "aws_autoscaling_policy" "clients_scale_down" {
...
 ##
 
+resource "aws_db_parameter_group" "postgres_parameters" {
+  name = "postgres-parameters-${var.user}-${var.stage}"
+  description = "Postgres Parameters ${var.user} ${var.stage}"
+  family = "postgres9.6"
+
+  parameter {
+    name = "deadlock_timeout"
+    value = "60000" # 60000ms = 60m
```
kurtwheeler: 60000ms = 1m

infrastructure/instances.tf
```diff
@@ -341,23 +341,43 @@ resource "aws_autoscaling_policy" "clients_scale_down" {
...
+  }
+
+  parameter {
+    name = "statement_timeout"
+    value = "60000" # 60000ms = 60m
```
kurtwheeler: Same comment. If you want 60m then you need to specify `3600000`.

Issue #273, opened 2018-05-22
Salmon output files will be overwritten by other samples


kurtwheeler:
### Context

The output directory for salmon jobs is on a per-experiment basis, however the processor jobs are run on a per-sample basis. This means that as each processor job finishes, it will overwrite the `quant.sf` file that was output by any jobs for that experiment which already finished.

### Problem or idea

We need to create per-sample or per-job output directories. The tricky part to this is making them discoverable from either ComputationalResult or ComputedFile objects.

### Solution or next step

Create separate directories for each salmon job so that files won't conflict. Find a way to track these down from database objects.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

dongbohu:
Are you saying that  the output files were saved in this directory? 
```<experiment_name>/processed/``` 

I thought they were in: 
```<experiment_name>/<sample_name>/processed/```

**Comment by kurtwheeler** [comment]
Sorry, I wanted to run a test to verify this and I forgot to come back and actually answer this question. `<experiment_name>/<sample_name>/processed/` does not seem to be the case:

```
$ ls volume/ERP016796/processed/
aux_info  cmd_info.json  lib_format_counts.json  libParams  logs  quant.sf
```

dongbohu:
Never mind. I confirmed this from Jackie yesterday. So I deleted my question. I am going to fix it in the upcoming PR.

kurtwheeler:
Oh sorry, I hadn't refreshed the page. Awesome that you're gonna fix it though!

dongbohu:
Fixed in PR #278.

Issue #243, opened 2018-05-02
Add a unit test for tximport


jaclyn-taroni:
### Context

Once `tximport` is implemented (#82), we'll need a unit test to ensure that we don't break this processor such that we get output that is dramatically different from what has been produced in the past.

### Problem or idea

There are a few challenges to keep in mind if this is part of an "end-to-end" test of the RNA-seq processing pipeline: 1) the genome version may change (at which point we probably need to update our test data)  2) the version of Salmon may change (which may affect either the preparation of the index or the quantification) and 3) `salmon quant` is nondeterministic. So, it will not be sufficient to run test data and compare the checksums.

We'll probably want something similar to what is done for the Affymetrix SCAN processor:
https://github.com/AlexsLemonade/refinebio/blob/60fdd5a4e2ed5a02ff466888914f4a5874ebfde5/workers/data_refinery_workers/processors/test_array_express.py#L206

And what Kurt mentioned [here](https://github.com/AlexsLemonade/refinebio/issues/117#issue-296499941).

Specifically, we'll want to make sure that the test data is highly correlated with the reference data.

### Solution or next step

To be revisited once #82 has been accomplished.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

dongbohu:
@jaclyn-taroni: Given a fixed set of `quant.sf` and the gene2txmap input files, how likely is the R package `tximport` (not our own wrapper script in `refinebio/workers/data_refinery_workers/processors/tximport.R`) going to generate different output files? I think this will only happen when `tximport` package changes its implementation. Since our Docker image of `salmon` pipeline will explicitly state the version number of `tximport`, I think we can just compare the checksums of the two output files with the expected output.

The issues you guys were talking about should be addressed by the unit tests of `salmon quant`, not `tximport`.

jaclyn-taroni:
> That is more like an integrated test than a unit test.

Gotcha, good to know 👍

dongbohu:
Done in PR #278.

Issue #82, opened 2018-01-16
Implement tximport for RNA-seq pipeline


kurtwheeler:
Use tximport along with the gene-to-transcript mapping already contained within the processor to implement this part of the salmon pipeline:

![gene_level_counts](https://user-images.githubusercontent.com/4026833/34995199-bd84f75c-faa3-11e7-8ee3-e28ee09b52a9.jpg)

This code does essentially what we need, we just need to include this into the Data Refinery's salmon pipeline: https://github.com/jaclyn-taroni/ref-txome/blob/79e2f64ffe6a71c5103a150bd3159efb784cddeb/4-athaliana_tximport.R
(Note that this script contains a link to a tutorial.)

Note that this should be done on a per-experiment basis, rather than a per-sample basis.

jaclyn-taroni:
If, in the future, we would like to enable users to perform differential gene expression analysis for an experiment processed through the refinery, we should consider generating files (e.g., [.RDS](https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=1&ved=0ahUKEwiUvp7Pg7PZAhVFpFkKHd2xCxYQFggpMAA&url=https%3A%2F%2Fstat.ethz.ch%2FR-manual%2FR-devel%2Flibrary%2Fbase%2Fhtml%2FreadRDS.html&usg=AOvVaw1sM0iizyqHpPKgF1nStYm6) or .Rda files) suitable for use with [`DESeq2::DESeqDataSetFromTximport`](https://www.rdocumentation.org/packages/DESeq2/versions/1.12.3/topics/DESeqDataSet-class). 

If so, we should probably run `tximport::summarizeToGene` twice:
1. with `countsFromAbundance = "no"` -- the resulting object is to be saved for use with `DESeq2` (also suitable for use with edgeR, will contain offset info)
2. with `countsFromAbundance = "lengthScaledTPM"`, as we're thinking that this will likely be most useful for downstream ML/data integration applications with some additional processing (also can be used with limma-voom)

See [the differential gene expression section of the tximport vignette](https://bioconductor.org/packages/devel/bioc/vignettes/tximport/inst/doc/tximport.html#use-with-downstream-bioconductor-dge-packages) for more information.

Running both ways will allow us to be the most flexible. It's also entirely possible that lengthScaledTPM (option 2) can be obtained using option 1, would require more thought on my part, though going that route may not be desirable. I wanted to get these initial thoughts down for the moment.

[...]

PR #278, opened 2018-05-23
Tximport in Salmon Pipeline

dongbohu:
## Issue Number
#82 
#243
#273

This PR is much larger than I expected. It includes:
- tximport data analysis in `Salmon` pipeline
- a unit test of `tximport`
- updated docker images for salmon pipeline
- #273 is fixed as a byproduct. 

## Types of changes
- new features

## Functional tests
Tested by the unit test in this PR

## Checklist


dongbohu:
@jaclyn-taroni: As you can see, I modified your R script a little bit. The main changes are:
- Add logic to handle the two arguments
- Changed the way to find `quant.sf` files and the corresponding sample names (due to the structure of files generated by the download and `salmon quant` step.

workers/data_refinery_workers/processors/tximport.R
```diff
@@ -0,0 +1,63 @@
...
+# <exp_dir>/<sample>/processed/quant.sf
+sf.files <- list.files(exp.directory, pattern="^quant.sf$", full.names = TRUE, recursive = TRUE)
+
+# Name each "sf.files" entry with the corresponding sample's name.
+names(sf.files) <- samples
+
+# Generate the gene to transcript mapping based on the input file
+# specified by "--gene2txmap" option:
+gene2tx.file <- opt$gene2txmap
+gene2tx.df <- readr::read_tsv(gene2tx.file, col_names = c("gene_id", "tx_name", "DROP"))
```
jaclyn-taroni: The way we're doing this now, there shouldn't be trailing tabs. So I think this should do it:
```R
gene2tx_df <- readr::read_tsv(gene2tx_file, col_names = c("gene_id", "tx_name))
```
And you won't need line 40.

workers/data_refinery_workers/processors/tximport.R
```diff
@@ -0,0 +1,63 @@
...
+# Two required arguments:
+# (1) --exp_dir    <experiement_dir> (directory where "quant.sf" files are found)
+# (2) --gene2txmap <path_of_gene2txmap_file>
+#
+# Two Output files:
+# (1) RDS file (saved as <experiment_dir>/txi_out.RDS)
+# (2) TPM file (saved as <experiment_dir>/gene_lengthScaledTPM.tsv)
+
+# Handle input arguments:
+library("optparse")
```
jaclyn-taroni: Where possible, I think we should go the [double colon route](https://stackoverflow.com/questions/23232791/is-it-a-good-practice-to-call-functions-in-a-package-via) (e.g., `optparse::`) rather than loading/attaching packages (to avoid namespace collisions). But it's only one library, so probably not a big deal. 

Below `make_option()` would become `optparse::make_option()`, `optparse::OptionParser`, etc.

workers/data_refinery_workers/processors/tximport.R
```diff
@@ -0,0 +1,63 @@
...
+option_list <- list(
+  make_option("--exp_dir", type = "character"),
+  make_option("--gene2txmap", type = "character")
+)
+
+opt_parser <- OptionParser(option_list = option_list)
+opt <- parse_args(opt_parser)
+
+# directory of the experiment data
+exp.directory <- opt$exp_dir
```
jaclyn-taroni: I think we should keep the style consistent and have this become `exp_directory`

workers/data_refinery_workers/processors/tximport.R
```diff
@@ -0,0 +1,63 @@
...
+
+# Each sample's data is in a sub-dir of experiment directory.
+samples <- list.dirs(exp.directory, full.names = FALSE, recursive = FALSE)
+
+# Get a list of "quant.sf" files in experiment directory.
+# The path of each "quant.sf" are in this format:
+# <exp_dir>/<sample>/processed/quant.sf
+sf.files <- list.files(exp.directory, pattern="^quant.sf$", full.names = TRUE, recursive = TRUE)
```
jaclyn-taroni: Same thing here, if we want to make this more pythonic style wise `sf_files`

workers/data_refinery_workers/processors/tximport.R
```diff
@@ -0,0 +1,63 @@
...
+gene2tx.file <- opt$gene2txmap
+gene2tx.df <- readr::read_tsv(gene2tx.file, col_names = c("gene_id", "tx_name", "DROP"))
+gene2tx.df <- dplyr::select(gene2tx.df, -DROP)
+# Reorder columns into: transcript, gene (which is required for tximport)
+tx2gene <- gene2tx.df[, c("tx_name", "gene_id")]
+
+# Run tximport and save RDS
+txi.out <- tximport::tximport(files = sf.files,
+                              type = "salmon",
+                              txOut = TRUE,
```
jaclyn-taroni: Based on the discussion over on #82, I think this should be `txOut  = FALSE`. (I would probably change the variable name to `txi` in that case.) Doing that & saving the RDS would mean we have:

* gene-level TPM
* gene-level counts
* gene-level offset [appropriate for use with differential expression analysis](http://bioconductor.org/packages/release/bioc/vignettes/tximport/inst/doc/tximport.html#use-with-downstream-bioconductor-dge-packages)

The way we have it here gives us [transcript-level estimates as a list of matrices](http://bioconductor.org/packages/release/bioc/vignettes/tximport/inst/doc/tximport.html#import-transcript-level-estimates), which we could extract from the `quants.sf` file should folks want that.

workers/data_refinery_workers/processors/tximport.R
```diff
@@ -0,0 +1,63 @@
...
+                              tx2gene = tx2gene,
+                              countsFromAbundance = "no")
+rds_filename <- file.path(exp.directory, "txi_out.RDS")
+saveRDS(txi.out, file = rds_filename)
+
+# Run tximport::summarizeToGene() and save length-scaled
+# TPM (transcripts per million) file.
+txi.sum <- tximport::summarizeToGene(txi = txi.out,
```
jaclyn-taroni: If we make the above change, I think it makes more sense to change this to
```R
txi_length_scaled <- tximport::tximport(files = sf_files,
                                        type = "salmon", 
                                        tx2gene = tx2gene,
                                        countsFromAbundance = "lengthScaledTPM")
```

workers/data_refinery_workers/processors/tximport.R
```diff
@@ -0,0 +1,63 @@
...
+                                     tx2gene = tx2gene,
+                                     countsFromAbundance = "lengthScaledTPM")
+# As data.frame, write to file
+lstpm.df <- as.data.frame(txi.sum$counts)
+lstpm.df <- cbind(rownames(lstpm.df), lstpm.df)
```
jaclyn-taroni: If there will be [tidyverse](https://www.tidyverse.org) packages (like `readr`, which is used above) on the Docker image, I believe we can convert lines 60-61 to

```R
lstpm_df <- tibble::rownames_to_column(df, var = "Gene")
```

workers/data_refinery_workers/processors/tximport.R
dongbohu: Sure.

workers/data_refinery_workers/processors/tximport.R
dongbohu: Sure.

workers/data_refinery_workers/processors/tximport.R
dongbohu: Sure.

workers/data_refinery_workers/processors/tximport.R
dongbohu: Sounds good.

dongbohu: You mean: 
```lstpm_df <- tibble::rownames_to_column(lstpm_df, var = "Gene")```
right? (since the first argument `df` is an undefined variable)

**Comment by dongbohu** [comment]
By the way, don't worry about the `All checks have failed` warning. I cancelled the CircleCI tests to save some resource for Kurt and Rich, because this PR has nothing to do with the other tests (yet).

**Comment by jaclyn-taroni** [comment]
Changes LGTM 👍 , thanks @dongbohu !

File: workers/data_refinery_workers/processors/tximport.R [hunk 8]
```diff
@@ -0,0 +1,63 @@
+# This script is based on the one written by J. Taroni:
```
dongbohu: Jackie has reviewed this file. So please feel free to skip this one.

workers/dockerfiles/Dockerfile.salmon
```diff
@@ -23,41 +30,29 @@ WORKDIR /home/user
 
 # Install Salmon
 ENV SALMON_VERSION 0.9.1
-
 RUN wget https://github.com/COMBINE-lab/salmon/releases/download/v${SALMON_VERSION}/Salmon-${SALMON_VERSION}_linux_x86_64.tar.gz
-
 RUN tar -xzf Salmon-${SALMON_VERSION}_linux_x86_64.tar.gz
-
-# Salmon can extract to a different directory than the name of the tar file.
```
dongbohu: The main reason for this change: when the pre-compiled salmon library files were copied into /usr/local/lib/, they have higher priorities than the library files in `/usr/lib/`, and messed up both JRE and Rscript. Alternatively, we can compile the source code of `salmon`, but they requires boost library, which makes the docker at least 150MB image larger. So I tweaked the PATH (see below) instead.

workers/dockerfiles/Dockerfile.salmon
```diff
@@ -23,41 +30,29 @@ WORKDIR /home/user
...
-RUN cp `tar -tzf Salmon-${SALMON_VERSION}_linux_x86_64.tar.gz | head -1 | cut -f1 -d"/"`/bin/salmon /usr/local/bin
-
-RUN cp `tar -tzf Salmon-${SALMON_VERSION}_linux_x86_64.tar.gz | head -1 | cut -f1 -d"/"`/lib/* /usr/local/lib
-
-RUN rm -r Salmon*
+RUN rm -f Salmon-${SALMON_VERSION}_linux_x86_64.tar.gz
 # End Salmon installation.
 
-# Package list based on:
```
dongbohu: I removed these packages because these packages were required to compile salmon, and we are using pre-compiled version. (As a matter of fact, most of these packages have already been installed implicitly because they were the dependencies of `r-base`.)

workers/dockerfiles/Dockerfile.salmon
```diff
@@ -23,41 +30,29 @@ WORKDIR /home/user
...
+# Install SalmonTools
 RUN git clone https://github.com/COMBINE-lab/SalmonTools.git
 RUN cd SalmonTools && cmake . -DCMAKE_INSTALL_PREFIX=/usr/local && make install
 
 RUN pip3 install --upgrade pip
 
 # We need JRE for QC tools
-RUN apt-get update -y
-RUN apt-get install -y unzip default-jre --fix-missing
+RUN apt-get update -y && apt-get install -y default-jre --fix-missing
```
dongbohu: I removed `unzip` here because it is already installed implicitly as a dependency of `r-base`.

workers/dockerfiles/Dockerfile.salmon
```diff
@@ -69,7 +64,7 @@ COPY common/dist/data-refinery-common-* common/
 RUN pip3 install common/$(ls common -1 | sort --version-sort | tail -1)
 
 USER user
-
+RUN echo "export PATH=`pwd`/Salmon-latest_linux_x86_64/bin:$PATH" >> ~/.bashrc
```
dongbohu: I didn't put it in `~/.profile` or `~/.bash_profile` because we would like the new PATH to be effective in all bash sessions (not only in BASH login shell).

dongbohu:
I added some extra comments in the PR to (hopefully) make the review easier.

dongbohu:
@jaclyn-taroni @kurtwheeler and @Miserlou: This PR is ready for review now! Thanks.

workers/data_refinery_workers/processors/salmon.py
```diff
@@ -373,6 +396,58 @@ def _run_fastqc(job_context: Dict) -> Dict:
...
+    cmd_tokens = [
+        "/usr/bin/Rscript", "--vanilla",
+        "/home/user/data_refinery_workers/processors/tximport.R",
+        "--exp_dir", exp_dir,
+        "--gene2txmap", genes_to_transcripts_path
+    ]
+    # TODO: Because this job is done at experiment level, it doesn't make
+    # sense to update job_context (which is at sample level right now).
+    # The DB model may have to be modified to associate this ComputationalResult
+    # record with the experiment (instead of samples in the experiment).
```
kurtwheeler: I disagree that we should modify the DB model. There's a many to many relationship between Samples and ComputationalResults for exactly this use case. I think every Sample from this experiment should be connected to this result.

workers/data_refinery_workers/processors/salmon.py
```diff
@@ -373,6 +396,58 @@ def _run_fastqc(job_context: Dict) -> Dict:
...
+        result.time_end = timezone.now()
+        result.command_executed = " ".join(cmd_tokens)
+        result.system_version = __version__
+        result.is_ccdl = True
+        result.save()
+    except Exception as e:
+        error_template = ("Encountered error in R code while running tximport.R"
+                          " pipeline during processing of {0}: {1}")
+        error_message = error_template.format(exp_dir, str(e))
+        logger.error(error_message, processor_job=job_context["job_id"])
```
kurtwheeler: I think we should mark the job as a failure if this doesn't succeed. With the way this is right now, if this fails the only way we'll have any idea is if we see this error message in the logs.

workers/dockerfiles/Dockerfile.salmon
kurtwheeler: Wait I'm confused, I don't actually see you tweaking the PATH at all. It looks like you're making `/usr/local/bin` into a softlink which points to the `bin` that came with salmon? It seems like if you wanted to add salmon's bin to the PATH you could just append it to the `$PATH` environment variable right? It also seems like another potential solution could be change `$PATH` to prioritize `/usr/lib` over `/usr/local/lib`?

No matter how we end up doing this, I do think the explanation should be a comment in the code instead of on the PR so that we can see it if we ever go to modify this later.

dongbohu: Originally I played with the $PATH, then I realized that no matter $PATH is defined in ~/.bashrc or ~/.profile, it will be effective only in BASH, so adding a soft link in `/usr/local/bin/` seems better. (Generally it is not a good idea to tweak the dynamic library path, because it may affect other programs.)

workers/data_refinery_workers/processors/salmon.py
dongbohu: Sure. I can add a comment line about it.

dongbohu: I didn't do it because even if `tximport` fails for whatever reason, the jobs that follows (salmontools, multiqc, etc) should still continue.

workers/data_refinery_workers/processors/salmon.py
dongbohu: One reason why I don't like this idea is because when a sample belongs to multiple experiments, then a sample will be associated with multiple tximport computational results. This seems confusing.

workers/dockerfiles/Dockerfile.salmon
dongbohu: I will add a comment line in this file.

workers/data_refinery_workers/processors/salmon.py
kurtwheeler: I don't know that I agree that they should, however I'm not sure. This actually seems like a good question for @jaclyn-taroni: if `tximport` fails for some reason, would we prefer to have the rest of the pipeline continue or prevent that sample/experiment from being added to the compendium? It's not entirely clear to me how critical `tximport` is to have for a sample.

However, regardless of the answer to that question, I think we should probably fail the jobs at least for a while until we run into some errors that we can't find a workaround for, and then only let that kind of error through. The reason I think this is that as we run this on millions of samples, we're going to run into tons of edge cases which will cause failures. The way we're gonna debug these will be to look through our jobs/samples/experiments and figure out which didn't make it all the way through the system successfully and then track down what went wrong using the logs. We probably won't be going through every error message in the log files trying to figure out what happened, so if a job doesn't fail it won't make it onto our radar, and we won't realize that it failed for a fixable reason.

workers/data_refinery_workers/processors/salmon.py
kurtwheeler: Hmm, that's a good point. I'm not sure which I like less. Maybe @Miserlou has thoughts on this?

workers/data_refinery_workers/processors/salmon.py
jaclyn-taroni: > if tximport fails for some reason, would we prefer to have the rest of the pipeline continue or prevent that sample/experiment from being added to the compendium? It's not entirely clear to me how critical tximport is to have for a sample.

Hmm, this is a great question. I would say at this stage (Keytar Kurt) just fail the job, as right now (to my knowledge) the only things made available to the users are 1) count-scale output of `tximport` and 2) `multiqc` reports. In the future, we will probably want to allow users to obtain different outputs from the pipeline (e.g., transcript-level abundances from `salmon quant`, etc).

jaclyn-taroni: Sounds good, thanks @dongbohu!

workers/data_refinery_workers/processors/salmon.py
kurtwheeler: I think we need to connect it to SOMETHING, as it is here these ComputationalResults can't even be queried for except by the `command_executed` field. I'm cool with picking whichever option is easier and going with that, opening an issue about it, and figuring out the best solution later, but I think we need to be able to find the `tximport` ComputationalResult for a given sample/experiment for these to be useful at all.

dongbohu: Unless we change the DB schema, the only table that  it can associate with right now is `SampleResultAssociate`. 

Suppose a user wants to find all experiments that have been processed by `tximport`. With the current DB schema, I can think of two approaches with pseudo-code like:
- Assume we have associated the tximport ComputationalResult record with each sample:
```Experiment.filter(samples[0].results.filter(command_executed.startwith('tximport ... <experiment_accession_code>).distinct()```

- Or: 
```
for each experiment:
    ComputationalResults.objects.filter(command_executed.startwith('tximport ... <experiment_accession_code>)
Then find the experiment based on <experiment_accession_code>
```
Neither of them is very satisfying to me.

dongbohu: Of course, if this is not a valid user case, then we can just link this result with the samples to keep track of it w/o modifying the DB at all.

dongbohu: Another question is: In the foreseeable future, is `tximport` going to be the only analysis that is at experiment level? If it is, then we can just treat as as a special case. If we will have other experiment-level jobs, then it is worth modifying the DB. Any thoughts, @jaclyn-taroni?

cgreene: We have tried, to the extent possible, to select methods that work on single samples because it streamlines the processing greatly. I don't think I could guarantee that we won't have anything else, but we are trying to avoid it to the extent possible.

dongbohu: Got it. I will associate this computational result record with all related samples for now. Although it may not help the query that much, at least it doesn't hurt. I will also add an issue to keep track of it.

dongbohu:
@jaclyn-taroni @kurtwheeler and @Miserlou: I think I have addressed all your comments and fixed all conflicts with `dev` branch. Please let me know if you have other questions or concerns. Thanks.

jaclyn-taroni:
I have no additional comments

kurtwheeler:
Looks good to me, good work Dongbo!

PR #281, opened 2018-05-24
Changes psycopg2 to psycopg2-binary to remove the warning about it. A…

kurtwheeler:
…lso kills a couple pip version warnings.

Every time we start a container the message:

```
/usr/local/lib/python3.5/dist-packages/psycopg2/__init__.py:144: UserWarning: The psycopg2 wheel package will be renamed from release 2.8; in order to keep installing from binary please use "pip install psycopg2-binary" instead. For details see: <http://initd.org/psycopg/docs/install.html#binary-install-from-pypi>.
```

Is the first thing out of the container. Some of our processors don't have many logs, so this message can for sections of our log files be about 25% of the log messages. This is really annoying so I'm finally fixing this the right way.

Also I got really sick of seeing
```
You are using pip version 9.0.1, however version 10.0.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
```
when building Docker images so I have also fixed this in `api` and `foreman.

## Types of changes

## Functional tests

I've build the `api` and `foreman` images and ran their tests locally. The workers' tests take a while longer so I'm going to rely on CI to do those for me. However I have rebuilt and pushed the `ccdl/dr_affymetrix:remove-psycopg2-warnings` image so that the CI tests will use the new image.

## Checklist

### DIFF
--- a/api/requirements.txt
+++ b/api/requirements.txt
@@ -22,7 +22,8 @@ itypes==1.1.0             # via coreapi
 jinja2==2.10              # via coreschema
 jmespath==0.9.3           # via boto3, botocore
 markupsafe==1.0           # via jinja2
-psycopg2==2.7.4
+psycopg2-binary==2.7.4
+psycopg2==2.7.4           # via djangorestframework-hstore
 python-dateutil==2.6.1    # via botocore
 python-nomad==0.6.1
 pytz==2018.3              # via django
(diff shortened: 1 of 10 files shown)


api/requirements.txt
```diff
@@ -22,7 +22,8 @@ itypes==1.1.0             # via coreapi
 jinja2==2.10              # via coreschema
 jmespath==0.9.3           # via boto3, botocore
 markupsafe==1.0           # via jinja2
-psycopg2==2.7.4
+psycopg2-binary==2.7.4
+psycopg2==2.7.4           # via djangorestframework-hstore
```
Miserlou: Rm this line since we have the bin version?

kurtwheeler: I'm not sure why `djangorestframework-hstore` wants it, but this no longer triggers the warning so since it works I'd rather not mess with it.

PR #267, opened 2018-05-21
Add Methods section to PR template

jaclyn-taroni:
## Issue Number

N/A

## Purpose/Implementation Notes

I am proposing the addition of a "methods section" to the pull request template. I don't think this will be necessary for all PRs, just for those that have important implications for data or metadata processing.

I believe this will have the following benefits (in no particular order):

* Timeliness of review from experts who are not developers when it is required
* Transparency for a wider audience of users and/or contributors
* Smoothes the way for any future scientific communication or documentation (as well as project planning in general, I'd imagine)

Here are some examples of the kinds of information that should be included:

* A description of any data processing steps (sequentially/chronologically)
* What data sources (e.g., ArrayExpress, SRA) and data types (e.g., microarray) does this work on?
* What assumptions were made? What are the limitations?
* What are the expected inputs and outputs?

It does not have to be very long. Here's an example of what I think is appropriate: https://github.com/AlexsLemonade/refinebio/pull/252#issuecomment-390286190 The notes around gender are an example of an important limitation. Note that the data sources are mentioned in the original post.

## Types of changes

N/A

## Functional tests

N/A

## Checklist

_Put an `x` in the boxes that apply._

- [ ] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

## Screenshots

N/A

Issue #222, opened 2018-04-26
Normalize platform accessions across ArrayExpress, GEO, and SRA


jaclyn-taroni:
### Context

We want to facilitate the display of platform information, the ability to filter by platform, and the scaling across an entire platform (chip). This means we have to identify each chip/sequencing instrument, which may have aliases within and across repositories.

Using **`Agilent 4x44K`** as an example:

* **ArrayExpress**
	* [`A-AGIL-28`](https://www.ebi.ac.uk/arrayexpress/arrays/A-AGIL-28/) on ArrayExpress
	* May be related (requires more investigation): [`A-GEOD-13497`](https://www.ebi.ac.uk/arrayexpress/arrays/A-GEOD-13497), [`A-GEOD-16876`](https://www.ebi.ac.uk/arrayexpress/arrays/A-GEOD-16876), [`A-MEXP-2183`](https://www.ebi.ac.uk/arrayexpress/arrays/A-MEXP-2183)
* **Gene Expression Omnibus**
	* [`GPL4133`](https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GPL4133)
	* [`GPL6480`](https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GPL6480)
	* [`GPL9822`](https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GPL9822)
	* [`GPL16022`](https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GPL16022)
	* [`GPL16297`](https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GPL16297)

### Problem or idea

This really has two steps 1) a human manually curates the list (at least until proven otherwise) and 2) the mapping occurs on the backend using said list  

_Note: the web interface for GEO does contain some notion of relationships between platform accessions (see: [`GPL6480`](https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GPL6480)). This information could be obtained when we work on #220. @kurtwheeler is investigating._

### Solution or next step

I will go through the lists generated [here](https://github.com/AlexsLemonade/refinebio/issues/115#issuecomment-364265976) and via #220 to satisfy step 1 above as I think this can be completed in tandem with #228  

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
[...]

Issue #159, opened 2018-03-12
Implement whitelist for sequencing platforms


jaclyn-taroni:
### New Issue Checklist
- [x] The title is short and descriptive
- [x] The issue contains an:
  - [ ] Idea (new feature, user story, etc)
  - [x] Problem
- [x] You have explained the:
  - [x] Context
  - [x] Problem or idea
  - [x] Solution or next step
  
### Context

Salmon, our RNA-seq quantification tool of choice, is not appropriate for every sequencing technology that's out there. For instance, Salmon should not be used for long-read sequencing (e.g. PacBio).

### Problem or idea

We'll need to come up with a whitelist of platforms for use with Salmon (somewhat related: #115 and #145). Here's a screenshot of the platform options in the European Nucleotide Archive search:

![127f0941c48647c492bfd382282a15fe](https://user-images.githubusercontent.com/19534205/37291363-d8a1af1e-25e4-11e8-9273-e751734abbc1.png)

### Solution or next step

Illumina and Ion Torrent are good to go. PacBio and Nanopore are long-read technology. I will need to further investigate the less familiar platforms.

jaclyn-taroni:
For Keytar Kurt, I think we should focus on running Illumina and Ion Torrent.

kurtwheeler:
> Illumina and Ion Torrent are good to go.

Since those are the only two we need for Keytar Kurt I think we should remove this from that milestone. This is still something we're gonna need to do later, but not before Keytar Kurt.

jaclyn-taroni:
So from my "make a list" point of view -- this is closed for now. Should another "implementation" ticket be filed?

jaclyn-taroni:
It occurred to me that a list of machines not be the only info required for this, from #231:

> Essentially, I am asking: Will a list of machines be sufficient to accomplish this? Do we need to create a summary list of sequencing machines from SRA similar to how we're doing so for GEO (#220)?

jaclyn-taroni:
We're interested in _whole_ transcriptome sequencing (RNA-seq). We'll be using the SRA library source metadata to help with this, i.e., only looking at `TRANSCRIPTOMIC` data

**Working off of this list:** https://github.com/AlexsLemonade/refinebio/issues/222#issuecomment-386704996 (note the open question [here](https://github.com/AlexsLemonade/refinebio/issues/222#issuecomment-386709966) as of the writing of this comment)

**I used the following documentation:**

[...]

PR #293, opened 2018-05-30
Kurtwheeler/normalize platforms

kurtwheeler:
## Issue Number

#222 

Platforms are messy. Array Express and GEO both have their own way of assigning accessions to them. This makes determining if a platform is one we support or not difficult. This PR adds some configuration files which have mappings generated by @jaclyn-taroni which specify which platforms we explicitly support, along with some additional information to support converting between different naming/accession systems. These configuration files account for over half the line count of this PR.

## Methods

I'll explain the accession lookups in a non-programmatic way:
* For the SRA surveyor, since we know it's always going to be RNASeq, we just use the platform name provided by SRA. It will be the responsibility of the routing logic to make sure it's supported before queuing downloader jobs for it.
* For the Array Express surveyor, we try to use the platform accession provided by Array Express to lookup our own internal accession. This is so that for experiments without raw data, we can determine if we want to NO_OP them or not. If there is raw data and the platform is Affy, the array express downloader will use the R Affy package to check the .CEL file to make sure that the platform wasn't incorrect.
* For the Array Express surveyor, we try to look up the readable name in the readable_names config file and if the accession is in there we set the platform name to that readable name.
* For Array Express we assume that anything we're getting is Microarray data, so that is always used as the `technology` property for those samples/experiments.
* For GEO, we have to determine the technology along with normalizing the platform. We do this by first checking the supported microarray platforms and using that if it's found. If it's not then we check the supported RNASeq platforms list and use that if it's found. If it's still not found, we just use the accession and title that GEO provides and mark the technology as UNKNOWN.

There's one other complication with GEO which is that the titles of platforms will often contain the organism, despite them being organism agnostic. To handle this, I check to see if the platform title provided by GEO contains the platform names that we support. For example GSE69572 has a platform title of  'Illumina Genome Analyzer IIx (Glycine max)' which should really just be 'Illumina Genome Analyzer IIx'. Since `"Illumina Genome Analyzer" in "Illumina Genome Analyzer IIx (Glycine max)"` would return true, I check the longest platform names first and stop once I find a match. Therefore even though `Illumina Genome Analyzer` would match, `Illumina Genome Analyzer IIx` is what is returned.

## Types of changes

## Functional tests

I've run a handful of different accessions, but I think the existing unit tests along with the ones I added do a pretty good job of covering the functionality.

## Checklist


jaclyn-taroni:
Really nice **Methods** section, @kurtwheeler! And very helpful, too, because it allowed me to identify some things we should to discuss easily:

#### RNA-seq from GEO & ArrayExpress

> For GEO, we have to determine the technology along with normalizing the platform. We do this by first checking the supported microarray platforms and using that if it's found. If it's not then we check the supported RNASeq platforms list and use that if it's found. If it's still not found, we just use the accession and title that GEO provides and mark the technology as UNKNOWN.

As I noted here https://github.com/AlexsLemonade/refinebio/issues/159#issuecomment-387502985:

> Processed RNA-seq data sometimes lives in GEO, where it will often be linked to its raw data counterpart in SRA. Here's an example: https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GSE72536

> Based on my understanding of the system, we are not capturing this relationship. As a result, I propose that we do not `NOOP` any sequencing data (i.e., use these platforms) from GEO or ArrayExpress for Keytar Kurt because we might be capturing a bunch of duplicates. (If we're somehow capturing this relationship at the sample level, I can be persuaded that I am wrong.)

Is what you have here likely to produce lots of duplicates in this case?

> For Array Express we assume that anything we're getting is Microarray data, so that is always used as the `technology` property for those samples/experiments.

This is a case where ArrayExpress has essentially only metadata (e.g, `idf.txt` and `sdrf.txt` files) but is linked to an ENA accession: https://www.ebi.ac.uk/arrayexpress/experiments/E-MTAB-4654/

Note, however, that the `Experiment type` on that page is `RNA-seq of coding RNA`, which could possibly be used to distinguish it from microarray data.

A few more examples: [`E-MTAB-4801`](https://www.ebi.ac.uk/arrayexpress/experiments/E-MTAB-4801/) which has processed data on ArrayExpress, [`E-MTAB-6204`](https://www.ebi.ac.uk/arrayexpress/experiments/E-MTAB-6204/)

How would these be handled right now?

kurtwheeler:
> Is what you have here likely to produce lots of duplicates in this case?
> How would these be handled right now?

So I actually would have to look at how the code would end up routing those experiments because I haven't addressed the routing logic yet. This PR is focused on determining platform information about samples, so that I when I refactor the routing logic I have all of the information about the samples that I'll need to actually be able to determine what should be done with it.

However I think that this PR sets me up in a good way to handle the cases you illustrated in my next PR. If the `technology` property ends up being UNKNOWN, I can have the routing logic choose to not do anything with that sample/experiment. In the case of ArrayExpress that you brought up, I can check the platform accession against our list of supported microarray platforms and only do something with the sample if that returns a hit (unless there's CEL files, in which case I can let the downloader check the CEL file to see if the platform accession was wrong). If the sample is actually RNASeq, it wouldn't have a CEL file nor would it match our supported platforms list so I would then just not do anything with it.

jaclyn-taroni:
Cool, thanks @kurtwheeler ! I had a feeling these points were beyond the scope, but wanted to make sure we had that discussion 🙂 

Everything else seems reasonable to me, I took a look at the configuration files and they LGTM 👍

api/data_refinery_api/tests.py
```diff
@@ -169,26 +171,26 @@ def test_search_and_filter(self):
         experiments = []
         for x in range(1, LOTS):
             ex = Experiment()
-            ex.accession_code = "".join(random.choice(words) for i in range(3)) + str(random.randint(0, 1000))[:64]
-            ex.title =  " ".join(random.choice(words) for i in range(10))
+            ex.accession_code = "".join(random.choice(words)
+                                        for i in range(3)) + str(random.randint(0, 1000))[:64]
+            ex.title = " ".join(random.choice(words) for i in range(10))
             ex.description = " ".join(random.choice(words) for i in range(100))
-            ex.technology = random.choice(["RNA-SEQ", "MICROARRAY"])
```
Miserlou: Removed why?

api/data_refinery_api/tests.py
```diff
@@ -203,15 +205,17 @@ def test_search_and_filter(self):
...
         self.assertEqual(response.json()['count'], 1)
         self.assertEqual(response.json()['results'][0]['accession_code'], 'FINDME')
 
     def test_create_update_dataset(self):
 
         # Good
         jdata = json.dumps({'data': {"A": ["B"]}})
-        response = self.client.post(reverse('create_dataset'), jdata, content_type="application/json")
+        response = self.client.post(
+            reverse('create_dataset'), jdata, content_type="application/json")
```
Miserlou: Did you get a new linter? I honestly don't think that this code looks improved this way, if you're going to line limit a function, then I think it's better to 
```python
response = self.client.post(
        endpoint=reverse('create_dataset'), 
        data=jdata, 
        content_type="application/json")
```
/opinion

api/data_refinery_api/tests.py
kurtwheeler: Because technology is no longer a property on Experiments.

api/data_refinery_api/tests.py
kurtwheeler: Kinda... mine was broken for a while without me realizing it. Once I realized and fixed it, it started correcting things all over the place. I agree that what you have there is more readable, I've been trying to catch places where it makes this correction and making it more like what you've got there.

api/data_refinery_api/tests.py
Miserlou: Ah, then it should be removed from the two Experiment objects created below this as well

foreman/data_refinery_foreman/surveyor/array_express.py
```diff
@@ -104,19 +111,17 @@ def create_experiment_from_api(self, experiment_accession_code: str) -> Dict:
...
 
             experiment_object = Experiment()
             experiment_object.accession_code = experiment_accession_code
             experiment_object.source_url = request_url
             experiment_object.source_database = "ARRAY_EXPRESS"
             experiment_object.name = parsed_json["name"]
-            experiment_object.description = parsed_json["description"][0]["text"]
-            experiment_object.platform_name = experiment["platform_accession_name"]
-            experiment_object.platform_accession_code = experiment["platform_accession_code"]
             experiment_object.technology = "MICROARRAY"
```
Miserlou: Can be removed?

foreman/data_refinery_foreman/surveyor/array_express.py
```diff
@@ -467,15 +463,17 @@ def create_samples_from_api(self,
 
             # Create associations if they don't already exist
             try:
-                assocation = ExperimentSampleAssociation.objects.get(experiment=experiment, sample=sample_object)
+                assocation = ExperimentSampleAssociation.objects.get(
```
Miserlou: mv to get_or_create?

foreman/data_refinery_foreman/surveyor/array_express.py
```diff
@@ -467,15 +463,17 @@ def create_samples_from_api(self,
...
+                    experiment=experiment, sample=sample_object)
             except ExperimentSampleAssociation.DoesNotExist:
                 association = ExperimentSampleAssociation()
                 association.experiment = experiment
                 association.sample = sample_object
                 association.save()
 
             try:
-                assocation = ExperimentOrganismAssociation.objects.get(experiment=experiment, organism=organism)
+                assocation = ExperimentOrganismAssociation.objects.get(
```
Miserlou: mv to get_or_create?

foreman/data_refinery_foreman/surveyor/geo.py
```diff
@@ -169,23 +262,25 @@ def create_experiment_and_samples_from_api(self, experiment_accession_code) -> (
...
 
                     original_file_sample_association = OriginalFileSampleAssociation()
                     original_file_sample_association.sample = sample_object
                     original_file_sample_association.original_file = original_file
                     original_file_sample_association.save()
 
                 try:
-                    assocation = ExperimentSampleAssociation.objects.get(experiment=experiment_object, sample=sample_object)
+                    assocation = ExperimentSampleAssociation.objects.get(
+                        experiment=experiment_object, sample=sample_object)
```
Miserlou: Same as above

foreman/data_refinery_foreman/surveyor/sra.py
```diff
@@ -338,42 +343,50 @@ def _generate_experiment_and_samples(self, run_accession: str) -> None:
...
+                experiment_object.protocol_description = metadata.get(
+                    "library_construction_protocol", "Protocol was never provided.")
 
             experiment_object.save()
 
-            ##
+            #
             # Experiment Metadata
-            ##
+            #
```
Miserlou: Two comments were in place as 'section indicators - if you don't want these, just remove the whole line rather than making it like 

```
#
# this
#
```

foreman/data_refinery_foreman/surveyor/test_geo.py
```diff
@@ -16,38 +16,109 @@
...
+    def prep_test(self, experiment_accession):
+        survey_job = SurveyJob(source_type="GEO")
         survey_job.save()
         self.survey_job = survey_job
 
         key_value_pair = SurveyJobKeyValue(survey_job=survey_job,
                                            key="experiment_accession_code",
-                                           value="GSE35186")
+                                           value=experiment_accession)
+                                           # value="GSE25624")
```
Miserlou: Can be removed

workers/data_refinery_workers/downloaders/array_express.py
```diff
@@ -108,20 +115,57 @@ def download_array_express(job_id: int) -> None:
...
+                                 file_name: filename,
+                                 sample=sample_id,
+                                 downloader_job=job_id)
+                    job.failure_reason = "Found a raw file with an unsupported platform."
+                    success = False
+
+                sample_object.platform_accession_code = platform_accession_code
+                sample_object.platform_name = get_readable_platform_names()[platform_accession_code]
+                sample_object.save()
+            except Exception:
```
Miserlou: Is warning enough here? Does this indicate that we should throw this file out?

api/data_refinery_api/tests.py
kurtwheeler: I made a mistake here, this property will actually be on both tables so I will readd this.

foreman/data_refinery_foreman/surveyor/array_express.py
kurtwheeler: I was wrong about this. I ended up keeping `technology` on Experiments so they could be filtered on that easily.

workers/data_refinery_workers/downloaders/array_express.py
kurtwheeler: TBH I'm not sure. It seems to make sense to me that if the header is corrupted then the rest of the file would be corrupted too. However, I suppose the header could also just be misformatted or otherwise mangled enough that something attempting to read the header couldn't, but the rest of the file could still be usable?

I figured that if the file wasn't usable then the processor job would fail and if we saw a bunch failing and never had one go through successfully, then we could change this. However I don't feel strongly about this either way so if you think we should throw this out I'm game to make the change.

workers/data_refinery_workers/downloaders/array_express.py
```diff
@@ -109,19 +116,54 @@ def download_array_express(job_id: int) -> None:
...
+                    original_file.absolute_file_path)
+
+                platform_accession_code = "UNSUPPORTED"
+                for platform in get_supported_microarray_platforms():
+                    if platform["external_accession"] == external_accession:
+                        platform_accession_code = platform["platform_accession"]
+
+                if platform_accession_code == "UNSUPPORTED":
+                    logger.error("Found a raw file with an unsupported platform!",
+                                 file_name: filename,
```
Miserlou: S Y N T A X E R R O R

Issue #296, opened 2018-05-31
Modify Samples end point to return detailed information


arielsvn:
### Context

The [experiments page](https://projects.invisionapp.com/d/main#/console/13557531/297243710/preview) needs to display a Table with detailed information about the Samples associated with an experiment.

### Problem or idea

We don't have an endpoint that returns detailed information for the samples associated with one experiment.

On https://github.com/AlexsLemonade/refinebio-frontend/pull/48 we're doing an individual request for each sample in the experiment. This means we're also not able to sort the samples there.

### Solution or next step

It would be good to modify the Samples List endpoint so that it returns detailed information for a sample, and add some parameter to sort the samples.

Also, it would be good if that endpoint returns the samples associated with some experiment, right now we have to send the individual ids in the url (`?ids=1,2,3,4`). 

@Miserlou @kurtwheeler @ramenhog

kurtwheeler:
This looks like it should be a simple enough change. The only question I have is whether we want the list samples endpoint to return detailed information by default, when passed an extra parameter, or to have a new endpoint for listing detailed samples. I suspect @Miserlou will have strong opinions about this.

Miserlou:
Are you asking for the Samples result to be modified, or for the Experiments endpoints to use the detailed sample serializer

PR #298, opened 2018-05-31
Use Detailed View and Add `order_by` to Sample List

Miserlou:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/296

Makes the samples endpoint more useful for the experiments page. Adds `order_by`, ex `order_by=title` and `order_by=-title` to reverse.

## Types of changes

## Functional tests

New tests added.

## Checklist


api/data_refinery_api/views.py
```diff
@@ -221,20 +221,23 @@ def get(self, request, format=None):
...
+            samples = samples.order_by(order_by)
 
         page = self.paginate_queryset(samples)
         if page is not None:
-            serializer = SampleSerializer(page, many=True)
+            serializer = DetailedSampleSerializer(page, many=True)
             return self.get_paginated_response(serializer.data)
         else:
-            serializer = SampleSerializer(samples, many=True)
+            serializer = DetailedSampleSerializer(samples, many=True)
```
kurtwheeler: If I'm not mistaken, these are the only two places we used SampleSerializer. Is there any point to having it around anymore? Although if we go the route of nixing it, I would think we'd probably want to get rid of DetailedSampleSerializer and make SampleSerializer into what it was.

Miserlou: I'm worried this is going to end up being very slow. They need it right now, okay, but let's hang on to the skinny one until we're sure we're not going to need it.

PR #301, opened 2018-06-01
count filters based on results of query

ramenhog:
## Issue Number
na

@Miserlou restricted the filter counts to the results in the query

## Types of changes

## Functional tests
* filter counts reflect what is expected from the query results

### DIFF
--- a/api/data_refinery_api/views.py
+++ b/api/data_refinery_api/views.py
@@ -22,8 +22,8 @@ from data_refinery_common.models import (
     ProcessorJob,
     Dataset
 )
-from data_refinery_api.serializers import ( 
-    ExperimentSerializer, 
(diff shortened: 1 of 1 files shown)


api/data_refinery_api/views.py
```diff
@@ -107,18 +107,20 @@ def list(self, request, *args, **kwargs):
         response.data['filters']['publication'] = {}
         response.data['filters']['organism'] = {}
 
-        techs = self.get_queryset().values('technology').annotate(Count('technology', unique=True))
+        qs = self.filter_queryset(self.get_queryset())
```
Miserlou: Is there a way to access this without calling the filter? Or does it not matter since the QS isn't calculated until serialization time.

PR #302, opened 2018-06-01
Makes association tables unique, adds organisms to geo experiments.

kurtwheeler:

Geo experiments had no organisms.

Association tables allowed duplicate rows.

This fixes both of those issues.

## Types of changes

### DIFF
--- a/foreman/data_refinery_foreman/surveyor/array_express.py
+++ b/foreman/data_refinery_foreman/surveyor/array_express.py
@@ -104,8 +104,8 @@ class ArrayExpressSurveyor(ExternalSourceSurveyor):
         try:
             experiment_object = Experiment.objects.get(accession_code=experiment_accession_code)
             logger.debug("Experiment already exists, skipping object creation.",
-                experiment_accession_code=experiment_accession_code,
-                survey_job=self.survey_job.id)
+                         experiment_accession_code=experiment_accession_code,
+                         survey_job=self.survey_job.id)
         except Experiment.DoesNotExist:
 
             experiment_object = Experiment()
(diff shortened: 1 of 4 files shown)


foreman/data_refinery_foreman/surveyor/array_express.py
```diff
@@ -460,11 +460,6 @@ def create_samples_from_api(self,
                 original_file_sample_association.sample = sample_object
                 original_file_sample_association.save()
 
-            association = ExperimentSampleAssociation()
```
Miserlou: Can you explain this removal? It isn't mentioned in the PR text.

foreman/data_refinery_foreman/surveyor/geo.py
```diff
@@ -124,17 +125,17 @@ def create_experiment_and_samples_from_api(self, experiment_accession_code) -> (
 
                 all_samples.append(sample_object)
 
-                association = ExperimentSampleAssociation()
-                association.experiment = experiment_object
-                association.sample = sample_object
-                association.save()
+                ExperimentSampleAssociation.objects.get_or_create(
```
Miserlou: If you want use this pattern, can you make it universal in the codebase?

foreman/data_refinery_foreman/surveyor/array_express.py
kurtwheeler: It was duplicated immediately below, except with in a try block.

kurtwheeler:
@Miserlou I used that pattern everywhere it made sense. I chose not to use it in places where the control flow of the `try/except` blocks were useful.

PR #304, opened 2018-06-05
Sorts sample dictionaries before trying to extract the title in harmony.

kurtwheeler:

While testing my routing refactor branch I ran into an experiment `E-TABM-38` which caused surveyor jobs to crash because of an uncaught KeyError. This is because `extract_title` was returning different titles for the samples which came from the JSON endpoint and samples which came from the SDRF.

## Types of changes

## Functional tests

I ran `./foreman/run_surveyor.sh survey_all --accession=E-TABM-381` and it finished successfully, whereas before my fix it crashed because of a KeyError.

## Checklist

### DIFF
--- a/foreman/data_refinery_foreman/surveyor/test_harmony.py
+++ b/foreman/data_refinery_foreman/surveyor/test_harmony.py
@@ -324,7 +324,6 @@ class HarmonyTestCase(TestCase):
         samples_endpoint = SAMPLES_URL.format(experiment_accession_code)
         r = utils.requests_retry_session().get(samples_endpoint, timeout=60)
         json_samples = r.json()["experiment"]["sample"]
-        # flat_json_sample = utils.flatten(json_samples[0])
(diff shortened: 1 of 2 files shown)


foreman/data_refinery_foreman/surveyor/test_harmony.py
```diff
@@ -303,3 +313,26 @@ def test_geo_leg_cancer(self):
...
+        """Makes sure that the order samples' keys are in does not affect the title chosen.
+
+        Related: https://github.com/AlexsLemonade/refinebio/pull/304
+        """
+        experiment_accession_code = "E-TABM-38"
+
+        samples_endpoint = SAMPLES_URL.format(experiment_accession_code)
+        r = utils.requests_retry_session().get(samples_endpoint, timeout=60)
+        json_samples = r.json()["experiment"]["sample"]
+        # flat_json_sample = utils.flatten(json_samples[0])
```
Miserlou: Remove dead line?

PR #303, opened 2018-06-04
Smasher, Part 1

Miserlou:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/221 (Main issue)
https://github.com/AlexsLemonade/refinebio/issues/238 (Past)
https://github.com/AlexsLemonade/refinebio/issues/186 (Future)

## Purpose/Implementation Notes

Implements the "Smasher". Uses `pandas` and `sklearn` to combine ComputedFiles into a file single matrix, transposes, applies a scaling function, retransposes, and writes the file out. Then creates a metadata file containing all of 

This PR only covers the naive use case where genes identifiers are already aligned, so essentially for the case where all samples come from the same experiment. Additionally, the metadata file is not yet on a per-experiment basis, and the metadata is only in JSON, not JSON and TSV (SDRF) format yet. (Sorry, Jackie.)

## Methods

This bit requires scientific review:

```
            # Merge all the frames into one
            all_frames = []
            for computed_file in input_files:
                data = pd.DataFrame.from_csv(str(computed_file.absolute_file_path), sep='\t', header=0)
                all_frames.append(data)
            i = 1
            merged = all_frames[0]
            while i < len(all_frames):
                merged = merged.merge(all_frames[i], left_index=True, right_index=True)
                i = i + 1
```

The merge statements is using pandas [DataFrame.merge](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.merge.html) is using `left_index=True` and `right_index=True`. This is using the default `how='inner'`.

Output examples:
```
(Pdb) all_frames[0]
                      GSM1237810_T09-1084.txt_Channel1  GSM1237810_T09-1084.txt_Channel2
GE_BrightCorner_1                             2.179542                          1.825372
DarkCorner_2                                 -0.511203                         -1.075368
DarkCorner_3                                 -0.710034                         -0.861985
DarkCorner_4                                 -0.659205                         -0.972789
DarkCorner_5                                 -0.571706                         -1.038416
[ .. ]

(Pdb) all_frames[1]
                      GSM1237812_S97-PURE.txt_Channel1  GSM1237812_S97-PURE.txt_Channel2
GE_BrightCorner_1                             3.448376                          1.407764
DarkCorner_2                                 -0.330210                         -1.140229
DarkCorner_3                                 -0.428655                         -1.115645
DarkCorner_4                                 -0.394874                         -1.131911
DarkCorner_5                                 -0.406034                         -1.126538
[ .. ]

[...]

Issue #232, opened 2018-04-30
Add Support for Broader Set of Illumina Beadchip Platforms


Miserlou:
### Context

Quoting @jaclyn-taroni, via https://github.com/AlexsLemonade/refinebio/pull/212#discussion_r184864928:

```
This only supports 2 human platforms, Illumina HumanWG-6 v2.0 (i.e., GPL13376) and Illumina HumanHT-12 V4.0 (i.e., GPL10558), using the relevant bioconductor packages at the moment.

We'll want to support a broader set of BeadChips. Specifically, there are bioconductor packages for the following additional platforms:

    illuminaHumanv1.db
    illuminaHumanv3.db
    illuminaMousev1.db
    illuminaMousev1p1.db
    illuminaMousev2.db
    illuminaRatv1.db

You also might consider getting the probe objects like so:

probeSequenceRef <- lazyeval::lazy_eval(paste0(platform, ".db::", platform, "PROBESEQUENCE"))
probeQualityRef <- lazyeval::lazy_eval(paste0(platform, ".db::", platform, "PROBEQUALITY"))
probeGeneRef <-  lazyeval::lazy_eval(paste0(platform, ".db::", platform, "ENSEMBLREANNOTATED"))

☝️ would also eliminate the need for loading the annotation library (line 356).
```

### Problem or idea

Until we explicitly add support for other platforms, we'll only be able to process results from two human Illumina platforms. 

### Solution or next step

Jackie has also ready described the solution, so I think it just needs to be implemented. There may be a little bit of work to detect the platform as well.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

Miserlou:
This ticket would be greatly enhanced by a small suite of example accessions.

jaclyn-taroni:
Will do today while wrapping up the microarray whitelist

jaclyn-taroni:
While working on #222/#228 today, I've found that relationship between Illumina (bead array) platforms (and therefore their annotation) is less straightforward than I initially imagined. Specifically, I thought we would just support the platforms that the Bioconductor packages are named for and that was that. However, there are a considerable amount of samples we'd miss out on that way. 

Here's a `v3.0` example:

[...]

PR #307, opened 2018-06-07
Smasher, Part 2: ENSG Conversion

Miserlou:
This PR depends on Smasher: Part 1.

## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/186 
https://github.com/AlexsLemonade/refinebio/pull/303

Converts Probe_IDs into Ensembl Gene Identifiers, allowing NO_OP'd data to be smashed with data we have processed.

## Methods

I think this is the interesting file for sci-reviewers:  https://github.com/AlexsLemonade/refinebio/pull/307/files#diff-89bb3c747fc413bb91fbbc9cb8f52a04

We use the R `Bioconductor` package to do this. I have never written R from scratch before, so I think there are some places where I'm using un-idiomatic R (loops rather than matrix applications) that could be improved.

Sciencewise, a few decisions were made: 

a) Probes which don't map to ENSG (i.e., return `NA`) are removed from the output, under the assumption that these are control probes. The OriginalFiles will still be available if these are ever needed.

b) Multiple probes can map to the same gene identifier. At *smash-time*, _not geneconvert-time_, these are squished down *to their mean*. We squish at smash-time in order to provide greater flexibility in the future, where we may want to squish to (min/max/etc) instead. The squshing is also done with `pandas` `groupby` rather than `R`.

c) (I don't think this will matter) - since different files have different headers, we make sure that they always have `ID_REF` and `VALUE` columns rather than detecting them dynamically. I don't know if these values have any further meaning but they're the most common that I see. 

d) We now automatically detect if data hasn't been `log2` scaled if `data.max() > 100).any()` (the same logic as with Illumina), and then use `numpy.log2` to transform before smashing. We do this at smashtime currently, we could also do it at scrapetime if we so choose.

## Types of changes

## Functional tests

Two new tests have been added.

I'm expecting that this PR will take a few revisions but I wanted to get it submitted now.

## Checklist


cgreene:
> We now automatically detect if data hasn't been log2 scaled if data.max() > 100).any() (the same logic as with Illumina)

As far as I know, all of the affy scanners are 16 bit. I think you could do > 16 unless there are Illumina scanners that have more range.

workers/data_refinery_workers/processors/gene_convert.R
```diff
@@ -0,0 +1,61 @@
...
+# Read the data file
+message("Reading data file...")
+suppressWarnings(data <- fread(filePath, stringsAsFactors=FALSE, sep="\t", header=TRUE, autostart=10, data.table=FALSE, check.names=FALSE, fill=TRUE, na.strings="", showProgress=FALSE))
+
+# Build a PROBEID:ENSEMBL map
+res <- select(get(paste0(platform, ".db", sep="")), keys=data$ID_REF, columns=c("ENSEMBL"), keytype="PROBEID")
+
+# Replace the probe IDs with Ensembles
+i <- 0
+for(m in data[, 1]){
```
cgreene: @jaclyn-taroni should take a gander at this. It can be made much more R-ish (and consequently likely much faster since R at least used to hate loops).

jaclyn-taroni:
This is a really good starting point @Miserlou. I can definitely take a look at [`gene_convert.R`](https://github.com/AlexsLemonade/refinebio/pull/307/files#diff-89bb3c747fc413bb91fbbc9cb8f52a04) and make it more "R-ish."

One thing I noticed looking at this R script is that it seems [to only support mapping between probe identifiers and Ensembl gene identifiers](https://github.com/AlexsLemonade/refinebio/pull/307/files#diff-89bb3c747fc413bb91fbbc9cb8f52a04R44). At this stage, we are mostly concerned with converting gene ids on submitter-processed data from supported platforms where no raw data is available/it is unsuitable (not probe ids; see discussion #232). (Side note: in the future, this will also be a concern in `SCAN.UPC::SCAN_TwoColor` processed data if we go that route.)  

We will certainly encounter submitter-processed data that uses probe identifiers, but there will also be data using other identifiers like Entrez ids and gene symbols, for example. 

[`GSE51182`](https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GSE51182) is one such example. `GSE51182_input_non-normalized.txt.gz` contains gene symbols, which we can't easily get probe sequences from = unsuitable for use with SCAN. Gene symbols are also used in the submitter-processed data (ftp://ftp.ncbi.nlm.nih.gov/geo/series/GSE51nnn/GSE51182/miniml/GSE51182_family.xml.tgz), so in order to get this to "play nicely" with the rest of the compendium we'd have to map from gene symbols to Ensembl gene ids.

I'll also note that packages like `hgu133plus2.db` contain _lots_ of different probe to gene id mappings and we're only concerned with Ensembl IDs. See this screenshot of the datasets included in `hgu133plus2.db` below; I think we should consider what including that package would do to Docker image size (may not be prohibitive, but worth thinking about).

screenshot from 2018-06-08 08-45-38




As an alternative approach, we might derive a bunch of data (e.g., JSON files?) from a collection of sources (likely to be Bioconductor packages as a starting point) that contain _only_ the mappings we care about. I bet this is likely to be more compact, but I don't know for sure. Since `hgu133plus2` is a supported platform, we could grab the mappings from `hgu133plus2ENSEMBL` and would repeat this
for all supported platforms, most likely using some of the code you have here. 

Here's an example how we could generate a mapping between Entrez IDs and Ensembl gene IDs using Bioconductor:

```R
library(AnnotationDbi)
ensg.obj <- org.Hs.eg.db::org.Hs.egENSEMBL
mapped.genes <- AnnotationDbi::mappedkeys(ensg.obj)
ensg.list <- as.list(ensg.obj[mapped.genes])
rlist::list.save(ensg.list, "entrez.json")
```

Where `entrez.json` looks like this (also note the one to many mapping):

```
{
  "1": [
    "ENSG00000121410"
  ],
  "10": [
    "ENSG00000156006"
  ],
  "100": [
    "ENSG00000196839"
  ],
  "1000": [
    "ENSG00000170558"
  ],
  "10000": [
    "ENSG00000117020",
    "ENSG00000275199"
  ],
  "100008586": [
    "ENSG00000236362"
  ],
  "100008587": [
    "ENSG00000277739",
    "ENSG00000276700",
    "ENSG00000278233",
    "ENSG00000275215",
    "ENSG00000274917",
    "ENSG00000278189",
    "ENSG00000275757",
    "ENSG00000273730"
  ],
  "100009676": [
    "ENSG00000256628"
  ],
  "10001": [
    "ENSG00000133997"
  ],
  [...]
  }
  ```
  
@cgreene and I could come up with a list of supported identifiers and we could have a **"conversion-refinery"** where we generate this set of files (mappings) for each supported organism within Docker containers that way we're on top of versioning. 

Here's how I'm thinking this would all come together (excuse the probably-not-great pseudocode):

```
if submitter-processed data
  check if the identifiers in the files are probe IDs using the platform label from the data source (and probably config/supported_microarray_platforms.csv?)
  if probe ids
    use probe ids to ENSG mapping
  else
    check against all other supported identifiers 
    if overlap > some reasonable threshold
      use detected identifier to ENSG mapping
    else
      fail
```

Just some thoughts - the best approach may be somewhere in between these two ideas. Happy to pair as needed.

**Comment by Miserlou** [comment]
Thanks for your comments Jackie!

I thiiiink there might be an easier way, since this line:

```R
res <- select(get(paste0(platform, ".db", sep="")), keys=data$ID_REF, columns=c("ENSEMBL"), keytype="PROBEID")
```

can probably be replaced with:

```R
res <- select(get(paste0(platform, ".db", sep="")), keys=data$ID_REF, columns=c("ENSEMBL"), keytypes=c(x)
```

where x is:

```
##  [1] "ENTREZID"     "PFAM"         "IPI"          "PROSITE"     
##  [5] "ACCNUM"       "ALIAS"        "CHR"          "CHRLOC"      
##  [9] "CHRLOCEND"    "ENZYME"       "PATH"         "PMID"        
## [13] "REFSEQ"       "SYMBOL"       "UNIGENE"      "ENSEMBL"     
## [17] "ENSEMBLPROT"  "ENSEMBLTRANS" "GENENAME"     "UNIPROT"     
## [21] "GO"           "EVIDENCE"     "ONTOLOGY"     "GOALL"       
## [25] "EVIDENCEALL"  "ONTOLOGYALL"  "PROBEID"
```

Then we can just try mapping to each of the resulting columns and get the one that has the highest number of remaining lines (most will be zero)? Then we don't have to muck about with any identifier-specific JSON. I think your pseudocode is correct though.

Re: Docker image size, it's currently at 3.4GB unpacked, which is large, but nothing compared to our Affy image, which is 21.5GB unpacked, so I think that worrying about shrinking the image right now would be a premature optimization, and one that should be done after shrinking Affy.

jaclyn-taroni:
I believe using a platform-specific package only supports conversion from `PROBEID` based on the documentation.

```R

library(hgu133plus2.db)
library(AnnotationDbi)

platform <- "hgu133plus2"

# gene symbols
data <- data.frame("ID_REF" = c("FBN1", "CD14", "CXCR5", "CDC7", 
                                "SERPINE1", "MPO", "ELANE", "CTSG"))
data$ID_REF <- as.character(data$ID_REF)

# we'l call these the supported identifiers
supported_ids <- c("ENTREZID", "SYMBOL", "PROBEID")

res <- select(get(paste0(platform, ".db", sep="")), 
              keys=data$ID_REF, columns=c("ENSEMBL"), 
              keytypes=supported_ids)
```
gives me the following error
```
Error in .testForValidKeys(x, keys, keytype, fks) : 
  None of the keys entered are valid keys for 'PROBEID'. Please use the keys method to see a listing of valid arguments.
```
There could be another way, though.

Miserlou:
Which documentation are you referring to? The AnnotationDB docs say this is valid:

```R
## lookup gene symbol and unigene ID based on unigene IDs by setting
## the keytype to "UNIGENE" and passing in unigene keys:
select(hgu95av2.db, keys=keyunis, columns = c("SYMBOL","UNIGENE"),
keytype="UNIGENE")
```
Or do you mean the `hgu133plus2` docs?

Does your code work with 
```
res <- select(get(paste0(platform, ".db", sep="")), 
              keys=data$ID_REF, columns=c("ENSEMBL"), 
              keytype="SYMBOL")
```

Either way, we can take a look together on Monday.

jaclyn-taroni:
I get the same error with that snippet. Yes, let's look together on Monday. We might be able to deal with some of the the Affymetrix Docker image size at the same time because some of those packages might need to move to this Docker image like [these guys](https://github.com/AlexsLemonade/refinebio/blob/3660d84a6b5f82721c70add05fc8bc2ee0447b25/workers/affymetrix_dependencies.R#L60), but we'll have to see if that's a good use of time.

jaclyn-taroni:
I think something like this will work
```R
# we'll call these the supported identifiers
supported_ids <- c("ENTREZID", "SYMBOL", "ENSEMBL", "UNIGENE")

res <- select(get(db.package), keys = keys(get(db.package)),
              columns = supported_ids)

# find % overlap by column
overlap <- apply(res, 2, function(x) length(intersect(data$ID_REF, x)) / length(data$ID_REF))

# threshold here can be adjusted
if (any(overlap > 0.9)) {
  # take column name with highest overlap value 
  # overlap is a named numeric vector
  detected_id <- names(which.max(overlap))
} else {
  stop("Not enough overlapping ids detected!")
}
```

Couple concerns I have with this that we can chat about tomorrow: 1) still has one-to-many mapping issue, which we may want to make some "data cleaning" type of decisions about before conversion 2) going this `AnnotationDbi::select()` route is probably slower than reading in a (potentially already cleaned) `data.frame` and we'll have to do it for every sample that requires conversion.

If I write `res` to a tsv like so `readr::write_tsv(res, "results.tsv")` and compare

```R
> system.time(df <- select(get(db.package), keys = keys(get(db.package)),
+                   columns = supported_ids))
'select()' returned 1:many mapping between keys and columns
   user  system elapsed 
  1.378   0.060   1.476 
```
vs.
```R
> system.time(df <- readr::read_tsv("results.tsv"))
Parsed with column specification:
cols(
  PROBEID = col_character(),
  ENTREZID = col_integer(),
  SYMBOL = col_character(),
  ENSEMBL = col_character(),
  UNIGENE = col_character()
)
   user  system elapsed 
  0.316   0.010   0.328 
```
or, since you have `data.table` here, vs.
```R
> system.time(df <- data.table::fread("results.tsv", sep = "\t",
+                                     data.table = FALSE))
   user  system elapsed 
  0.075   0.008   0.103 
```

jaclyn-taroni:
Note to self for tomorrow’s discussion: trailing “_at” for BrainArray processed data

jaclyn-taroni:
Okay, want to summarize the outcome of discussions I've had with @Miserlou and @cgreene here.

* For Keytar Kurt, we will be supporting gene identifier conversion for Affymetrix submitter-processed data _only_ for Brainarray-supported platforms since this is easier to get right at this time.
* We'll use a combination of Brainarray (e.g., `hgu133plus2hsensgprobe` and Bioconductor (e.g., `hgu133plus2.db` packages to generate "master conversion" tables for each of these supported platforms and we'll be filtering by what probes and Ensembl gene identifiers are represented in the Brainarray package. Recall that the bulk of these data will be processed from raw by us and therefore be constrained to these identifiers.

I've put this together to do that filtering and tried it out with a `hgu133plus2` CEL file:

```R
# J. Taroni 2018

# magrittr pipe
`%>%` <- dplyr::`%>%`

option_list = list(
  optparse::make_option(c("-p", "--platform"),
                        type = "character",
                        default = "hgu133plus2",
                        help = "Platform",
                        metavar ="character"),
  optparse::make_option(c("-s", "--species"),
                        type = "character",
                        default = "hs",
                        help = "Species",
                        metavar ="character"),
  optparse::make_option(c("-i", "--inputFile"),
                        type = "character",
                        default = "",
                        help = "inputFile",
                        metavar = "character"),
  optparse::make_option(c("-o", "--outputFile"),
                        type = "character",
                        default = "",
                        help = "outputFile",
                        metavar = "character")
)

opt_parser <- optparse::OptionParser(option_list=option_list)
opt <- optparse::parse_args(opt_parser)

platform <- opt$platform
org_code <- opt$species
cel_path <- opt$inputFile
out_file_path <- opt$outputFile

# Does platform in CEL header match platform passed as an argument?
header_platform <- unlist(affyio::read.celfile.header(cel_path)[1])
header_platform <- tolower(gsub("[[:punct:]]", "", header_platform))
# if not, throw error
if (header_platform != platform) {
  stop("CEL file header does not match platform argument")
}

# load brainarray package
ba_package_name <- paste0(platform, org_code, "ensgprobe")
library(ba_package_name, character.only = TRUE)

# load bioconductor .db package
db_package_name <- paste0(platform, ".db")
library(db_package_name, character.only = TRUE)

# generate a ExpressionFeatureSet from inputFile -- this step will automatically
# install platform design packages from bioconductor if not already installed
# adapted from SCAN.UPC one color
affy_feature_set <- oligo::read.celfiles(cel_path)
# this data.frame has the x, y coords and the manufacturer ID
probe_info_df <- oligo::getProbeInfo(affy_feature_set, 
                                     field = c("x", "y"), 
                                     probeType = "pm")

# data.frame contains x, y coord and the gene IDs from brainarray package
brainarray_df <- as.data.frame(get(ba_package_name), stringsAsFactors = FALSE)
# we'll drop the trailing "_at" to get the Ensembl gene IDs
brainarray_df <- brainarray_df %>%
  dplyr::mutate(Probe.Set.Name = sub("_at", "", Probe.Set.Name))

# join brainarray and probe info by x,y coord & clean up a bit
ba_probe_ensg_df <- dplyr::inner_join(probe_info_df, brainarray_df,
                                      by = c("x", "y")) %>%
  dplyr::select(c("man_fsetid", "Probe.Set.Name")) %>%
  dplyr::mutate(PROBEID = man_fsetid, ENSEMBL = Probe.Set.Name) %>%
  dplyr::select(PROBEID:ENSEMBL) %>%
  dplyr::mutate(ENSEMBL = as.character(ENSEMBL))

# what identifiers will we support conversion from? we'll supply these to the
# .db package
supported_identifiers <- c("SYMBOL", "ENTREZID", "UNIGENE", "ENSEMBL",
                           "REFSEQ")
# only those that are actually supported by the package
supported_identifiers <- base::intersect(supported_identifiers, 
                                         keytypes(get(db_package_name)))

# extract the supported identifiers and probe IDs from .db package
db_map_df <- select(get(db_package_name), 
                    keys = keys(get(db_package_name)),
                    columns = supported_identifiers)

# filtering join on probe IDs and Ensembl gene IDs in brainarray
master_map_df <- dplyr::left_join(x = ba_probe_ensg_df,
                                  y = db_map_df,
                                  by = c("PROBEID", "ENSEMBL"))

# output file, using the path supplied as an argument
output_file <- file.path(out_file_path, 
                         paste0(platform, "_master_conversion.tsv.gz"))
# write compressed file
readr::write_tsv(master_map_df, 
                 path = output_file)
```
---

Note that this is what I think needs to live on the Docker image used to generate these mappings:

```
bioc_pkgs <- c(
  'oligo_1.42.0.tar.gz',
  'Biobase_2.38.0.tar.gz',
  'affy_1.56.0.tar.gz',
  'affyio_1.48.0.tar.gz',
  'AnnotationDbi_1.40.0.tar.gz',
)
```
**For each platform:** brainarray package, `*.db` package from bioconductor, the `org.*` package from bioconductor (e.g., `org.Hs.eg.db` for humans; required for some `*.db` packages), and the platform design package (`pd.*`) from bioconductor

[`tidyverse`](https://www.tidyverse.org): `readr`, `dplyr`

`optparse`

---

Because that R code requires an `ExpressionFeatureSet` to be generated from a CEL file, we should run this for multiple CEL files from each platform and compare the results.

Using [`supported_microarray_platforms.csv`](https://github.com/AlexsLemonade/refinebio/blob/220537e621d6355b2e54faa125ee84edeef3b55e/config/supported_microarray_platforms.csv), we should do the following:

1. filter this to `is_brainarray == 'y'`
2. for each unique `internal_accession`, take the first `external_accession` which should correspond to the _most popular_ (by way of sample count) GEO platform accession
3. identify 10 random samples (CEL files), ideally from 10 distinct experiments
4. run that R script for each CEL file in platform to generate 10 of these "master conversion" tables for us to compare

jaclyn-taroni:
@Miserlou let's make sure we talk about removing trailing `_at` from Brainarray-supported SCAN-processed data and dealing with the [ENSG version information](https://useast.ensembl.org/Help/Faq?id=488) in the identifiers from RNA-seq data today. Should both be fairly simple to solve, it's a matter of _at what point_ we do so.

Miserlou:
@jaclyn-taroni, handled by adding a safety `data.index = data.index.str.replace('_at', '')` to the smasher.

Because of [a bug in pandas](https://github.com/pandas-dev/pandas/pull/21098), this requires `pandas>=0.23.1`, which was released only two days ago. Good timing!

jaclyn-taroni:
Here's a more fleshed out toy example for ID conversion @Miserlou: [id_conversion_toy.zip](https://github.com/AlexsLemonade/refinebio/files/2105825/id_conversion_toy.zip)
I put this together somewhat quickly so it might need some tweaking. We can probably chat about it this afternoon if need be!

jaclyn-taroni:
For the record, in that notebook ☝️ I separate out the joining and dropping the detected IDs for illustrative purposes. In practice, it can be done in one step:
```R
converted_exprs <- identifier_df %>%  
  dplyr::select(c("GENEID", detected_id)) %>%  # sub "ENSEMBL" for "GENEID"
  dplyr::inner_join(exprs, by = detected_id) %>%
  dplyr::select(-rlang::UQ(rlang::sym(detected_id)))
```

workers/data_refinery_workers/processors/gene_convert.R
```diff
@@ -0,0 +1,92 @@
...
+
+platform <- opt$platform
+filePath <- opt$inputFile
+outFilePath <- opt$outputFile
+
+# Read the data file
+message("Reading data file...")
+suppressWarnings(data <- fread(filePath, 
+					stringsAsFactors=FALSE, 
+					sep="\t", header=TRUE,
```
jaclyn-taroni: Will these files have headers in our system? Sample tables from at least GEO do not generally have them. Processed data from ArrayExpress also tend to follow a bit of a different (multi-sample) format as far as I can tell.

workers/data_refinery_workers/processors/gene_convert.R
```diff
@@ -0,0 +1,92 @@
...
+					autostart=10, 
+					data.table=FALSE, 
+					check.names=FALSE, 
+					fill=TRUE, 
+					na.strings="", 
+					showProgress=FALSE)
+				)
+
+# find % overlap by column
+overlap <- apply(index_data, 2, function(x) length(intersect(data$ID_REF, x)) / length(data$ID_REF))
```
jaclyn-taroni: This is also related to the header question -- if you assume that the first column will contain identifiers, could do `data[[1]]` instead of `data$ID_REF`

workers/data_refinery_workers/processors/gene_convert.R
```diff
@@ -0,0 +1,92 @@
...
+library("rlang")
+
+option_list = list(
+  make_option(c("-p", "--platform"), type="character", default="", 
+              help="Platform", metavar="character"),
+  make_option(c("-i", "--inputFile"), type="character", default="", 
+              help="inputFile", metavar="character"),
+  make_option(c("-o", "--outputFile"), type="character", default="", 
+              help="outputFile", metavar="character")
+);
```
jaclyn-taroni: I don't think `;` should be necessary here.

workers/data_refinery_workers/processors/gene_convert.R
```diff
@@ -0,0 +1,92 @@
...
+  	detected_id <- names(which.max(overlap))
+} else {
+	# Less than 50%? Get outta here!
+  	stop("Not enough overlapping ids detected!")
+}
+
+message("Merging..")
+colnames(data)[1] <- detected_id
+converted_exprs <- index_data %>%  
+  dplyr::select(c("ENSEMBL", detected_id)) %>%  # sub "ENSEMBL" for "GENEID"
```
jaclyn-taroni: Remove comment

workers/data_refinery_workers/processors/no_op.py
```diff
@@ -25,16 +30,96 @@ def _no_op_processor_fn(job_context: Dict) -> Dict:
...
+        result = subprocess.check_output([
+                "/usr/bin/Rscript", 
+                "--vanilla", 
+                "/home/user/data_refinery_workers/processors/gene_convert.R",
+                "--platform", job_context["internal_accession"],
+                "--inputFile", job_context['input_file_path'],
+                "--outputFile", job_context['output_file_path']
+            ])
+    except Exception as e:
+        error_template = ("Encountered error in R code while running illumina.R"
```
jaclyn-taroni: `gene_convert.R`?

workers/data_refinery_workers/processors/gene_convert.R
jaclyn-taroni: I _think_ the answer is yes based on `no_op.py` but I'm gonna leave this here so it can be confirmed.

workers/data_refinery_workers/processors/gene_convert.R
```diff
@@ -0,0 +1,92 @@
...
+converted_exprs <- converted_exprs %>%
+  dplyr::select(-rlang::UQ(rlang::sym(detected_id)))
+
+message(head(converted_exprs))
+
+# Data here can have duplicate rows that need to be squished together (via mean/max/etc),
+# but this should be done at smash-time so that we can provide options on the squish method.
+
+# Save to output file
+write.table(converted_exprs, outFilePath, row.names=FALSE, col.names=TRUE, quote=FALSE, sep="\t")
```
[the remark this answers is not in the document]
Miserlou: Do you mean `write_tsv`? Or does read write?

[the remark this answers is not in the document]
jaclyn-taroni: Ha yes I meant write_tsv oops

workers/dockerfiles/Dockerfile.no_op
```diff
@@ -1,14 +1,27 @@
 FROM ubuntu:16.04
 
 RUN \
+  apt-get update -qq && \
+  apt-get install -y lsb-release && \
+  echo "deb http://archive.linux.duke.edu/cran/bin/linux/ubuntu $(lsb_release -sc)/" \
+      >> /etc/apt/sources.list.d/added_repos.list && \
+  apt-key adv --keyserver keyserver.ubuntu.com --recv-keys E084DAB9 && \
```
kurtwheeler: Can you switch this to the more reliable key server?

workers/dockerfiles/Dockerfile.no_op
```diff
@@ -20,8 +33,16 @@ COPY workers/data_refinery_workers/downloaders/requirements.txt .
 
 RUN pip3 install --upgrade pip
 
+# Noop-specific requirements
+ENV R_LIBS "/usr/local/lib/R/site-library"
+COPY workers/gene_indexes gene_indexes/
+COPY workers/install_gene_convert.R .
+RUN Rscript install_gene_convert.R
+# End Noop-specific
```
kurtwheeler: Do these commands have any python dependencies? If not it might be nice to move it above line 32 so adding a python dependency doesn't require this step to be rerun.

workers/data_refinery_workers/processors/no_op.py
```diff
@@ -25,16 +30,96 @@ def _no_op_processor_fn(job_context: Dict) -> Dict:
...
+    if 'ID_REF' not in joined:
+        try:
+            float(row[1])
+            # Okay, there's no header so can just prepend to the file.
+            with open(job_context["input_file_path"], 'r') as f:
+                all_content = f.read()
+
+            job_context["input_file_path"] = job_context["input_file_path"] + ".fixed"
+            with open(job_context["input_file_path"], 'w+') as f:
+                f.seek(0, 0)
```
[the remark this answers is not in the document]
Miserlou: We are inserting at the beginning of a file. Without being able to _read_ and write (`w+`), we wouldn't be able to seek back to the beginning for our insertion.

workers/dockerfiles/Dockerfile.no_op
Miserlou: Done

workers/data_refinery_workers/processors/test_smasher.py
```diff
@@ -106,56 +104,58 @@ def test_smasher(self):
...
         for scale_type in ['MINMAX', 'STANDARD']:
-            dataset = Dataset.objects.get(id=dsid)
+            dataset = Dataset.objects.filter(id__in=relations.values('dataset_id')).first()
             dataset.aggregate_by = 'EXPERIMENT'
             dataset.scale_by = scale_type
             dataset.save()
 
+            print("###")
+            print("# " + scale_type)
+            print('###')
```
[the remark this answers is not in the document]
Miserlou: I like them.

workers/data_refinery_workers/processors/smasher.py
```diff
@@ -74,7 +75,43 @@ def _smash(job_context: Dict) -> Dict:
...
+                # Detect if this data hasn't been log2 scaled yet.
+                # Ideally done in the NO-OPPER, but sanity check here.
+                if (data.max() > 100).any():
+                    logger.info("Detected non-log2 data.", file=computed_file)
+                    data = np.log2(data)
+
+                # Ensure that we don't have any dangling Brainarray-generated probe symbols.
+                # BA likes to leave '_at', signifying probe identifiers,
+                # on their converted, non-probe identifiers. It makes no sense.
+                # So, we chop them off and don't worry about it.
```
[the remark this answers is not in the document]
Miserlou: 👍

.circleci/config.yml
```diff
@@ -54,21 +58,21 @@ jobs:
...
-
-        # Run API Tests.
+      
+      # Run API Tests.
       - run: sudo chown -R circleci:circleci workers/test_volume/
       - run: ./api/run_tests.sh
 
+      # Run Illumna tests.
+      - run: sudo chown -R circleci:circleci workers/test_volume/
+      - run: .circleci/filter_tests.sh -t illumina
```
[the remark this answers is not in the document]
Miserlou: 🎉

api/data_refinery_api/tests.py
```diff
@@ -342,7 +342,7 @@ def test_create_update_dataset(self, mock_send_job):
                                     jdata,
                                     content_type="application/json")
         self.assertEqual(response.status_code, 400)
-
+
```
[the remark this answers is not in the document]
Miserlou: Not sure how this happened.

.circleci/config.yml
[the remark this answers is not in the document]
kurtwheeler: Wait I'm confused, I know you already removed one extra instance of these, but before you did so there were actually 3 of them so at this point we're down to 2 instead of 1.

jaclyn-taroni:
> > We now automatically detect if data hasn't been log2 scaled if data.max() > 100).any() (the same logic as with Illumina)

> As far as I know, all of the affy scanners are 16 bit. I think you could do > 16 unless there are Illumina scanners that have more range.

@cgreene do you have any objection to or know of an issue with using `> 100`? If not, we can use this threshold and we don't have to dig into the Illumina scanner issue.

cgreene:
@jaclyn-taroni : no objection. Seems fine. I don't think we'd see a case with un-transformed data where there was no value > 100.

Issue #196, opened 2018-04-10
Appropriately Handle Pre-Processed Illumina Outputs


Miserlou:
### Context

We are trying to collect and harmonize all of the available Illumina samples we can find on GEO.

### Problem or idea

There are two types of Illumina output files that I can find on GEO: raw outputs (I think) and outputs which have already been post-processed. The raw outputs are a single file with multiple sample expression columns, the processed outputs have individual txt files per sample with only a single column, like this:

```
ILMN_1792173    6.211704822
ILMN_1668111    6.011355366
ILMN_1805104    6.225849152
ILMN_1766054    7.000323919
ILMN_1701642    7.501428409
ILMN_1747627    6.549195354
ILMN_1742051    6.364110564
ILMN_1743205    6.56023115
ILMN_1812070    6.105929072
ILMN_1653933    6.157854609
```

For instance, [GSE14428](https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GSE14428) only has individual samples, [found in "MINiML" format](ftp://ftp.ncbi.nlm.nih.gov/geo/series/GSE14nnn/GSE14428/miniml/GSE14428_family.xml.tgz). (The `GSE14428_RAW.tar` supplementary file contains only a `.bgx` file - I don't know what this is.)

On the other hand, [GSE22427](https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GSE22427) contains in `GSE22427_non-normalized.txt.gz` the kind of raw data which our R-based Illumina processor expects:

```
# The identifiers in the ID_REF column must match the identifiers in the ID column of the referenced platform (GPLxxxx).
# The Matrix table should include normalized (scaled) signal count data and detection p-value
"# Values that should be disregarded may either be left blank or labeled as ""null""."

ID_REF  LV-C&si-Control-1       Detection Pval  LV-C&si-Control-2       Detection Pval  LV-C&si-Control-3       Detection Pval  LV-C&si-EZH2-1  Detection Pval  LV-C&si-EZH2-2  Detection Pval  LV-C&si-EZH2-3  Detection Pval  LV-EZH2&si-EZH2-1       Detection Pval  LV-EZH2&si-EZH2-2       Detection Pval  LV-EZH2&si-EZH2-3       Detection Pval  LV-T350A&si-EZH2-1      Detection Pval  LV-T350A&si-EZH2-2      Detection Pval  LV-T350A&si-EZH2-3      Detection Pval
[...]

Issue #188, opened 2018-04-06
Refactor logic which determines what should be done with a given experiment


kurtwheeler:
### Problem or idea

Currently the logic of what should be done with an experiment is scattered across the surveyors which makes it hard to track down what's going on when there's unexpected behavior.

### Context

There's two different ways that the surveyor logic is scattered:
* One is that there are three management commands based on what source the data is coming from.
* The other is that once a survey job starts for a given experiment, logic is performed to see what should be done with it.

### Solution or next step

I think there should be a single component which is responsible for determining what should be done with a given experiment. I can think of two options to accomplish this:
1. Having a single point of entry which makes a request, parses out enough information to decide what to do, and then performs logic to determine what should be done and then fire off the correct surveying code.
2. Have a single module which contains similar logic, but can be called from various surveyors and then return what should be done.

I think I like the latter method for managing pipeline logic more I think. However I think it's actually useful to have different management commands for different sources so we can have different help texts for them and already with only three sources they have different parameters. Shoehorning them into a single command would probably be messier than it is now. Therefore what I propose is to create a function which accepts:
* source_type
* platform
* has_raw
* has_processed
   
as arguments and returns a pipeline that should be run on that experiment along with a field which indicates whether raw or processed data should be downloaded.

kurtwheeler:
@Miserlou and @dongbohu I think I am going to try to implement my proposed solution this weekend. If either of you have any strong feelings that I should do this in a different way it'd be great to hear about them before I implement this. However it is Friday evening so if you don't get back to me I'll just do this and if there's changes I need to make I'll make them.

cgreene:
When you say surveyor here, do you mean processor: "once a survey job starts for a given experiment"?

My impression is that surveyors look at endpoints and identify what's there. Downloaders download. Then things would get kicked to one or more processors. Some of the "what should we do" logic is going to have to rest within a processor because it will depend, in part, on some of the analyses of the data itself.

kurtwheeler:
> Then things would get kicked to one or more processors.

[...]

PR #308, opened 2018-06-08
Kurtwheeler routing refactor

kurtwheeler:
## Issue Number

#188
#196

This PR refactors, consolidates, and clarifies the routing logic we perform to determine what should be done with any given kind of data. The general strategy I took here was to set the properties on `sample_objects` during surveying (and some additional checks during downloader jobs as needed) and then just have my routing functions take a Sample as an input. One bug feature of this approach is that if we ever are tracking down a sample that didn't get routed correctly, we'll be able to look at the database record for it and exactly what the routing logic used to route it.

## Methods

The methods used in this PR build on those discussed in #293 

Most of the routing logic is now contained in `job_lookup.determine_downloader_task` and `job_lookup.determine_processor_pipeline`, so looking at those functions will be a good way to see what's being done here.

Those functions mostly are just using the metadata that was already discussed in #293. The one piece of additional metadata that they're using is the manufacturer, which I realized was necessary after I started this PR. 

[Here's one place](https://github.com/AlexsLemonade/refinebio/pull/308/files#diff-23de673953caefc5b0fc9030bd21ff20R94) the manufacturer gets set.

[Here's another](https://github.com/AlexsLemonade/refinebio/pull/308/files#diff-d9117937cb121a9f4a2ba51b656bf3e5R402)

[And one more](https://github.com/AlexsLemonade/refinebio/pull/308/files#diff-187890faf2c06f308f803eabf9e9eb64R88)

I think this PR addresses #196. I'm not 100% sure what the upshot of that discussion was other than:
>  in a case like GSE14786, we'll want to NO-OP the submitter processed data.

I tested that accession specifically and it got NO-OP'd just like we wanted.

## Types of changes

## Functional tests

I ran a lot of surveyor jobs to make sure that jobs got routed correctly. At least one job of each type was allowed to queue processor jobs, but for most of them I prevented even downloader jobs to get queued to keep the local network speedy ;)
```
./foreman/run_surveyor.sh survey_all --accession=E-GEOD-17114
./foreman/run_surveyor.sh survey_all --accession=E-MTAB-1944
./foreman/run_surveyor.sh survey_all --accession=E-GEOD-19617
./foreman/run_surveyor.sh survey_all --accession=E-GEOD-32413
./foreman/run_surveyor.sh survey_all --accession=E-MEXP-32
./foreman/run_surveyor.sh survey_all --accession=E-MTAB-11
./foreman/run_surveyor.sh survey_all --accession=E-GEOD-15573
./foreman/run_surveyor.sh survey_all --accession=E-GEOD-24742
./foreman/run_surveyor.sh survey_all --accession=E-MEXP-3390
./foreman/run_surveyor.sh survey_all --accession=GSE19274
./foreman/run_surveyor.sh survey_all --accession=GSE62564
./foreman/run_surveyor.sh survey_all --accession=E-TABM-38
./foreman/run_surveyor.sh survey_all --accession=GSE49710
./foreman/run_surveyor.sh survey_all --accession=GSE113945
./foreman/run_surveyor.sh survey_all --accession=GSE79910
./foreman/run_surveyor.sh survey_all --accession=GSE65303
./foreman/run_surveyor.sh survey_all --accession=GSE73517
./foreman/run_surveyor.sh survey_all --accession=GSE108682
./foreman/run_surveyor.sh survey_all --accession=GSE14786
./foreman/run_surveyor.sh survey_all --accession=ERP001942
./foreman/run_surveyor.sh survey_all --accession=SRP024369
./foreman/run_surveyor.sh survey_all --accession=ERP000263
./foreman/run_surveyor.sh survey_all --accession=DRP000161
./foreman/run_surveyor.sh survey_transcriptome Ensembl 1 "Mus Musculus"
```

## Checklist


common/data_refinery_common/job_lookup.py
```diff
@@ -23,13 +34,103 @@ class ProcessorPipeline(PipelineEnums):
...
     """An enumeration of downloaders for downloader_task."""
     ARRAY_EXPRESS = "ARRAY_EXPRESS"
     SRA = "SRA"
     TRANSCRIPTOME_INDEX = "TRANSCRIPTOME_INDEX"
     GEO = "GEO"
+    NONE = "NONE"
+
+
+def _is_platform_supported(platform: str) -> bool:
+    upper_platform = platform.upper()
```
Miserlou: Docstring

common/data_refinery_common/job_lookup.py
```diff
@@ -23,13 +34,103 @@ class ProcessorPipeline(PipelineEnums):
...
+    for supported_platform in utils.get_supported_rnaseq_platforms():
+        # Spacing can be inconsistent, easiest to just remove it entirely.
+        if supported_platform.upper().replace(" ", "") in upper_platform.replace(" ", ""):
+            return True
+
+    return False
+
+
+def determine_downloader_task(sample_object: Sample) -> Downloaders:
+    if _is_platform_supported(sample_object.platform_accession_code):
```
Miserlou: This seems like a very important function, so should a docstring describing its high-level purpose

common/data_refinery_common/job_lookup.py
```diff
@@ -23,13 +34,103 @@ class ProcessorPipeline(PipelineEnums):
...
+        original_files = OriginalFile.objects.filter(id__in=relations.values('original_file_id'))
+        for original_file in original_files:
+            if original_file.source_filename[-4:].upper() == ".CEL":
+                return Downloaders[sample_object.source_database]
+
+    return Downloaders.NONE
+
+
+def determine_processor_pipeline(sample_object: Sample) -> ProcessorPipeline:
+    if not _is_platform_supported(sample_object.platform_accession_code):
```
Miserlou: Same as above - this seems like a very important function, so should a docstring describing its high-level purpose

common/make_migrations.sh
```diff
@@ -27,12 +27,12 @@ docker run \
        --env-file common/environments/dev \
        --interactive \
        --link drdb:postgres\
-       dr_models python3.6 manage.py makemigrations data_refinery_common --noinput
```
Miserlou: Why was this removed? This is necessary for running this command in CI. That's why we added it in the first place.

common/make_migrations.sh
```diff
@@ -27,12 +27,12 @@ docker run \
...
+       dr_models python3.6 manage.py makemigrations data_refinery_common
 
 docker run \
        --volume $script_directory/data_refinery_common:/home/user/data_refinery_common \
        --add-host=database:$DB_HOST_IP \
        --add-host=nomad:$HOST_IP \
        --env-file common/environments/dev \
        --link drdb:postgres \
-       dr_models python3.6 manage.py migrate --noinput
```
Miserlou: Why was this removed? This is necessary for running this command in CI. That's why we added it in the first place.

foreman/data_refinery_foreman/surveyor/external_source.py
```diff
@@ -90,65 +140,21 @@ def queue_downloader_jobs(self, experiment: Experiment):
...
-                send_job(downloader_job.downloader_task, downloader_job.id)
+                            downloader_job=downloader_job.id,
+                            downloaded_urls=downloaded_urls)
+                message_queue.send_job(downloader_task, downloader_job.id)
             except Exception as e:
                 # If the task doesn't get sent we don't want the
                 # downloader_job to be left floating
-                logger.exception("Failed to enqueue downloader job for URL: "
-                                 + original_file.source_url,
+                logger.exception("Failed to enqueue downloader job.",
```
Miserlou: Can we check how this logger.exception field works? I'm not actually sure that these ever appear in our nomad logs.

foreman/data_refinery_foreman/surveyor/external_source.py
```diff
@@ -90,65 +140,21 @@ def queue_downloader_jobs(self, experiment: Experiment):
                 downloaded_urls.append(original_file.source_url)
 
             try:
-                logger.info("Queuing downloader job for URL: " + original_file.source_url,
+                logger.info("Queuing downloader job.",
                             survey_job=self.survey_job.id,
-                            downloader_job=downloader_job.id)
-                send_job(downloader_job.downloader_task, downloader_job.id)
+                            downloader_job=downloader_job.id,
+                            downloaded_urls=downloaded_urls)
```
Miserlou: 👍

foreman/data_refinery_foreman/surveyor/management/commands/survey_all.py
```diff
@@ -70,3 +77,17 @@ def handle(self, *args, **options):
                             surveyor.survey_sra_experiment(accession)
                     except Exception as e:
                         logger.exception(e)
+
+        if options["accession"]:
```
Miserlou: This is just a copy paste of the code right above it, you should functionalize this

foreman/data_refinery_foreman/surveyor/sra.py
```diff
@@ -187,6 +187,12 @@ def gather_run_metadata(run_accession: str) -> Dict:
         response = utils.requests_retry_session().get(
             ENA_METADATA_URL_TEMPLATE.format(run_accession))
         run_xml = ET.fromstring(response.text)
+
+        # Necessary because ERP000263 has only one ROOT element containing this error:
+        # Entry: ERR15562 display type is either not supported or entry is not found.
+        if len(run_xml) == 0:
+            return {}
```
Miserlou: Is this enough? Should this be explicitly handled as a fail case?

foreman/data_refinery_foreman/surveyor/sra.py
```diff
@@ -267,10 +273,11 @@ def gather_study_metadata(metadata: Dict) -> None:
     def gather_all_metadata(run_accession):
         metadata = SraSurveyor.gather_run_metadata(run_accession)
 
-        SraSurveyor.gather_experiment_metadata(metadata)
-        SraSurveyor.gather_sample_metadata(metadata)
-        SraSurveyor.gather_study_metadata(metadata)
-        SraSurveyor.gather_submission_metadata(metadata)
+        if metadata != {}:
```
Miserlou: Again, if metadata is {} we have nothing of value to work with and this is probably a fail..

foreman/data_refinery_foreman/surveyor/test_external_source.py
```diff
@@ -25,9 +25,19 @@ def test_queue_downloader_jobs(self, mock_send_task):
 
         sample_object_1 = Sample()
         sample_object_1.accession_code = "Sample1"
+        sample_object_1.platform_accession_code = "Illumina Genome Analyzer"
+        sample_object_1.platform_accession_name = "Illumina Genome Analyzer"
```
Miserlou: Duplicated line

foreman/data_refinery_foreman/surveyor/test_external_source.py
```diff
@@ -25,9 +25,19 @@ def test_queue_downloader_jobs(self, mock_send_task):
...
+        sample_object_1.technology = "RNA-SEQ"
+        sample_object_1.manufacturer = "ILLUMINA"
+        sample_object_1.source_database = "SRA"
         sample_object_1.save()
         sample_object_2 = Sample()
         sample_object_2.accession_code = "Sample2"
+        sample_object_2.platform_accession_code = "Illumina Genome Analyzer"
+        sample_object_2.platform_accession_name = "Illumina Genome Analyzer"
```
Miserlou: Duplicated duplicated line

workers/data_refinery_workers/downloaders/geo.py
```diff
@@ -138,11 +145,11 @@ def download_geo(job_id: int) -> None:
 
     file_assocs = DownloaderJobOriginalFileAssociation.objects.filter(downloader_job=job)
 
-    original_file = file_assocs[0].original_file 
+    original_file = file_assocs[0].original_file
```
Miserlou: If these are unique now you can use `.get`

workers/data_refinery_workers/downloaders/geo.py
```diff
@@ -225,21 +233,26 @@ def download_geo(job_id: int) -> None:
                 unpacked_sample_files.append(actual_file)
             except Exception as e:
                 # TODO - is this worth failing a job for?
-                logger.warn("Found a file we didn't have an OriginalFile for! Why did this happen?: " + og_file['filename'],
-                    exc_info=1, file=og_file['filename'], sample_id=sample_id, accession_code=accession_code)
+                logger.warn("Found a file we didn't have an OriginalFile for! Why did this happen?: "
+                            + og_file['filename'],
+                            exc_info=1, file=og_file['filename'],
```
Miserlou: newline

workers/data_refinery_workers/downloaders/sra.py
```diff
@@ -107,20 +114,41 @@ def _download_file_aspera(  download_url: str,
...
+                                     target_file_path,
+                                     attempt + 1
+                                     )
+    return True
+
+
+def _are_downloads_ready(original_files: List[OriginalFile]) -> bool:
+    """RNASeq reads can be paired. Makes sure both files are downloaded if applicable.
+    """
+    original_file = original_files[0]
```
Miserlou: Can we change the function signature rather than always taking the first of a list

common/make_migrations.sh
kurtwheeler: This ticket I opened with Django sums up my beef with that flag: https://code.djangoproject.com/ticket/29470

foreman/data_refinery_foreman/surveyor/test_external_source.py
kurtwheeler: Not quite, one is setting the code the other is setting the name.

foreman/data_refinery_foreman/surveyor/sra.py
kurtwheeler: I agree. I supposed I could have raised an exception and then caught it higher up, but instead I added some additional logic to make sure that the dictionary returned by these functions isn't empty. If you feel strongly about it can I refactor it to use exceptions.

foreman/data_refinery_foreman/surveyor/test_external_source.py
kurtwheeler: Not quite, one is setting the code the other is setting the name.

Miserlou: "The bond's Name. James Name" 

Pleased to... what? 

"Bond Name's the james" 

Are you alright? 

"Bames Nond's having a stronk, call a Bondulance"

foreman/data_refinery_foreman/surveyor/external_source.py
kurtwheeler: They appear as log level ERROR.

foreman/data_refinery_foreman/surveyor/sra.py
```diff
@@ -379,13 +391,22 @@ def _generate_experiment_and_samples(self, run_accession: str) -> None:
...
             sample_object.organism = organism
 
-            sample_object.platform_name = metadata.get("platform_instrument_model", "No model.")
+            sample_object.platform_name = metadata.get("platform_instrument_model", "UNKNOWN")
             # No platform accession nonsense with RNASeq, just use the name:
             sample_object.platform_accession_code = sample_object.platform_name
             sample_object.technology = "RNA-SEQ"
+            if "ILLUMINA" in sample_object.platform_name.upper():
+                sample_object.manufacturer = "ILLUMINA"
+            elif "NEXTSEQ" in sample_object.platform_name.upper():
```
jaclyn-taroni: NextSeq is an Illumina product: https://www.illumina.com/systems/sequencing-platforms/nextseq.html

Not exactly sure why that is not reflected in the platform name from data sources. Is the distinction important here or should this also be `sample_object.manufacturer = "ILLUMINA"`?

workers/data_refinery_workers/downloaders/geo.py
kurtwheeler: Mmm I didn't change any of the logic around this, just deleted a trailing whitespace. It does seem like they should be unique, however before I changed it I guess I'd just be curious if there was any reason you didn't do it like that in the first place.

foreman/data_refinery_foreman/surveyor/sra.py
kurtwheeler: It is not important (at least not yet). At this point manufacturer is pretty much ignored for RNA-SEQ technologies, I just figured that since we could we may as well record it now in case we ever did have a use for it later. I will update this so NEXTSEQ gets marked as  having ILLUMINA for a manufacturer.

common/data_refinery_common/models/models.py
```diff
@@ -54,6 +55,7 @@ def __str__(self):
     platform_accession_code = models.CharField(max_length=256, blank=True)
     platform_name = models.CharField(max_length=256, blank=True)
     technology = models.CharField(max_length=256, blank=True)
+    manufacturer = models.CharField(max_length=256, blank=True)
```
Miserlou: It'd be great if you could update the serializers in api to add these two new fields

common/data_refinery_common/job_lookup.py
```diff
@@ -23,13 +34,129 @@ class ProcessorPipeline(PipelineEnums):
...
+                    return ProcessorPipeline.AGILENT_ONECOLOR_TO_PCL
+            elif sample_object.manufacturer == "UNKNOWN":
+                logger.error("Found a Sample on a supported platform with an unknown manufacturer.",
+                             sample=sample_object.id,
+                             platform_accession=sample_object.platform_accession_code)
+                return ProcessorPipeline.NONE
+
+    else:
+        if not sample_object.has_raw:
+            # Only NO_OP RNASeq data if it's from SRA, because
```
jaclyn-taroni: To my knowledge, SRA doesn't contain any processed data. Have you found an example of processed data there? I don't think it will matter because of the way you have the logic here, but we wouldn't want to NOOP anything from SRA if there's no processed data (unless there's a metadata consideration here I am unaware of).

jaclyn-taroni:
Does this also address #159?

common/data_refinery_common/job_lookup.py
kurtwheeler: Ok, so I'll change the RNA-SEQ logic block to just check if there is raw data and if so to SALMON, otherwise don't do anything.

common/data_refinery_common/models/models.py
kurtwheeler: Done

kurtwheeler:
@Miserlou and @jaclyn-taroni I believe I have addressed your comments/concerns. Are we good here?

jaclyn-taroni:
Re: #196 we may "more fully" address that when we take care of #232. So I think it's reasonable at this stage to just ensure that particular accession is NOOP (probably it should still be even if we tackle #232).

PR #309, opened 2018-06-08
Slice tximport Output Into Smashable Individual Files

Miserlou:
## Issue Number

Depends on: https://github.com/AlexsLemonade/refinebio/pull/303

`tximport` outs to a single massive TSV. This splits that massive file into individual files which can be used by the smasher.

## Methods

There aren't really any scientific changes to the data here. However, we may want to change the mode that salmon operates in future for the tsv that gets split up. From Slack:

> Rich [9:57 AM]
> okay but it makes sense like scientifically
> they can be treated as individuals after tximport
> 
> jaclyn.taroni [9:58 AM]
> I think it is fine for now
> we might want different output from tximport (e.g., not `lengthScaledTPM`) to do this properly
> also it will probably depend quite a bit on the downstream application

## Types of changes

## Functional tests

Additional assertions were added to the tximport test.

## Checklist


workers/data_refinery_workers/processors/tximport.R
```diff
@@ -59,5 +59,3 @@ lstpm_df <- as.data.frame(txi_length_scaled$counts)
 lstpm_df <- tibble::rownames_to_column(lstpm_df, var = "Gene")
 tpm_filename <- file.path(exp_dir, "gene_lengthScaledTPM.tsv")
 readr::write_tsv(lstpm_df, path = tpm_filename)
-# Compress this output file
-R.utils::gzip(tpm_filename)
```
dongbohu: Since this line is deleted, `R.uitls` package is not needed, so we can also delete the following line:
https://github.com/AlexsLemonade/refinebio/blob/220537e621d6355b2e54faa125ee84edeef3b55e/workers/install_tximport.R#L12

workers/dockerfiles/Dockerfile.salmon
```diff
@@ -61,6 +61,9 @@ COPY workers/data_refinery_workers/processors/requirements.txt .
 
 RUN pip3 install -r requirements.txt
 
+# Install our python data tools
+RUN pip3 install numpy scipy matplotlib pandas scikit-learn sympy nose
```
dongbohu: Do you think it will be better to include these packages in `requirements.txt`?

workers/run_tests.sh
```diff
@@ -51,6 +51,7 @@ if [[ -z $tag || $tag == "salmon" ]]; then
     # Download tximport test data
     rm -rf $volume_directory/tximport_test/
     git clone https://github.com/dongbohu/tximport_test.git $volume_directory/tximport_test
+    gunzip $volume_directory/tximport_test/expected_output/gene_lengthScaledTPM.tsv.gz
```
dongbohu: I have replaced the gzipped file with unzipped one in my `tximport_test` github repo. So you can remove this line now.

Issue #274, opened 2018-05-23
Determine ensembl version via API call rather than hardcoding versions.


kurtwheeler:
### Context

Ensembl released a new version of its API's and this broke our code:https://github.com/AlexsLemonade/refinebio/pull/266#pullrequestreview-122598723

### Problem or idea

The ensembl mailing list replied to the concern I raised about it: http://lists.ensembl.org/pipermail/dev/2018-May/013014.html

The upshot is that they do archive the last five releases, but at special URLs. The more useful information is that the current release can be found by hitting http://rest.ensembl.org/info/software?content-type=application/json. So we can harden our code by using that rather than hardcoded version numbers.

### Solution or next step

Rather than having `MAIN_DIVISION_ASSEMBLY_VERSION` and `OTHER_DIVISIONS_ASSEMBLY_VERSION` hardcoded in `foreman/data_refinery_foreman/surveyor/transcriptome_index.py`, first hit the endpoint to determine the correct version to specify.

I tagged this as a SCIENCE issue because I think we need @jaclyn-taroni and @cgreene to make sure that automatically using to the latest version of ensembl won't be undesired behavior. I'll note that it's not like we'll automatically reprocess all the organism indices when a new version is released, we'll just use the latest version whenever we do decide to rerun the Index Refinery.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

cgreene:
Using the latest version whenever we decide to update the indices seems fine. We need to know which version was used for a specific organism though, and we also need to store the transcriptome indices that we have used and make them available (potentially) to users. Doesn't have to be a fancy approach - sticking them in a bucket with a link to them should be fine.

jaclyn-taroni:
I agree

jaclyn-taroni:
Was this closed by #310?

cgreene:
Looks like it. https://github.com/AlexsLemonade/refinebio/pull/310/files#diff-aad5724ec88845dd627fd8338f38ed27R127

Should be safe to close.

Issue #158, opened 2018-03-08
OrganismIndex should be versioned and include genome build information


kurtwheeler:
### New Issue Checklist
- [x] The title is short and descriptive
- [ ] The issue contains an:
  - [ ] Idea (new feature, user story, etc)
  - [x] Problem
- [x] You have explained the:
  - [x] Context
  - [x] Problem or idea
  - [x] Solution or next step
  
### Context

In the refactor we are adding a table named OrganismIndex which will be used for storing information about indices built by the so called "Index Refinery". The first one we're building is for Salmon indices. Every now and then as Ensembl updates their genomes we'll want to rebuild these indices.

### Problem or idea

We need to give each build of these indices a different version so we can track which version was used when processing RNAseq data.

### Solution or next step

Come up with a way of versioning different pre-built indices. I would propose using a date since it's simple and this really isn't a good application of semantic versioning.

cgreene:
Is this closed #310

kurtwheeler:
Yes, the `organism_index` table now has a `source_version` field which is set to the version of the Ensembl  release which was used to build it.

PR #310, opened 2018-06-08
Use assembly version from Ensembl's API rather than hardcoding it. St…

kurtwheeler:
## Issue Number

#274 
#158

We were hardcoding the Ensembl assembly version rather than determining it via their API. This will future-proof us for those (since you can't query for metadata about past versions using the same REST interface). Additionally, and this is the part that is more useful for Keytar Kurt, we're storing these versions on the organism_index tables now, so that we can manage versions for them.

## Methods

Scientific discussion already happened on #274.

## Types of changes

## Functional tests

I built the transcriptome index for Mus Musculus and it worked.

## Checklist


foreman/data_refinery_foreman/surveyor/transcriptome_index.py
```diff
@@ -65,7 +60,7 @@ def __init__(self, species: Dict):
         self.url_root = "ensemblgenomes.org/pub/release-{assembly_version}/{short_division}"
         self.short_division = DIVISION_LOOKUP[species["division"]]
         self.assembly = species["assembly_name"].replace(" ", "_")
-        self.assembly_version = OTHER_DIVISIONS_ASSEMBLY_VERSION
+        self.assembly_version = requests.get(DIVISION_RELEASE_URL).json()["release"]
```
Miserlou: Use the retrier in util?

I wish `requests` came with a warning (it actually does in the upcoming v3): NEVER USE .GET IN PRODUCTION WITHOUT A `,timeout=`! The retrier we already have will take care of this.

foreman/data_refinery_foreman/surveyor/transcriptome_index.py
```diff
@@ -130,7 +125,7 @@ def __init__(self, species: Dict):
         self.species_sub_dir = species["name"]
         self.filename_species = species["name"].capitalize()
         self.assembly = species["assembly"]
-        self.assembly_version = MAIN_DIVISION_ASSEMBLY_VERSION
+        self.assembly_version = requests.get(MAIN_RELEASE_URL).json()["release"]
```
Miserlou: Same as above

workers/data_refinery_workers/processors/transcriptome_index.py
```diff
@@ -87,6 +87,33 @@ def _prepare_files(job_context: Dict) -> Dict:
...
+    transcriptome index's only database model is the OriginalFiles
+    until processing is complete, there's no other way to pass this
+    information through to this processor without modifying the
+    OriginalFile model.
+    """
+    original_files = job_context["original_files"]
+
+    for og_file in original_files:
+        if "gtf.gz" in og_file.source_filename:
+            extensionless_url = og_file.source_url[:-7]
```
dongbohu: I suppose `-7` here is used because the length of `.gtf.gz` is 7? If it is, it might be better to use `.gtf.gz` on line 110.

cgreene: Is this preferred over something like os.path.basename? What if somehow gtf.gz isn't the final seven characters but happens to be somewhere else in the name (not sure why that would be, but just imagining corner cases).

dongbohu: No matter which implementation we use to extract the version number, I think it will be helpful to include the current gtf.gz file's exact format in the docstring of this function: `<species>.<version>.gtf.gz` (in case it may be changed in the future). It will also make the code more readable.

kurtwheeler: `os.path.basename` won't help much here:
```python
>> import os
>>> os.path.basename("/some/really/long/path.with.file.extensions")
'path.with.file.extensions'
```

> What if somehow gtf.gz isn't the final seven characters but happens to be somewhere else in the name (not sure why that would be, but just imagining corner cases).

We don't have many corner cases to worry about, because we're the ones constructing these URLs in `foreman/data_refinery/surveyor/transcriptome_index.py` by formatting our templates. The reason I'm checking for `.gtf.gz` is because there are two OriginalFile records associated with the job and I want to parse out the assembly version from the correct one (the fasta file URL doesn't have the assembly version in it).

In general, this isn't really an approach I'd be happy with (and I even mention so in the docstring), but this is the only way we're storing the assembly version at all in our database and this is a relatively minor issue to rework the transcritome-index model over so close to Keytar Kurt.

However I will make both of the changes Dongbo suggested to improve clarity.

Issue #117, opened 2018-02-12
Add a unit test for Salmon 


kurtwheeler:
I asked Rob Patro, creator of Salmon, this question:

> My first question is about what an appropriate unit test would be. I have unit tests in place to make sure that the code that runs salmon works, however at the moment it is just checking that salmon ran and output something. Can you think of any tests which could be used as unit tests which would verify that the output of salmon is meaningful or is the fact that salmon runs without an error evidence enough that it was successful? (Note that since this will be run as a unit test it will only be run on a single sample to make sure that the code hasn't broken, not run on every sample as a QC check.) 

And his response was:

> Great question.  There are basically 2 options I can think of.  First is what we do internally for regression testing etc.  We have some simulated data (which you can generate with Polyester), and then we have a script (https://github.com/COMBINE-lab/salmon/blob/master/.drone/test_quant.sh) that runs Salmon on the samples, and then calls a python script to compute and print out the correlations, to ensure they are high.  This by no means covers everything, but it is a reasonable basic sanity check. 

> Another option, if this will only be for a single sample.  Would be to do something similar on real data.  For example, compute gene-level estimates using an alternative pipeline (e.g STAR + featureCounts), and then keep the actual gene-level quantification results as part of your test suite.  Then, the test could run Salmon on the same sample, aggregate the quantifications to the gene level using tximport, and check that the two different pipelines generally concord.  Of course, there will be differences (hence why we developed Salmon in the first place :)), but the correlation should be reasonably high.

I think there may be a third option, which is to do something similar to SCAN.UPC's unit test. What we do there is have an output file which @jaclyn-taroni ran using proper parameters included in the tests, and then we re-process the same input files and compare the output to make sure they're well correlated. This seems like it would be just about as effective as Rob's first option, but perhaps not quite as robust as his second. However his second option includes implementing a new pipeline just to check the results, and I think what we really want to do with this unit test is make sure that we don't break the method of processing such that it starts outputting results that are drastically different than what we've gotten in the past.

Miserlou:
Related?:  https://github.com/COMBINE-lab/salmon/issues/189

kurtwheeler:
I'd say that they're related, but ideally we'd have two tests around this. One which performs the dry-run all the time, and another which actually verifies that Salmon's output is consistent with its historical output. (This second would be marked with `@tag('slow')`)

jaclyn-taroni:
[...]

PR #311, opened 2018-06-11
Salmon quant unit tests

dongbohu:
## Issue Number
#117 

Include two experiment test cases for `salmon_.run_salmon` function. One fires `tximport` and the other doesn't.

## Methods
The code and tests depend heavily on how the sample data are organized in a certain experiment directory. Unfortunately the sample dataset is too large to be hosted in a github repository (like the tests cases of `salmon_tools` and `tximport`). Instead, it is provisioned as a tarball file in AWS S3 bucket. I plan to create a doc file somewhere to describe the required file system hierarchy.

Thanks to the test cases, this PR also includes fixes of a couple of minor bugs in transition from `salmon quant` step to `tximport` step. 

## Types of changes

- Bug fix 

## Functional tests

## Checklist


workers/data_refinery_workers/processors/salmon.py
```diff
@@ -327,6 +330,7 @@ def _run_salmon(job_context: Dict, skip_processed=SKIP_PROCESSED) -> Dict:
         with transaction.atomic():
             ComputationalResult.objects.select_for_update()
             result.save()
+            SampleResultAssociation.objects.create(sample=job_context['sample'], result=result)
```
Miserlou: We are trying to use `SampleResultAssociation.objects.get_or_create` as these are unique now and this will fail if the object already exists

dongbohu: Sure. Good point!

Issue #229, opened 2018-04-27
Add database tables to track reproducibility information.


kurtwheeler:
### Context

Reproducibility is one of the primary requirements of our system. An important part of reproducibility is knowing what versions of programs and dependencies were used to produce our outputs. We have taken a lot of measures to make sure this information is tracked _in some way_, but we currently don't have a good way to expose this information to users other than providing a system version and pointing them to sift through github.

### Problem or idea

We need database tables dedicated to storing this information. Me, @jaclyn-taroni, and @dvenprasad had a discussion about this and came up with an initial specification of these tables which we think we will make displaying this information easy.

##### programs
This table tracks the programs we run. Every combination of command, program_version, and system_version should be unique in the table. Its columns are:
__id__
__program__: the name of the program (ex: salmon)
__program_version__
__command__: the way we run this command so people know what options we used
__system_version__

##### processing_steps
This table tracks the order of steps used to achieve a computational result. Every combination of computationalresult_id and program_id should be unique. Its columns are:
__computationalresult_id__
__program_id__: an id relating to the `programs` table.
__order__: the step in processing that this was run to produce the computational result

##### dependencies
This table tracks all of the dependencies our system has that are not programs. It is functionally the same as programs, but is split off so that the database enforces that a dependency can never be listed as a `processing_step`. Each combination of all of its fields except id should be unique. Its columns are:
__id__
__package_name__: the name of the program (ex: salmon)
__package_version__
__system_version__

These tables will allow us to display the full pipeline for a computational result with detailed information about each step. We probably won't do too much in the UI with the `dependencies` table other than just display the full thing on a page if someone wants to look for it. However exposing it in the API will allow a savvy user to ask "hey for refine bio 1.2.6, what version of annotation array was used" with a call like `curl http://refine.bio/api/dependencies?system_version=1.2.6` or to potentially build this into our clients later.

### Solution or next step

I think that there should probably just be a CSV file for both the `programs` and `dependencies` tables which gets loaded into the database upon each deploy. Note that when I say loaded I do not mean to empty the table and reload it, but just add the data on top of what is already there. We will need to make sure that these csvs are accurate for every release. If we want to get fancy with it we could have our dependency files read it in to ensure they are installing the versions specified.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

[...]

PR #313, opened 2018-06-13
New tables to keep track of reproducibility 

dongbohu:
## Issue Number

#229 
As discussed above, the current implementation is more like a prototype, which will provide some backend data to unblock a front end issue: AlexsLemonade/refinebio-frontend#65. We need to come back later to get a better design.

I will file another PR to address the API.

Add new tables to keep track of programs and other packages used in processors. 

## Types of changes

## Functional tests

## Checklist

### DIFF
--- a/common/data_refinery_common/management/commands/programs.tsv
+++ b/common/data_refinery_common/management/commands/programs.tsv
@@ -1,7 +1,236 @@
-# progam_name	program_version	command
+# progam	N/A	name	program	N/A	version	command
+
+#################################################
+# Part 1: programs executed in "salmon" pipeline
+#################################################
+
 salmon	0.9.1	salmon --no-version-check quant -l A --biasSpeedSamp 5 -i <index_dir> -1 <input_file_1> -2 <input_file_2> -p 20 -o <output_dir> --seqBias --gcBias --dumpEq --writeUnmappedNames
 salmon	0.9.1	salmon --no-version-check quant -l A --biasSpeedSamp 5 -i <index_dir> -r <file> -p 20 -o <output_dir> --seqBias --dumpEq --writeUnmappedNames
 salmontools	0.1.0	salmontools extract-unmapped -u <unmapped_file> -o <output> -1 <input_file_1> -2 <input_file_2>
 salmontools	0.1.0	salmontools extract-unmapped -u <unmapped_file> -o <output> -r <input_file>
 FastQC	0.11.7	fastqc --outdir=<qc_dir> <files>
 tximport	1.6.0	tximport::tximport(files = quant_files, type = "salmon", txOut = FALSE, tx2gene = <transcript_to_gene_mapping>, countsFromAbundance = "no" or "lengthScaledTPM")
+
+################################################
+# Part 2: R packages:
+################################################
+
+r-base	>=3.4	N/A
+
+# tximport dependencies in "salmon" pipeline
+optparse	1.4.4	N/A
+rjson	0.2.19	N/A
+readr	1.1.1	N/A
+
+# affymetrix dependencies
+###########################
+ff	2.2-13	N/A
+XML	3.98-1.10	N/A
+RCurl	1.95-4.10	N/A
+RSQLite	2.0	N/A
+tibble	1.4.2	N/A
+xtable	1.8-2	N/A
+pkgconfig	2.0.1	N/A
+# Illumina
+doParallel	1.0.11	N/A
+data.table	1.11.0	N/A
+# Bioc v3.6 packages
+oligo	1.42.0	N/A
+Biobase	2.38.0	N/A
+SCAN.UPC	2.20.0	N/A
+affy	1.56.0	N/A
+affyio	1.48.0	N/A
+AnnotationDbi	1.40.0	N/A
+zlibbioc	1.24.0	N/A
+preprocessCore	1.40.0	N/A
+genefilter	1.60.0	N/A
+sva	3.26.0	N/A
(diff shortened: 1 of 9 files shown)


common/data_refinery_common/management/commands/programs.tsv
```diff
@@ -0,0 +1,7 @@
+# progam_name	program_version	command
```
dongbohu: This is an overly simple file for demo only. Obviously more programs should be added later.

common/data_refinery_common/models/models.py
```diff
@@ -705,3 +706,48 @@ class SampleResultAssociation(models.Model):
...
+    """
+
+    computational_result = models.ForeignKey(ComputationalResult, blank=False, null=False,
+                                    on_delete=models.CASCADE)
+    program = models.ForeignKey(Program, blank=False, null=False, on_delete=models.CASCADE)
+    order = models.PositiveIntegerField(blank=False, null=False)
+
+    class Meta:
+        db_table = "process_steps"
+        unique_together = ('computational_result', 'order')
```
dongbohu: @kurtwheeler: Instead of setting `computational_result` and `program` unique together, I am setting ('computational_result', 'order'), because a program might be called multiple times to generate a certain final result, but the order in that result must be unique value. Does this make sense to you?

kurtwheeler: That seems reasonable to me :+1:

common/data_refinery_common/management/commands/programs.tsv
```diff
@@ -0,0 +1,7 @@
...
+salmon	0.9.1	salmon --no-version-check quant -l A --biasSpeedSamp 5 -i <index_dir> -1 <input_file_1> -2 <input_file_2> -p 20 -o <output_dir> --seqBias --gcBias --dumpEq --writeUnmappedNames
+salmon	0.9.1	salmon --no-version-check quant -l A --biasSpeedSamp 5 -i <index_dir> -r <file> -p 20 -o <output_dir> --seqBias --gcBias --dumpEq --writeUnmappedNames
```
jaclyn-taroni: No GC-bias correction on single-end reads, right? https://github.com/AlexsLemonade/refinebio/blob/caa2fd630de8d9e6e8bc1b2dfbedf24455e36baa/workers/data_refinery_workers/processors/salmon.py#L314

common/data_refinery_common/management/commands/programs.tsv
```diff
@@ -0,0 +1,7 @@
...
+salmontools	0.1.0	salmontools extract-unmapped -u <unmapped_file> -o <output> -1 <input_file_1> -2 <input_file_2>
+salmontools	0.1.0	salmontools extract-unmapped -u <unmapped_file> -o <output> -r <input_file>
+FastQC	0.11.7	fastqc --outdir=<qc_dir> <files>
+tximport	1.6.0	tximport::tximport(files = sf_files, type = "salmon", txOut = FALSE, tx2gene = tx2gene, countsFromAbundance = "no" or "lengthScaledTPM")
```
jaclyn-taroni: We could potentially sub `<quant_files>` for `sf_files` and `<transcript_to_gene_mapping>` for `tx2gene`. I'm not entirely sure that would improve clarity, but just a thought.

common/data_refinery_common/management/commands/programs.tsv
dongbohu: @jaclyn-taroni You are definitely right> I made a mistake during copy/paste. Thanks!

common/data_refinery_common/management/commands/programs.tsv
dongbohu: @jaclyn-taroni Sure. (I have to confess that I created this file w/o putting too much thought.)

dongbohu:
Good point! I will modify the code as soon as  possibles.

You are also right about the API part. It is in my plan!

dongbohu:
@Miserlou: I modifed this PR based on the feedback from Kurt and Jackie. I also modified a few files in `api/data_refinery_api/` to add the API support. Can you take a look at the API changes? (I am not an expert on django-rest-framework and the changes are mainly copy/paste.)

dongbohu:
@Miserlou: Forgot to mention, you can skip the other changes for now. The proposal that Kurt gave in his comment yesterday is not flawless either. We need further discussion on it later.

Miserlou:
The program serializer looks good, but I don't think it's actually going to be used by the result serializer, so it'll never actually be exposed unless it's looked up directly. You might also want to add a test which checks that the expected value of the response is correct in addition to the HTTP status code.

dongbohu:
@Miserlou Yes, I didn't link the program serializer with result serializer, because it is very likely that we will modify the current models for the reproducibility feature. I just want to provide some API for front end to work with.

I will confirm the values in the test case. Thanks.

dongbohu:
@jaclyn-taroni: I have updated `programs.tsv` files to include the R dependencies. Originally I used to the R script snippet to get all R packages, but the output includes so many dependencies that are not directly related to data analysis. So I include only the R packages that are listed explicitly in either docker files or the R scripts that were called directly.

dongbohu:
I am using "N/A" as command for the dependencies in `programs.tsv`. I also added a workaround in `import_programs.py` to skip these dependencies for now. The script will be modified later once we decide the exact database schema.

dongbohu:
I am closing this PR because the new approach to #229 is very different from the code in this PR.
See #320, #321, #322, #323 and #324.

Issue #199, opened 2018-04-11
Create additional environments for local/staging and make environments better!


kurtwheeler:
### Context

We do not have enough environments! We currently have `test`, `dev`, and `prod`. `test` is appropriate for running tests, `dev` is appropriate for local development, and `prod` is appropriate for cloud development.

### Problem or idea

In https://github.com/AlexsLemonade/refinebio/pull/192#pullrequestreview-110516826 Rich points out that this is not sufficient. We need to have the following environments in total:

```
test
local
dev
staging
prod
```

### Solution or next step

Create the `local` and `staging` environments, make them work correctly, and test that everything works. I think that at the current `dev` environment is actually what we want `local` to be, and that the current `prod` environment is what we actually want `dev` to be. Therefore the way I think this issue should be approached is to change `dev` to `local`, change `prod` to `dev`, and then create `staging` and `prod` from scratch.

jaclyn-taroni:
Was this closed by #315? If not, let's update this to note what still needs to be done.

kurtwheeler:
This was closed... I thought I had closed this before. Oh whelp.

PR #315, opened 2018-06-14
Add more envs, improve deployment, stand up staging stack

kurtwheeler:
## Issue Number

#199 
https://github.com/AlexsLemonade/refinebio-frontend/issues/5

We needed test, local, dev, staging, and prod environments. Before we only had test, dev (which was really local) and prod. Now we have them all!

This also greatly improves our deploy process. I created a new Dockerhub organization `ccdlstaging` where our staging images go. We have a staging stack! It's pointed to by `staging.refine.bio` and `api.staging.refine.bio`! Since we now have legit stacks I've also generated real Django secret keys and database passwords which are in the `*.tfvars` files.

I've also ensured that we're installing the binaries that Hashicorp is distributing and not ones that have been meddled with by a man in the middle.

I also took out the terraform configuration for the S3 buckets that store terraform state files, because that was a bit much and those aren't really going to change ever so there's not too much point to them.

Unfortunately at least one limitation related to the fact that AWS doesn't support `*.bio` domains. Because of this, we can't have AWS/terraform automagically update the IP address for the API. I don't know how to make this work.

## Types of changes

## Functional tests

I've tested deploys, redeploys, docker image builds, and the actual website itself! It all works and seems to be rather robust!

## Checklist

## Screenshots

image


infrastructure/users.tf
```diff
@@ -88,3 +97,35 @@ resource "aws_iam_user_policy" "data_refinery_user_foreman_policy" {
...
 }
+
+data "aws_iam_policy_document" "data-refinery-deployment" {
+  statement {
+    actions = [
+      "s3:ListObjects",
+      "s3:GetObject",
+      "s3:PutObject",
+      "s3:DeleteObject",
+      "s3:PutObjectAcl",
```
kurtwheeler: This bucket is private but this permission makes it so the objects in it can be pushed as public, which makes the site work. Idk if we should just switch the bucket to being public though.

kurtwheeler:
One thing I should mention about this is that I couldn't figure out how to point react at the staging api. @ramenhog

ramenhog:
@kurtwheeler I just updated our app to use an env variable to set the API domain: https://github.com/AlexsLemonade/refinebio-frontend/pull/87

So now you should be able to run `REACT_APP_API_HOST=http://something.com yarn run build` to build the app with the env variable for staging. However, it will run into CORS issues, so we'll need to enable CORS.

dongbohu:
By the way, the failed test `affymetrix_common_agilent_tests` is probably due to some database migration was not applied.

PR #327, opened 2018-06-18
Hopefully better Linux IP address detection

Willv19:
The old command to detect IP addresses on Linux was `ip route get 8.8.8.8 | awk '{print $NF; exit}'`,
but that did not work on my machine or the default Ubuntu docker image I tested. It turns out that `ip route` does not
appear to have standardized output across different versions and distros. I found this alternate option that is hopefully
standardized and installed by any sane Linux distro to give us the internal ip address for the setup scripts.

## Issue Number

N/A

Make the install scripts work on more people's hardware.

## Methods

N/A

## Types of changes
- Bugfix

## Functional tests

Running the get_ip_address function on my laptop gave the correct internal ip address, and it made the install scripts start working properly.

## Checklist

## Screenshots

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,17 @@ Instructions for installing Docker and Nomad can be found by
 following the link for each service. git-crypt and jq can be installed via
 `sudo apt-get install git-crypt jq`.
 
+When installing pip packages later in the install, you might get an error saying you need sudo permissions.
+In order to fix this you have to edit your `~/.config/pip/pip.conf` to add this:
+
+```
+[install]
+user = yes
+no-binary = :all:
+```
+
+This sets pip to install all packages in your user directory so sudo is not required for pip intalls.
+
 #### Mac
 
 The following services will need to be installed:
@@ -170,6 +181,10 @@ Finally, to make the migrations to the database, use:
 ```bash
 ./common/make_migrations.sh
 ```
+
+Note: there is a small chance this might fail with a `can't stat`, error. If this happens, you have
+to manually change permissions on the volumes directory with `sudo chmod -R 740 volumes_postgres`
+then re-run the migrations.
 
 If you need to access a `psql` shell for inspecting the database, you can use:
 
--- a/README.md
+++ b/README.md
@@ -82,10 +82,11 @@ so Docker does not need sudo permissions.
 - [Nomad](https://www.nomadproject.io/docs/install/index.html#precompiled-binaries) can be installed on Linux clients with `sudo ./install_nomad.sh`.
 - git-crypt
 - jq
+- iproute2
 
 Instructions for installing Docker and Nomad can be found by
-following the link for each service. git-crypt and jq can be installed via
-`sudo apt-get install git-crypt jq`.
+following the link for each service. git-crypt, jq, and iproute2 can be installed via
+`sudo apt-get install git-crypt jq iproute2`.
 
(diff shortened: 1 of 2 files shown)


kurtwheeler:
Looks like this one won't work across distros. On my Ubuntu machine:
```
$ hostname -i
127.0.1.1
```

This one is unfortunate because it requires network capabilities, but I think it's pretty cross platform: 
```
dig +short myip.opendns.com @resolver1.opendns.com
```

dongbohu:
I am not sure whether we should add explicit support to Linux distributions other than Ubuntu. It is like Pandora's box. Once it is open, we may ask ourselves: Should we also support other popular distributions such as CentOS/Fedora/RHEL/SUSE/Gentoo...? 

By the way, the `ip` command is part of `iproute2` package. On Arch Linux, if we install this package by `pacman`. The `ip route ...` command will work too.

common.sh
```diff
@@ -1,7 +1,9 @@
 #!/bin/bash
 
 get_ip_address () {
-    if [ `uname` == "Linux" ]; then
+    if [ -e /etc/arch-release ]; then
+        echo $(hostname -i | awk '{print $1; exit}')
```
dongbohu: I think this line can be simplified into: `hostname -i`, right?

dongbohu:
I think a better solution is probably to use `ip route ...` command to find the IP address on all Linux platform, but also list `iproute2` package explicitly in "Installation > Linux" section of `README` file. (The `README` file in this  PR may give users the impression that refinebio can only be run on Ubuntu and Arch Linux.)

Willv19:
I reset the command to the `ip route ...`, but that alone didn't fix the issue. It turns out that, rather annoyingly, even though I had `iproute2` installed it had a different output than the version on Ubuntu 16.04. I was curious, so I also ran the command on a Docker image of Ubuntu 18.04 and the output matched my machine, not the 16.04 docker image. See below:

Ubuntu 16.04
```
$ ip route get 8.8.8.8
8.8.8.8 via 172.17.0.1 dev eth0  src 172.17.0.3 
    cache
```

Ubuntu 18.04
```
$ ip route get 8.8.8.8
8.8.8.8 via 172.17.0.1 dev eth0 src 172.17.0.3 uid 0 
    cache
```

So, as it turns out, because of this difference in output the `awk` command we were using did not work on all versions of `iproute2`, just the one that comes installed on Ubuntu 16.04. I rewrote the command to `grep -oE` for this regex first: `src ([0-9]{1,3}\.){3}[0-9]{1,3}`, since it is common to all the outputs I saw, then `sed` for the ip address. This should hopefully mean that our script supports a much wider variety of `iproute2` versions.

README.md
```diff
@@ -82,10 +82,22 @@ so Docker does not need sudo permissions.
...
+When installing pip packages later in the install, you might get an error saying you need sudo permissions.
+In order to fix this you have to edit your `~/.config/pip/pip.conf` to add this:
+
+```
+[install]
+user = yes
+no-binary = :all:
+```
+
+This sets pip to install all packages in your user directory so sudo is not required for pip intalls.
```
kurtwheeler: Was sudo necessary for installing some pip packages even within the virtualenvironment?

Willv19: The issues is when installing `virtualenv` itself. If the pip directory is in a place where only root has write access, `pip install virtualenv` fails, which can happen on some distros since the package manager wants complete control over installed software.

PR #331, opened 2018-06-18
add accession code filtering and tests

ramenhog:
* Filter samples by `accession_code`

## Types of changes

## Functional tests
* passing in a list of accession codes as a parameter to the `samples` endpoint will return a filtered list of samples by those accession codes

## Checklist

### DIFF
--- a/api/data_refinery_api/views.py
+++ b/api/data_refinery_api/views.py
@@ -232,7 +232,7 @@ class SampleList(PaginatedAPIView):
 
     Pass in a list of pk to an ids query parameter to filter by id.
 
-    Append the pk to the end of this URL to see a detail view.
+    Append the pk or accession_code to the end of this URL to see a detail view.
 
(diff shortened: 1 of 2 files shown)


api/data_refinery_api/views.py
```diff
@@ -242,11 +242,16 @@ def get(self, request, format=None):
...
         order_by = filter_dict.pop('order_by', None)
         ids = filter_dict.pop('ids', None)
+        accession_codes = filter_dict.pop('accession_codes', None)
 
         if ids is not None:
             ids = [ int(x) for x in ids.split(',')]
             filter_dict['pk__in'] = ids
 
+        if accession_codes is not None:
+            accession_codes = [ x for x in accession_codes.split(',')]
```
Miserlou: I think this can be replaced with just `accession_codes.split(',')`

Issue #336, opened 2018-06-18
Remove unnecessary packages from Affymetrix Docker image


jaclyn-taroni:
### Context

The Affymetrix (/Agilent) Docker image is by far the largest image we have and I've noticed some packages I think can be removed.

### Problem or idea

Here's what I've noticed that may be unnecessary:

* [Illumina packages](https://github.com/AlexsLemonade/refinebio/blob/dev/workers/affymetrix_dependencies.R#L53) - necessary for Illumina SCAN and gene ID conversion, but not Affymetrix SCAN
* [`org.*` annotation packages](https://github.com/AlexsLemonade/refinebio/blob/dev/workers/affymetrix_dependencies.R#L60) - I believe these might only be necessary for gene identifier conversion and/or might be a vestige of when we were using Entrez gene ID Brainarray packages.
* [`tximport`](https://github.com/AlexsLemonade/refinebio/blob/dev/workers/affymetrix_dependencies.R#L47) - probably leftover from before we split up the processor images

### Solution or next step

We should try to remove these. I will leave the details of the exact approach to the backend devs. 

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

PR #338, opened 2018-06-19
Remove unnecessary R packages from affymetrix docker image

dongbohu:
## Issue Number
#336 

Remove unnecessary packages from Affymetrix Docker image

## Functional tests
Tested on my local desktop.

### DIFF
--- a/workers/affymetrix_dependencies.R
+++ b/workers/affymetrix_dependencies.R
@@ -44,32 +44,9 @@ bioc_pkgs <- c(
   'preprocessCore_1.40.0.tar.gz',
   'genefilter_1.60.0.tar.gz',
   'sva_3.26.0.tar.gz',
-  'tximport_1.6.0.tar.gz',
   'limma_3.34.9.tar.gz'
 )
 install_with_url(bioc_url, bioc_pkgs)
-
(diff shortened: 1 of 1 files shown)


dongbohu:
@Miserlou: I checked with @jaclyn-taroni and those `pd.*` packages are required.

jaclyn-taroni:
One thought that came up in my discussion with @dongbohu: make _organism-specific_ Affymetrix images. So all the platform design and Brainarray packages from one organism would live on the same image and the organism for an assay would dictate what Docker image is used for that processor. The rationale for having an organism-centric image rather than one for every single platform is because, as has been noted elsewhere, sometimes the CEL header does not match the platform listed in the data source but we can probably assume that the organism label is correct the majority of the time.

dongbohu:
Hi @jaclyn-taroni: I was about to close this PR, then I noticed lines 19-22:
https://github.com/AlexsLemonade/refinebio/blob/dev/workers/affymetrix_dependencies.R#L19
Should we delete them too?

jaclyn-taroni:
Sure. I have not evaluated whether or not every single package will be required. I just listed the ones that jumped out at me in #338.

Issue #157, opened 2018-03-06
Optimize Salmon index length calculation


kurtwheeler:
### New Issue Checklist
- [x] The title is short and descriptive
- [x] The issue contains an:
  - [ ] Idea (new feature, user story, etc)
  - [X] Problem
- [x] You have explained the:
  - [x] Context
  - [x] Problem or idea
  - [x] Solution or next step
  
### Context

Depending on the average transcript read length for FASTA files, we should use either a salmon index that is optimized for long or short reads. Therefore when running salmon we first calculate the average read length and then run salmon. However we are doing this with Python and it is single-core gunziping the FASTA files line by line. 

### Problem or idea

@Miserlou  thinks there is a faster way of doing this, maybe by shelling and using UNIX! It 'd be faster to gunzip with multicores and then count lines that way, ex:
```bash
$ pigz file.fasq.gz;
$ cat file.fastq | echo $((`wc -l`/4))
```

### Solution or next step

Try shelling out and using multiple cores to speed up the average read length calculation.

jaclyn-taroni:
Salmon assumes that the reads you give it are in a random order -- see the first **Note** [here](https://salmon.readthedocs.io/en/latest/salmon.html#using-salmon) in the Salmon docs. Do you think it would behoove us to shuffle the `fastq.gz` file around the same time we calculate the read lengths just in case they are ordered in some way? (This is just something that has been in the back of my mind that might be worth discussing -- can file another ticket, let me know.)

Miserlou:
This ticket is related to a comment around a single line of code that has to do with how unzipping in Python is slow. If your idea is something we should investigate @jaclyn-taroni , I think it'd be best to open a new ticket.

jaclyn-taroni:
I believe this was closed via #339

PR #339, opened 2018-06-20
Optimize Salmon Index Calculations

Willv19:
## Issue Number

#157 

Replace calls to `gzip.open()` with shells out to the native `gzip` command for better performance in unzipping testq files. I also added two tests for this function that make sure that the output of this new approach matches that of the old salmon index calculation.

The original issue mentions using `pigz` instead of `gzip`, but I chose `gzip` for a couple of reasons. First, tests on my laptop and my desktop at home both show that for unzipping files the two have pretty identical performance. `pigz` does note this in their documentation, saying that `Decompression can’t be parallelized, at least not without specially prepared deflate streams for that purpose. Asaresult, pigz uses a single thread (the main thread) for decompression, but will create three other threads for reading, writing, and check calculation, which can speed up decompression under some circumstances.` Also, `gzip` is already included in the docker debian images so no modifications to those images are required with my implementation.

## Methods

N/A

## Types of changes

## Functional tests

Tested on my machine.

## Checklist


Miserlou:
Wow, looking great! Do you have any metrics for how much of an improvement this creates?

Willv19:
I made a 6gb gzipped fastq file and ran both this and the old algorithm on it, and I saw ~50% time reduction.

workers/data_refinery_workers/processors/salmon.py
```diff
@@ -1,8 +1,8 @@
 from __future__ import absolute_import, unicode_literals
 
+import gzip
```
Miserlou: `gzip` isn't used any more, so this can be removed.

workers/data_refinery_workers/processors/salmon.py
```diff
@@ -99,29 +98,10 @@ def _determine_index_length(job_context: Dict) -> Dict:
...
-    ###         total_base_pairs += len(line.replace("\n", ""))
-    ###         number_of_reads += 1
-    ###     counter += 1
-
-    with gzip.open(job_context["input_file_path"], "rt") as input_file:
-        for line in input_file:
+    # zcat unzips the file provided and dumps the output to STDOUT.
+    # It is installed by default in Debian so it should be included
+    # in every docker image already.
+    for line in os.popen("zcat %s" % job_context["input_file_path"]):
```
dongbohu: Did you try `subprocess` module (which is supposed to replace `os.popen` methods)?
https://docs.python.org/3.5/library/os.html#os.popen

Did you test whether the better performance is at the cost of loading the whole file into the memory?

kurtwheeler:
@Willv19 These tests are failing because the downloader tests started taking long enough to timeout. This PR has the fix for this, so once its approved and merged you'll just be able to pull dev and this should work: https://github.com/AlexsLemonade/refinebio/pull/349

PR #340, opened 2018-06-20
Add Support for Broader Set of Illumina Beadchip Platforms

Miserlou:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/232

WIP PR!

Adds support for more than just Human2 and Human4 platforms. Uses auto-detection to correct for mislabeled platforms (ex., GSE54661, with test).  Platforms now supported are:

```python
    all_databases = {
        'HOMO_SAPIENS': [
            'illuminaHumanv1',
            'illuminaHumanv2',
            'illuminaHumanv3',
            'illuminaHumanv4',
        ],
        'MUS_MUSCULUS': [
            'illuminaMousev1',
            'illuminaMousev1p1',
            'illuminaMousev2',
        ],
        'RATTUS_NORVEGICUS': [
            'illuminaRatv1'
        ]
    }
```

Also now splits Illumina results into smashable individual files. Necessitates the creation of a `SampleComputedFileAssociation` field and hooks that up throughout the application.

## Methods

Methods are described in #232 and are implemented in the new `detect_database.R`. The one thing worth checking here is we're determining platform via the best of

```
common_probes <- intersect(unlist(expr_probes), database_probes)
percent <- ( length(common_probes) / length(database_probes) ) * 100.0
```
which might not be right, although it is in tests so far.

_TODO (?)_:

* I've forgetten the difference between `calculate the % of IDs in the probes for current package`
and `calculate the % of probes in the IDs for current package` now. This implementation uses the percentage of the database probes which are in both for the thresholding. I don't know what the other value would be used for.

* NO_OP'ing the data we already have probably won't work and may require re-surveying, or perhaps just directly downloading and processing from the same location. I need an example GSE where all values are below our processing threshold so we know what to do here. 

## Types of changes

## Functional tests

New tests added with `illumina` tag.

## Checklist


jaclyn-taroni:
What you have in the methods section here is a bit different from what is noted on #232.

[Where I describe](https://github.com/AlexsLemonade/refinebio/issues/232#issuecomment-398188047): 

`calculate the % of IDs in the probes for current package`
`calculate the % of probes in the IDs for current package`

I mean

 `IDs = identifiers used by submitter`  which I believe is the equivalent of `expr_probes`
and my `probes` is equivalent to `database_probes`

So if I understand correctly, you are calculating `the % of probes in the IDs for current package`. That metric is likely to be most useful for distinguishing whole genome (so "WG" or "HT") from RefSeq ("Ref") in my opinion. This is because the Ref chips should have around half the whole genome probes on them and the `probes`/`database_probes` should contain all the whole genome probes. Essentially, I would expect a Ref chip to have a value ~0.5 for  `% of probes in the IDs for current package` based on 1) what I've read about the chips and 2) the `GSE25580` `GPL6104` results [here](https://github.com/jaclyn-taroni/beadarray-platform-detection/tree/61bc2700e34b07f78672e32a9aa3544f150a4718#results).

If you have used _whole genome_ chips in your tests, I would expect this metric to give you a very high value (probably > 0.95).

The rationale for calculating `the % of IDs in the probes for current package` is that, in the case of both whole genome and ref chips, we would expect this to be quite high for the correct version of the platform (e.g., `illuminaHumanv2.db` vs. `illuminaHumanv4.db`) and 

> probe sequence information will be required to perform the normalization, so if many of the [`IDs`/ `exprs_probes`] cannot be mapped to [`probes` / `database_probes`] and, therefore, their sequences, we will not be able to normalize them effectively.

which is why I suggested using this metric to determine whether or not to proceed with processing.

A limitation of the analyses on #232 is that I was solely focused on human whole genome chips, used the major/official GEO platforms, and any Ref chips I found were by accident. So I think it makes sense to implement both calculations and test some random experiments from additional platforms that I note [here]( https://github.com/AlexsLemonade/refinebio/issues/232#issuecomment-390421762).

Miserlou:
I think this is addressed now,

```
percent <- ( length(common_probes) / length(database_probes) ) * 100.0
mapped_percent <- ( length(common_probes) / length(unlist(expr_probes)) ) * 100.0
```
`percent` is used for platform detection and `mapped_percentage` is used for inclusion/non-inclusion. 

Values for GSE54661 are `{'detected_platform': 'illuminaHumanv3', 'detection_percentage': '98.43674', 'mapped_percentage': '100.0'}`

Miserlou:
It appears to be unable to use `illuminaRatv1.db`, though I'm not sure why. Otherwise I think this is ready for review.

jaclyn-taroni:
> It appears to be unable to use illuminaRatv1.db, though I'm not sure why. Otherwise I think this is ready for review.

You are referring to `detect_database.R`, is that correct? Or are you referring to the modifications to `illumina.R`?

The good news is since there is only one rat platform, we don't have to _detect_ the platform. But we probably still want some kind of filter on `mapped_percentage` for rat. That gets checked in `illumina.R` here: https://github.com/AlexsLemonade/refinebio/pull/340/files#diff-ad337bc3f96b5b4756d39909ac5de5aaR406 so we could modify that.

Miserlou:
> You are referring to detect_database.R, is that correct? Or are you referring to the modifications to illumina.R?

Yeah, I just get:
```
Error in library(paste(platform, ".db", sep = ""), character.only = TRUE) :
  there is no package called 'illuminaRatv1.db'
```

jaclyn-taroni:
Does it work for mouse but not rat? Just wondering if this: 
https://github.com/AlexsLemonade/refinebio/pull/340/files#diff-c15bd38730f46e138ebc2e5daf4a0801R37

needs to be `illuminaRatv1.db_1.26.0.tar.gz`

Miserlou:
Yep. Thanks, I really should have seen that..

jaclyn-taroni:
If you set `options(warn=2)` at the beginning of `illumina_dependencies.R`, that might help. Apparently the bioconductor installer will throw a warning instead of an error if a package fails to install.

https://github.com/AlexsLemonade/refinebio/blob/5e721f097338c97c4297783d4482217bb3c0b032/workers/affymetrix_dependencies.R#L1

Miserlou:
Cool! Is there anyway that I can suppress all the library "welcome" messages that happen as well?

jaclyn-taroni:
Perhaps `suppressPackageStartupMessages` https://www.rdocumentation.org/packages/base/versions/3.5.0/topics/message

Have never used it myself

workers/data_refinery_workers/processors/illumina.py
```diff
@@ -156,6 +173,99 @@ def _detect_columns(job_context: Dict) -> Dict:
...
+    if high_mapped_percent > 75.0:
+        job_context['platform'] = high_db
+
+        for sample in job_context['samples']:
+            sa = SampleAnnotation()
+            sa.sample = sample
+            sa.data = {
+                "detected_platform": high_db, 
+                "detection_percentage": highest,
+                "mapped_percentage": high_mapped_percent
```
kurtwheeler: These seem like some useful statistics to have around even if we're just going to NO_OP the sample. I could see us wanting to look into the stats of all the samples we weren't confidant enough to process at some point.

workers/data_refinery_workers/processors/illumina.py
```diff
@@ -216,50 +320,55 @@ def _create_result_objects(job_context: Dict) -> Dict:
...
+    big_tsv = job_context["output_file_path"]
+    data = pd.read_csv(big_tsv, sep='\t', header=0, index_col=0)
+    individual_files = []
+    frames = np.split(data, len(data.columns), axis=1)
+    for frame in frames:
+        frame_path = job_context["output_file_path"] + '.' + frame.columns.values[0].replace('&', '') + '.tsv'
+        frame.to_csv(frame_path, sep='\t', encoding='utf-8')
+
+        # This needs to be the same as the ones in the job context!
+        sample = Sample.objects.get(accession_code=frame.columns.values[0])
```
kurtwheeler: When you build the path above you do a `frame.columns.values[0].replace('&', '')`. Will the accession codes in the database contain these ampersands?

workers/data_refinery_workers/processors/smasher.py
```diff
@@ -139,6 +139,7 @@ def _smash(job_context: Dict) -> Dict:
         shutil.make_archive(final_zip, 'zip', smash_path)
         job_context["output_file"] = final_zip + ".zip"
     except Exception as e:
+        logger.error(e)
```
kurtwheeler: Could you add a little more context to this log statement?

workers/data_refinery_workers/processors/test_geo.py
```diff
@@ -87,8 +89,64 @@ def test_illumina_to_pcl(self):
...
         from data_refinery_workers.processors import illumina
         job = prepare_illumina_job()
-        illumina.illumina_to_pcl(job.pk)
-        self.assertTrue(os.path.isfile('/home/user/data_store/raw/TEST/processed/GSE22427_non-normalized.PCL'))
+        final_context = illumina.illumina_to_pcl(job.pk)
+
+        for sample in final_context['samples']:
+            print(sample)
+            smashme = sample.get_most_recent_smashable_result_file()
+            print(smashme)
```
kurtwheeler: This prints look like they'll be pretty chatty.

workers/data_refinery_workers/processors/test_geo.py
```diff
@@ -87,8 +89,64 @@ def test_illumina_to_pcl(self):
...
+        og_file.save()
+
+        assoc1 = ProcessorJobOriginalFileAssociation()
+        assoc1.original_file = og_file
+        assoc1.processor_job = pj
+        assoc1.save()
+
+        sample_names = [
+            "hypoxia_Signal",
+        ]
```
kurtwheeler: Were there going to be more of these? I'm a bit confused why we're defining a list of length one to loop over.

workers/data_refinery_workers/processors/illumina.py
jaclyn-taroni: I agree.

Miserlou:
I think these are addressed now.

Issue #342, opened 2018-06-21
Create a static IP address for the prod/staging API and enable HTTPS


kurtwheeler:
### Context

In #315 I was able to get the API up and running and have our DNS point to it. However every time the instance is recreated during deploys its IP address changes so the DNS record becomes out of date and `http://api.staging.refine.bio` breaks.

Additionally when the DNS is working, it doesn't support HTTPS. The reason for both of these issues is that the top level domain `.bio` is not supported by AWS' Route53, so the methods we used previously for Cognoma's DNS, SSL, and ELBs aren't able to be reused here.

### Problem or idea

We should get a static IP address for the API servers so the DNS stops breaking and we should make them support HTTPS.

### Solution or next step

- [x] Put a load balancer in front of the API with an elastic IP address associated with it so that the IP address won't change.
- [ ] Create and install an SSL cert onto the API servers so they can support HTTPS.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

jaclyn-taroni:
Closed by #343

PR #343, opened 2018-06-21
Make API persistent and support HTTPS

kurtwheeler:
## Issue Number

#342 

Because AWS doesn't support the top level domain `.bio`, we cannot use Route53 to manage the domain and therefore cannot use the Amazon Certificate Manager like we did in cognoma. This enables HTTPS (and forces redirection from HTTP). Additionally this uses an AWS Network Load Balancer with an Elastic IP to make it so that when the API server goes down and comes back up with a new IP, our DNS settings are still valid for forwarding traffic to it.

## Types of changes

## Functional tests

I ran the new configuration with the Load Balancer from CCI and made sure that the elastic IP address was properly working especially when forwarded to via DNS. I also ran the `certbot` commands directly on the API instance to verify that they worked.

I am going to use the merge into dev to both verify that `dev` merges properly cause the staging infrastructure to be updated and that the change to the API server's startup script works properly.

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,7 +37,9 @@ jobs:
       # Running these in the same job as the common tests is good
       # because their dockerfiles are very similar so a lot of the
       # build time is saved by only building those layers once.
-      - run: .circleci/filter_tests.sh -t downloaders
+      - run:
+          command: .circleci/filter_tests.sh -t downloaders
+          no_output_timeout: 36000
 
       # Push the no_op and downloader images to the local docker repo so Nomad
       # can pull from there when running end-to-end tests.
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -161,7 +161,7 @@ jobs:
     steps:
       - checkout
       - run: bash .circleci/git_decrypt.sh
-      # - run: bash .circleci/update_docker_img.sh
+      - run: bash .circleci/update_docker_img.sh
       - run: bash .circleci/run_terraform.sh
 
 workflows:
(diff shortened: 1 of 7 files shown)


.circleci/run_terraform.sh
```diff
@@ -56,7 +56,7 @@ sudo mv nomad /usr/local/bin/
 if [ $CIRCLE_BRANCH == "master" ]; then
     ENVIRONMENT=prod
     BUCKET_NAME="refinebio-tfstate-deploy-production"
-elif [[ $CIRCLE_BRANCH == "dev" ]]; then
+elif [[ $CIRCLE_BRANCH == "dev" || $CIRCLE_BRANCH == "kurtwheeler-make-api-persistent" ]]; then
```
Miserlou: remove or comment?

.circleci/config.yml
```diff
@@ -161,7 +161,7 @@ jobs:
     steps:
       - checkout
       - run: bash .circleci/git_decrypt.sh
-      - run: bash .circleci/update_docker_img.sh
+      # - run: bash .circleci/update_docker_img.sh
```
Miserlou: Lots and lots is commented out..

infrastructure/nomad-configuration/client.tpl.hcl
```diff
@@ -22,5 +22,5 @@ consul {
 telemetry {
   publish_allocation_metrics = true
   publish_node_metrics       = true
-  statsd_address = "${nomad_lead_server_ip}:8125"
+  statsd_address = "${nomad_server_address}:8125"
```
Miserlou: There are two values in the TF with the same value, I'd suggest we collapse to just one, and that `nomad_lead_server_ip` is more accurate and specific.

kurtwheeler: I'm a bit confused about this comment because there is only one variable available to this template and it's defined here: https://github.com/AlexsLemonade/refinebio/blob/dev/infrastructure/instances.tf#L214.

However I definitely can rename this variable to keep things more consistent TF-wide.

infrastructure/networking.tf
```diff
@@ -80,3 +80,90 @@ resource "aws_db_subnet_group" "data_refinery" {
...
+  load_balancer_type = "network"
+
+  enable_deletion_protection = true
+
+  # Only one subnet is allowed and the API lives in 1a.
+  subnet_mapping {
+    subnet_id = "${aws_subnet.data_refinery_1a.id}"
+    allocation_id = "${aws_eip.data_refinery_api_ip.id}"
+  }
+
```
Miserlou: rm whitespace

infrastructure/api-configuration/api-server-instance-user-data.tpl.sh
```diff
@@ -21,6 +21,21 @@ apt-get install nginx -y
...
+if [[ ${stage} == "staging" || ${stage} == "prod" ]]; then
+    # Create and install SSL Certificate
+    # Only necessary on staging and prod
+    apt-get install software-properties-common
+    add-apt-repository ppa:certbot/certbot
+    apt-get update
+    apt-get install python-certbot-nginx
+
+    if [[ ${stage} == "staging" ]]; then
+        certbot --nginx -d api.staging.refine.bio -n --agree-tos --redirect -m g3w4k4t5n3s7p7v8@alexslemonade.slack.com
```
Miserlou: can you add a comment about what this email is/is being used for?

PR #345, opened 2018-06-22
Improve scripts and documentation

Willv19:
## Issue Number

N/A

I added more documentation on how to run individual tests rather than the full test suites. The tests now also make sure that Nomad and Docker are running before starting rather than failing silently afterwards. The help documentation in the test scripts was also rewritten in an explicit man pages-like format. I also added URLs to the dependencies in the README that were missing them before.

## Types of changes

## Functional tests

I reran all the scripts that I changed to make sure that they work exactly like the old ones when used correctly, and warn the user when used incorrectly.

## Checklist

### DIFF
--- a/api/run_tests.sh
+++ b/api/run_tests.sh
@@ -14,7 +14,7 @@ cd $script_directory
 cd ..
 
 # Ensure that postgres is running
-if ! docker ps | tail -n +2 | awk '{ print $NF }' | grep drdb > /dev/null; then
+if [[ $(docker ps --filter NAME=drdb -q) ]]; then
     echo "You must start posgres first with './run_postgres.sh'" >&2
     exit 1
 fi
--- a/api/run_tests.sh
+++ b/api/run_tests.sh
@@ -14,7 +14,7 @@ cd $script_directory
 cd ..
 
 # Ensure that postgres is running
-if [[ $(docker ps --filter NAME=drdb -q) ]]; then
+if ! [[ $(docker ps --filter name=drdb -q) ]]; then
     echo "You must start posgres first with './run_postgres.sh'" >&2
     exit 1
 fi
--- a/api/run_tests.sh
+++ b/api/run_tests.sh
@@ -15,7 +15,7 @@ cd ..
 
 # Ensure that postgres is running
 if ! [[ $(docker ps --filter name=drdb -q) ]]; then
-    echo "You must start posgres first with './run_postgres.sh'" >&2
+    echo "You must start Postgres first with './run_postgres.sh'" >&2
     exit 1
 fi
 
--- a/api/run_tests.sh
+++ b/api/run_tests.sh
@@ -13,6 +13,12 @@ cd $script_directory
 # move up a level
 cd ..
 
+# Ensure that postgres is running
+if ! docker ps | tail -n +2 | awk '{ print $NF }' | grep drdb > /dev/null; then
+    echo "You must start posgres first with './run_postgres.sh'" >&2
+    exit 1
+fi
+
 ./prepare_image.sh -i api_local -s api
 
 source common.sh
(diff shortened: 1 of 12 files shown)


workers/run_tests.sh
```diff
@@ -34,6 +49,17 @@ cd $script_directory
 # move up a level
 cd ..
 
+# Ensure that Nomad is running first
```
kurtwheeler: This is why all the tests are failing. Only the `foreman` tests require Nomad to be running because that's where we have our end-to-end tests. All of the `workers` tests mock out `send_job` so that we never actually hit Nomad. If you've seen anywhere where this isn't the case, I think the correct fix is to mock that function out.

The reason we do this is that in order for a local test Nomad agent to work we need a local Docker registry running which contains the images we want to use. Some of these images are large and pushing them to even a local registry is slow (thanks Docker) so we only do this in the circle tests once.

api/run_tests.sh
```diff
@@ -13,6 +13,12 @@ cd $script_directory
 # move up a level
 cd ..
 
+# Ensure that postgres is running
+if ! docker ps | tail -n +2 | awk '{ print $NF }' | grep drdb > /dev/null; then
```
dongbohu: I suppose you are using this command to test whether `drdb` container is running or not. This is probably better:
`docker ps --filter NAME=drdb -q`
If the command above gives nothing, `drdb` is not running.

api/run_tests.sh
```diff
@@ -13,6 +13,12 @@ cd $script_directory
...
+    echo "You must start posgres first with './run_postgres.sh'" >&2
```
dongbohu: Typo: `postgres`. To be more professional, you may want to use `Postgres`.

run_all_tests.sh
```diff
@@ -5,6 +5,18 @@
...
 # sudo -E ./run_nomad.sh -e test
 # This should mirror what happens in the CircleCI config (.circleci/config).
+
+# Ensure that Nomad is running first
+if ! ps aux | grep test_nomad | grep -v grep > /dev/null; then
+    echo "You must start the nomad test environment first with" >&2
+    echo "'sudo -E ./run_nomad.sh -e test'" >&2
+    exit 1
+# Then ensure postgres is running
+elif ! docker ps | tail -n +2 | awk '{ print $NF }' | grep drdb > /dev/null; then
```
dongbohu: See my earlier comment.

run_postgres.sh
```diff
@@ -13,8 +13,15 @@ if [ ! -d $VOLUMES ]; then
   mkdir $VOLUMES
 fi
 
-
-# via https://hub.docker.com/_/postgres/
-# 9.6.6 is the current (as of Jan 23 2018) RDS most recent version.
-# Password can be exposed to git/CI because this is only for dev/testing purposes, not real data.
-docker run -p 5432:5432 --name drdb -v "$VOLUMES":/var/lib/postgresql/data -e POSTGRES_PASSWORD=mysecretpassword -d postgres:9.6.6
+# Check if a docker database named "drdb" exists, and if so just run it
+if [[ $(docker ps -a --filter name=drdb | wc -l) > 1 ]]; then
```
dongbohu: You can simplify this line into:
```if [[ $(docker ps -a -q --filter name=drdb  ]]; then```

File: update_models.sh [hunk 6]
```diff
@@ -1,3 +1,9 @@
 #! /bin/bash -e
 # Makes migrations and re-installs so Docker images update locally
+
+if ! docker ps | tail -n +2 | awk '{ print $NF }' | grep drdb > /dev/null; then
+    echo "You must start posgres first with './run_postgres.sh'"
```
dongbohu: See my earlier comments.

dongbohu:
@Willv19: According to docker documentation: https://docs.docker.com/engine/reference/commandline/ps/#parent-command
It seems more reasonable to use the filter this way:
```docker ps --filter name=drdb -a```
(It was my fault using `NAME=drdb`.) Thanks.

dongbohu:
@Willv19: All occurrences of `posgres` in the echo lines should be changed to `Postgres` (note the extra letter `t`).

Issue #324, opened 2018-06-18
Create and Associatate 'Processor' and 'Pipeline' objects with 'ComputationalResult' objects


Miserlou:
### Context

In the interest of provenance and reproducibility, we need to associate `ComputationalResults` with an explicit `Processor` model which tracks the purpose, program, image and environment of a processor, and a `Pipeline` object which defines an explicit association of the various `ComputationalResult` objects which are produced during the transformation of a sample into a smashable ComputedFile.

### Problem or idea

Processor objects should contain a docker image string in the form `accountname/imagename:systemversion`, and an `environment` JSONField which contains information about key packages, extracted at runtime, as well as a SHA1 hash of that field's contents which can be used for fast lookup. In addition, it should have a short name and a human-readable name that can be used for compiling a Pipeline's human readable name.

The Pipeline object should have an ArrayField containing the PKs of ComputationalResult objects. (ArrayFields in Postgres preserve order, so no step_number field is necessary.) Pipelines should have a human_readable_name, which can either be explicit or a dynamic property compiled from the Processor objects it is related to through ComputationalResults.

### Solution or next step

A hypothetical schema from the whiteboard:

![20180618_121031](https://user-images.githubusercontent.com/139987/41548932-4e56a4a8-72f2-11e8-88a5-7ec09afb88e1.jpg)

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

Miserlou:
Related/superset of:

https://github.com/AlexsLemonade/refinebio/issues/320
https://github.com/AlexsLemonade/refinebio/issues/321
https://github.com/AlexsLemonade/refinebio/issues/322
https://github.com/AlexsLemonade/refinebio/issues/323

dongbohu:
Fixed in #346.

Issue #322, opened 2018-06-18
Create `processor` table


cgreene:
### Context

We have been having the reproducibility and provenance meeting.

### Problem or idea

We would like to have a way to store the version information for a processor. We would like this to be populated automatically.

### Solution or next step

We should implement a processor table that is automatically populated. When a processor's docker image is started it will check this table for an existing row that matches it. If this is not found, it will extract the version information for "key dependencies" and store this in a JSON field. We should store the `primary program` (either as a field or in a table - if in a table we need to update it programmatically as the processor field is updated). As this conversation is ongoing we are moving a little bit away from a separate table, so let's go with the field route unless, on reflection, we come up with a strong reason to do something different.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

dongbohu:
Right now the `Processor` model is defined like this:
```
class Processor(models.Model):
    """Processor associated with a certain ComputationalResult."""

    name = models.CharField(max_length=255)
    docker_image = models.CharField(max_length=255)
    environment = JSONField(default={})

    class Meta:
        db_table = "processors"
        unique_together = ('name', 'docker_image')

    def __str__(self):
        return "Processor: %s (docker_image: %s" % (name, docker_image)
```

`Salmon` pipeline includes the following 4 processors:
- Salmon Quant
- Salmontools
- Tximport
- MultiQC

What should be included in each processor's `environment` field?

cgreene:
This YAML file would be expected at a defined location with a defined set of categories (`ubuntu`, `R`, `python` initially).

dongbohu:
@cgreene: Originally I was thinking of using JSON format, but as far as I nkow, JSON doesn't allow comment lines, which is very annoying. YAML looks good to me. By the way, since we will populate the YAML fields to the `environment` of each processor, does it make more sense to define the YAML file as an array of different processors? The example you listed above is a little confusing to me.

cgreene:
I'm thinking that the YAML file for each processor would define what, exactly, we want versioned at the database level (as opposed to just "go check the docker image" which is what we offer for all other packages). Then we'd have to write some code that goes at the end of the spin up process (if I understand the conversation with @kurtwheeler and @Miserlou correctly) that extracts the version information via apt, pip? and R. That information would then get stored in the JSON field in the database and be made available to the front end for styling.

[...]

Issue #320, opened 2018-06-18
Change ComputationalResult command field to array


jaclyn-taroni:
### Context

During a meeting about provenance, we discussed updating the ComputationalResult table command field to be an array rather than a string.

### Problem or idea

`command` field should include all commands involved to generate this result.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.

dongbohu:
Done by #346.

PR #346, opened 2018-06-22
Updated models for reproducibility

dongbohu:
## Issue Number

#320 
#322
#324 

Updated models for the two issues above

## Types of changes

## Functional tests

## Checklist

## Screenshots


dongbohu:
@kurtwheeler and @Miserlou: This PR includes the updated models that we discussed on Monday. Since the field in `ComputationResult` is renamed from `command_executed` to `commands` and the removal of `program_version` and `system_version`, the unit tests will fail. Before I fix the lines that are affected by the new models, I want to make sure that this is what we were talking about. Thanks.

common/data_refinery_common/models/models.py
```diff
@@ -366,6 +364,29 @@ def save(self, *args, **kwargs):
         self.last_modified = current_time
         return super(ComputationalResultAnnotation, self).save(*args, **kwargs)
 
+
+class Pipeline(models.Model):
+    "Pipeline that is associated with a series of ComputationalResult records."""
+
+    name = models.CharField(max_length=255)
+    steps = ArrayField(models.IntegerField(), default=[])
```
dongbohu: Ideally we would want to have something like this instead:
```
steps = ArrayField(
        models.ForeignKey(ComputationalResult, on_delete=models.CASCADE),
        default=[]
) 
```
But it is not allowed in Django, see:
https://docs.djangoproject.com/en/2.0/ref/contrib/postgres/fields/#django.contrib.postgres.fields.ArrayField.base_field

So I am using the regular integer to refer to the primary key of related ComputationalResult record.

common/data_refinery_common/models/models.py
```diff
@@ -306,13 +306,11 @@ def __str__(self):
...
     public_objects = PublicObjectsManager()
+    commands = ArrayField(models.TextField())
 
-    command_executed = models.TextField(blank=True)
-    program_version = models.TextField(blank=True)
-    system_version = models.CharField(
-        max_length=255)  # Generally defined in from data_refinery_workers._version import __version__
     is_ccdl = models.BooleanField(default=True)
 
+    # TODO: This field should be changed into "processor" (foreign key to "Processor").
```
dongbohu: This is a reminder to ourselves for further changes of the model to fix #324.

kurtwheeler:
This looks correct to me.

common/data_refinery_common/models/models.py
```diff
@@ -306,13 +306,11 @@ def __str__(self):
     # Managers
     objects = models.Manager()
     public_objects = PublicObjectsManager()
+    commands = ArrayField(models.TextField())
```
Miserlou: This should be spatially separated from the Manger objects here, which perform a different function

common/data_refinery_common/models/models.py
```diff
@@ -366,6 +364,29 @@ def save(self, *args, **kwargs):
...
+
+    class Meta:
+        db_table = "pipelines"
+
+
+class Processor(models.Model):
+    """Processor associated with a certain ComputationalResult."""
+
+    name = models.CharField(max_length=255)
+    docker_img = models.CharField(max_length=255)
```
Miserlou: We should use pythonic rather than C-like names here, so `docker_image` and `environment`, please.

It'd also be great if these could have a `__str__` method so that they display nicely in the Django admin interface.
Looks good otherwise!

dongbohu: Thank you @Miserlou!

dongbohu:
I think this PR is ready for review. I have removed "Work in progress" for its title. It also addressed isse #324 partially. With the new `processor` field added in `ComputationalResult` as a foreign key, `pipeline` field in `ComputationalResult` becomes redundant. I tried to remove it but then realized that it will break some other code that @Miserlou is working on. (As you can see, this PR already conflicts with a few files in `dev` branch.) So I am leaving both fields in for now. When @Miserlou is done with his work, one of us needs to file another PR to resolve the redundancy.

I didn't include the YAML config files (and `Processor` table population based on them) here, because:
(1) This PR is already big enough;
(2) I need more time to finalize the YML files. 

I am taking next week off, but I will be accessible on the Internet. So please feel free to add your comments.

common/data_refinery_common/models/models.py
```diff
@@ -294,6 +294,30 @@ def save(self, *args, **kwargs):
         self.last_modified = current_time
         return super(ExperimentAnnotation, self).save(*args, **kwargs)
 
+class Pipeline(models.Model):
+    "Pipeline that is associated with a series of ComputationalResult records."""
+
+    name = models.CharField(max_length=255)
```
Miserlou: After review the code, I'm thinking that this should maybe be a ChoiceField or at least use an Enum or something somwhere so we don't have the problem of having to know `No Op` versus `noop` versus `NO_OP` vs etc. etc. etc. 

This goes for everything

dongbohu: I can definitely see the advantages of using Enum or ChoiceField. There is one disadvantage though: Every time when we add a new pipeline, we will have to modify the model and migrate the database.

workers/data_refinery_workers/processors/array_express.py
```diff
@@ -144,19 +153,17 @@ def _create_result_objects(job_context: Dict) -> Dict:
     """ Create the ComputationalResult objects after a Scan run is complete """
 
     result = ComputationalResult()
-    result.command_executed = "'SCAN.UPC', 'SCANfast'" # Need a better way to represent this R code.
+    result.commands = ['SCAN.UPC', 'SCANfast'] # Need a better way to represent this R code.
```
kurtwheeler: This should be `"SCAN.UPC::SCANfast"` because this is the same as `"SCAN.UPC::SCAN_TwoColor"`, just a different function from the SCAN.UPC package.

workers/data_refinery_workers/processors/no_op.py
```diff
@@ -34,11 +42,13 @@ def _no_op_processor_fn(job_context: Dict) -> Dict:
 
     # This is a NO-OP, but we make a ComputationalResult regardless.
     result = ComputationalResult()
-    result.command_executed = "" # No op!
+    result.commands.append("") # No op!
```
kurtwheeler: I think this should actually have a command name, especially since once Rich's recent work goes out the NO_OP command is actually going to be more of a CONVERT_GENE_IDS command.

common/data_refinery_common/models/models.py
kurtwheeler: Well we don't necessarily need to have it be at the database level. I think we need it somewhere, it could be an Enum class or even just a namespace full of `SALMON_QUANT = "SALMON_QUANT"` so that you could do `PipelineEnums.SALMON_QUANT` instead of hardcoding the string. I think this might help looking up what the possible values are along with keeping them consistent. We have at least the following values:
`"Salmon Quant", "Tximport", "MultiQC", "Salmontools", "Submitter-processed", "Agilent SCAN TwoColor"` which aren't very consistent case-wise.

I think I would prefer a naming convention like `ALL_CAPS_v1.2.3`. 

However I was also thinking that it might make sense to actually have the version be a separate field so that if we ever wanted to get all the pipelines for `AGILENT_SCAN_TWOCOLOR` we could just query with `WHERE name='AGILENT_SCAN_TWOCOLOR' rather than `WHERE name LIKE 'AGILENT_SCAN_TWOCOLOR%'`. Also since we're using the version of the system for the version number we could query for all results from a given version.

workers/data_refinery_workers/processors/no_op.py
dongbohu: Are you saying that this line should be changed to:
```
result.commands.append("CONVERT_GENE_IDS")
```

kurtwheeler: Maybe? At the moment all the gene conversion work is still using NO_OP as the name, but it's becoming a misnomer because it's actually doing a decent amount of work.

Let's keep it as NO_OP for now, and then we can think about using a CONVERT_GENE_IDS name in a broader sense at some point.

common/data_refinery_common/models/models.py
dongbohu: In the new database model, the terms such as "Salmon Quant", "Tximport", "MultiQC", "Salmontools", "Submitter-processed" and "Agilent SCAN TwoColor" are actually names of **processors**, not names of **pipelines**. 

The possible values of pipelines are hardcoded at the end of each pipeline module (`salmon.py`, `agilent_twocolor.py`, etc)  right before the call of `utils.run_pipeline()`. 

The naming convention of `ALL_CAPS_v1.2.3` is a good point. But I'm not sure whether these processor names will be shown on web UIs. If they will, then a name with white spaces seems to be more friendly to the front end because it can be displayed directly.  

I agree that having the processor and pipelines hardcoded everywhere is not a good idea. Let me put them at the beginning of each pipeline module.

Having a separate `version` field is a good idea too. I will add it in the model. Thanks.

kurtwheeler: Ah yeah that's a good point about user-friendly names.

workers/data_refinery_workers/processors/array_express.py
```diff
@@ -153,15 +153,15 @@ def _create_result_objects(job_context: Dict) -> Dict:
     """ Create the ComputationalResult objects after a Scan run is complete """
 
     result = ComputationalResult()
-    result.commands = ['SCAN.UPC', 'SCANfast'] # Need a better way to represent this R code.
+    result.commands.append('SCAN.UPC::SCANfast') # Need a better way to represent this R code.
```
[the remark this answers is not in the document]
dongbohu: Agree. I have removed this comment. If there's nothing else, I will merge this PR. Thanks.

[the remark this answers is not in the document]
kurtwheeler: Do it!

Issue #332, opened 2018-06-18
Deliver sample metadata as a TSV


jaclyn-taroni:
### Context

As indicated in the discussion around sample metadata on https://github.com/AlexsLemonade/refinebio/issues/221 and by the TODO on https://github.com/AlexsLemonade/refinebio/pull/303, we'll want to deliver sample metadata in tab-delimited text format.

### Problem or idea

The smasher needs to be modified to deliver sample metadata in TSV form as well as JSON, which is currently implemented.

### Solution or next step

This should be addressed after #307 gets merged. More information can be found on the original ticket: #221 

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

jaclyn-taroni:
Closed by #347

PR #347, opened 2018-06-22
Add TSV Metadata to Smasher Output

Miserlou:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/332

For some peculiar reason, some people prefer TSVs over JSON files for metadata. This adds those files to smasher-produced archives.

## Methods

Fields are provided by the Sample object's `to_metadata_dict` class, so any missing fields should be added there in the future.

## Types of changes

## Functional tests

New test assertions have been added.

## Checklist

### DIFF
--- a/workers/data_refinery_workers/processors/smasher.py
+++ b/workers/data_refinery_workers/processors/smasher.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import, unicode_literals
 
 import boto3
+import csv
 import json
 import os
 import shutil
@@ -114,7 +115,7 @@ def _smash(job_context: Dict) -> Dict:
 
         # Create metadata file.
         metadata = {}
-        metadata['files'] = os.listdir(smash_path)
+
         metadata['num_samples'] = num_samples
         metadata['num_experiments'] = job_context["experiments"].count()
         metadata['aggregate_by'] = job_context["dataset"].aggregate_by
@@ -130,6 +131,18 @@ def _smash(job_context: Dict) -> Dict:
             experiments[experiment.accession_code] = experiment.to_metadata_dict()
         metadata['experiments'] = experiments
 
+        # Metadata to TSV
+        with open(smash_path + 'metadata.tsv', 'w') as output_file:
+            sample_ids = [k for k in metadata['samples'].keys()]
+            keys = [k for k in metadata['samples'][sample_ids[0]].keys()] + ['sample_id']
+            dw = csv.DictWriter(output_file, keys, delimiter='\t')
+            dw.writeheader()
+            for key, value in metadata['samples'].items():
+                value['sample_id'] = key
+                dw.writerow(value)
+
+        metadata['files'] = os.listdir(smash_path)
+        # Metadata to JSON
         metadata['created_at'] = datetime.utcnow().strftime('%Y-%m-%dT%H:%M:%S')
         with open(smash_path + 'metadata.json', 'w') as metadata_file: 
             json.dump(metadata, metadata_file, indent=4, sort_keys=True)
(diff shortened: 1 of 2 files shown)


workers/data_refinery_workers/processors/smasher.py
```diff
@@ -130,6 +131,18 @@ def _smash(job_context: Dict) -> Dict:
             experiments[experiment.accession_code] = experiment.to_metadata_dict()
         metadata['experiments'] = experiments
 
+        # Metadata to TSV
+        with open(smash_path + 'metadata.tsv', 'w') as output_file:
+            sample_ids = [k for k in metadata['samples'].keys()]
+            keys = [k for k in metadata['samples'][sample_ids[0]].keys()] + ['sample_id']
```
kurtwheeler: This line is super confusing. I'm not entirely sure what `keys` is a list of after this line except some amount of keys. Any chance you could make this a bit more readable?

workers/data_refinery_workers/processors/smasher.py
```diff
@@ -130,6 +131,18 @@ def _smash(job_context: Dict) -> Dict:
...
+            dw = csv.DictWriter(output_file, keys, delimiter='\t')
+            dw.writeheader()
+            for key, value in metadata['samples'].items():
+                value['sample_id'] = key
+                dw.writerow(value)
+
+        metadata['files'] = os.listdir(smash_path)
```
kurtwheeler: After #351, this looks like it could be potentially problematic at some point. Looks like you're using `str(job_context["dataset"].pk)` to make this directory specific to this dataset. However couldn't there be eventually be a scenario where a job which puts one or more files into this directory, gets either interrupted or fails for a reason which doesn't happen a second time, then the foreman restarts it causing a non-empty directory to be used. While it seems like we probably won't put extra files in there cause the names would collide, potentially we could append to an existing file? It maybe wouldn't ever be an issue, but I think we eliminate the potential for it to happen if up where we do:
```
# Prepare the output directory
smash_path = "/home/user/data_store/smashed/" + str(job_context["dataset"].pk) + "/"
os.makedirs(smash_path, exist_ok=True)
```
We first make sure the directory doesn't exist and remove it and then recreate it if it does.

workers/data_refinery_workers/processors/test_smasher.py
```diff
@@ -162,6 +164,17 @@ def test_smasher(self):
             final_frame.std(axis=1)
             final_frame.median(axis=1)
 
+            final_context['output_file']
```
kurtwheeler: This line looks like it's not doing much. Also I'm now curious about the lines above it where we're calling methods on `final_frame`. Are those doing anything? It doesn't look like we're testing any of the outputs or using `final_frame` in the rest of the test.

Miserlou:
I think these are all addressed now.

workers/data_refinery_workers/processors/smasher.py
```diff
@@ -133,8 +135,8 @@ def _smash(job_context: Dict) -> Dict:
 
         # Metadata to TSV
         with open(smash_path + 'metadata.tsv', 'w') as output_file:
-            sample_ids = [k for k in metadata['samples'].keys()]
-            keys = [k for k in metadata['samples'][sample_ids[0]].keys()] + ['sample_id']
+            sample_ids = list(metadata['samples'].keys())
+            keys = list(metadata['samples'][sample_ids[0]].keys()) + ['sample_id']
```
kurtwheeler: The list comprehension wasn't the confusing part of this syntax. It was `metadata['samples'][sample_ids[0]].keys()` which is so many layers of lookups at once it's very hard to tell what is happening, especially without having the structure of `metadata` fresh in my head. If you can find a way to clarify this cool, if not it's not that big of a deal.

PR #351, opened 2018-06-26
Prevent duplicate Array Express processor jobs

kurtwheeler:
## Issue Number

N/A

This PR prevents array express from queuing multiple processor jobs per original_file, associates all original files to their downloader jobs, and improves logging a bit.

We were queuing lots of extra processor jobs because for experiments with so much data that multiple zip files are necessary, we would extract the zip files into a common directory and then iterate over the files in it and queue processor jobs for each of them. As each zip was added to the directory, all the files already in the directory were queued again.

## Types of changes

## Functional tests

I have rerun the surveying and downloading steps for E-GEOD-57805 while preventing downloader jobs from actually queuing processor jobs yet letting them still create the database records for them. I then verified that the number of samples was the same as the number of original files was the same as the number of original files. Previously this experiment would queue far too many processor jobs.

## Checklist


foreman/data_refinery_foreman/surveyor/external_source.py
```diff
@@ -53,11 +53,17 @@ def queue_downloader_jobs(self, experiment: Experiment):
...
 
             # We don't need to create multiple downloaders for the same file.
-            if original_file.source_url in downloaded_urls:
+            # However, we do want to associate original_files with the
+            # DownloaderJobs that will download them.
+            if original_file.source_url in download_urls_with_jobs.keys():
+                asoc = DownloaderJobOriginalFileAssociation()
+                asoc.downloader_job = download_urls_with_jobs[original_file.source_url]
+                asoc.original_file = original_file
+                asoc.save()
```
Miserlou: use `get_or_create`?

cgreene:
I agree with @Miserlou that a test for the correct # of jobs queued would be 🥇

kurtwheeler:
@Miserlou I added some tests, wanna take a look at that code?

Miserlou:
Much better, thanks for doing that, and for including your steps in the DEBUGGING file.

Issue #317, opened 2018-06-15
Use Sentry to capture errors


cgreene:
### Context

Sentry ( http://getsentry.com/ ) is a service that helps to capture errors and their context, notify a team or team member, and track the error from discovery through resolution. The frontend repository is now using Sentry. We should make sure we're taking advantage of this resource in the backend as well.

There are packages for R ( https://github.com/mfrasca/r-logging ) and Python ( https://github.com/getsentry/raven-python ).

### Problem or idea

We do not have a way to track errors from when they occur through resolution.

### Solution or next step

We should implement logging to sentry for errors.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

PR #359, opened 2018-06-29
Kurtwheeler sentry

kurtwheeler:
## Issue Number

#317 

We want to use sentry to record and track errors and bugs! This PR adds an integration to do so!

 This PR adds three different mechanisms which send errors to https://sentry.io/greenelab/refinebio/:
  * Any log message with level WARNING or higher will also be sent to sentry via an additional log handler configured in the `common` project.
  * Any uncaught exception will be sent to sentry: https://raven.readthedocs.io/en/stable/integrations/django.html#setup
  * Any request our API responds with a status code 400 or higher will cause a message to be sent to sentry 5% of the time so we can watch trends without having a deluge of issues come from the API.

## Types of changes

## Functional tests

I have triggered errors in the workers and foreman projects by calling `raise ValueError()` within a `try` block so it would be logged and that log message would be sent to sentry. **WORKS!**

I put `1 / 0` at the top of a management command to make sure that uncaught exceptions get sent to sentry. **WORKS!**

I modified the ListSamples endpoint to raise a 403 error to make sure it made it into Sentry. **WORKS!**

I have triggered errors that should be sent to sentry but without the `RAVEN_DSN` env var being set and made sure that Raven didn't complain in any way about not being configured. **WORKS**

## Checklist

## Screenshots

sentry_issues_list

sentry_403


Miserlou:
Is there any way to test this? If this introduces new middleware that all of our API responses will be going through, I'd really like to make sure that there is a well-covered test. Looks good otherwise.

api/data_refinery_api/middleware.py
```diff
@@ -0,0 +1,32 @@
...
+
+        from raven.contrib.django.models import client
+
+        if not client.is_enabled():
+            return response
+
+        data = client.get_data_from_request(request)
+        data.update({
+            "level": logging.WARN,
+            "logger": "BadRequestCatcher?",
```
Miserlou: What does the `?` in `BadRequestCatcher?` mean?

api/data_refinery_api/settings.py
```diff
@@ -159,6 +161,25 @@
...
+# /usr/local/lib/python3.6/site-packages/raven/conf/remote.py:91:
+# UserWarning: Transport selection via DSN is deprecated. You should
+# explicitly pass the transport class to Client() instead.
+raven_dsn = get_env_variable('RAVEN_DSN', "not set")
+if raven_dsn != "not set":
+    RAVEN_CONFIG = {
+        'dsn': raven_dsn,
+        # Only send 5% of errors for the API, since we aren't going to
+        # be interested in any single one.
+        'sampleRate': 0.05
```
Miserlou: Let's crank this to 15/25 for the first release then lower it, I don't want to miss anything while we're still low traffic.

common/data_refinery_common/logging.py
```diff
@@ -41,4 +41,15 @@ def get_and_configure_logger(name: str) -> logging.Logger:
         fmt=FORMAT_STRING, keywords=[]))
     logger.logger.addHandler(handler)
 
+    # This is the Sentry handler
+    raven_dsn = get_env_variable("RAVEN_DSN", "not set")
```
Miserlou: String matching here is a bit weird - we can't just use False?

foreman/data_refinery_foreman/settings.py
```diff
@@ -140,3 +141,20 @@
...
 
 STATIC_URL = '/static/'
+
+
+# Setting the RAVEN_CONFIG when RAVEN_DSN isn't set will cause the
+# following warning:
+# /usr/local/lib/python3.6/site-packages/raven/conf/remote.py:91:
+# UserWarning: Transport selection via DSN is deprecated. You should
+# explicitly pass the transport class to Client() instead.
+raven_dsn = get_env_variable('RAVEN_DSN', "not set")
```
Miserlou: Same comment as above

workers/data_refinery_workers/settings.py
```diff
@@ -145,3 +139,20 @@ def get_env_variable(var_name):
...
 
 STATIC_URL = '/static/'
+
+
+# Setting the RAVEN_CONFIG when RAVEN_DSN isn't set will cause the
+# following warning:
+# /usr/local/lib/python3.6/site-packages/raven/conf/remote.py:91:
+# UserWarning: Transport selection via DSN is deprecated. You should
+# explicitly pass the transport class to Client() instead.
+raven_dsn = get_env_variable('RAVEN_DSN', "not set")
```
Miserlou: And again

Miserlou:
ACTUALLY now that I'm using Sentry, is there anyway that we can use different settings for API and Crunch endpoints? They're all mixed together.

Miserlou:
It looks like the API responses aren't getting the correct remote IP address, it returns `"172.17.0.1"` which is the docker bridge network.

kurtwheeler:
You're definitely correct about that IP. I've now spent quite a bit of time looking into this. It seems like this actually isn't the most simple issue to resolve. There's an issue that's almost three years old about it: https://github.com/moby/moby/issues/15086 and another that's close to two years old: https://github.com/docker/for-mac/issues/180 (it's not actually specific to mac)

I was able to get the actual IP by using the `--net=host \` option that was suggested in that issue. However this conflicts with the `--link drdb:postgres` flag and perhaps has some other implications. I'm not exactly sure what this would do to the prod configuration, if anything. Therefore I think I'll wait till we can discuss this in person on Monday to make sure that there's nothing else that would need to be done along with that change.

In the meantime I'm going to merge this PR so that I can kick off some crunch with Sentry enabled. This remaining issue is only going to affect the API, and we don't have too many users there anyway. I'm opening an issue to keep track of this IP issue.

kurtwheeler:
I've made a new issue to track the IP issue here: https://github.com/AlexsLemonade/refinebio/issues/368

Issue #341, opened 2018-06-20
Add Capabilities to Process and Smash Externally Supplied Data


Miserlou:
### Context

It's quite likely that there will be researchers who will want to combine `refine.bio`-provided data with their own private data repositories.

It'd be great if we provided some functionality to make this easier for them by either running through our processors, gene-converters and smasher. Off the top of my head, I'm imagining that this could be new management command to live inside a processor/smasher Dockerfile and a small script with documentation to wire everything up together. But I'm certainly open to suggestions here!

This would also improve our own testing abilities.

### Problem or idea

I think this could be a good project for @Willv19  to take on if @kurtwheeler doesn't have a grander plan in mind.

### Solution or next step

Next step would probably to put some bounds on the requirements here - what kind of private data is a researcher most likely to have? Will it need to be processed, or just gene-conversion, or even just smashing?

kurtwheeler:
We had a meeting to discuss the right way to implement this. The upshot of the discussion was that we should end up with a command like:
```
./external_smasher.sh -o 'GALLUS_GALLUS' -i smashed.zip -v /tmp/my_data
```
Where `-o` specifies the organism, `-i` specifies a zipfile that was downloaded from `refine.bio`, and `-v` specifies a directory which contains their data and is where the smashed data will be output.

We came up with a few ways to limit the complexity of the issue, at least initially but perhaps also longterm depending on what requests we receive:
  * We will only support a single organism
  * We will allows users to run this on raw data only.
  * We will only support Affymetrix and RNA-SEQ platforms.

This should also include a nice tutorial both explaining how to use the tool along with examples. The instructions/tutorial will need a very large and obvious warning that when users are downloading data from the UI they NEED TO SELECT NO TRANSFORMATIONS FOR THE DATA. It should explain that transforming the data before the download and after it gets resmashed with their private data will be a double transformation which will make their data wrong.

kurtwheeler:
@jaclyn-taroni @Miserlou I believe I just wrote up everything we discussed. One thing that occurred to me as I did so was that we haven't specified a way to allow users to pick their transformation for the smashing. Should we just have a `-t` options which allows them to specify one of the valid values? I believe [these are the options](https://github.com/AlexsLemonade/refinebio/blob/dev/workers/data_refinery_workers/processors/smasher.py#L61) they'd have to pick from, am I correct? Is standard the same as picking none?

cgreene:
Regarding
[...]

Issue #268, opened 2018-05-21
Allow users to download transcriptome indices


kurtwheeler:
### Context

We use `salmon` to build indices which are then used while running `salmon` to process RNASeq data. These indices are based on data which we retrieve from ensembl.org.

### Problem or idea

Ensembl releases a new version every now and then. In April they released version 92. This broke our transcriptome code because rest.ensembl.org does not accept a release number as a query parameter. This means that we cannot rebuild old transcriptome indices using the same code that built them in the first place. In order for the data processed using those indices to be reproducible, we'll need to allow users to download the version of the transcriptome index which was used for the data they are interested in.

### Solution or next step

Since we cannot rebuild old transcriptome indices, we should store them and allow users to download any of them.

I'll note that this could become moot if Ensembl adds a `release` parameter to their REST API. I have emailed the mailing list requesting this.

### New Issue Checklist

- [X] The title is short and descriptive.
- [X] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

kurtwheeler:
See http://lists.ensembl.org/pipermail/dev/2018-May/013012.html for a more in depth description of the problem and (hopeful) followup from the mailing list.

jaclyn-taroni:
Possibly related: #79

jaclyn-taroni:
During tech team meeting this morning, we decided to remove this from Keytar Kurt

kurtwheeler:
Ok @Willv19 is going to tackle this before he gets to https://github.com/AlexsLemonade/refinebio/issues/341 since that will need these indices. I've come up with a rough outline of how to accomplish this:
 1. Create an S3 bucket within terraform to store the OrganismIndex files.
 2. Modify the OrganismIndex model to include an `s3_url` field which points to where the index is stored within S3.
 3. Modify the `transcriptome_index` processor so that it uploads the processed indices to the correct S3 bucket.
4. Deploy these changes.
5. Create a script to trigger surveyor jobs for all organisms in http://ensemblgenomes.org/  http://useast.ensembl.org/index.html.
6. Run the script within the staging or a dev stack and debug it until all of the organisms are processed.
7. Add an API route to retrieve the URLs for these indices. Basically a route like `/transcriptome_indices?organism=GALLUS_GALLUS&length=short` should do it.

jaclyn-taroni:
@Willv19 was this closed by #363?

Willv19:
Yes

PR #363, opened 2018-07-02
Willv19 download transcriptome indices

Willv19:
## Issue Number

#268 

Now, whenever we download a transcriptome index we also upload a copy of it to an S3 bucket. The S3 bucket lets us download the index later, particularly things like the external smasher in #341. There is also a new API endpoint `/transcriptome-indices` that returns the S3 url to the requested transcriptome index.

## Types of changes

## Functional tests

- I spun up an S3 bucket and ran the surveyor locally to get the transcriptome indices for 10 samples, and the processors successfully uploaded those indices to the correct S3 bucket.
- I ran the API locally and it gave me back the correct S3 URL for the transcriptome indices that I processed.

## Checklist


api/data_refinery_api/views.py
```diff
@@ -485,3 +490,45 @@ def get(self, request, format=None):
...
+        Length must either be "long" or "short"
+        """
+
+        if ORGANISM_INDEX_BUCKET == "NO_BUCKET":
+            raise Http404
+
+        transcription_length = "TRANSCRIPTOME_" + transcription_length.upper()
+
+        try:
+            organism_index = (OrganismIndex.public_objects.exclude(s3_key__exact="")
```
Miserlou: I would suggest moving this functionality onto the Organism object itself as get_index, in fact I'd even take that a step further and add a get_index_s3_url to move all of this model logic off of the API view function.

api/data_refinery_api/views.py
```diff
@@ -485,3 +490,45 @@ def get(self, request, format=None):
...
+        if "organism" not in params:
+            errors["organism"] = "You must specify the organism of the index you want"
+        if "length" not in params:
+            errors["length"] = "You must specify the length of the transcriptome index"
+
+        if len(errors) > 0:
+            raise ValidationError(errors)
+
+        url = self.get_s3_url(params["organism"], params["length"])
+        return Response(url)
```
Miserlou: If you make the s3_url a property of the Index, and we add a proper serializer, then we can serialize the entire object (or alternately make this a sub feature of the Organism itself). It seems a bit unusual for one of our endpoints to only return a single url value with no other information about the object.

api/environments/local
```diff
@@ -21,3 +21,4 @@ DATABASE_TIMEOUT=5
 RUNNING_IN_CLOUD=False
 SERVICE=api
 
+S3_TRANSCRIPTOME_INDEX_BUCKET_NAME=NO_BUCKET
```
Miserlou: I suggest making this None so we can use it programmatically if need be.

common/data_refinery_common/models/models.py
```diff
@@ -15,6 +18,8 @@
 
 from data_refinery_common.models.organism import Organism
 
+S3 = boto3.client('s3', config=Config(signature_version='s3v4'))
```
Miserlou: `signature_version='s3v4'` is voodoo, so you might want to include a comment line about why this is necessary for connecting to certain regions.

api/environments/local
Willv19: Our current implementation of get_env_variable fails whenever a falsey value is set as the environment variable. I initially had it set as an empty string but then the program would error. Is there a circumstance where this behaviour is needed/expected? If not, should I just rewrite the get_env_variable function?

common/data_refinery_common/models/models.py
```diff
@@ -407,11 +412,24 @@ def __str__(self):
     # This matters, for instance salmon 0.9.0 indexes don't work with 0.10.0
     salmon_version = models.CharField(max_length=255)
 
+    # S3 Key
+    s3_key = models.CharField(max_length=255, default="")
```
Miserlou: You can't store `s3_key` without storing `s3_bucket` along with it as well, since both are needed to actually retrieve the value. It's okay to get the bucket location from an envvar at runtime when _uploading_, but it must be retrieved from a datastore when the value is being _retrieved_, else they could desync and we could lose the value.

common/data_refinery_common/models/models.py
```diff
@@ -407,11 +412,24 @@ def __str__(self):
...
+
     # Common Properties
     is_public = models.BooleanField(default=True)
     created_at = models.DateTimeField(editable=False, default=timezone.now)
     last_modified = models.DateTimeField(default=timezone.now)
 
+    def upload_to_s3(self, absolute_file_path, bucket_name, logger):
+        if bucket_name != "NO_BUCKET":
```
Miserlou: Check against `None` rather than string matching

common/data_refinery_common/models/models.py
```diff
@@ -407,11 +412,24 @@ def __str__(self):
...
+            s3_key = self.organism.name + '_' + self.index_type + '.tar.gz'
+            S3.upload_file(absolute_file_path, bucket_name, s3_key,
```
Miserlou: If possible, we should `util`-ize all of the S3 logic so that it can be repeated for all of the different objects which need to be stored and retrieved from S3.

If you need a reference: https://github.com/Miserlou/Zappa/blob/192948eef9cfaaa58cd683c0c9a9afa0ecb31fb4/zappa/core.py#L902 has a bit more sanding on it for weird regional quirks.

Miserlou: This will also help with testability.

workers/data_refinery_workers/processors/test_transcriptome_index.py
```diff
@@ -11,6 +11,7 @@
     ProcessorJobOriginalFileAssociation,
     ProcessorJob
 )
+from unittest.mock import patch
```
Miserlou: For AWS testing, I suggest that we maybe try to use the [placebo](https://github.com/garnaat/placebo) library rather than simply mocking calls out.

workers/data_refinery_workers/processors/transcriptome_index.py
```diff
@@ -29,6 +30,8 @@
 # Removes each occurrance of ; and "
 IDS_CLEANUP_TABLE = str.maketrans({";": None, "\"": None})
 
+ORGANISM_INDEX_BUCKET = get_env_variable("S3_TRANSCRIPTOME_INDEX_BUCKET_NAME", "NO_BUCKET")
```
Miserlou: None

PR #355, opened 2018-06-27
Dualsmash Part 1: Bug Fixes and Test Files

Miserlou:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/330

## Purpose/Implementation Notes

Provides some important fixes necessary for building the files for dual-smashing. Adds the fixes and some test files generated for them. The rest of the changes are applied above the Smasher Part 2 PR.

Also adds an `affymetrix_local` image to avoid having to rebuild affy image so many times/not use updated code with affy processors.

## Types of changes

- Bugfix (non-breaking change which fixes an issue)

## Functional tests

New test added.

## Checklist

_Put an `x` in the boxes that apply._

- [ ] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

Issue #330, opened 2018-06-18
Initial version of cross-technology aggregation


jaclyn-taroni:
### Context

We'd like to provide the ability to combine samples from microarray and RNA-seq. I imagine this is going to be a concern almost exclusively when "smashing" by species rather than by experiment.

### Problem or idea

For Marimba Miso, from https://github.com/AlexsLemonade/refinebio/issues/186#issuecomment-395187013:
>if there are a mix of technologies:

> aggregating by experiment -> return untransformed output from tximport
> aggregating by species -> log2(x + 1) tximport output

In the near future, we'd like to do something [like this](https://github.com/AlexsLemonade/refinebio/issues/186#issuecomment-395218310).

### Solution or next step

The smasher logic should take the technology and the aggregation type into account when smashing RNA-seq data and transform when appropriate.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

Miserlou:
From the parent:

> The Troyanska lab had something a little more complicated that I think Chad Myers wrote but we should do this for now.

Does anybody have a paper on this to share for my files?

cgreene:
It may be earlier than this, but it's roughly described in the hefalmp paper: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2694471/

> We collected 635 human microarray data sets from the NCBI Gene Expression Omnibus (GEO) repository (Barrett et al. 2005) comprising 14,671 conditions; see Supplemental Table 1 for a complete list. These were processed largely as in Huttenhower and Troyanskaya (2008), with additional manipulation to handle single-channel data and the ambiguity of human probe mapping. Within each data set, negative and very small (less than two) single-channel values were removed, genes with missing values in >30% of the conditions were removed, and the remaining missing values were imputed using KNNImpute (Troyanskaya et al. 2001) with k = 10 (cutoff values recommended by Hibbs et al. [2007] and Troyanskaya et al. [2001], respectively).
> 
> Probe IDs were mapped to HGNC symbols using the appropriate GEO platform files. When multiple probes mapped to a single HGNC symbol, a consensus set of probes was generated by finding pairwise Euclidean distances more likely to have been generated from the data set's distribution of intragene probe pairs than from the distribution of intergene probe pairs. If this consensus set contained at least half of the probes mapping to a gene symbol, the consensus set's average value became the expression vector for that gene.

PR #364, opened 2018-07-02
Dualsmash Part 2: Transform and Smash

Miserlou:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/330
Depends on: https://github.com/AlexsLemonade/refinebio/pull/355
Conflicts with: https://github.com/AlexsLemonade/refinebio/pull/307

Adds ability to smash RNASeq/Microarray data together. 

## Methods
Jackie recommended: 

```
aggregating by experiment -> return untransformed output from tximport
aggregating by species -> log2(x + 1) tximport output
```

For `SPECIES`, this is accomplished by: 

```
data = data + 1
data = np.log2(data)
```

I don't understand the inclusion of `aggregating by experiment`, since there are no experiments (as far as I am aware) which use both technologies at once. Still, if it does occur, the transformation above won't.

## Types of changes

## Functional tests

New tests and new test data is provided.

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,7 +58,10 @@ jobs:
       - run: sudo chown -R circleci:circleci workers/test_volume/
       - run: .circleci/filter_tests.sh -t illumina
 
-        # Run Smasher tests.
+      # Run Smasher tests.
+      - run: sudo apt-get update && sudo apt-get install -qq -y python-pip libpython-dev
+      - run: curl -O https://bootstrap.pypa.io/get-pip.py && sudo python get-pip.py
+      - run: sudo pip install -q awscli --upgrade
       - run: sudo chown -R circleci:circleci workers/test_volume/
       - run:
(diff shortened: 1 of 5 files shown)


jaclyn-taroni:
To clarify: If a user has selected both microarray and RNA-seq experiments but wants to aggregate by experiment (i.e., they are not actually combining technologies), then do not perform the transformation. It sounds like we're on the same page 👍

Miserlou:
Yes. However, this conflicts with some behavior in #307: https://github.com/AlexsLemonade/refinebio/pull/307/files#diff-04738c4482565d676e6b7a40dc060df9R87

Does that check need to be a) made technology-aware or b) removed entirely?

jaclyn-taroni:
> Does that check need to be a) made technology-aware

As in check only for microarray data? I like that idea.

Miserlou:
Okay, I added that to the other PR.

PR #367, opened 2018-07-03
More Useful Log Lines

Miserlou:
## Issue Number

n/a

There were quite a few errors in the logs from the recent crunch that are pesky to debug because they don't provide enough filterable information to `awslogs`. This adds more information to the relevant log lines.

## Types of changes

## Functional tests

n/a

### DIFF
--- a/workers/data_refinery_workers/downloaders/utils.py
+++ b/workers/data_refinery_workers/downloaders/utils.py
@@ -65,7 +65,18 @@ def create_processor_jobs_for_original_files(original_files: List[OriginalFile],
     """
     for original_file in original_files:
         sample_object = original_file.samples.first()
-        pipeline_to_apply = determine_processor_pipeline(sample_object)
+
+        if original_file.filename[-4:] == ".xml":
+            logger.info("Skipping useless file",
+                file=original_file.id,
+                filename=original_file.filename)
+            continue
+
+        # We NO_OP processed data. It's what we do.
+        if '.processed' in original_file.source_url:
+            pipeline_to_apply = ProcessorPipeline.NO_OP
+        else:
+            pipeline_to_apply = determine_processor_pipeline(sample_object)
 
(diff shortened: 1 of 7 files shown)


Miserlou:
I've now also included two fixes: an explicit check to ensure we NO-OP processed files, and to make sure that we skip useless XML files that come in source archive zips.

workers/data_refinery_workers/downloaders/utils.py
```diff
@@ -65,7 +65,18 @@ def create_processor_jobs_for_original_files(original_files: List[OriginalFile],
...
+            logger.info("Skipping useless file",
+                file=original_file.id,
+                filename=original_file.filename)
+            continue
+
+        # We NO_OP processed data. It's what we do.
+        if '.processed' in original_file.source_url:
+            pipeline_to_apply = ProcessorPipeline.NO_OP
+        else:
+            pipeline_to_apply = determine_processor_pipeline(sample_object)
```
kurtwheeler: I think this logic should be in `determine_processor_pipeline` so that anywhere else that calls this also has this check (like the other call to this function in `create_processor_job_for_original_files`).

Miserlou: The difference is that `determine_processor_pipeline` is entirely Sample based, and can only return a valid pipeline. This is a check based on a file, and particularly a file which will _never_ have a valid associated pipeline. I see this more as a preprocessor check. In `create_processor_job_for_original_files`, we always want to include all of the supplied files. There isn't anywhere else where `determine_processor_pipeline` is used.

kurtwheeler:
Ok, I think I see what you mean. Just to note, this means that there could be some samples which get processed with a pipeline, but have some files get NO-OP'ed. This also means that there could be samples where we only NO-OP some of the files associated with the sample. However you're more experienced with the content of these files and which are more useful so if that seems like the ideal behavior to you I'm with it.

PR #369, opened 2018-07-06
Get HTTPS working for the frontend.

kurtwheeler:
This also allows the frontend to successfully make requests to the API in staging. Additionally this fixes the bug where refreshing the page causes a 403.

## Issue Number

I'm not I actually made an issue for this.

This sets up cloudfront so that we can get HTTPS working for staging.refine.bio. I didn't know much about cloudfront, so I had to learn a lot and there's some settings that I'm not at all confident are correct. However this in combination with the current DNS settings makes everything work so it's at least close.

Additionally I have changed the nginx configuration to have a wide open CORS policy. Once again I'm not sure this is something we want to do, but I don't think we're really restricting who can access us so it might be?

## Types of changes

## Functional tests

https://staging.refine.bio/results?q=human

## Checklist


infrastructure/disk.tf
```diff
@@ -82,3 +82,30 @@ resource "aws_s3_bucket" "data-refinery-static" {
...
+
+resource "aws_s3_bucket" "data-refinery-static-access-logs" {
+  bucket = "data-refinery-static-access-logs"
+
+  tags {
+    Name = "data-refinery-static-access-logs-${var.user}-${var.stage}"
+    Environment = "${var.stage}"
+  }
+
+
```
Miserlou: space

infrastructure/networking.tf
```diff
@@ -166,3 +166,83 @@ resource "aws_lb_target_group_attachment" "api-https" {
...
+    forwarded_values {
+      query_string = false
+
+      cookies {
+        forward = "none"
+      }
+    }
+
+    viewer_protocol_policy = "allow-all" # Change this to redirect-to-https
+    # XXX: This might make things pricey, but what should they actually be?
```
Miserlou: Let's get rid of this XXX and explain what this behavior does and what the possibilities here can be.

infrastructure/api-configuration/nginx_config.conf
```diff
@@ -60,6 +60,34 @@ http {
             proxy_set_header Host $http_host;
             proxy_set_header X-Cluster-Client-Ip $remote_addr;
             proxy_pass http://app;
+
+            if ($request_method = 'OPTIONS') {
```
Miserlou: I am surprised to see this here. Normally CORS is handled at the application layer, and we already have CORS middleware installed for the API.

I'd rather this stuff not be site wide in case we want to have parts of the site un-CORS (admin pages, internal tools and dashboards, etc).

I was expecting this to be a plain http -> https rewrite - more generally, can you explain what's actually happening in this PR?

I see that this includes both Let'sEncrypt's CertBot and AWS's Certificate Manager - which service is generating the SSL certificate? How is it being verified? How is it being renewed? Where is this process documented?

kurtwheeler:
OK @Miserlou I've addressed your comments and the tests are passing again.

Issue #328, opened 2018-06-18
Conversion of submitter-supplied gene identifiers to Ensembl gene ID: Illumina BeadArray


jaclyn-taroni:
### Context

A subset of what was originally envisioned for #186. As of the writing of this ticket/#307, we add support for the detection of submitter-supplied gene identifiers and conversion to Ensembl gene IDs. Currently, this requires us to generate and store a mapping between the identifiers we will support conversion from and the Ensembl gene ids (see: https://github.com/AlexsLemonade/refinebio/pull/307#issuecomment-396364904). We have only generated this data for Brainarray-supported Affymetrix.

### Problem or idea

We'll need to generate this mapping for the supported Illumina BeadArray platforms: https://github.com/AlexsLemonade/refinebio/issues/232#issuecomment-389302061 and https://github.com/AlexsLemonade/refinebio/issues/232#issuecomment-389609871 (also related: https://github.com/AlexsLemonade/refinebio/issues/222#issuecomment-389571978).

### Solution or next step

I will work with the backend team to generate this mapping for supported Illumina BeadArray platforms. The initial solution to this likely involves using the `.db` Illumina packages and filtering by probe quality as [is done in the Illumina SCAN processor](https://github.com/AlexsLemonade/refinebio/blob/e9900fa42459ab1b06e8980f5d8849705351f338/workers/data_refinery_workers/processors/illumina.R#L424). I think it makes sense to tackle this or at least discuss it at the same time as #232 because the same packages will be required.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

jaclyn-taroni:
@Miserlou was this closed by #375?

Miserlou:
Closed by #375!

PR #375, opened 2018-07-09
NO_OP Preprocessed Illumina Data

Miserlou:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/328

For Illumina data we can't process, convert the probe identifiers into smashable Ensembl gene identifiers.

## Methods

We use the existing `detect_database.R` script to find the correct database to use for a given file, then use AnnotationDbi to convert the identifiers like so:

```R
exprs[, 1] <- mapIds(get(db_name), keys=exprs[, 1], column="ENSEMBL", keytype="PROBEID")
```

## Types of changes

## Functional tests

New tests, in `test_convert_processed_illumina`, are provided.

## Checklist


Miserlou:
The branch this was based on has been merged, so this should be reviewable now.

workers/data_refinery_workers/processors/gene_convert_illumina.R
```diff
@@ -0,0 +1,71 @@
...
+message(head(exprs))
+
+# Build a PROBEID:ENSEMBL map
+message("Here's the db..")
+db_name <- paste(platform, ".db", sep="")
+message(db_name)
+library(db_name, character.only=TRUE)
+
+# Replace the probe IDs with Ensembles
+exprs[, 1] <- mapIds(get(db_name), keys=exprs[, 1], column="ENSEMBL", keytype="PROBEID")# , multiVals="first")
```
jaclyn-taroni: If we use [`mapIds` with the default behavior (we don't set `multiVals` argument)](https://www.rdocumentation.org/packages/AnnotationDbi/versions/1.34.4/topics/AnnotationDb-objects) and there are multiple matches (so here it would be a single probe maps to multiple Ensembl gene identifiers), it selects the whatever comes back first. I'm not sure how frequent one-to-multiple mappings are, have you come across that?

workers/data_refinery_workers/processors/gene_convert_illumina.R
```diff
@@ -0,0 +1,71 @@
...
+res <- exprs
+message(head(res))
+
+# Remove all of the unmapped (NA) values (likely control probes?)
+res <- res[complete.cases(res), ]
```
jaclyn-taroni: Note that this will also drop missing expression values, which is not necessarily a bad thing.

workers/data_refinery_workers/processors/gene_convert_illumina.R
Miserlou: It doesn't yell about one-to-multiple mappings are far as I can see, but I'm not sure if that error would make it to the logs.

What mode for multiVals would you prefer?

jaclyn-taroni: What if we did something like (relies on tidyverse packages):
```R
# need what the first column is called
exprs_id_name <- colnames(exprs)[1]
# don't replace the identifiers in exprs yet
mapped_list <- mapIds(get(db_name), keys=exprs[, 1], column="ENSEMBL", keytype="PROBEID", multiVals="list")
# get into data.frame form, should capture one to multiple mapping
mapped_df <- reshape2::melt(mapped_list)
# not 100% sure this bit is correct -- might be the other way around
colnames(mapped_df) <- c("ENSEMBL", "PROBEID")
annot_exprs_df <- mapped_df %>%
  dplyr::filter(!is.na(ENSEMBL)) %>%   # dropping unmapped -- mapIds might do this by default!
  dplyr::inner_join(y = exprs, by = c("PROBEID" = exprs_id_name)) %>%  # annotation heavy lifting
  dplyr::select(-PROBEID)  # drop the probe ids
```
And then you'd write `annot_exprs_df` to file. I haven't fully tested this out, but I think it's close.

workers/data_refinery_workers/processors/no_op.py
```diff
@@ -110,6 +122,78 @@ def _convert_genes(job_context: Dict) -> Dict:
...
+    high_db = None
+    for platform in databases:
+        try:
+            result = subprocess.check_output([
+                    "/usr/bin/Rscript", 
+                    "--vanilla", 
+                    "/home/user/data_refinery_workers/processors/detect_database.R",
+                    "--platform", platform,
+                    "--inputFile", job_context['input_file_path'],
+                    "--column", str("Reporter Identifier"), # XXX: Detect 0 col name
```
kurtwheeler: This seems like an XXX we might want to X out now? Or is it blocked by something?

workers/data_refinery_workers/processors/no_op.py
```diff
@@ -110,6 +122,78 @@ def _convert_genes(job_context: Dict) -> Dict:
...
+                ])
+
+            results = result.decode().split('\n')
+            cleaned_result = float(results[0].strip())
+
+            if cleaned_result > highest:
+                highest = cleaned_result
+                high_db = platform
+                high_mapped_percent = float(results[1].strip())
```
kurtwheeler: What's the difference between `highest` and `high_mapped_percent`. `high_mapped_percent` doesn't actually seem to be referenced anywhere else. Also would this be good to store on the result as an annotation so we can later see how good the mapping we went with actually was?

workers/data_refinery_workers/processors/no_op.py
```diff
@@ -110,6 +122,78 @@ def _convert_genes(job_context: Dict) -> Dict:
...
+
+        except Exception as e:
+            logger.exception(e)
```
kurtwheeler: Will these log statements have much info we can work on? I'm not sure if we really need to log these or not, but if we do I think we should include enough context to know what job generated the logs.

PR #378, opened 2018-07-10
Prevent Duplicate Items in Dataset

Miserlou:
## Issue Number

https://github.com/AlexsLemonade/refinebio-frontend/issues/139

Adds stricter dataset validation to prevent allowing a current bug in the front end.

## Types of changes

## Functional tests

New assertion added.

## Checklist


Miserlou:
I actually tried that but it gets caught by the JSON parser since JSON doesn't allow overlapping values in the dictionary. The behavior is that it accepts the last definition of a duplicate keyvalue.

kurtwheeler:
Ah perfect then, my approval stands!

PR #386, opened 2018-07-13
Sync Original and Result Files to S3_IA

Miserlou:
## Issue Number

n/a

Uploads results of scrapes and computations to AWS S3 Infrequent Access. Downloads ComputedFiles as necessary at smashtime.

OriginalFiles in addition to ComputedFiles are both synced. For the 3-million crunch, we'll want to switch this to simply deleting the OriginalFiles on disk once they're not needed anymore, without any S3 syncing, as it's actually cheaper to rescrape than it is to store them on disk. We will need special considerations for Salmon, as `tximport` requires OriginalFiles survive until all the samples in an experiment are available (I think).

This change will save a projected $22,000,000 annually over the previous method.

## Types of changes

## Functional tests

New test case added.

## Checklist

### DIFF
--- a/common/data_refinery_common/models/models.py
+++ b/common/data_refinery_common/models/models.py
@@ -689,8 +689,6 @@ class ComputedFile(models.Model):
             os.remove(self.absolute_file_path)
         except OSError:
             pass
-        self.is_downloaded = False
-        self.save()
 
     def get_synced_file_path(self):
         """ Fetches the absolute file path to this ComputedFile, fetching from S3 if it
--- a/common/data_refinery_common/models/models.py
+++ b/common/data_refinery_common/models/models.py
@@ -570,6 +570,29 @@ class OriginalFile(models.Model):
         else:
             return self.filename
 
+    def sync_from_s3(self):
+        """ Downloads a file from S3 to the local file system.
+        Returns the absolute file path.
+        """
+        try:
+            S3.download_file(
+                        self.s3_bucket, 
+                        self.s3_key,
+                        self.absolute_file_path
+                    )
+            return self.absolute_file_path
+        except Exception as e:
+            logger.exception(e)
+            return None
+
+    def get_synced_file_path(self):
+        """ Fetches the absolute file path to this ComputedFile, fetching from S3 if it
+        isn't already available locally. """
+        if os.path.exists(self.absolute_file_path):
+            return self.absolute_file_path
+        else:
+            return self.sync_from_s3()
+
     def delete_local_file(self):
         """ Deletes a file from the path and actually removes it from the file system,
         resetting the is_downloaded flag to false. Can be refetched from source if needed. """
--- a/common/data_refinery_common/models/models.py
+++ b/common/data_refinery_common/models/models.py
@@ -16,12 +16,17 @@ from django.db import transaction
 from django.db import models
 from django.utils import timezone
 
+from data_refinery_common.logging import get_and_configure_logger
 from data_refinery_common.models.organism import Organism
-from data_refinery_common.utils import get_s3_url
+from data_refinery_common.utils import get_env_variable, get_s3_url
 
 # We have to set the signature_version to v4 since us-east-1 buckets require
 # v4 authentication.
 S3 = boto3.client('s3', config=Config(signature_version='s3v4'))
+
+logger = get_and_configure_logger(__name__)
+LOCAL_ROOT_DIR = get_env_variable("LOCAL_ROOT_DIR", "/home/user/data_store")
+CHUNK_SIZE = 1024 * 256 # chunk_size is in bytes
 
 """
 # First Order Classes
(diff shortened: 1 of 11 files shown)


jaclyn-taroni:
> We will need special considerations for Salmon, as tximport requires OriginalFiles survive until all the samples in an experiment are available (I think).

We want to keep the Salmon quant output for research purposes.

Miserlou:
Quant output is computed, so should never be deleted entirely. It's only the original input files that I'm talking about discarding.

common/data_refinery_common/models/models.py
```diff
@@ -610,13 +685,48 @@ def save(self, *args, **kwargs):
...
+                        s3_bucket, 
+                        s3_key,
+                        ExtraArgs={
+                            'ACL': 'public-read',
+                            'StorageClass': 'STANDARD_IA'
+                        }
+                    )
+            self.save()
+        except Exception as e:
+            logger.exception(e)
```
kurtwheeler: Same comment about an id.

common/data_refinery_common/models/models.py
```diff
@@ -610,13 +685,48 @@ def save(self, *args, **kwargs):
...
+
+        try:
+            S3.download_file(
+                        self.s3_bucket, 
+                        self.s3_key,
+                        self.absolute_file_path
+                    )
+            return self.absolute_file_path
+        except Exception as e:
+            logger.exception(e)
```
kurtwheeler: Same comment about an id.

common/data_refinery_common/models/models.py
```diff
@@ -634,6 +744,24 @@ def calculate_size(self) -> None:
         self.size_in_bytes = os.path.getsize(self.absolute_file_path)
         return self.size_in_bytes
 
+    def delete_local_file(self):
+        """ Deletes a file from the path and actually removes it from the file system,
+        resetting the is_downloaded flag to false. Can be refetched from source if needed. """
```
kurtwheeler: This comment looks like it was copied from the OriginalFile function because I don't think ComputedFile has  a `is_downloaded` field nor will they be re-fetched from the source.

workers/data_refinery_workers/processors/smasher.py
```diff
@@ -49,6 +49,10 @@ def _prepare_files(job_context: Dict) -> Dict:
         all_sample_files = list(set(all_sample_files))
         job_context['input_files'][key] = all_sample_files
 
+    # So these get delted from disk after..
```
kurtwheeler: delted -> deleted

workers/data_refinery_workers/processors/test_smasher.py
```diff
@@ -629,3 +629,31 @@ def test_sanity_imports(self):
...
+        result = ComputationalResult()
+        result.save()
+
+        computed_file = ComputedFile()
+        computed_file.s3_key = "all_the_things.jpg"
+        computed_file.s3_bucket = "data-refinery-test-assets"
+        computed_file.filename = "all_the_things.jpg"
+        computed_file.absolute_file_path = "/home/user/data_store/PCL/" + computed_file.filename
+        computed_file.result = result
+        computed_file.size_in_bytes = 9000
```
kurtwheeler: Are you sure this shouldn't be OVER 9000?!?!?

workers/data_refinery_workers/processors/utils.py
```diff
@@ -66,6 +71,7 @@ def start_job(job_context: Dict):
 
         # Just in case
         job_context["original_files"] = []
+        job_context["computed_files"] = []
```
kurtwheeler: It seems like this line could be outside of the `if` instead of both of its branches.

workers/data_refinery_workers/processors/utils.py
```diff
@@ -96,6 +102,22 @@ def end_job(job_context: Dict, abort=False):
...
+        if result:
+            original_files.delete_local_file()
+
+    # S3-sync Computed Files
+    for computed_file in job_context['computed_files']:
+        # Ensure even distribution across S3 servers
+        nonce = ''.join(random.choice(string.ascii_lowercase + string.digits) for _ in range(8)) 
+        result = computed_file.sync_to_s3(S3_BUCKET_NAME, nonce + "_" + computed_file.filename)
+        if result:
+            computed_file.delete_local_file()
```
kurtwheeler: I remember you telling me about this trick, but it might be good to include a link to something explaining what and why you're doing it.

Miserlou: There is already in a comment in the line above this.

Miserlou:
Should be addressed now.

workers/data_refinery_workers/processors/test_smasher.py
```diff
@@ -743,7 +743,7 @@ def test_get_synced_files(self):
         computed_file.filename = "all_the_things.jpg"
         computed_file.absolute_file_path = "/home/user/data_store/PCL/" + computed_file.filename
         computed_file.result = result
-        computed_file.size_in_bytes = 9000
+        computed_file.size_in_bytes = 9001
```
kurtwheeler: Yes!

PR #394, opened 2018-07-17
Backend Changes for Frontend Issue 124

Willv19:
## Issue Number

Backend changes required to fix https://github.com/AlexsLemonade/refinebio-frontend/issues/124

Add the organisms and the technology to the DetailedExperimentSerializer so the frontend can use the values.

## Types of changes

## Checklist

## Screenshots

### DIFF
--- a/api/data_refinery_api/serializers.py
+++ b/api/data_refinery_api/serializers.py
@@ -255,7 +255,6 @@ class DetailedExperimentSerializer(serializers.ModelSerializer):
                     'submitter_institution',
                     'last_modified',
                     'created_at',
-                    'technology',
                     'organisms',
                 )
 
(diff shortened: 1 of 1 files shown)


Willv19:
@Miserlou The original issue mentions "instrument" instead of technology. Are those the same thing? If not, where in the Experiment model is the "instrument" stored?

Miserlou:
Technology is going to be either Microarray or RNA Seq. Instrument information is on the individual sample, which is a combination of manufacturer information and model number, found formatted nicely by the `pretty_platform` function, ex `Affymetrix HT HG-U133+ PM Array Plate (hthgu133pluspm)`. `technology` actually lives on both, since there are on rare occasions experiments which combine technologies, but generally we use whatever the experiment self describes as.

PR #415, opened 2018-07-18
Sets Smash Priority to 100 and All Other Priorities to 50

Miserlou:
## Issue Number

n/a

Scraping and processing can happen at their own pace, but people who are smashing need the data as soon as possible. This will create a priority for Smash over Crunch.

I have no idea what these values actually mean, and I can't find any documentation about it. I assume assuming that 100 good, 50 less good.

## Types of changes

## Functional tests
No

## Checklist


kurtwheeler:
Higher numbers are higher priority. However 100 is the max so by choosing 100 instead of 75 or even 90 you remove our ability to queue a job with even higher priority.

PR #421, opened 2018-07-19
Various Fixes - R bug, Pandas Fail Tolerance, Split Upload and Notify

Miserlou:
## Issue Number

n/a

R bug, Pandas Fail Tolerance, Split Upload and Notify into separate functions.

## Types of changes

## Functional tests

`¯\_(ツ)_/¯`


workers/data_refinery_workers/processors/smasher.py
```diff
@@ -189,6 +194,7 @@ def _smash(job_context: Dict) -> Dict:
         metadata['num_experiments'] = job_context["experiments"].count()
         metadata['aggregate_by'] = job_context["dataset"].aggregate_by
         metadata['scale_by'] = job_context["dataset"].scale_by
+        metadata['unsmashable_files'] = unsmashable_files
```
kurtwheeler: The user will say this, so I think we should call this property "non_aggregated_files"

PR #429, opened 2018-07-20
Adds and hopefully uses ProcessedPublicObjectsManager

Miserlou:
## Issue Number

n/a

Only return Experiments with processed Samples via Search

## Types of changes

## Functional tests

New assertions.

## Checklist

### DIFF
--- a/common/data_refinery_common/models/models.py
+++ b/common/data_refinery_common/models/models.py
@@ -205,6 +205,17 @@ class SampleAnnotation(models.Model):
         return super(SampleAnnotation, self).save(*args, **kwargs)
 
 
+class ProcessedPublicObjectsManager(models.Manager):
+    """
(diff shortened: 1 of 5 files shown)


common/data_refinery_common/models/models.py
```diff
@@ -205,6 +205,17 @@ def save(self, *args, **kwargs):
...
 
 
+class ProcessedPublicObjectsManager(models.Manager):
+    """
+    Only returns Experiments that are is_public and have related is_processed Samples.
+    """
+    def get_queryset(self):
+        return super().get_queryset().filter(
+            is_public=True, 
+            samples__is_processed=True,
```
kurtwheeler: Is this true if there exists at least one processed sample or if all samples are processed?

PR #444, opened 2018-07-26
Add a Simple DatasetStats Endpoint

Miserlou:
## Issue Number

https://github.com/AlexsLemonade/refinebio-frontend/issues/192

The Download page currently has a very costly stats table which requires fetching and tabulating statistics about a dataset. This adds an endpoint so those stats can be done on the server without fetching the whole set at once.

## Types of changes

## Functional tests
New test case added.

## Checklist

### DIFF
--- a/api/data_refinery_api/tests.py
+++ b/api/data_refinery_api/tests.py
@@ -326,7 +326,6 @@ class APITestCases(APITestCase):
                                     'organisms__name': 'Extra-Terrestrial-1982'})
         self.assertEqual(response.json()['count'], 1)
 
-
     @patch('data_refinery_common.message_queue.send_job')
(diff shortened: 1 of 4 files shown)


Willv19:
The frontend also needs an endpoint where it can get a dict of the accession codes of every sample, where the key is the name of the organism in the form 'GALLUS_GALLUS'. For example: `{'HOMO_SAPIENS': ['320230', '120301'], 'GALLUS_GALLUS': ['23133', '32132']}`

Miserlou:
What view is that for?

Willv19:
That is for viewing the list of samples. It needs a list of all of the accession codes sorted by organism so when trying to load the list of samples for a specific organism it knows which samples' information to fetch.

Issue #448, opened 2018-07-27
Create Management Command For Compiling Compendia


Miserlou:
### Context

We want to create _The Compendia_: all of the samples for every species smashed in a single gigantic matrix.

### Problem or idea

If we define a `create_compendia_ management command, we can build those datasets and dispatch the smash jobs without having it to do it manually. 

### Solution or next step

Make the thing!

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

cgreene:
Minor edit: all of the samples for each species is probably better. There are ways to map across species but they are dicey and definitely open research questions. We did hear some approaches presented at CSIG 2018, so hopefully good answers will start to come along soon!

Miserlou:
PR'd here: https://github.com/AlexsLemonade/refinebio/pull/449#event-1758115124

PR #449, opened 2018-07-27
Adds Create and Fetch Compendia Commands

Miserlou:
## Issue Number
https://github.com/AlexsLemonade/refinebio/issues/448

Adds `create_compendia` and `fetch_compendia` management commands.

## Methods

Get all experiments for each species, smash by species. Fetch when finished.

## Types of changes

## Functional tests
Tested locally, needs more data though to make sure smashers have enough RAM.

## Checklist

### DIFF
--- a/workers/run_command.sh
+++ b/workers/run_command.sh
@@ -2,6 +2,21 @@
 
 # Script for executing Django management commands within a Docker container.
 
+while getopts "hi:" opt; do
+    case $opt in
+        i)
+            image=$OPTARG
+            ;;
+        \?)
+            echo "Invalid option: -$OPTARG" >&2
+            exit 1
+            ;;
+    esac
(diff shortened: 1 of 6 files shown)


workers/run_command.sh
```diff
@@ -0,0 +1,47 @@
...
+fi
+
+source common.sh
+HOST_IP=$(get_ip_address)
+DB_HOST_IP=$(get_docker_db_ip_address)
+
+chmod -R a+rwX $volume_directory
+
+./prepare_image.sh -i smasher -s workers
+image_name=ccdlstaging/dr_smasher
```
kurtwheeler: So the rest of this script seems like it could be used for most any image, but these lines make this smasher specific. It seems like these should be more generalized.

Issue #461, opened 2018-07-31
Enable state locking for terraform remote backend


kurtwheeler:
### Context

We're now using terraform's remote backend feature to use our state directly from S3. However we aren't utilizing their state locking feature. This means that if multiple deploys try to run at the same time Bad Things :tm: could happen. However we deploy infrequently and only when someone triggers a deploy via a tag so this should be fine for the time being. I tried to set up the state locking but it just didn't seem to work and we've got more important issues to tackle first.

### Problem or idea

Long term it'd be good to utilize the state locking feature.

### Solution or next step

Set up state locking and make sure it works. I tested this by running terraform twice from the same machine and having each run do conflicting things to S3 buckets.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

Miserlou:
Is there an upstream ticket related to "it just didn't seem to work"?

kurtwheeler:
> Is there an upstream ticket related to "it just didn't seem to work"?

No, I didn't dig into it deep enough to really figure out what was going wrong. I'm assuming there was something that I did wrong. However there wasn't any output which gave me something to work on and I didn't wanna spend the time to really dig in at this point.

Miserlou:
Did #462 close this?

kurtwheeler:
No, #462 closed https://github.com/AlexsLemonade/refinebio/issues/207 and spawned this because state locking was a black box it didn't seem worth debugging at the time.

kurtwheeler:
This was closed by #1326

Issue #207, opened 2018-04-18
Use remote backend to store terraform state files


dongbohu:
### Problem or idea

Use `terraform remote config` to ease statefile maintenance. 

### Context

Right now the CircleCI deployment is using a S3 bucket to store terraform state files, but the process of downloading/decrypt/deploy/encrypt/uploading is manual and error prone. I just noticed that it is possible 
to configure a remote backend (such as AWS S3) in terraform to store the state files and the whole process can be done automatically:
https://www.terraform.io/intro/getting-started/remote.html
This will make the CircleCI deployment script `run_terraform.sh` mush simpler.
 
### Solution or next step

Use the same S3 bucket, but configure the remote backend before deployment in CircleCI.

cgreene:
Tagging for `Miso` as this may touch on stability / reliability.

PR #462, opened 2018-07-31
Adds the configuration changes necessary to use terraform's remote S3…

kurtwheeler:
… backend feature.

## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/207

Rather than manually storing the terraform state in S3, this makes it so we have terraform do this directly. This means that our script can no longer be interrupted potentially causing our terraform state to be lost. This also means that our terraform state is never even written to disk.

However I didn't get terraform's state locking feature to work. I'm not sure what was going on but it just wasn't working for me. I've created this issue: https://github.com/AlexsLemonade/refinebio/issues/461 to implement that at some point. However for reasons laid out in that issue I don't think this is important enough to work on at the moment.

## Types of changes

## Functional tests

I ran `deploy.sh -e dev` and stood up a stack for myself and the state file was only in S3 and not on my disk.

## Checklist


Miserlou:
👍 👍 👍

PR #471, opened 2018-08-01
Expose scale_by to the API

### LINKED ISSUE #469: Expose transformation option to API
### Context

https://github.com/AlexsLemonade/refinebio-frontend/issues/208

### Problem or idea

We want a dropdown to change the transformation option, but the API currently doesn't support changing that value.

### Solution or next step

I think transformation just needs to be added to the DataSetSerializer

### New Issue Checklist

Willv19:
## Issue Number

Resolves #469 

Adds `scale_by` to the list of fields in the `DataSet` serializer.

## Types of changes

## Functional tests

I ran the frontend and API locally and confirmed that these are the correct changes for the frontend issue associated with #469 

## Checklist

### DIFF
--- a/api/data_refinery_api/serializers.py
+++ b/api/data_refinery_api/serializers.py
@@ -416,6 +416,7 @@ class DatasetSerializer(serializers.ModelSerializer):
                     'id',
                     'data',
                     'aggregate_by',
+                    'scale_by',
                     'is_processing',
                     'is_processed',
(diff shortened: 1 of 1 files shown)


Willv19:
I don't think it makes since to put it in the create serializer since we don't choose an option when the dataset is created, which is why `aggregate_by` is also not in the create serializer.

Issue #466, opened 2018-07-31
Process some small experiments explicitly for "kicking the identifier conversion tires"


jaclyn-taroni:
### Context

Once https://github.com/AlexsLemonade/refinebio/pull/447 lands, we should be able to do the following identifier conversions for submitter-processed data:

* Illumina BeadArray: Probe ID to Ensembl gene ID (note: not in that PR)
* Affymetrix: Probe ID,  Entrez ID, RefSeq, Unigene, or gene symbol to Ensembl gene ID (see: https://github.com/AlexsLemonade/identifier-refinery)

### Problem or idea

@Miserlou suggested we do some analyses with data that's been processed with this pipeline before we close #186, #328, #329, and #348. 

### Solution or next step

I've picked out a few experiments for this purpose.

* Brainarray-supported Affymetrix, has raw data [`GSE12021`](https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GSE12021)
* Brainarray-supported Affymetrix, no raw data: [`GSE56409`](https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GSE56409)
* Illumina, has raw data: [`GSE44719`](https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GSE44719)
* Illumina, no raw data: [`E-GEOD-44719`](https://www.ebi.ac.uk/arrayexpress/experiments/E-GEOD-44719/) (note that this is the same experiment from above, but from ArrayExpress -- the raw data might be an "additional data" file at AE; let me know if this won't work)
* Affymetrix PM Array Plate, has raw data: [`GSE77094`](https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GSE77094)
* Affymetrix PM Array Plate, no raw data: [`E-GEOD-77094`](https://www.ebi.ac.uk/arrayexpress/experiments/E-GEOD-77094/) (similar to above, where the AE accession doesn't have raw data)

These should be run after #447 lands. They don't need to be run in production, I just need to get my hands on them.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

Miserlou:
Excellent test list, thank you Jackie. We should probably be able to take care of this for you tomorrow.

The list for our file:
```
GSE12021
GSE56409
GSE44719
[...]

PR #473, opened 2018-08-02
Fixes Two Issues Which Caused Extra Jobs of the Wrong Type

Miserlou:
## Issue Number

Discovered during: https://github.com/AlexsLemonade/refinebio/issues/466

Two issues were combining here to create superfluous `AFFY_TO_PCL` jobs being created for files that should have been `NO_OP`'d if they were to be processed at all, but shouldn't have been.

One fix is a really stupid one, the file type dispatcher wasn't using slices properly so it just wasn't working at all.

The second fix is a behavior change: we don't create OriginalFile:Sample associations if the sample is comes from a `_family` (processed) file if that Sample _also_ has raw data. Is that the correct behavior? I'm not sure, but it stops us from creating additional `NO_OP` jobs in the situation where we're already going to have an `AFFY_TO_PCL` anyway. 

We might end up with an extra file lying around though.

## Types of changes

## Functional tests

Tried locally with `GSE12021`

## Checklist

### DIFF
--- a/common/data_refinery_common/job_lookup.py
+++ b/common/data_refinery_common/job_lookup.py
@@ -124,9 +124,9 @@ def determine_processor_pipeline(sample_object: Sample, original_file=None) -> P
 
     # Optional explicit filetype checks
     if original_file:
-        if original_file.filename[-3].upper() == "CEL":
+        if original_file.filename[-3:].upper() == "CEL":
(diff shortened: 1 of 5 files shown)


Miserlou:
The other change is that we don't collapse entirely when failing to dispatch a Nomad job, we just log the error and move on. The foreman should pick them all up later, so the job is a success otherwise.

PR #476, opened 2018-08-02
Moves from HStoreField to JSONField

Miserlou:
## Issue Number

n/a

We were using HStoreField rather than JSONField for storing our remotely source annotation data. This changes it to use JSONField, so the data is more natural when served through the API. We don't query on these fields anyway so I don't think this will have any negative impact in that department.

## Types of changes

May break frontend! Don't think so though.

## Functional tests
Tested locally with real GSE.

## Checklist


Miserlou:
I've tested the API locally and it seems fine, but I don't know if it will break the frontend or not. I don't thinkkkkk so, and I wanted to get this change in before Will(?) works on https://github.com/AlexsLemonade/refinebio-frontend/issues/215

Issue #361, opened 2018-06-29
Populate Processor table dynamically 


dongbohu:
### Context

According to this conversation:
https://github.com/AlexsLemonade/refinebio/issues/322#issuecomment-400808316

We need to create a YAML config file for each processor (`Salmon Quant`, `Salmontools`, `Tximport`, etc), then inside each processor's corresponding docker container, query versions of programs/packages listed in this config file, and populate these information into `Processor` table in the database.

### Problem or idea

The population process should be done **after** docker images have been updated by the deployment process and Terraform has created RDS instance.

### Solution or next step

(1) Create a YAML file for each processor;
(2) Write a Django management command to read the YAML file and populate the records;
(3) Call the management command inside Terraform after RDS has been created.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

dongbohu:
Here is the structure of the YAML file for each processor that I have been thinking about:

```
---

# Operating system distribution (such as Ubuntu 16.04, CentOS 7.2, etc)
os_distribution:

# OS-level packages (retrieved by "apt-query")
os_pkg:
  - python
  - r-base
  - java
  - g++
  - etc ...

# Command line programs (retrieved by "<command> --version")
cmd_line:
  - salmon
  - samontools
  - etc ...

# Python packages (retrieved by "pip")
python:
  - Django
  - request
  - etc ...

# R packages (retrieved by R command)
R:
  - tximport
  - etc ...

# Any other package/program category? 
```

@cgreene and @jaclyn-taroni: Please let me know whether it makes sense to you. Thanks.

Miserlou:
@dongbohu - You are saying `retrieved by "pip"` - does that mean that these files will be created dynamically before being loaded in? Or vice versa? This seems like we are significantly increasing the amount of dependencies we have to maintain without any system for checking that they're in sync.

dongbohu:
What I meant is, when the database server and docker container are both up, a management command will be launched inside the docker container to read in this YAML file, retrieve the version information for each processor, build a JSON object, and update the processor records in the database (if necessary).

dongbohu:
Forgot to mention: The YAML files are static.

Miserlou:
[...]

PR #481, opened 2018-08-02
Update Processor Info Dynamically

dongbohu:
## Issue Number
#361 

This PR includes functions to read a YAML file into database and a bunch of YAML files, each of which corresponds to a processor. The structure of YAML file is based on the discussion in #361.

## Types of changes

## Functional tests

I have tested parsing functions and YAML files in corresponding docker images.

## Checklist


dongbohu:
@jaclyn-taroni The packages listed in each YAML file is based on your comments at https://github.com/AlexsLemonade/refinebio/issues/322#issuecomment-400860143
I also reviewed the source code of all processors and Dockerfiles.  Having too many dependencies may be distracting, so I tried to list the packages that are most useful to users (instead of to developers).

workers/data_refinery_workers/processors/multiqc.yml
```diff
@@ -0,0 +1,22 @@
...
+os_pkg:
+  - python3
+  - python3-pip
+  - default-jre   # required by fastqc
+
+
+# Command lines (assuming "<cmd> --version" is supported)
+cmd_line:
+  - salmon
+  - /home/user/FastQC/fastqc
```
dongbohu: I confess that this is a little awkward. I tried to create a soft link in `/usr/local/bin`, but the command complains that it has two many levels of symbolic links.

kurtwheeler:
I was under the impression that we wanted these processor records checked at the start of every job. We use django management commands to run all of our jobs in production. Therefore I would think we would want that code to be run at the start of every management command used to run jobs via nomad, rather than just at deploy time via its own management command.

workers/data_refinery_workers/processors/no_op.yml
```diff
@@ -0,0 +1,30 @@
+---  # "Submitter-processed" processor
```
jaclyn-taroni: Because the gene identifier conversion for submitter-processed Affymetrix data is done with mappings we've put together in [AlexsLemonade/identifier-refinery](https://github.com/AlexsLemonade/identifier-refinery) and we get those mappings from Zenodo when we build this image: https://github.com/AlexsLemonade/refinebio/blob/3939050215468cf517680119f2c521d127de6fc5/workers/dockerfiles/Dockerfile.no_op#L43

Is it possible to keep track of the version of the Zenodo record?

dongbohu:
@kurtwheeler As far as I understand, `Processor` model is not associated with any job. Instead, it is supposed to keep track of the information of a certain processor that is being used to generate a certain ComputationalResult (as you may have seen, `ComputationalResult` model has a foreign key to `Processor`). This is why I think the `Processor` table should be populated once and only once at the deployment time right before any data processing jobs. Does this make sense?

cgreene:
I don't want to dive into the specifics but I want our solution to:
* Track the association of a processor to dependencies (the detailed ones - i.e., docker; and the less detailed ones listed in this YAML file).
* Not be able to change the information associated with a processor once it's created (no "UPDATE" operation), so our provenance doesn't get out of whack.
* Tie a job to the specific processor used to process it that gets to the docker and YAML-file specified dependency levels.

Is this aligned with what you are both thinking @kurtwheeler and @dongbohu?

dongbohu:
@cgreene: Right now the processor is associated with each `ComputationalResult`, not with each job. I think it makes more sense, because each result is generated by a certain processor. The `ProcessorJob` is indirectly associated with the processor by `ProcessorJobOriginalFileAssociation`.

@kurtwheeler: I see your point on why you want to generate new `Processor` record at the beginning of each job. One question: When a new job is launched, is it possible to get the docker image based on which the current container was created? (I noticed that nomad-specs do keep track of the docker image before starting the container. But inside the container I am not sure whether the docker image is saved as a environment variable.)

cgreene:
ComputationalResult seems better than job. Thanks.

I don't understand why we want to generate a new processor record. Wouldn't we just want to check it at most?

kurtwheeler:
> I don't understand why we want to generate a new processor record. Wouldn't we just want to check it at most?

We would only create a new one if we checked it and it was wrong. I guess the other thing we could do is fail loudly.

dongbohu:
@cgreene and @kurtwheeler: When you guys say **check**, do you mean checking only the processor name, version and docker image name, or checking these three PLUS the versions of packages listed in the YAML file?

dongbohu:
If the docker container includes the corresponding docker image name, I think we can only check processor name, version and the docker image name. That will save some time.

kurtwheeler:
> When a new job is launched, is it possible to get the docker image based on which the current container was created?

I haven't tried this but it looks like this might do it: https://groups.google.com/forum/#!topic/docker-user/qhfJzJD72fs

> When you guys say check, do you mean checking only the processor name, version and docker image name, or checking these three PLUS the versions of packages listed in the YAML file?

So unfortunately docker images names and versions can be overwritten, so one Bad Thing that could happen is we push a new image to the same `repo/name:tag` in dockerhub, overwriting the docker image with an image containing new/different dependencies. I know we have code in our deploy process to prevent this, but bugs happen and making sure what we ran is very important.

However docker images also have IDs, so if we also stored that then we could use that to make sure its actually the same image so we didn't have to check all the dependency info.

dongbohu:
@kurtwheeler Thanks. I will play with the trick.

cgreene:
@kurtwheeler : does there _need_ to be a change to accomplish our goals? That's what I still don't understand from the conversation. I realize there is a different way to do it, but is there a _why_ to doing it that way?

dongbohu:
@kurtwheeler: I talked to @cgreene briefly about the best place to create `Processor` records. Since we have the full control of the deployment process, how likely you think it is that we create `Processor` records at deployment time but it messes up during data processing time? Unless the job is run w/o nomad, it shouldn't happen, right?

kurtwheeler:
> Unless the job is run w/o nomad, it shouldn't happen, right?

This is true.

> but is there a _why_ to doing it that way

I gave a reason for the why:

> one Bad Thing that could happen is we push a new image to the same repo/name:tag in dockerhub, overwriting the docker image with an image containing new/different dependencies. I know we have code in our deploy process to prevent this, but bugs happen and making sure what we ran is very important.

I'll agree that as long as we're careful and we don't introduce any bugs in our deployment process we _shouldn't_ end up in a state where we're running a different docker image than we think we are. I just thought that:
* Doing it at run time rather than deploy time would be safer, and provenance is something we tend to try to be very safe about.
* We had originally planned out this process to happen at run time (@Miserlou had the same impression as I did about when this would be run)
* We already have to make a change to run this at deploy time:
> I will file a followup PR to launch this management command during deployment.

If we do a this with a `check and only update if something is out of date` at runtime then we don't need to do this at deploy time at all, so instead of having to make an additional change we're making a different change.

However if we want to de-prioritize the runtime check then that's totally something we can do. I just thought we should discuss it since it's different than what was originally planned.

cgreene:
I lacked this context:
> If we do a this with a check and only update if something is out of date at runtime then we don't need to do this at deploy time at all, so instead of having to make an additional change we're making a different change.

I should specify that my goal with asking was to prevent additional work.

kurtwheeler:
OK, well then I should probably also note that while I'm not 100% sure about the amount of work involved in either change, I would imagine that the one I proposed is probably more work than just running an additional set of management commands at deploy time. However @dongbohu could probably give a better estimate what the difference between the two implementation times would be.

cgreene:
Also, I agree this is important, but I needed it laid out this way to understand how the proposed versions met the objectives.
> Doing it at run time rather than deploy time would be safer, and provenance is something we tend to try to be very safe about.

dongbohu:
It is easier to do it at deployment time: We just need to create a new task to launch the management in each nomad-spec file. 

On the other hand, it is not that bad to change the step inside each job. Just move the functions from management command to utils.py and call them inside each job. 

I put more thoughts and agree that it is much more reliable to do it at runtime. This piece of information is too important to get wrong. And if it did get wrong, it would be hard to detect from when it did.

Let me update the code as soon as possible.

workers/data_refinery_workers/processors/smasher.yml
```diff
@@ -0,0 +1,18 @@
...
+# OS-level packages (installed by OS package manager)
+os_pkg:
+  - python3
+  - python3-pip
+
+# Python pip packages
+python:
+  - Django
+  - data-refinery-common
+  - matplotlib
```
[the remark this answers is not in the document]
dongbohu: I added it because `matplotlib` was listed in `processors/requirements.in` and imported by `test_smasher.py`. But I just realized that it was not being used at all. I will remove it from all files that include this package.

workers/data_refinery_workers/processors/affymetrix.yml
```diff
@@ -0,0 +1,29 @@
...
+  - Django
+  - data-refinery-common
+  - rpy2
+
+# User-installed R packages
+R:
+  - affy
+  - affyio
+  - AnnotationDbi
+  - Biobase
```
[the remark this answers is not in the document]
dongbohu: I will remove it if you think it is unnecessary.

workers/data_refinery_workers/processors/illumina.yml
```diff
@@ -0,0 +1,32 @@
+---  # "Illumina SCAN" processor
```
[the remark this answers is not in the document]
dongbohu: Are you saying to keep track of the commit hash of `processors/illumina.R`?

workers/data_refinery_workers/processors/no_op.yml
dongbohu: So in this case you want to keep track of `1322711`?

workers/data_refinery_workers/processors/illumina.yml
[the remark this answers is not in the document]
jaclyn-taroni: Yes

workers/data_refinery_workers/processors/no_op.yml
cgreene: I think you'd want to do that with a `cmd_line` specification. Same with the git hash tracking above.

dongbohu: @cgreene: I talked to @jaclyn-taroni about the best way to retrieve the version `1322711`. The best approach I can think of is to read `Dockerfile.no_op` and parse that `curl ...` line. Is there any other way?

cgreene: Can you make the Dockerfile have that as a settable variable and also dump that variable to a tiny file, which you could then read easily?

cgreene: Also, what's up with the digits around the zip file? Can they be known in advance or calculated.

Miserlou: It's a UNIX timestamp of the file creation time, so `1532646318` is `Thursday, July 26, 2018 11:05:18 PM (UTC)`. This number increases with updated releases.

dongbohu: If we want to keep track of the digits in zip file as well, we probably should create a field such as `zenodo_url` in YAML file to record the whole URL.

cgreene: It seems like "what version should we use" for the no_op processor should be a variable. If there isn't a way to make it a variable because of the unix timestamp, then I think that URL should be stored somewhere and read from both places. As it is, it's sort of a magic number in the Dockerfile.

dongbohu: If it is a variable, it will have to be defined in some env file and passed to `docker build --envfile <filename> ...`. So it will depend on the docker building command, and we can not guarantee whoever builds the docker image will remember this option. I think it is better to have a static one-line `no_op_zenodo_url.txt` file, which will be read by both Dockerfile.no_op and the program that dynamically collects the version info.

cgreene: Works for me 👍

cgreene: check with @Miserlou for thoughts before implementing. Tagging so he gets a ping

Miserlou: @dongbohu IRL discussion w/ @cgreene  has come up with an even ~~~lazier~~~ shorter path such that this could be done inside the dockerfile itself (ex:)

```
RUN ID_REFINERY_URL=https://zenodo.org/record/1322711/files/all_1532646318.zip
RUN curl -O $ID_REFINERY_URL
RUN echo $ID_REFINERY_URL > id_refinery_url.txt
```

dongbohu: This looks good. I think it will be better to save `id_refinery_url.txt` using absolute path, something like:
```RUN echo $ID_REFINERY_URL > /etc/id_refinery_url.txt```

File: workers/data_refinery_workers/processors/illumina.yml (hunk 5 above)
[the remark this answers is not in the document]
Comment (jaclyn-taroni) [reply workers/data_refinery_workers/processors/illumina.yml]: In my mind, this is a unique case because it is an implementation of what is in the `SCAN.UPC` package but for a different platform and it is not yet included in the package itself. It's the only spot that I can think of where a script is doing the heavy lifting of data processing rather than just calling functions from a package. That being said, I'm not dead set on this if it turns out to be too much of a hassle.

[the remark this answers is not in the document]
Comment (dongbohu) [reply workers/data_refinery_workers/processors/illumina.yml]: I wrote a function to get its md5sum for this file. Stay tuned for the updated PR. (May not be able to finish it by today due to some unexpected complexities though.)

**Comment by dongbohu** [comment]
Code not ready for review yet. I need to run some tests.

**Comment by dongbohu** [comment]
@cgreene @jaclyn-taroni @kurtwheeler @Miserlou This PR is ready for review. Thanks.

**Comment by cgreene** [comment]
The questions that I had have been addressed. 👍

File: workers/data_refinery_workers/processors/no_op.yml [hunk 6]
```diff
@@ -0,0 +1,34 @@
...
+  - illuminaHumanv2.db
+  - illuminaHumanv3.db
+  - illuminaHumanv4.db
+  - illuminaMousev1.db
+  - illuminaMousev1p1.db
+  - illuminaMousev2.db
+  - illuminaRatv1.db
+
+# Any other stuff
+other:
```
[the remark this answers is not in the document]
cgreene: For what it's worth, I can see why you did this. I wonder if it really makes sense to write code for each edge case instead of grabbing `stdout` by running `cat` (or allowing another arbitrary command line - this could be a superset of the things above where you run --version - could that grab arbitrary output and then you specify the version command then).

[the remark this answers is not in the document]
dongbohu: In the case of `zenodo_url`, we just need to grab `cat <filename>` output. What if there are other information that we want to keep track of but it is not saved in a file? In those cases we may need to launch some other commands. I am using `cmd_line` to specify any user-installed programs that are launched by a command line, and `other` for anything else.

Does this make sense?

dongbohu:
Just talked with @cgreene about the implementation of parsing `cmd_line` section. I am going to rewrite that part. So please stay tuned.

workers/dockerfiles/Dockerfile.no_op
```diff
@@ -42,7 +42,9 @@ COPY workers/install_gene_convert.R .
 RUN Rscript install_gene_convert.R
 RUN mkdir -p gene_indexes
 WORKDIR /home/user/gene_indexes
-RUN curl -O https://zenodo.org/record/1327563/files/all_1533307773.zip
+ENV ID_REFINERY_URL https://zenodo.org/record/1322711/files/all_1532646318.zip
+RUN curl -O $ID_REFINERY_URL
+RUN echo $ID_REFINERY_URL > /etc/id_refinery_url.txt
```
cgreene: Chatted briefly with @dongbohu. Since we're going to make the parsing much simpler, the commands need to be something that is at least googleable. If this filename was `/etc/identifier_refinery_url` it may be a little bit friendlier. Then a user can google for "identifier refinery".

dongbohu:
@cgreene @jaclyn-taroni @kurtwheeler @Miserlou: Back online for review, thanks.

dongbohu:
@jaclyn-taroni I think I have addressed all comments that you posted on Friday. Please note that the `cmd_line` section is different now. It lists the EXACT commands that will be executed inside the docker container.

jaclyn-taroni:
Yes @dongbohu, what we're keeping track of LGTM 👍

workers/data_refinery_workers/processors/agilent_twocolor.py
```diff
@@ -106,8 +106,15 @@ def _create_result_objects(job_context: Dict) -> Dict:
     result.time_start = job_context['time_start']
     result.time_end = job_context['time_end']
     result.pipeline = "Agilent SCAN TwoColor"  # TODO: should be removed
-    # result.processor = Processor.objects.get(name=utils.ProcessorEnum.AGILENT_TWOCOLOR.value,
-    #                                          version = __version__)
+    try:
+        result.processor = utils.find_processor("AGILENT_TWOCOLOR")
+    except Exception as e:
+        err_str = "Failed to set processor: %s" % e
+        logger.error(err_str)
```
kurtwheeler: Could you add some additional context to this log message. Maybe like `job_id=job_context["job"].id, processor="AGILENT_TWOCOLOR"`?

workers/data_refinery_workers/processors/array_express.py
```diff
@@ -160,8 +160,15 @@ def _create_result_objects(job_context: Dict) -> Dict:
     result.time_start = job_context['time_start']
     result.time_end = job_context['time_end']
     result.pipeline = "Affymetrix SCAN"  # TODO: should be removed
-    # result.processor = Processor.objects.get(name=utils.ProcessorEnum.AFFYMETRIX_SCAN.value,
-    #                                          version=__version__)
+    try:
+        result.processor = utils.find_processor("AFFYMETRIX_SCAN")
+    except Exception as e:
+        err_str = "Failed to set processor: %s" % e
+        logger.error(err_str)
```
kurtwheeler: Same comment about additional context for this logger call and the other ones just like it.

workers/data_refinery_workers/processors/test_salmon.py
```diff
@@ -41,7 +37,8 @@ def prepare_job():
     samp.organism = c_elegans
     samp.save()
 
-    computational_result = ComputationalResult(processor=Processor.objects.first())
+    #computational_result = ComputationalResult(processor=Processor.objects.first())
```
kurtwheeler: Looks like we don't need this dead line anymore.

dongbohu:
@kurtwheeler Thank you for your comments. I did test the runtime proncessor environment manually inside each docker container. Let me add some test cases.

dongbohu:
@kurtwheeler Here are the main changes in today's update on this PR:

* Added some test runtime processor test cases in `test_salmon.py`. The test code for each processor look similar. I did it on purpose so that it will be easier to customize these tests in the future. 

* Added a function in `utils` to handle the exception raised by `util.find_processor`. The new function bundles job_id and processor name in the log message too.

* Changed some of the new functions' return values from `list` to `dict` type to make the JSON `environment` variable simpler.

* Polished some comments and removed some that are obsolete.

I hope I have addressed all your concerns. Please let me know if you have any other thoughts.

kurtwheeler:
@dongbohu This is looking pretty solid. I'll take a final look once the tests are passing again.

dongbohu:
I just committed a few more minor changes to synchronize the `Processor` model with the serializer and API test case.

PR #482, opened 2018-08-03
Fixes uploading to s3, improves resiliency of Foreman, and fixes a bl…

kurtwheeler:
…ock of code being a loop that shouldn't have.

## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/466

While processing a crunch in a dev stack for #466 some issues were identified. This should address three of them:
* Files weren't being uploaded to S3 because of an inverted boolean check.
* Some errors from the Nomad client caused Foreman threads to crash
* A block of code got indented so it was being run in a loop when it didn't need to.

## Types of changes

## Functional tests

This will be tested when we go to rerun the Jackie Crunch with the bug fixes we've made thus far.

## Checklist

### DIFF
--- a/common/data_refinery_common/models/models.py
+++ b/common/data_refinery_common/models/models.py
@@ -216,8 +216,8 @@ class ProcessedPublicObjectsManager(models.Manager):
     """
     def get_queryset(self):
         return super().get_queryset().filter(
-            is_public=True, 
-            samples__is_processed=True, 
+            is_public=True,
+            samples__is_processed=True,
             samples__is_public=True).distinct()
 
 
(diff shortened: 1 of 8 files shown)


common/data_refinery_common/models/models.py
```diff
@@ -577,7 +577,7 @@ def save(self, *args, **kwargs):
     def sync_to_s3(self, s3_bucket=None, s3_key=None) -> bool:
         """ Syncs this OriginalFile to AWS S3.
         """
-        if settings.RUNNING_IN_CLOUD:
+        if not settings.RUNNING_IN_CLOUD:
```
Miserlou: _Nice._

kurtwheeler:
@Miserlou I had to make some additional changes to fix the tests. While I was at it I noticed a bunch of dead code to remove. Wanna take a look at all that?

PR #484, opened 2018-08-03
Add sample_metadata to experiments endpoint

Willv19:
## Issue Number

https://github.com/AlexsLemonade/refinebio-frontend/issues/211

This PR adds a field to the ExperimentSerializer that gives all of the metadata fields that at least one associated Sample has. This is so that this can be displayed on the frontend.

## Types of changes

## Functional tests

I ran the API locally to make sure that the output was correct.

## Checklist

### DIFF
--- a/common/data_refinery_common/models/models.py
+++ b/common/data_refinery_common/models/models.py
@@ -310,9 +310,15 @@ class Experiment(models.Model):
         See https://github.com/AlexsLemonade/refinebio-frontend/issues/211 for why this is needed.
         """
         fields = []
-        for sample in self.samples.all():
-            metadata = sample.to_metadata_dict()
-            fields += [k for k, v in metadata.items() if v and k not in fields]
+
+        possible_fields = ['title', 'accession_code', 'sex', 'age', 'specimen_part', 'genotype',
+                           'disease', 'disease_stage', 'cell_line', 'treatment', 'race', 'subject',
+                           'compound', 'time']
+
+        for field in possible_fields:
+            filter = {"age__isnull": True} if field == 'age' else {'%s__exact' % field: ''}
+            if len(self.samples.exclude(**filter)) > 0:
+                fields.append(field)
         return fields
 
     @property
--- a/common/data_refinery_common/models/models.py
+++ b/common/data_refinery_common/models/models.py
@@ -127,7 +127,7 @@ class Sample(models.Model):
         metadata = {}
         metadata['title'] = self.title
         metadata['accession_code'] = self.accession_code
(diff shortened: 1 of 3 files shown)


common/data_refinery_common/models/models.py
```diff
@@ -305,6 +305,16 @@ def to_metadata_dict(self):
...
         return metadata
 
+    def get_sample_metadata_fields(self):
+        """ Get all metadata fields that are non-empty for at least one sample in the experiment.
+        See https://github.com/AlexsLemonade/refinebio-frontend/issues/211 for why this is needed.
+        """
+        fields = []
+        for sample in self.samples.all():
+            metadata = sample.to_metadata_dict()
+            fields += [k for k, v in metadata.items() if v and k not in fields]
```
kurtwheeler: > if v and k not in fields

Are you sure we should be checking for both v and k? Seems like that would cause the key to be added once for every different value it has.

Willv19: `not in` has higher precedence than `and`, so with parenthesis the comparison would be `if (v) and (k not in fields)`, which is correct. It's checking if `v` is truthy and `k` is not already in `fields`.

Miserlou: I'm also not crazy about this bit, surely this should be doable entirely at the query level so we don't need to loop at all?

kurtwheeler: For my part:

Duh, of course that's how that logic is working... Sorry 'bout that.

common/data_refinery_common/models/models.py
```diff
@@ -305,6 +305,22 @@ def to_metadata_dict(self):
 
         return metadata
 
+    def get_sample_metadata_fields(self):
+        """ Get all metadata fields that are non-empty for at least one sample in the experiment.
+        See https://github.com/AlexsLemonade/refinebio-frontend/issues/211 for why this is needed.
+        """
+        fields = []
+
+        possible_fields = ['title', 'accession_code', 'sex', 'age', 'specimen_part', 'genotype',
```
Miserlou: rm `'title', 'accession_code'`

common/data_refinery_common/models/models.py
```diff
@@ -305,6 +305,22 @@ def to_metadata_dict(self):
...
+                           'disease', 'disease_stage', 'cell_line', 'treatment', 'race', 'subject',
+                           'compound', 'time']
+
+        for field in possible_fields:
```
Miserlou: We can probably do this in a single mega-query..

Also, needs a robust test!

PR #485, opened 2018-08-03
De-duplicate Rows from NO_OP Processor

Miserlou:
## Issue Number

Jackiecrunch.

Some of the files output by the NO_OP processor were significantly larger than they should have been. This deduplicates those lines (although it doesn't adequately explain _why_ those were duplicated in the first place.)

## Types of changes

## Functional tests

Tested locally, and a new assertion is added to make sure that the file is the correct size.

## Checklist

### DIFF
--- a/workers/data_refinery_workers/processors/test_no_op.py
+++ b/workers/data_refinery_workers/processors/test_no_op.py
@@ -104,7 +104,7 @@ class NOOPTestCase(TestCase):
         final_context = no_op.no_op_processor(job.pk)
         self.assertTrue(final_context['success'])
         self.assertTrue(os.path.exists(final_context['output_file_path']))
-        self.assertEqual(os.path.getsize(final_context['output_file_path']), 346325)
+        self.assertEqual(os.path.getsize(final_context['output_file_path']), 346535)
 
     @tag('no_op')
     def test_convert_processed_illumina(self):
--- a/workers/data_refinery_workers/processors/test_no_op.py
+++ b/workers/data_refinery_workers/processors/test_no_op.py
@@ -103,6 +103,8 @@ class NOOPTestCase(TestCase):
 
         final_context = no_op.no_op_processor(job.pk)
         self.assertTrue(final_context['success'])
+        self.assertTrue(os.path.exists(final_context['output_file_path']))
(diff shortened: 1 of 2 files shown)


jaclyn-taroni:
Did this happen in all Affymetrix files? I'm wondering if it is a particular identifier that's causing the problem.

workers/data_refinery_workers/processors/test_no_op.py
```diff
@@ -103,6 +103,8 @@ def test_convert_simple_pcl(self):
 
         final_context = no_op.no_op_processor(job.pk)
         self.assertTrue(final_context['success'])
+        self.assertTrue(os.path.exists(final_context['output_file_path']))
+        self.assertEqual(os.path.getsize(final_context['output_file_path']), 346325)
```
kurtwheeler: Is it really this deterministic?!? That's pretty cool.

kurtwheeler:
I'll note that the line of R that changed I do not understand. @jaclyn-taroni should make sure it's legit.

jaclyn-taroni:
I think I know _why_. If the detected identifier (`detected_id`) is not RefSeq (possibly even if it is), there's likely to be lots of duplicate rows. See the below image from [AlexsLemonade/identifier-refinery](https://github.com/AlexsLemonade/identifier-refinery)

68747470733a2f2f692e696d6775722e636f6d2f4770685572326d2e706e67

So another way to accomplish this is probably adding `%>% dplyr::distinct()` either after [this line](https://github.com/AlexsLemonade/refinebio/pull/485/files#diff-89bb3c747fc413bb91fbbc9cb8f52a04R85) or the one after it (86).

jaclyn-taroni:
`distinct` docs: https://dplyr.tidyverse.org/reference/distinct.html

cgreene:
@jaclyn-taroni good catch! Adding the distinct check early seems good, and then I think the current line 97 can be removed too:
https://github.com/AlexsLemonade/refinebio/pull/485/files#diff-89bb3c747fc413bb91fbbc9cb8f52a04R97

jaclyn-taroni:
If this gives us the expected unduplicated output, it LGTM

PR #487, opened 2018-08-03
Improvements for Different Illumina Headers in NO_OP

Miserlou:
## Issue Number

Jackiecrunch.

Not all of Illumina files have the same header. This attempts to normalize them so that they can succesfully gene convert.

This doesn't support _all_ of the scenarios we've encountered (particularly, the one with lots of samples in a single file where the sample titles don't match the metadata), but does improve the amount we can support. 

## Types of changes

## Functional tests

New test file and tests added. More assertions for existing tests.

## Checklist

### DIFF
--- a/workers/data_refinery_workers/processors/no_op.py
+++ b/workers/data_refinery_workers/processors/no_op.py
@@ -73,7 +73,6 @@ def _prepare_files(job_context: Dict) -> Dict:
                 with open(job_context["input_file_path"], 'w+', encoding='utf-8') as f:
                     f.seek(0, 0)
                     f.write('ID_REF\tVALUE' + '\n' + ("\n".join(all_content[1:])))
-                job_context["input_file_path"] = job_context["input_file_path"] + ".fixed"
             except Exception as e:
                 logger.exception("Unable to read input file or header row.",
                     input_file_path=job_context["input_file_path"])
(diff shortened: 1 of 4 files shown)


workers/data_refinery_workers/processors/no_op.py
```diff
@@ -82,7 +81,35 @@ def _prepare_files(job_context: Dict) -> Dict:
...
+
+                    # These exist, but they're too hard for us to handle right now.
+                    if len(row) > 3:
+                        e_msg = "We found an Illumina file to NO_OP that we're not set up to process yet. Tell Rich!"
+                        logger.error(e_msg,
+                            job_id = job_context["job"].pk,
+                            file=job_context["input_file_path"]
+                        )
+                        job_context['success'] = False
+                        job_context["job"].failure_reason = str(e)
```
kurtwheeler: I think `e` is undefined. You probably meant `e_msg` instead of `str(e)`

workers/data_refinery_workers/processors/no_op.py
```diff
@@ -82,7 +81,35 @@ def _prepare_files(job_context: Dict) -> Dict:
...
+                    with open(job_context["input_file_path"], 'w+', encoding='utf-8') as f:
+                        f.seek(0, 0)
+                        if len(row) == 2:
+                            f.write('Reporter Identifier\tVALUE' + '\n' + all_content)
+                        elif len(row) == 3:
+                            f.write('Reporter Identifier\tVALUE\tDetection Pval' + '\n' + all_content)
+
+                except ValueError:
+                    # Okay, there's a header column, we're good.
+                    job_context['column_name'] = row[0]
```
kurtwheeler: If there's no header, `job_context['column_name']` won't be set at all. Now, I know this will still work because in https://github.com/AlexsLemonade/refinebio/pull/487/files#diff-094bf86f6e84a5a7a3b57705efa7ad62R208 you make the default value to `get` the same as what you make the column name in the `try` block here: `Reporter Identifier`. However it seems like it could be a bit brittle like this in a scenario where the value changed in one place but not the other. Could you either explicitly set `job_context['column_name']` with the `try` block or create a constant that you use rather than the literal `Reporter Identifier`?

PR #494, opened 2018-08-06
Change database instance type, change ASG thresholds, and remove pers…

kurtwheeler:
…istent client instance.

## Issue Number

N/A

Having a persistent client instance is expensive, so we wanna nix that. However doing so means that we won't have anything to run a surveyor job which will spawn more jobs, pushing us to the point of scaling up. Therefore the thresholds needed to change in order to support scaling up based on a single job.

Also our database instance was no where near large enough so that is getting bumped up too.

## Types of changes

## Functional tests

Can't really functional test infra changes.

## Checklist


infrastructure/variables.tf
```diff
@@ -116,11 +116,11 @@ variable "max_clients" {
 }
 
 variable "scale_up_threshold" {
-  default = "40"
+  default = "1"
 }
 
 variable "scale_down_threshold" {
-  default = "10"
+  default = "0"
```
Miserlou: I think this will require a corresponding change to `nomad_queue_length_alarm_down`:

```git 
- comparison_operator = "LessThanThreshold"
+ comparison_operator = "LessThanOrEqualToThreshold"
```

Issue #451, opened 2018-07-30
We need a mechanism to retry previously failed experiments/samples


kurtwheeler:
### Context

The Foreman currently will retry jobs up to a certain number of times. After that number, it then assumes that the job is failing for a non-transient reason and gives up. The job may have failed because the data or metadata were mangled or wrong, but it also could fail because of a bug.

### Problem or idea

If there is a bug which causes one or more jobs to fail which is then fixed, we need a mechanism to retry the affected experiments/samples.

Currently there is a hacky workaround to get jobs to be retried: change the `num_retries` field for the last job that was tried to 1. This causes the Foreman to think the job hasn't been retried enough and it queues it up once again. This isn't a great solution for a few reasons:
  * It's very manual and requires direct manipulation of the database tables
  * It only works at a job level. If you want to retry experiment X you need to determine the samples for the experiment, then the files for the samples, then find which jobs were associated with those files.
  * The job is only retried once so if there's a transient error it will fail without trying again.

### Solution or next step

Create a Django management command to queue fresh jobs based on Experiment Accessions, Sample Accessions, or Job IDs. Ideally these could be specified via a file similar to the way the  `survey_all` management command works.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

kurtwheeler:
So I've thought about this a bit more and I've decided that two separate mechanisms would probably be best:
  * One to delete an experiment and anything associated with it. (Which means then we can just re-run a surveyor job for that experiment)
  * One to retry a downloader or processor job which is no longer being retried (either because it hit the limit or because of https://github.com/AlexsLemonade/refinebio/issues/431)

Miserlou:
> One to delete an experiment and anything associated with it

I'm not crazy about this. If an experiment fails to scrape, but it contains samples that are used by another experiment which did scrape.. there might be some problems.

kurtwheeler:
Hmm, that's true. However when a surveyor job fails, there may be undiscovered samples associated with it, but yet the job won't rerun because the experiment already exists. Perhaps rather than `anything associated with it` we could delete `anything associated with it and not anything else`?

Issue #452, opened 2018-07-30
Audit our data cleanup/management practices throughout the project


kurtwheeler:
### Context

https://github.com/AlexsLemonade/refinebio/pull/386 will move processed files to S3 and cleans up the raw files that were direct inputs to the process. However if we download any supplemental or extraneous files or create any temporary files they may just end up sitting on disk in EFS, which is expensive.

### Problem or idea

Look through the various surveyors, downloaders, and processors we have and determine if they are downloading or creating any temporary files which should be cleaned up.

### Solution or next step

Once files have been identified that should be cleaned up but are not being cleaned up, add code to delete them.

A potential test to verify that data is being cleaned properly would be run some jobs and see that the EFS volume is empty after everything completes: all of the processed data should be in S3 and everything else should be deleted.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

Miserlou:
Related upstream ticket: https://github.com/guma44/GEOparse/issues/41

jaclyn-taroni:
@Miserlou: may shake out during collaborative code review

jaclyn-taroni:
@kurtwheeler: may not be required before 1% crunch

jaclyn-taroni:
Would you say this can be closed now that you've gone through the majority of the code base @Miserlou @kurtwheeler @dongbohu ?

Miserlou:
Yep we're good now

PR #513, opened 2018-08-09
Kurtwheeler/unsurveyor

kurtwheeler:
## Issue Number

#451 (although only the half related to being able to resurvey an experiment)

When we have a surveyor bug that we fix and so we want to rerun the affected experiments, we don't have a good way of doing so. This adds the ability to "unsurvey" an experiment, so that a new survey job can be run.

## Types of changes

## Functional tests

I surveyed, downloaded, and processed an experiment locally. I then ran the unsurveyor and verified that all the tables were emptied.

## Checklist

### DIFF
--- a/common/data_refinery_common/models/models.py
+++ b/common/data_refinery_common/models/models.py
@@ -784,16 +784,29 @@ class ComputedFile(models.Model):
         self.size_in_bytes = os.path.getsize(self.absolute_file_path)
         return self.size_in_bytes
 
-    def delete_local_file(self):
+    def delete_local_file(self, force=False):
         """ Deletes a file from the path and actually removes it from the file system,
         resetting the is_downloaded flag to false. Can be refetched from source if needed. """
-        if not settings.RUNNING_IN_CLOUD:
+        if not settings.RUNNING_IN_CLOUD and not force:
             return
 
         try:
             os.remove(self.absolute_file_path)
         except OSError:
             pass
+
+    def delete_s3_file(self, force=False):
+        if not settings.RUNNING_IN_CLOUD and not force:
+            return False
+
+        try:
+            S3.delete_object(self.s3_bucket, self.s3_key)
+            return True
+        except:
+            logger.exception("Failed to delete S3 object for Computed File.",
+                             "computed_file"=self.id,
+                             "s3_object"=self.s3_key)
+            return False
 
     def get_synced_file_path(self, force=False):
         """ Fetches the absolute file path to this ComputedFile, fetching from S3 if it
--- a/common/data_refinery_common/models/models.py
+++ b/common/data_refinery_common/models/models.py
@@ -796,6 +796,8 @@ class ComputedFile(models.Model):
             pass
 
     def delete_s3_file(self, force=False):
+        # If we're not running in the cloud then we shouldn't try to
+        # delete something from S3 unless force is set.
         if not settings.RUNNING_IN_CLOUD and not force:
             return False
 
--- a/common/data_refinery_common/models/models.py
+++ b/common/data_refinery_common/models/models.py
@@ -804,8 +804,8 @@ class ComputedFile(models.Model):
             return True
         except:
             logger.exception("Failed to delete S3 object for Computed File.",
-                             "computed_file"=self.id,
-                             "s3_object"=self.s3_key)
+                             computed_file=self.id,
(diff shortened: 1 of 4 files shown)


kurtwheeler:
I just took another look at this and realized I need more comments. I'm adding those.

common/data_refinery_common/models/models.py
```diff
@@ -784,17 +784,30 @@ def calculate_size(self) -> None:
...
         """ Deletes a file from the path and actually removes it from the file system,
         resetting the is_downloaded flag to false. Can be refetched from source if needed. """
-        if not settings.RUNNING_IN_CLOUD:
+        if not settings.RUNNING_IN_CLOUD and not force:
             return
 
         try:
             os.remove(self.absolute_file_path)
         except OSError:
             pass
```
dongbohu: Not really in this PR: If the file can't be deleted, should we create a log message?

dongbohu:
@kurtwheeler: Since the model is modified, you may want to add the migration file in the PR too.

common/data_refinery_common/models/models.py
```diff
@@ -784,17 +784,30 @@ def calculate_size(self) -> None:
...
 
+    def delete_s3_file(self, force=False):
+        if not settings.RUNNING_IN_CLOUD and not force:
```
dongbohu: Since you are deleting a file in S3, it is supposed to be:
```
if not settings.RUNNING_IN_CLOUD or not force:
```

dongbohu: (I am not familiar with the calling context of this method, so my comment above could be totally wrong.)

kurtwheeler: `force` defaults to false so `or not force` would make the conditional true most of the time. The idea here is that if RUNNING_IN_CLOUD is false but we want to force the S3 code anyway, then we shouldn't return early. (I agree the conditional  is a bit confusing because it's dealing with a case where we want to exit the function early rather than only running the function if it ends up being true.

kurtwheeler:
>  I think that the command should have a confirmation input since this is such a destructive command.

The reason I didn't already include this is because I was unsure how we want to run this long term. If we want to ssh onto the foreman box, exec onto the image and run this, then I agree we should have a confirmation. However if we'd rather trigger this being run via a Nomad dispatch command similar to how we kick off surveyor jobs, then I don't know how we would provide the confirmation input.

@Miserlou what do you think? I guess I could only have it ask for confirmation if `accession` is specified rather than `file`...

common/data_refinery_common/models/models.py
kurtwheeler: I don't think so... at least not yet. I think there's some good reasons for this to fail, such as the file getting deleted by something else. I think that if we notice our EFS isn't getting cleaned up properly once we've addressed #452, then we could improve logging around data cleanup. I could be persuaded otherwise though.

Miserlou:
I think destructive commands should have a confirmation. You can pass in a `--yes` if you want to auto-confirm although I don't really see why we'd want to run this as a dispatch.

common/data_refinery_common/models/models.py
dongbohu: Thanks. It will be great if you can add some docstring to clarify its usage.

kurtwheeler:
>  Since the model is modified, you may want to add the migration file in the PR too.

The models.py file is changed, but the database model isn't actually changing because the changes are only additional methods.

kurtwheeler:
> I think destructive commands should have a confirmation. You can pass in a --yes if you want to auto-confirm although I don't really see why we'd want to run this as a dispatch.

I like that idea. Just a note, the reason I would think we'd maybe run it as a dispatch would be if we ever made the deploy process automatically unsurvey experiments the code it's deploying fixes. That could be a nice way to lock down our production stack from manual interventions.

common/data_refinery_common/models/models.py
dongbohu: I see your point. I am a little worried that the failure to delete a local may indicate something serious, such as the crash of the file system. But if EFS is robust enough, then we can ignore it for now.

kurtwheeler: I think that EFS crashes the rest of the system will be freaking out enough that it should be obvious to us that's what's happening.

PR #518, opened 2018-08-10
Call utils.handle_processor_error properly

dongbohu:
## Issue Number

NA

## Purpose/Implementation Notes

`utils.handle_processor_error` function was not called correctly by processors.

## Types of changes

What types of changes does your code introduce?

- Bugfix (non-breaking change which fixes an issue)

## Functional tests

List out the functional tests you've completed to verify your changes work locally.

## Checklist

_Put an `x` in the boxes that apply._

- [ ] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

## Screenshots

Please attach any screenshots that illustrate these changes.

Issue #488, opened 2018-08-03
Implement quantile normalization


jaclyn-taroni:
### Context

We'll be combining multiple platforms and even multiple technologies in this project, particularly in the species-wide compendium case. We'll want the underlying distributions of these platforms/technologies to be as similar as possible. 

We intend to use [quantile normalization](https://en.wikipedia.org/wiki/Quantile_normalization) to accomplish this.

By doing so, we assume, then, that the difference between samples arises solely from _technical differences_. Of course this is not always going to be the case. We'll use as many samples as possible to generate the reference or target distribution. By including as diverse biological conditions as we have available to us to inform the reference distribution, we hope that this will mitigate the effect of the "dampening" of biological signal that is likely to occur.

### Problem or idea

As I see it, this has three components:

1. Identify the most popular microarray platform for _each species_ (that we can process)
2. Calculate the reference distribution from the entire "most popular platform" compendium for each species
3. Just in time quantile normalization during smashing

_I'll note that this may be better split into multiple tickets for tracking, but we can make that determination once the team takes a look at it._

#### Most popular platform

We'll use microarray platforms because we, and others, have experience using microarray data as a reference distribution for RNA-seq.

Using the GEO data from #220, [`config/supported_microarray_platforms.csv`](https://github.com/AlexsLemonade/refinebio/blob/13444092d7c5f186612dd8228f3e656b42e3f895/config/supported_microarray_platforms.csv), and only including platforms with `num_samples > 100`, I get this list: [affymetrix_platforms_for_QN_reference.txt](https://github.com/AlexsLemonade/refinebio/files/2258673/affymetrix_platforms_for_QN_reference.txt)

We'll probably want to prioritize crunching samples on these platforms.

#### Calculate reference distribution

We should aggregate (smash) all samples together from a platform _that we have processed_ (e.g., do not include submitter-processed data) without any row-wise transformation.

We have some options here:

* With rpy2, we can use [`normalize.quantiles.determine.target`](https://www.rdocumentation.org/packages/preprocessCore/versions/1.34.0/topics/normalize.quantiles.target) from the `preprocessCore` R package. A downside raised here is reading the whole matrix into memory.
* I believe @kurtwheeler has an implementation in Python handy that works on a matrix and that includes unit tests. Not sure if/how it handles missing values, but I don't believe this will be too much of a concern if using `SCAN.UPC` processed Affymetrix data.
[...]

PR #519, opened 2018-08-10
Adds Quantile Normalization

Miserlou:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/488

This ticket will probably need some discussion.

Two things of importance in here: 1) creating new QN targets (via a new management command) and 2) applying QN with those targets at smash time. 

This has currently only been tested with toy data so far and should be run for an entire organism before we can use it confidently everywhere.

Also worth noting is that uses a new image `qn`, but it's very similar to the (updated) `smasher` image, so I think this can be refactored to just use a single beefy `smasher` image.

Also I think it needs https://github.com/AlexsLemonade/refinebio/pull/518 to land before tests can pass. The processor pipeline stuff for QN target object creation probably needs updating since it was written before the new pipeline PR landed but I need somebody to explain that process to me a bit.

## Methods

Described in the related ticket above.

## Types of changes

## Functional tests

Lots of new tests with toy data, and ran locally with other sample data.

## Checklist


workers/data_refinery_workers/processors/utils.py
```diff
@@ -402,11 +410,22 @@ def get_bioc_version():
...
+    file = annotation.result.computedfile_set.first()
+    return file
+
+
 def get_r_pkgs(pkg_list):
     """Returns a dictionary in which each key is the name of a R package
     and the corresponding value is the package's version.
     """
 
+    for label in ProcessorEnum:
```
dongbohu: Lines 426-428 should be removed.

kurtwheeler:
Just a heads up that this will conflict with the migration in https://github.com/AlexsLemonade/refinebio/pull/505 once that gets approved.

dongbohu:
@Miserlou: I compared `Dockerfile.qn` and `Dockerfile.smasher`. They look pretty close. Here are the differences that I noticed:

1. `Dockerfile.smasher` doesn't add apt-key of CRAN repo, so it installs the R packages that are bundled with the original Ubuntu (which is R 3.2.3), while `Dockerfile.qn` adds the apt-key, so it always installs the latest R (3.4.4).

2. `Dockerfile.qn` includes two extra pip packages: `rpy2` and `tzlocal`.

I also built these two docker images, whose sizes are both 1.64 GB (uncompressed). Considering the overlap between these two images, I think it makes sense to merge them into one. I bet the size of merged Docker image will be still ~1.64 GB.

I am not sure whether we should keep the separate `QN_REFERENCE` processor (and its YAML file)  or merge it into `SMASHER` processor too.

Miserlou:
Have merged the two Dockerfiles into `dr_smasher` now.

workers/data_refinery_workers/processors/smasher.py
```diff
@@ -188,10 +210,57 @@ def _smash(job_context: Dict) -> Dict:
...
+
+                        # And finally convert back to Pandas
+                        ar = np.array(reso)
+                        new_merged = pd.DataFrame(ar, columns=merged.columns, index=merged.index)
+                        job_context['merged_no_qn'] = new_merged
+                        merged = new_merged
+                except Exception as e:
+                    logger.exception("Problem occured during quantile normalization",
+                        dataset_id=job_context['dataset'].id,
+                        dataset_data=job_context['dataset'].data,
```
kurtwheeler: This is another place we'll return a un-normalized dataset without alerting the user right?

workers/data_refinery_workers/processors/smasher.py
```diff
@@ -321,7 +391,8 @@ def _upload(job_context: Dict) -> Dict:
         logger.exception("Failed to upload smash result file.", file=job_context["output_file"])
         job_context['job'].success = False
         job_context['job'].failure_reason = str(e)
-        job_context['success'] = False
+        # Delay failing this pipeline until the failure notify has been sent
+        # job_context['success'] = False
```
kurtwheeler: Is this line meant as a way of conveying that it wasn't included on purpose?

workers/data_refinery_workers/processors/smasher.py
```diff
@@ -339,11 +410,18 @@ def _notify(job_context: Dict) -> Dict:
...
 
+            if job_context['job'].failure_reason not in ['', None]:
+                SUBJECT = "Your refine.bio Dataset is Ready!"
+                BODY_TEXT = "Hot off the presses:\n\n" + job_context["result_url"] + "\n\nLove!,\nThe refine.bio Team"
+                BODY_HTML = "Hot off the presses:<br /><br />" + job_context["result_url"] + "<br /><br />Love!,<br />The refine.bio Team"
+            else:
+                SUBJECT = "There was a problem processing your refine.bio dataset :("
+                BODY_TEXT = "We tried but were unable to process your requested dataset. Error was: \n\n" + str(job_context['job'].failure_reason) + "\nDataset ID: " + str(dataset.id) + "\n We have been notified and are looking into the problem. \n\nSorry!"
+                BODY_HTML = BODY_TEXT = "We tried but were unable to process your requested dataset. Error was: <br /><br />" + job_context['job'].failure_reason + "<br />Dataset: " + str(dataset.id) + "<br /> We have been notified and are looking into the problem. <br /><br />Sorry!"
+                job_context['success'] = False
```
kurtwheeler: Since you're doing this regardless of whether this is cloud or not below, does this really need to be here?

workers/data_refinery_workers/processors/utils.py
```diff
@@ -272,6 +274,12 @@ class ProcessorEnum(Enum):
         "yml_file": "transcriptome_index.yml"
     }
 
+    QN_REFERENCE = {
+        "name": "Quantile Normalization Reference",
+        "docker_img": "dr_qn",
```
kurtwheeler: Aren't we using the smasher image for this?

workers/data_refinery_workers/processors/test_qn_reference.py
```diff
@@ -0,0 +1,119 @@
...
+    ProcessorJobDatasetAssociation
+)
+from data_refinery_workers.processors import qn_reference, smasher, utils
+
+
+class QNRefTestCase(TestCase):
+
+    @tag('qn')
+    def test_sanity(self):
+        print("Hey!")
```
kurtwheeler: I can assure you you're sane :D

workers/data_refinery_workers/processors/smasher.py
```diff
@@ -188,10 +210,57 @@ def _smash(job_context: Dict) -> Dict:
...
+                                                            x=merged_matrix,
+                                                            target=target_vector,
+                                                            copy=True
+                                                        )
+
+                        # And finally convert back to Pandas
+                        ar = np.array(reso)
+                        new_merged = pd.DataFrame(ar, columns=merged.columns, index=merged.index)
+                        job_context['merged_no_qn'] = new_merged
+                        merged = new_merged
```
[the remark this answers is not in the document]
jaclyn-taroni: This is also similar to what I was expecting.

workers/data_refinery_workers/processors/qn.yml
```diff
@@ -0,0 +1,22 @@
+---  # "QN" processor
```
jaclyn-taroni: These are what I'd want to know about, LGTM

workers/data_refinery_workers/processors/qn_reference.R
```diff
@@ -0,0 +1,59 @@
...
+# Get and use this DB
+message("Loading...")
+source("http://bioconductor.org/biocLite.R")
+
+# Read the data file
+message("Reading data file...")
+suppressWarnings(data <- fread(filePath, 
+					stringsAsFactors=FALSE, 
+					sep="\t", header=TRUE, 
+					autostart=10,
```
jaclyn-taroni: This means we're skipping the first 10, is that correct? Is that because of how the data is getting passed to this script in our system?

workers/data_refinery_workers/processors/qn_reference.R
```diff
@@ -0,0 +1,59 @@
...
+              help="outputFile", metavar="character")
+); 
+
+opt_parser = OptionParser(option_list=option_list);
+opt = parse_args(opt_parser);
+
+filePath <- opt$inputFile
+outFilePath <- opt$outputFile
+
+# Get and use this DB
```
jaclyn-taroni: I don't think we need lines 29-31 since we won't be using any annotation `.db` packages or installing bioconductor packages.

workers/data_refinery_workers/processors/qn_reference.R
```diff
@@ -0,0 +1,59 @@
...
+					fill=TRUE, 
+					na.strings="", 
+					showProgress=FALSE)
+				)
+
+message(data)
+
+# Create the target result
+data_matrix <- data.matrix(data, rownames.force = NA)
+data_matrix<-data_matrix[,-1] # We don't want the index column
```
jaclyn-taroni: Will the reference data that goes through our smasher always have some kind of identifier in the first column? Since this was developed with the toy data I posted, that had row indices in the first column and would cause an issue. Something to keep an eye on when working with real data.

workers/data_refinery_workers/processors/qn_reference.R
jaclyn-taroni: Ah, no it does not mean that. Was looking at newer documentation where it says it's depreciated.

workers/data_refinery_workers/processors/smasher.py
Miserlou: There is a small window where it hasn't been sent yet and it could be interruped. It shouldn't really matter but I don't want to re-do a job if the email is sent. I don't think it matters too much.

workers/data_refinery_workers/processors/smasher.py
Miserlou: Yes

common/data_refinery_common/message_queue.py
```diff
@@ -43,6 +43,8 @@ def send_job(job_type: Enum, job) -> None:
         nomad_job = ProcessorPipeline.ILLUMINA_TO_PCL.value
     elif job_type is ProcessorPipeline.SMASHER:
         nomad_job = ProcessorPipeline.SMASHER.value
+    elif job_type is ProcessorPipeline.QN_REFERENCE:
+        nomad_job = ProcessorPipeline.QN_REFERENCE.value
```
kurtwheeler: I believe this means that this will use the `QN_REFERENCE` image. Take a look at the agilent `elif` to see how we get it to use the affy image.

PR #525, opened 2018-08-14
Makes downloader and processor jobs complain when they don't have fai…

kurtwheeler:
…lure_reason set.

## Issue Number

N/A came up during Jackie Crunch

I keep seeing jobs that have `end_time` but no `failure_reason`. This shouldn't happen. If we know a job is failing we should say why. This will hopefully make cases where we don't easier to find and track down and alert us if we accidentally introduce any new ones.

While I tested this I once again noticed that all of our log messages appear twice, once with our nice context and one with nothing. I fixed this while I was there.

While testing I also noticed that if `end_job` is ever called with an incomplete `job_context` we would throw an error. If this ever were to happen it would probably be because the job didn't make it very far and so this new error would be hiding whatever caused the job to fail so early.

## Types of changes

## Functional tests

```
>>> from data_refinery_workers.processors.utils import end_job
>>> from data_refinery_common.models import ProcessorJob
>>> pj = ProcessorJob(success=False, pipeline_applied="AFFY_TO_PCL") 
>>> pj.save()
>>> job_context = {"job": pj, "success": False}
>>> end_job(job_context)
2018-08-14 15:48:06,324 local/MainProcess data_refinery_workers.processors.utils INFO [pipeline_applied: AFFY_TO_PCL] [processor_job: 6]: Processor job failed!
2018-08-14 15:48:06,326 local/MainProcess data_refinery_workers.processors.utils ERROR [pipeline_applied: AFFY_TO_PCL] [processor_job: 6]: Processor job failed without having failure_reason set. FIX ME!!!!!!!!
{'success': False, 'job': <ProcessorJob: ProcessorJob 6: AFFY_TO_PCL>}
```

I ran a similar functional test for downloaders.

## Checklist

### DIFF
--- a/workers/data_refinery_workers/downloaders/utils.py
+++ b/workers/data_refinery_workers/downloaders/utils.py
@@ -52,6 +52,14 @@ def end_downloader_job(job: DownloaderJob, success: bool):
     if success:
         logger.info("Downloader Job completed successfully.",
                     downloader_job=job.id)
+    else:
+        logger.info("Downloader job failed!",
+                    downloader_job=job.id,
+                    downloader_task=downloader_task)
+        if not job.failure_reason:
+            logger.error("Downloader job failed without having failure_reason set. FIX ME!!!!!!!!",
+                         downloader_job=job.id,
+                         downloader_task=downloader_task)
 
     job.success = success
     job.end_time = timezone.now()
--- a/workers/data_refinery_workers/downloaders/utils.py
+++ b/workers/data_refinery_workers/downloaders/utils.py
@@ -53,13 +53,15 @@ def end_downloader_job(job: DownloaderJob, success: bool):
         logger.info("Downloader Job completed successfully.",
                     downloader_job=job.id)
     else:
-        logger.info("Downloader job failed!",
-                    downloader_job=job.id,
-                    downloader_task=job.downloader_task)
         if not job.failure_reason:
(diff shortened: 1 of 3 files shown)


workers/data_refinery_workers/downloaders/utils.py
```diff
@@ -52,6 +52,14 @@ def end_downloader_job(job: DownloaderJob, success: bool):
     if success:
         logger.info("Downloader Job completed successfully.",
                     downloader_job=job.id)
+    else:
+        logger.info("Downloader job failed!",
+                    downloader_job=job.id,
+                    downloader_task=job.downloader_task)
+        if not job.failure_reason:
```
Miserlou: Can this be refactored such that if the failure reason is present, it gets passed to the logger?

workers/data_refinery_workers/processors/utils.py
```diff
@@ -107,30 +107,41 @@ def end_job(job_context: Dict, abort=False):
...
     job.save()
 
     if success:
-        logger.info("Processor job completed successfully.", processor_job=job.id)
+        logger.info("Processor job completed successfully.",
+                    processor_job=job.id,
+                    pipeline_applied=job.pipeline_applied)
     else:
-        logger.info("Processor job failed!", processor_job=job.id)
+        logger.info("Processor job failed!",
```
Miserlou: Can this be refactored such that if the failure reason is present, it gets passed to the logger?

Issue #430, opened 2018-07-20
OperationalError: FATAL:  remaining connection slots are reserved for non-replication superuser connections


kurtwheeler:
https://sentry.io/greenelab/staging-refinebio/issues/614186488/

```
OperationalError: FATAL:  remaining connection slots are reserved for non-replication superuser connections

  File "django/db/backends/base/base.py", line 216, in ensure_connection
    self.connect()
  File "django/db/backends/base/base.py", line 194, in connect
    self.connection = self.get_new_connection(conn_params)
  File "django/db/backends/postgresql/base.py", line 168, in get_new_connection
    connection = Database.connect(**conn_params)
  File "psycopg2/__init__.py", line 130, in connect
    conn = _connect(dsn, connection_factory=connection_factory, **kwasync)

OperationalError: FATAL:  remaining connection slots are reserved for non-replication superuser connections

(17 additional frame(s) were not displayed)
...
  File "data_refinery_workers/processors/management/commands/run_processor_job.py", line 35, in handle
    affy_to_pcl(options["job_id"])
  File "data_refinery_workers/processors/array_express.py", line 218, in affy_to_pcl
    utils.end_job])
  File "data_refinery_workers/processors/utils.py", line 199, in run_pipeline
    job = ProcessorJob.objects.get(id=job_id)

OperationalError: FATAL:  remaining connection slots are reserved for non-replication superuser connections

```

Miserlou:
```
Events: 10.4k
```

This is the most frequent error in the whole system by a wide margin (the next being a `FileAlreadyDownloaded` error-non-error at 5.8k events). The problem is that addressing it properly requires completely rearchitecting the whole job system into something API driven, since we're not going to be able to scale up to our desired size while still making direct simultaneous unpooled database connections.

There are some smaller hacks we could try in the meantime, like using named database users with a limited of number of connections, so at least one subsystem can't overwhelm the rest and we'll always have some spare room for the API and Smasher. However, that'd just be a bandaid on a fatal wound.

Another option is just to keep buying beefier database servers and hope the problem goes away. The current staging deployment is using a `db.t2.micro`, which is way, way too small for our size (our previous crunches were run on a `db.m4.large` and still had this error, I don't know how we got switched down to to a smaller instance type). I suggest we at least use a `db.m4.4xlarge` and cross our fingers.

jaclyn-taroni:
Related: #453

Miserlou:
[...]

PR #531, opened 2018-08-15
Adds PgBouncer Instance

Miserlou:
## Issue Number
https://github.com/AlexsLemonade/refinebio/issues/430

Attempts to solve the DB connection limit nightmare.

This is really a stop-gap measure rather than the major refactor to an API-driven approach, but the effect might be enough to make it a non-issue.

Also: I can't test this. There is a bug in Terraform (reported upstream) that crashes TF during deploy for me, both for this patchset and for the current /infrastructure in dev. I've run the scripts here manually on an EC2 and they work, but I need somebody else to actually confirm the connections work across instances.

Also sneaks in an assertion that checks an error we got in Sentry.

## Types of changes

## Functional tests

See above


infrastructure/nomad-configuration/pg-bouncer-instance-user-data.tpl.sh
```diff
@@ -0,0 +1,65 @@
...
+
+# Change to home directory of the default user
+cd /home/ubuntu
+
+service postgresql stop
+
+apt-get -y update
+# sh -c 'echo "deb http://apt.postgresql.org/pub/repos/apt/ $(lsb_release -cs)-pgdg main" > /etc/apt/sources.list.d/pgdg.list'
+# wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key -y add -
+# apt-get -y update
```
kurtwheeler: These lines look pretty dead.

infrastructure/nomad-configuration/pg-bouncer-instance-user-data.tpl.sh
```diff
@@ -0,0 +1,65 @@
...
+apt-get -y install pgbouncer postgresql-client --allow-unauthenticated
```
kurtwheeler: `--allow-unathenticated` isn't insecure?

infrastructure/nomad-configuration/pg-bouncer-instance-user-data.tpl.sh
```diff
@@ -0,0 +1,65 @@
...
+listen_addr = *
+max_client_conn = 500
+default_pool_size = 20
+listen_port = 5432
+auth_type = md5
+auth_file = /etc/pgbouncer/userlist.txt
+pool_mode = transaction
+server_reset_query = DISCARD ALL
+logfile = /tmp/pgbouncer.log
+logfile = /var/log/postgresql/pgbouncer.log
```
kurtwheeler: We don't have anything set up to rotate these logs. Do you know if there's any chance that they'll grow too large at some point?

infrastructure/nomad-configuration/pg-bouncer-instance-user-data.tpl.sh
```diff
@@ -0,0 +1,65 @@
...
+
+# Set up PG Bouncer
+cat << FOE >> /etc/pgbouncer/pgbouncer.ini
+[databases]
+${database_name} = host=${database_host} port=5430 dbname=${database_name}
+
+[pgbouncer]
+listen_addr = *
+max_client_conn = 500
+default_pool_size = 20
```
kurtwheeler: Could you add a comment explaining how you came up with these values so if we ever have to tweak it we have some context?

infrastructure/security.tf
```diff
@@ -214,6 +223,85 @@ resource "aws_security_group_rule" "data_refinery_db_outbound" {
...
+}
+
+resource "aws_security_group_rule" "data_refinery_pg_api_icmp" {
+  type = "ingress"
+  from_port = -1
+  to_port = -1
+  protocol = "icmp"
+  source_security_group_id = "${aws_security_group.data_refinery_api.id}"
+  security_group_id = "${aws_security_group.data_refinery_pg.id}"
+}
```
kurtwheeler: Two questions about all these security group rules:
* I think we already have these in place for communicating directly with the rds instance. Is there a reason our clients would need access to both the rds instance and the pgbouncer instance?
* The workers and api have both tcp and icmp rules, is there a reason the foreman doesn't need icmp?

Miserlou: For debugging, I think it'd be useful to have direct psql access. We mostly just want PGB for Django.

infrastructure/instances.tf
```diff
@@ -349,6 +352,52 @@ resource "aws_db_instance" "postgres_db" {
 
 }
 
+resource "aws_instance" "pg_bouncer" {
+  ami = "${data.aws_ami.ubuntu.id}"
+  instance_type = "${var.nomad_server_instance_type}"
+  availability_zone = "${var.region}a"
```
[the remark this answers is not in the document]
Miserlou: I don't think this will matter too much.

infrastructure/nomad-configuration/pg-bouncer-instance-user-data.tpl.sh
Miserlou: rm'd

infrastructure/nomad-configuration/pg-bouncer-instance-user-data.tpl.sh
Miserlou: rm'd since we're just using default Django instance now

infrastructure/nomad-configuration/pg-bouncer-instance-user-data.tpl.sh
Miserlou: moved to syslog

infrastructure/nomad-configuration/pg-bouncer-instance-user-data.tpl.sh
Miserlou: commented

PR #533, opened 2018-08-17
R error -> logger

cgreene:
## Issue Number

N/A

There are hundreds of R issues in sentry. These get annotated by the R script and the file that failed, but not the error from R. This edit reports the R script and R error in the logger. 

This does change the behavior slightly: The specific file that was processed doesn't get put into the `job_context["job"].failure_reason`. Instead the R stderr does. It looks like the `job_context['input_file_path']` already gives us the file that was processed. This is the only potentially behavior-breaking change that I know of.

## Methods

N/A

## Types of changes

## Functional tests

I tested running an R script in python via `subprocess.check_output` and `stderr` from R gets captured and put into the error message with this approach. I did not test running this specific script 😨 . Still trying to figure out the best way to do that.

## Checklist

## Screenshots

### DIFF
--- a/workers/data_refinery_workers/processors/no_op.py
+++ b/workers/data_refinery_workers/processors/no_op.py
@@ -245,12 +245,19 @@ def _convert_illumina_genes(job_context: Dict) -> Dict:
                 "--platform", high_db,
                 "--inputFile", job_context['input_file_path'],
                 "--outputFile", job_context['output_file_path']
-            ])
+            ], stderr=subprocess.PIPE)
+    except subprocess.CalledProcessError as e:
+        error_template = "Status code {0} from gene_convert_illumina.R: {1}"
+        error_message = error_template.format(e.returncode, e.stderr)
+        logger.error(error_message, context=job_context)
+        job_context["job"].failure_reason = error_message
+        job_context["success"] = False
+        return job_context
     except Exception as e:
         error_template = ("Encountered error in R code while running gene_convert_illumina.R"
                           " pipeline during processing of {0}: {1}")
         error_message = error_template.format(job_context['input_file_path'], str(e))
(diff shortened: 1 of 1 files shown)


cgreene:
If this looks sensible I can modify other calls to R.

cgreene:
This should do the two calls to R that follow this pattern. We have 5 total in the current project. Two in illumina.py and three in no_op.py.

I debated writing a utility function for this (which could be more easily tested) but it looks like the output of subprocess.check_output is used differently there and I didn't want to risk raising new issues.

kurtwheeler:
I tested this by breaking the invocation of the R script in `no_op.py` and then running the tests and looking at the log message.

kurtwheeler:
FYI it looks like:
```
ERROR:data_refinery_workers.processors.no_op:Status code 1 from gene_convert.R: b'\nAttaching package: \'dplyr\'\n\nThe following objects are masked from \'package:data.table\':\n\n    between, first, last\n\nThe following objects are masked from \'package:stats\':\n\n    filter, lag\n\nThe following objects are masked from \'package:base\':\n\n    intersect, setdiff, setequal, union\n\n\nAttaching package: \'rlang\'\n\nThe following object is masked from \'package:data.table\':\n\n    :=\n\nLoading...\nBioconductor version 3.6 (BiocInstaller 1.28.0), ?biocLite for help\nA new version of Bioconductor is available after installing the most recent\n  version of R; see http://bioconductor.org/install\nReading data file...\nError in fread(filePath, stringsAsFactors = FALSE, sep = "\\t", header = TRUE,  : \n  File \'asd\' does not exist; getwd()==\'/home/user\'. Include correct full path, or one or more spaces to consider the input a system command.\nExecution halted\n'
```

Issue #535, opened 2018-08-17
Illumina no-op detection


cgreene:
Noticed this error while scanning through the code. Wondered how many there were. Only one! Asked Jackie and she says:
> that file should not be no-op'd it's raw data

https://sentry.io/greenelab/staging-refinebio/issues/648682435/

```
We found an Illumina file to NO_OP that we're not set up to process yet. Tell Rich!
```

jaclyn-taroni:
It could be related to the fact that the headers of the non-normalized file (`GSE60102_non-normalized.txt`; [here's the GEO accession](https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GSE60102)) are the sample _descriptions_ rather than the accession numbers or sample names? Related: https://github.com/AlexsLemonade/refinebio/pull/487

cgreene:
😮 ! I filed the issue just as something to track for the future and within an hour there's PR #536! 🎉

PR #536, opened 2018-08-17
 Check for Permuatations of Non-Normalized Data Filenames 

Miserlou:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/535

Turns out that `_non_normalized.txt` wasn't uniform across the data. This adds more permutations of the filename so that `has_raw` gets set appropriately and the correct processors are dispatched.

## Types of changes

## Functional tests
Tested locally


Miserlou:
Tests are failing, I think because Zenodo is down? Passes locally for me.

PR #540, opened 2018-08-17
Kurtwheeler/tx index on efs

kurtwheeler:
## Issue Number

N/A came up during zebrafish crunch.

Seems like most salmon jobs failed because of file path collisions either related to transcriptome indices or the files the jobs themselves were generating.

This creates a unique path for each sample of each job so they cannot collide. It also changes the way we handle transcriptome indices. Since we want them downloaded and extracted on disk we now don't delete the directory and instead just store the path to it.

## Types of changes

## Functional tests

I ran a crunch with the transcriptome index changes and only saw errors related to file paths not existing or not being able to be written to. The job directory changes were tested with unit tests.

## Checklist


Miserlou:
Will these test fails be fixed by https://github.com/AlexsLemonade/refinebio/pull/545 ?

workers/data_refinery_workers/processors/salmon.py
```diff
@@ -197,7 +197,7 @@ def _find_or_download_index(job_context: Dict) -> Dict:
     except FileExistsError:
         # Someone already installed the index or is doing so now.
         pass
-    except e:
+    except Exception as e:
```
cgreene: This issue also crops up in utils.py. If you get that one as well, I think this should close #569.

kurtwheeler:
@Miserlou the tests are finally passing so this is ready to be re-reviewed. It's definitely gotten large because I kept deploying from it and finding minor fixes and adding them into it. It also needed additional changes to handle the change from EFS to EBS. Finally, the collab review changes to salmon and transcriptome indices were made in the branch rather than collab review.

Also I pulled collab review at some point so some of the changes in this branch are actually from there and have already been reviewed. However this is now a big PR and I apologize about that.

Miserlou:
Looks like changes from `35a25c7` broke the tests, and there's a merge conflict. I think the agreement during collab-review was that we should keep that as the same key, not chance it the way you did in `35a25c7` (your idea at the time :-P)

kurtwheeler:
Whoops, I had that change sitting uncommitted and forgot why so I just added it. Merge resolution incoming.

foreman/data_refinery_foreman/surveyor/array_express.py
```diff
@@ -129,27 +125,30 @@ def create_experiment_from_api(self, experiment_accession_code: str) -> (Experim
...
+            # We aren't sure these fields will be populated, or how many there will be.
+            # Try to join them all together, or set a sensible default.
+            experiment_descripton = ""
+            if "description" in parsed_json and len(parsed_json["description"]) > 0:
+                for description_item in parsed_json["description"]:
+                    if "text" in description_item:
+                        experiment_descripton = experiment_descripton + description_item["text"] + "\n"
+
+            if experiment_descripton == "":
+                experiment_descripton = "Description not available.\n"
```
Miserlou: Why the newline?

Miserlou:
I think this is good but if it's so big and contains a lot of our collab-review changes, it'd be great if you could point me to the core of what you'd like reviewed.

foreman/data_refinery_foreman/surveyor/array_express.py
kurtwheeler: Honestly, I don't remember. This was pulled in via the collab review so I guess we decided this is how we should do it.

I'm guessing the reason is that if `experiment_description` does actually get populated we join all the lines of it together into a string with a newline inbetween each one (see line 134) so it would end in a newline so this is consistent with that.

PR #544, opened 2018-08-22
Move from 1xEFS to NxEBS High Throughput

Miserlou:
## Issue Number

n/a

Writing to a single EFS instance, even with the extremely expensive provisioned-throughput turned on, was a significant bottleneck of the system.

This moves us from using a single EFS instance to using an EBS per-processor instance. This lowers costs and most importantly removes the disk writing bottleneck that otherwise would have sunk the ship.

Unfortunately, this comes with other tradeoffs, as we lose the benefits that came from using EFS. Notably, downloaders and processors are now tied to using the same instance, since an EBS can only be attached to a single instance at a time.

The magic solution that avoided a massive rewrite or combining processor and downloader images is that we now explode our number of Nomad job specifications per EBS instance, and then include a the EBS Volume Index as a `constraint` for job placement, like so:

```
      constraint {
        attribute = "${meta.volume_index}"
        operator  = "="
        value     = "${{INDEX}}"
      }
```

such that jobs will only be scheduled on instances which have the correct EBS volume attached.

Miscellaneous notes:
    - In order to ensure an even distribution of a files across volumes, we should consider "pre-scaling" the system before we begin large crunches, or we should rewrite a little bit of the surveyor so that it runs in a more distributed fashion.
    - Although all of our volumes are maxed out in terms of available size, there is no elasticity, so we must still be mindful of cleaning up after ourselves. (See #452 ). We also pay for what we don't use.
    - All the `jq` stuff in the new client init script seems really fickle to me. It might break if there are multiple stages running the same code, it needs a second set of eyes on it.

## Methods

Speed tests were conducted for EFS, S3 and EBS. EBS came out on top by a wide margin.

EBS write speed is actually an AWS-limited function of instance type, which you can see on [this chart](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSOptimized.html). For our desired X1.32large, we get a maximum throughput of `1,750 MB/s`, and we'll have ten running in parallel. 

Doing some napkin-math, for a 10PB dataset, with 10x`1,750 MB/s` and assuming unlimited upload bandwidth and no other overhead, we should be able to download the whole bundle in 6.97 days at a cost of $2,000, which is significantly better than with provisioned EFS.

## Types of changes

## Functional tests

Mostly ops changes that have been tested manually, one new test for the small code change.

## Checklist


infrastructure/nomad-configuration/client-instance-user-data.tpl.sh
```diff
@@ -17,11 +17,44 @@
...
 apt-get update -y
 apt-get upgrade -y
-apt-get install --yes nfs-common
+apt-get install --yes nfs-common jq iotop dstat
+
+# EFS
+# mkdir -p /var/efs/
+# echo "${file_system_id}.efs.${region}.amazonaws.com:/ /var/efs/ nfs4 nfsvers=4.1,rsize=1048576,wsize=1048576,hard,timeo=600,retrans=2 0 0" >> /etc/fstab
+# mount -a -t nfs4
+# chown ubuntu:ubuntu /var/efs/
```
kurtwheeler: Rip this bandaid!!!

infrastructure/nomad-configuration/client-instance-user-data.tpl.sh
```diff
@@ -17,11 +17,44 @@
...
+FILE_RESULT=`file -s /dev/$ATTACHED_AS`
+
+if file -s /dev/$ATTACHED_AS | grep data; then
+	mkfs -t ext4 /dev/$ATTACHED_AS # This is slow
+fi
+mount /dev/$ATTACHED_AS /var/efs/
+
 chown ubuntu:ubuntu /var/efs/
+echo $EBS_VOLUME_INDEX >  /var/efs/VOLUME_INDEX
+chown ubuntu:ubuntu /var/efs/VOLUME_INDEX
```
kurtwheeler: We don't use efs. I know it's not a big deal but this could be confusing especially if we ever have a new dev join the team.

infrastructure/permissions.tf
```diff
@@ -72,6 +72,33 @@ resource "aws_iam_role_policy_attachment" "s3" {
...
+         ],
+         "Resource": "*"
+      }
+   ]
+}
+EOF
+}
+
+resource "aws_iam_role_policy_attachment" "ec2" {
+  role = "${aws_iam_role.data_refinery_instance.name}"
```
kurtwheeler: Wait is this giving all ec2 permissions to all of our instances? I'm guessing they need some of these to mount the EBS volumes themselves, but can't it be any more locked down than this?

infrastructure/users.tf
```diff
@@ -66,6 +66,13 @@ resource "aws_iam_user_policy" "data_refinery_user_worker_policy" {
               "SES:SendRawEmail"
             ],
             "Resource": "*"
+        },
+        {
+            "Effect": "Allow",
+            "Action":[
+              "ec2:*"
+            ],
+            "Resource": "*"
```
kurtwheeler: Same question here: can this be locked down any more?

Miserlou:
Changes should all be addressed now. Retested to AWS and it works.

Issue #543, opened 2018-08-21
Update template for email sent after a data set is processed


arielsvn:
### Context

When a new download is processed we send an email to users. @dvenprasad created a design for it in https://github.com/AlexsLemonade/refinebio/issues/539

### Problem or idea

We should send emails with that template instead, I pasted the html code she created here https://gist.github.com/arielsvn/cb128188579d78b8b8de7faf69d1799e

### Solution or next step

We'll probably have to update the html https://github.com/AlexsLemonade/refinebio/blob/e1e5b8e0269ef58988468257418c05ec50f0daaf/workers/data_refinery_workers/processors/smasher.py#L359-L361

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

cgreene:
Updating to note that there is a PR: https://github.com/AlexsLemonade/refinebio/pull/546

PR #546, opened 2018-08-22
Adds HTML Email Template

Miserlou:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/543

Adds the email HTML template

## Types of changes

## Functional tests
¯\\\_(ツ)\_/¯


workers/data_refinery_workers/processors/smasher.py
```diff
@@ -447,3 +447,1508 @@ def smash(job_id: int, upload=True) -> None:
                         _update_result_objects,
                         _delete_local_files,
                         utils.end_job])
+
+# Look ma, I'm React!
```
kurtwheeler: Can you move this to another file and read in the string?

Miserlou:
Moved to file.

workers/data_refinery_workers/processors/smasher.py
dongbohu: I agree with @kurtwheeler. It will be much easier to review and maintain the content of `BODY_HTML` if it is in a separate html file.

By the way, I copied the HTML into another file and checked it in firefix. The web UI looks pretty simple:
screenshot from 2018-08-24 12-00-47
Why the html code has over 1500 lines with multiple tables and all kinds of CSS styles?

workers/data_refinery_workers/processors/smasher_email.html
```diff
@@ -0,0 +1,1501 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
```
dongbohu: Not sure whether you saw my previous comment. So I am posting it here again:
Here is how this web page looks like:
screenshot from 2018-08-24 12-00-47
It looks very simple. Are you sure you need all those CSS lines and tables in the code?

workers/data_refinery_workers/processors/smasher.py
Miserlou: Email has been moved to a file.

Email HTML and Web HTML are very different beasts. Different mail services and clients have different constraints. So, all style has to be inlined, the HTML feature set is limited, etc. etc. etc. We use Foundation Email and Zurb Inliner to make this. I was given this as an asset by Ariel/Deepa.

workers/data_refinery_workers/processors/smasher_email.html
```diff
@@ -0,0 +1,1501 @@
...
+                </tr>
+              </tbody>
+            </table>
+          </center>
+        </td>
+      </tr>
+    </table>
+  </body>
+
+</html>
```
dongbohu: Better to have a new line char at the end.

workers/data_refinery_workers/processors/smasher_email.html
```diff
@@ -0,0 +1,1501 @@
...
+      }
+    </style>
+  </head>
+
+  <body>
+    <!-- <style> -->
+    <table class="body" data-made-with-foundation="">
+      <tr>
+        <td class="float-center" align="center" valign="top">
+          <center data-parsed="">
```
dongbohu: `<center>` is not supported in html5. You may want to use CSS.

workers/data_refinery_workers/processors/smasher_email.html
```diff
@@ -0,0 +1,1501 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
+<html xmlns="http://www.w3.org/1999/xhtml">
+
+  <head>
+    <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
+    <meta name="viewport" content="width=device-width">
+    <title>Title</title>
+    <style>
```
dongbohu: I took a closer look at the style sheet in this file. You may want to correct the indentations, and some of the lines are redundant. For example, `<h6>` is not being used, but the style has some customization. Maybe we should ask @dvenprasad or @arielsvn to simplify this file?

workers/data_refinery_workers/processors/smasher_email.html
arielsvn: @dvenprasad built this template using foundation I think, and that was the recommended structure. `center` might have better support with old email clients.

Miserlou: The point is that we are not targeting only HTML5. The DOCTYPE is `XHTML 1.0 Strict` but even that is generally outside of our control. You'll notice that there are also centering `class` and `align` tags above - the whole point here is that we're trying to target all of the different target clients we can, which aren't standardized. HTML email is still like the old wild west days of the web.

workers/data_refinery_workers/processors/smasher_email.html
Miserlou: @arielsvn you can make your changes directly to this branch, just be mindful of the `REPLACEME` links

workers/data_refinery_workers/processors/smasher_email.html
dongbohu: If Deepa is using some tools to generate this page automatically, then I guess we don't need to review the HTML code at all. Am I right?

arielsvn: I think so, still I'm looking into the logo and the button not being centered. 

Should we minify the code of the template??

arielsvn:
@Miserlou @dongbohu I updated the email template, and also added a minified version `smasher_email.min.html`. This is the one that should be used in production since it has all the styles inlined. Both have the `REPLACEME` links.

PR #562, opened 2018-08-29
Sets the ulimit option for files for nomad clients.

kurtwheeler:
## Issue Number

N/A came up during crunch

During the zebrafish crunch, many jobs failed due to 

```
Driver Failure  failed to start task "affy_to_pcl" for alloc "8882c9e9-6f1f-6270-dc0b-910de0f527e5": error creating rpc client for executor plugin: pipe2: too many open files
```

[Here,](https://github.com/hashicorp/nomad/issues/3686#issuecomment-372283062) someone describes an essential property we should set. I have done so in this PR.

## Types of changes

## Functional tests

N/A

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -76,7 +76,6 @@ jobs:
   affymetrix_common_agilent_tests:
     working_directory: ~/refinebio
     machine:
-      docker_layer_caching: true
       pre:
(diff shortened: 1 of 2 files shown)


kurtwheeler:
I also just pushed a commit to try to fix the build here as well.

Miserlou:
https://www.youtube.com/watch?v=RkEXGgdqMz8

PR #585, opened 2018-09-05
Make the get_volume_index function return a default value that is valid.

kurtwheeler:
## Issue Number

N/A Ariel was trying to test the new emails and could not trigger a smash job:

https://sentry.io/greenelab/staging-refinebio-api/issues/612495404/?query=is:unresolved

> Dispatching Nomad job of type ProcessorPipeline.SMASHER for job spec SMASHER_0_4096 to host 10.0.167.44 and port 4646 failed.

I fixed this in two ways:
  * First I changed the default value of get_volume_index so that it returns an index that is actually used. We were returning 0 by default when the volume indices go from 1-10. The API doesn't have that set up, so it always chooses the default.
  * However this means that if the system is at rest, i.e. no spot instances are spun up, then we have to spin up instances until one of them chooses to use the EBS volume with index 1. This means that for a single smash job we could have to spin up 10 instances. However the smash job is not preceded by a downloader job and it has to download all of its data anyway so there's really no need for it to be tied to a volume index. Therefore I have also made it so that it's not parameterized on a volume index. Also we're just always using 4096 memory for smasher jobs so I took out it's different RAM amounts. If we end up needing them later, we can add a special case to make an unvolume-indexed-yet-RAM-multiplexed set of nomad jobs for the smasher then.

## Types of changes

## Functional tests

I was able to queue Ariel's smasher job by setting the volume_index of the job to '1' before sending (via the django shell) so I know the default volume was the problem.

I've also ran `run_nomad.sh` to make sure the smasher nomad job gets templated correctly.

## Checklist

### DIFF
--- a/api/data_refinery_api/views.py
+++ b/api/data_refinery_api/views.py
@@ -120,10 +120,10 @@ class ExperimentFilter(django_filters.FilterSet):
 
     class Meta:
         model = Experiment
-        fields = ['has_publication', 
-                        'submitter_institution', 
+        fields = ['has_publication',
+                        'submitter_institution',
                         'technology',
-                        'source_first_published', 
+                        'source_first_published',
                         'organisms__name',
                         'samples__platform_accession_code']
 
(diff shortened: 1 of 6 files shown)


Miserlou:
This looks great to me, excellent catch. I'm not 100% sure about the change to remove the RAM striation since we haven't realllly investigated Smasher RAM usage, but I don't think it'll be a problem in the short term and we can put it back if we want to make different smash jobs more RAM aware in the future.

Did Ariel get an email about the failure, or was it only reconstructed from the logs?

kurtwheeler:
> Did Ariel get an email about the failure, or was it only reconstructed from the logs?

He found it from the sentry error. I would imagine our system didn't send any email because the smasher jobs are what send the emails and it couldn't even start because it couldn't get queued.

Miserlou:
Ah, right, that's what I figured. Okay, a think to be mindful of for manual or automated integration testing.

Miserlou:
I'm actually more concerned that this error didn't float up sooner. If the error occurred in send_job, then the API should have returned an error status code and the frontend should have been able to handle that. I don't like user-waits-for-an-email-that-never-comes scenarios.

kurtwheeler:
Mmm good point, I just pushed a commit to address that.

Miserlou:
@arielsvn - can we support displaying an error message in this scenario on the front end?

arielsvn:
> can we support displaying an error message in this scenario on the front end?

Yes I think that's a good idea, but how do you suggest that we do it??

The problem is the call to start the dataset doesn't fail when this error happens. What do you think of adding a `has_error` field to the dataset, and then while the dataset is being processed, I can fetch this information every X segs, and show a message when `has_error == true`.

Miserlou:
Kurt's above change should make it such that it'll error code when failing to dispatch: https://github.com/AlexsLemonade/refinebio/pull/585/commits/1fdfa4fc3f2191c20802f74940f47777d0ade4c1#diff-6f07013017b6975f386a8b609128ca1dR275

PR #588, opened 2018-09-06
Purge EFS, Better eu-west-2 support

Miserlou:
## Issue Number

n/a

Purges last remnants of EFS, adds better support for the `eu-west-2` region by adding logic to the `aws_ami` TF object.


Miserlou:
It doesn't, but we have updated the client instance script to install it already.

PR #590, opened 2018-09-06
Heal Our PGP Wound - R Documentation's Fault

Miserlou:
@jwilk [burned us](http://keyserver.ubuntu.com:11371/pks/lookup?op=get&search=0x4359ED62E084DAB9).

Uses a longer key identifier (the good `0x51716619E084DAB9`, not the naughty `0x4359ED62E084DAB9`).


jwilk:
Duplicating 64-bit key IDs is not trivial, but it could be still within budget of a large criminal organization.
Please consider using full fingerprint instead, i.e.: `E298A3A825C0D65DFD57CBB651716619E084DAB9`

Miserlou:
You got it! Thank you for disclosing this in an attention-getting but not-malicious way!

PR #603, opened 2018-09-10
[WIP] Adds a temporary non-spot client instance so the smasher can run.

kurtwheeler:
## Issue Number

Getting downloads from the UI is broken

We can't give people their downloads because we can't run smasher jobs because we don't have any instances up to run them. This adds a single instance to stay up so it can run smasher jobs. We will add a more permanent solution to this issue later this week.

## Types of changes

## Functional tests

N/A yet, I gotta spin this up to make sure it works still.

## Checklist


kurtwheeler:
@Miserlou could you give this a quick look to make sure there's nothing obvious before I start a lengthly devops test for this?

Issue #530, opened 2018-08-15
Track version number using Docker's ARG instruction for Dockerfiles


kurtwheeler:
### Context

After running the Zebrafish Crunch @cgreene requested we enable sentry's releases feature like so:

```
RAVEN_CONFIG = {
    'dsn': 'REDACTED',
    # If you are using git, you can also automatically configure the
    # release based on the git info.
    'release': raven.fetch_git_sha(os.path.abspath(os.pardir)),
}
```

### Problem or idea

The problem is that we don't run our project from a git repo. We copy the relevant code into a Docker image, then run that. The git hash is not available from within the docker container.

We can put the git hash into the docker container using the [ARG instruction](https://docs.docker.com/engine/reference/builder/#arg). However if we're going to do that, we may as well just use the same instruction to manage our version number.

This would solve a couple other issues I've been concerned about for a bit:
* We have `_version.py` files in our project we used at one point, I'm not even sure if they're used anymore but because they're in the code they haven't been getting updated. Keeping the version actually in the code means that in order to do a release, we need to update the code first.
* We specify docker images in our Nomad Job Specifications. However we are always using the `latest` tag. This means that if a new image is pushed, the latest tag will probably update and we could end up running new code without running a deployment. Specifying exactly the tag we want to run will mean that unless we repush to that tag, the image won't change under us. We have code in our deploy process that prevents us from overwriting existing docker images for this reason.

### Solution or next step

In our code that updates the docker images, we should pass the version as an ARG to `docker build` and use that for the sentry release, our system version number, and then when we format the Nomad Job Spec templates, specify that version as the tag for the docker image.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

jaclyn-taroni:
Closed by #606 @kurtwheeler ?

kurtwheeler:
Yes!

PR #606, opened 2018-09-11
Kurtwheeler/docker system version

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/530

We haven't been keeping our system version numbers up to date. This builds them into our deploy workflow.

It does this by using the CIRCLE_TAG as the version which is how we kick off deployments. In non-deployment scenarios we instead default to a version of `local`.

## Types of changes

## Functional tests

The tests pass. I won't actually know if the deployment process works with this until we try it but it should work.

## Checklist


common/setup.py
```diff
@@ -5,19 +5,14 @@
...
-    version_string = match_object.group(1)
-else:
-    raise RuntimeError("Unable to find version string in %s." % (VERSION_FILE,))
+VERSION_FILE = "version"
+try:
+    with open(VERSION_FILE, "rt") as version_file:
+        version_string = version_file.read().strip()
+except:
+    print("Cannot read version.config to determine System Version."
+          " Please create a file common/version.config containing an up to date System Version.")
```
Miserlou: This should be `common/version`, not .config

prepare_image.sh
```diff
@@ -68,6 +68,11 @@ if [[ -z $dockerhub_repo ]]; then
     dockerhub_repo="ccdlstaging"
 fi
 
+# Default to "local" for system version if we're not running in the cloud.
+if [[ -z $SYSTEM_VERSION ]]; then
+    SYSTEM_VERSION="local"
```
Miserlou: Can we make this something like `local + \`date +%s\``

update_models.sh
```diff
@@ -7,4 +7,12 @@ if ! docker ps | tail -n +2 | awk '{ print $NF }' | grep drdb > /dev/null; then
     exit 1
 fi
 
+# Default to "local" for system version if we're not running in the cloud.
+if [[ -z $SYSTEM_VERSION ]]; then
+    export SYSTEM_VERSION="local"
```
Miserlou: Same again

workers/setup.py
```diff
@@ -6,19 +6,15 @@
...
-# code found at: https://stackoverflow.com/a/7071358/6095378
-VERSION_FILE = "data_refinery_workers/_version.py"
-version_line = open(VERSION_FILE, "rt").read()
+VERSION_FILE = "version"
+try:
+    with open(VERSION_FILE, "rt") as version_file:
+        version_string = version_file.read().strip()
+except:
+    print("Cannot read version.config to determine System Version."
+          " Please create a file workers/version.config containing an up to date System Version.")
```
Miserlou: Not .config again

PR #607, opened 2018-09-12
Adds updating email_address field without sending it back out to the …

Miserlou:
…serializer

## Issue Number

n/a

Fixes setting this field, which was changed for security reasons, without re-implementing the problem.

## Functional tests

New assertions

## Checklist

### DIFF
--- a/api/data_refinery_api/views.py
+++ b/api/data_refinery_api/views.py
@@ -213,6 +213,9 @@ class CreateDatasetView(generics.CreateAPIView):
 class DatasetView(generics.RetrieveUpdateAPIView):
     """ View and modify a single Dataset. Set `start` to `true` along with a valid
     activated API token (from /token/) to begin smashing and delivery.
+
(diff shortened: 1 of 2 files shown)


api/data_refinery_api/views.py
```diff
@@ -258,6 +264,11 @@ def perform_update(self, serializer):
                 pjda.save()
 
                 job_sent = False
+
+                obj = serializer.save()
+                if obj.email_address != supplied_email_address:
```
kurtwheeler: Could we add a check to make sure supplied_email_address isn't empty? This would make it possible to accidentally overwrite the email address if it was set in a previous request.

Miserlou: That'd also prevent you from removing an email address. I've been operating with the idea that it's okay not to supply an email if you just want to use the website and not be notified. If we're going to require an email address, we should put some bounds on that and require it everywhere.

PR #617, opened 2018-09-14
Bumps nomad timeout, adds a few missing dependencies, speeds up Nomad…

kurtwheeler:
… job killing in deploy process.

## Issue Number

N/A came up during crunch.

Most of these are pretty minor/obvious things. The one significant change is in `deploy.sh`. This makes it so we kill all the Nomad jobs in the background and then wait for them so we can do it a lot faster. 

Here is an image of the CPU utilization of the lead nomad server: 
image

The long period at the start at ~18% was us killing them one by one. After the change to `deploy.sh` you can see it peak up to the mid fifties as it kills all the jobs very quickly. I think the last high point which then trails off is probably us re-registering the base jobs.

## Types of changes

## Functional tests

I have tested all of this by running it against production.

## Checklist


Miserlou:
Is this PR going to including killing the telemetry stuff? Ideally just a comment out since it shouuuuuld hopefully be improved in a future Nomad, but we want to disable it for prod for now.

kurtwheeler:
I have added commenting out the graphite code for Nomad lead server along with reverting the EBS-volume discovery changes I made.

Miserlou:
Related: https://github.com/hashicorp/nomad/issues/4422

Miserlou:
You have removed the telemetry statsd servers, but not the telemetry configuration from Nomad itself

PR #626, opened 2018-09-18
give fasterqdump its own tempdir

Miserlou:
Discussed IRL, prevents fasterq-dump from using the local FS.


Miserlou:
Also implements a bizarre suggestion from SRA themselves, so let's see..

PR #632, opened 2018-09-19
Move to Hard Limited Pagination on Jobs, add Harakiri

Miserlou:
Ariel thinks he can reliably crash the API server by sending it requests for gigantic querysets. This sets a hard limit on the request size with `max`, and makes the `uswgi` settings more likely to end requests which take too long.

### DIFF
--- a/api/data_refinery_api/views.py
+++ b/api/data_refinery_api/views.py
@@ -561,21 +561,17 @@ class SurveyJobList(PaginatedAPIView):
 	  - ?start_time__lte=2018-03-23T15:29:40.848381Z&start_time__gte=2018-03-23T14:29:40.848381Z
 	  - ?success=True
 
+      Works with required 'limit' and 'offset' params.
+
     """
 
     def get(self, request, format=None):
         filter_dict = request.query_params.dict()
-        filter_dict.pop('limit', None)
-        filter_dict.pop('offset', None)
-        jobs = SurveyJob.objects.filter(**filter_dict)
-
-        page = self.paginate_queryset(jobs)
-        if page is not None:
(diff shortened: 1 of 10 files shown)


api/data_refinery_api/views.py
```diff
@@ -603,17 +593,11 @@ class ProcessorJobList(PaginatedAPIView):
...
-        jobs = ProcessorJob.objects.filter(**filter_dict)
-
-        page = self.paginate_queryset(jobs)
-        if page is not None:
-            serializer = ProcessorJobSerializer(page, many=True)
-            return self.get_paginated_response(serializer.data)
-        else:
-            serializer = ProcessorJobSerializer(jobs, many=True)
-            return Response(serializer.data)
+        limit = max(int(filter_dict.pop('limit', 100)), 100)
```
arielsvn: I think it's a good idea to add a maximum for `limit` but this might affect the dashboard.

Right now we're requesting ALL Survey/Processor/Downloader jobs in a specified time interval, and then calculating their state in the front-end (pending/failed/completed/open). At some point we'll probably have many jobs and this may not be feasible.

Miserlou:
We already have way too many jobs for that approach to be reasonable,
depending on what the time interval is. 25,000 downloader jobs created in
an hour is perfectly normal now.

arielsvn:
> We already have way too many jobs for that approach to be reasonable, depending on what the time interval is. 25,000 downloader jobs created in an hour is perfectly normal now.

So what do you suggest we do in the dashboard? the only solution that I can think is moving these calculations to the backend. Either to the stats endpoint or to another one

Issue #608, opened 2018-09-12
We need to link samples to the OrganismIndex that was used to process them.


kurtwheeler:
### Context

We build OrganismIndex objects for every organism and use them when we run salmon. However we will have multiple versions of these and we should have a link from the sample to the OrganismIndex.

### Problem or idea

The link doesn't necessarily need to be between the sample and the OrganismIndex, it could instead be to the computed file. However the upshot is that if I want to know what OrganismIndex object was used to process a certain sample, I should be able to determine that.

### Solution or next step

Create the link!

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

jaclyn-taroni:
@kurtwheeler if we wanted to use this sample information on the front end, would it require changes to the API? cc @arielsvn  I believe we're currently using the most recent transcriptome index assembly (https://github.com/AlexsLemonade/refinebio-frontend/pull/280/commits/19c17e32d611338a142ff3f94802d125a59fa31a)

Miserlou:
We can tackle this at the same time: https://github.com/AlexsLemonade/refinebio/issues/460

jaclyn-taroni:
Closed by #636

PR #636, opened 2018-09-21
Link samples to the organism index that was used to process them if a…

kurtwheeler:
…plicable.

## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/608

We don't have a way to tell which organism index was used to process a sample with salmon. This fixes that by creating a link.

## Types of changes

## Functional tests

The tests cover this case fairly well.

## Checklist


dongbohu:
Don't we need to set this `organism_index` field for each sample instance?

dongbohu:
Never mind, I noticed that it was set in `salmon.py`.

PR #638, opened 2018-09-21
Modify stats endpoint to include all data we need on the dashboard

arielsvn:
## Issue Number

https://github.com/AlexsLemonade/refinebio-frontend/issues/140

This adds all the statics required by the dashboard to the `/stats` endpoint.

When the parameter `range` is included (can be one of `day`, `week`, `month` or `year`), the endpoint will also return statics for several time intervals that we use to build the graphs in the front-end.

I didn't add any tests for this, should we?

## Types of changes

## Functional tests

Check the dashboard in the front-end.

## Checklist

## Screenshots

This is the dashboard running locally.

image

### DIFF
--- a/api/data_refinery_api/views.py
+++ b/api/data_refinery_api/views.py
@@ -699,11 +699,16 @@ class Stats(APIView):
         return [self._get_job_interval(jobs, start, end) for (start, end) in self._get_time_intervals(range_param)]
 
     def _created_timeline(self, objects, range_param):
-        return [({
-            'start': start, 
-            'end': end, 
-            'total': objects.filter(created_at__gte=start, created_at__lte=end).count()
-        }) for (start, end) in self._get_time_intervals(range_param)]
+        results = []
+        for start, end in self._get_time_intervals(range_param):
+            total = objects.filter(created_at__gte=start, created_at__lte=end).count()
+            stats = {
+                'start': start, 
+                'end': end, 
+                'total': total
+            }
+            results.append(stats)
+        return results
 
 ###
 # Transcriptome Indices
--- a/api/data_refinery_api/views.py
+++ b/api/data_refinery_api/views.py
@@ -57,6 +57,9 @@ from data_refinery_api.serializers import (
     APITokenSerializer
 )
 
+import datetime as dt
+from django.utils import timezone
+
 ##
 # Custom Views
 ##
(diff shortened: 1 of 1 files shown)


api/data_refinery_api/views.py
```diff
@@ -57,6 +57,9 @@
     APITokenSerializer
 )
 
+import datetime as dt
```
Miserlou: Generally avoid doing this please

api/data_refinery_api/views.py
```diff
@@ -606,34 +609,99 @@ def get(self, request, format=None):
...
+            'total': jobs.count(),
+            'pending': jobs.filter(start_time__isnull=True).count(),
+            'completed': jobs.filter(end_time__isnull=False).count(),
+            'open': jobs.filter(start_time__isnull=False, end_time__isnull=True).count(),
+            # via https://stackoverflow.com/questions/32520655/get-average-of-difference-of-datetime-fields-in-django
+            'average_time': jobs.filter(start_time__isnull=False, end_time__isnull=False).aggregate(
+                average_time=Avg(F('end_time') - F('start_time')))['average_time']
+        }
+
+        if result['average_time'] is None:
```
Miserlou: This can be refactored to `if not result['average_time']` etc

api/data_refinery_api/views.py
```diff
@@ -606,34 +609,99 @@ def get(self, request, format=None):
...
+            result['average_time'] = 0
+        else:
+            result['average_time'] = result['average_time'].total_seconds()
+
+        if range_param is not None:
```
Miserlou: same again

api/data_refinery_api/views.py
```diff
@@ -606,34 +609,99 @@ def get(self, request, format=None):
...
+    interval_timestep = {
+        'day': dt.timedelta(hours=1),
+        'week': dt.timedelta(days=1),
+        'month': dt.timedelta(days=2),
+        'year': dt.timedelta(weeks=4)
+    }
+
+    def _get_time_intervals(self, range_param):
+        current_date = dt.datetime.now(tz=timezone.utc)
+        time_step = self.interval_timestep.get(range_param)
```
Miserlou: This is a bit of weird pattern, but it's okay, just move the `interval_*` objects before the function definitions - however, since they're only used be a single function, you can just declare them at the start of the function

api/data_refinery_api/views.py
```diff
@@ -606,34 +609,99 @@ def get(self, request, format=None):
...
+            'pending': pending.count(),
+            'failed': failed.count(),
+            'open': open.count()
+        }
+
+    def _jobs_timeline(self, jobs, range_param):
+        return [self._get_job_interval(jobs, start, end) for (start, end) in self._get_time_intervals(range_param)]
+
+    def _created_timeline(self, objects, range_param):
+        return [({'start': start, 'end': end, 'total': objects.filter(created_at__gte=start, created_at__lte=end).count()}) for (start, end) in self._get_time_intervals(range_param)]
```
Miserlou: This and the return line above it have bad [JonesComplexity](https://github.com/Miserlou/JonesComplexity)

arielsvn:
@Miserlou thanks for the review! I addressed your comments. Let me know if you have any other tips on how to make this better.

api/data_refinery_api/views.py
Miserlou: I think the simplest way to make this line more maintainable is something like

```
        results = []
        for start, end in self._get_time_intervals(range_param):
            total = objects.filter(created_at__gte=start, created_at__lte=end).count()
            stats = {   'start': start, 
                        'end': end, 
                        'total': total
                    }
            results.append(stats)

```

arielsvn: Good Idea! Thanks!

PR #639, opened 2018-09-24
Kurtwheeler/fix tx index race

kurtwheeler:
## Issue Number

N/A came up during crunch

I thought that having one processor job download and install the transcriptome index would be enough, and that then the rest could use the index that job installed. This is true, but it takes a while so the other jobs were getting to processing before the first one properly installed the index. Therefore this makes all jobs try to install the index until one of them finishes the process and creates symlinks denoting that it's done.

This also bumps up the downloader RAM requirement to 12GB.

## Types of changes

## Functional tests

I have tested this by running some processors in the dev stack.

## Checklist

### DIFF
--- a/infrastructure/nomad-configuration/client-instance-user-data.tpl.sh
+++ b/infrastructure/nomad-configuration/client-instance-user-data.tpl.sh
@@ -46,10 +46,11 @@ while [  $COUNTER -lt 99 ]; do
 done
 
 sleep 15
-ATTACHED_AS=`lsblk -n | grep 500G | cut -d' ' -f1`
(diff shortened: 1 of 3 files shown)


infrastructure/nomad-configuration/client-instance-user-data.tpl.sh
```diff
@@ -46,10 +46,11 @@ while [  $COUNTER -lt 99 ]; do
 done
 
 sleep 15
-ATTACHED_AS=`lsblk -n | grep 500G | cut -d' ' -f1`
+ATTACHED_AS=`lsblk -n | grep 8.8T | cut -d' ' -f1`
```
Miserlou: If you're going to change this value in the code can you also change it in the TF

PR #641, opened 2018-09-24
Correctly Mark RNA-Seq is_processed Before tximport, RDS Backups in Prod, Send Fail Notification Upon QN Failure

Miserlou:
## Issue Number
na

Some get-ready-for-prod changes, and some changes to the Smasher logic based on using the results from the current dev deployment. 

## Methods

Previously a failure to QN would send the un-QN'd dataset, not we send the fail message. "It is dangerous to send people un-QN'd data".

## Types of changes


workers/data_refinery_workers/processors/utils.py
```diff
@@ -121,16 +121,23 @@ def end_job(job_context: Dict, abort=False):
...
+                # This handles most of our cases
+                for sample in job_context["samples"]:
+                    sample.is_processed = True
+                    sample.save()
+
+                # Explicitly for the single-salmon scenario
+                if 'sample' in job_context:
+                    sample = job_context['sample']
+                    sample.is_processed = True
+                    sample.save()
```
kurtwheeler: So I think that for Salmon, `job_context["samples"] == [job_context['sample']]` because both of those fields are populated based on the original files associated with the job, which I believe only relate to a single sample. I think what you want to be doing here is looping over all the samples associated with all of the experiments returned by `_get_tximport_inputs()`.

Issue #610, opened 2018-09-12
Aggregated by Experiment: Change file names


dvenprasad:
### Context

This came up when @jaclyn-taroni and I were chatting about illustrations for download files (for docs)

### Problem or idea

Currently, all the files are just dumped in the download folder. If someone downloads 10 experiments, it will be messy for them find their files. The files should be grouped together in folders based on Experiment. 

For example:
![agg-experiments](https://user-images.githubusercontent.com/15315514/45447294-18620d80-b69d-11e8-9f07-496236ba8361.png)

Related to #609. But this should be given a higher priority. 

### Solution or next step

Tagging @Miserlou for input on how best move this forward

jaclyn-taroni:
Grabbed data from the dev stack this morning. The structure looks good. File names are a bit different. 
```
single_species_by_experiment_example/
├── E-GEOD-39842
│   ├── E-GEOD-39842_metadata.tsv
│   └── E-GEOD-39842.tsv
├── LICENSE.TXT
├── metadata.json
├── README.md
├── SRP062067
│   └── SRP062067_metadata.tsv
└── SRP111705
    └── SRP111705_metadata.tsv

3 directories, 7 files
```
`metadata.json` -> `aggregated_metadata.json`
`SRP111705_metadata.tsv` -> `metadata_SRP111705.tsv`

cgreene:
Is there a reason to put the word metadata first? I am imagining a world in which I am sorting by file name, and it seems like having the experiment/metadata pair show up together (instead of all the experiments followed by all the metadata) would be preferred. I may not be thinking of why having `metadata_` first is better though. I realize that is the way it is designed, but I'm just trying to think about why.

jaclyn-taroni:
Question for @dvenprasad

Miserlou:
That's what I was thinking @casey

jaclyn-taroni:
I will say after working with a bunch of this data this weekend with the new folder structure, I do think putting `metadata` first in the file name would reduce some friction.

dongbohu:
@jaclyn-taroni and @dvenprasad: So why when the dataset is aggregated by species, we want to have species-level json file, but when it is aggregated by experiment, we don't need experiement-level json file?

dvenprasad:
The folder structure already groups the files by experiment. Adding ```meatadata``` first will help them distinguish between the files for the same experiment easily.

dongbohu:
[...]

Issue #609, opened 2018-09-12
Aggregated by Species: Organize files in folders


dvenprasad:
### Context

This came up when @jaclyn-taroni and I were chatting about illustrations for download files (for docs)

### Problem or idea

Currently, all the files are just dumped in the download folder. The files should be grouped together in folders based on species. 

For example:
![agg-species](https://user-images.githubusercontent.com/15315514/45446641-70981000-b69b-11e8-8d4e-a3e17a4ee3e5.png)

### Solution or next step

Tagging @Miserlou for input on how best move this forward

jaclyn-taroni:
I obtained files from the dev stack this morning. The files are now grouped into folders, but there are some differences from what is in @dvenprasad's illustration here.

**Multispecies aggregated by species example:**
```
multispecies_by_species_example/
├── DANIO_RERIO
│   ├── DANIO_RERIO_metadata.tsv
│   └── DANIO_RERIO.tsv
├── GALLUS_GALLUS
│   ├── GALLUS_GALLUS_metadata.tsv
│   └── GALLUS_GALLUS.tsv
├── LICENSE.TXT
├── metadata.json
├── MUS_MUSCULUS
│   ├── MUS_MUSCULUS_metadata.tsv
│   └── MUS_MUSCULUS.tsv
└── README.md

3 directories, 9 files
``` 
**Single species aggregated by species example:**
```
single_species_by_species_example/
├── DANIO_RERIO
│   ├── DANIO_RERIO_metadata.tsv
│   └── DANIO_RERIO.tsv
├── LICENSE.TXT
├── metadata.json
└── README.md

1 directory, 5 files
```
Specifically:
* There are no species level metadata json files
* The "overall" metadata json file is `metadata.json` rather than `aggregated_metadata.json`

dongbohu:
@jaclyn-taroni and @dvenprasad: I can think of two possible structures of the species-level json file. One is simply an array like this:
```
[ 
  { 1st_sample_metadata }, 
  { 2nd_sample_metadata }, 
  ... 
]
```

The other option is to add some extra fields and make it a dictionary:
```
{
  species_name: "homo sapiens",
  created_at: "...",
  # ... other fields ...

  samples: [
    { 1st_sample_metadata ... }, 
    { 2nd_sample_metadata ... }, 
    ... 
  ]
}
```
Which one do you prefer?

dvenprasad:
[...]

PR #648, opened 2018-09-25
Fine tuning of smasher's output filenames

dongbohu:
## Issue Number

#609 
#610 

The files generated by smasher are not named exactly like what they were supposed to be. This PR corrects them. 

## Types of changes

## Checklist


workers/data_refinery_workers/processors/smasher.py
```diff
@@ -264,18 +267,30 @@ def _write_tsv(job_context, metadata, smash_path):
...
                         row_data = _get_tsv_row_data(sample_metadata)
                         dw.writerow(row_data)
+                        samples_in_species.append(sample_metadata)
+
+            # Writes a json file for current species:
+            if len(samples_in_species):
+                species_metadata = {
+                    'species': species,
+                    'samples': samples_in_species,
+                    'created_at': datetime.utcnow().strftime('%Y-%m-%dT%H:%M:%S')
```
Miserlou: I'm not sure about this keyvalue, as it it will look like that's when the samples were created, not the metadata, and it will change slightly throughout the loop. I think we can remove it unless it's been specifically asked for?

dongbohu: That's a good point. I added it mainly because it was added in the json file that includes all samples:
https://github.com/AlexsLemonade/refinebio/blob/cbef527aca99f9d30efd8ebd5d8aa739d1002d46/workers/data_refinery_workers/processors/smasher.py#L612
What do you think @dvenprasad?

dongbohu: :+1:

PR #650, opened 2018-09-25
Add Support for Smashing `SRP051449`

Miserlou:
## Issue Number

n/a

We can't rely on externally supplied `title` fields for smash column titles as they aren't guaranteed to be unique. Switches to use `accession_code`, which is. Also fixes an email bug case.

## Types of changes

## Functional tests

New test files and assertions added.

## Checklist

### DIFF
--- a/workers/data_refinery_workers/processors/smasher.py
+++ b/workers/data_refinery_workers/processors/smasher.py
@@ -497,10 +497,11 @@ def _smash(job_context: Dict) -> Dict:
                         ks_test = rlang("ks.test")
 
                         set_seed(123)
-                        n = ncol(reso)
+
+                        n = ncol(reso)[0]
                         m = 2
                         if n < m:
-                            raise Exception("Found fewer columns that required for QN combinatorial - bad smash?")
+                            raise Exception("Found fewer columns than required for QN combinatorial - bad smash?")
                         combos = combn(ncol(reso), 2)
 
                         # Convert to NP, Shuffle, Return to R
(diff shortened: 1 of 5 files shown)


workers/data_refinery_workers/processors/smasher.py
```diff
@@ -393,14 +393,14 @@ def _smash(job_context: Dict) -> Dict:
 
                     # Explicitly title this dataframe
                     try:
-                        data.columns = [computed_file.samples.all()[0].title]
+                        # Unfortuantely, we can't use this as `title` can cause a collision
+                        #data.columns = [computed_file.samples.all()[0].title]
+                        # So we use this, which also helps us support the case of missing SampleComputedFileAssociation
+                        data.columns = [computed_file.filename]
                     except ValueError:
```
kurtwheeler: Is this except block still necessary? I think the only exception that could be raised here would be an AttributeError.

workers/data_refinery_workers/processors/smasher.py
```diff
@@ -495,6 +497,10 @@ def _smash(job_context: Dict) -> Dict:
                         ks_test = rlang("ks.test")
 
                         set_seed(123)
+                        n = ncol(reso)
+                        m = 2
+                        if n < m:
```
kurtwheeler: This could just be `if ncol(reso) < 2:` right?

workers/data_refinery_workers/processors/smasher.py
```diff
@@ -495,6 +497,10 @@ def _smash(job_context: Dict) -> Dict:
...
+                            raise Exception("Found fewer columns that required for QN combinatorial - bad smash?")
```
kurtwheeler: `than required`*

workers/data_refinery_workers/processors/utils.py
```diff
@@ -152,7 +152,7 @@ def end_job(job_context: Dict, abort=False):
         # S3-sync Computed Files
         for computed_file in job_context.get('computed_files', []):
             # Ensure even distribution across S3 servers
-            nonce = ''.join(random.choice(string.ascii_lowercase + string.digits) for _ in range(8))
+            nonce = ''.join(random.choice(string.ascii_lowercase + string.digits) for _ in range(24))
```
kurtwheeler: Just curious why this change?

workers/data_refinery_workers/processors/test_smasher.py
```diff
@@ -610,6 +610,103 @@ def test_no_smash_dupe(self):
         for column in final_context['original_merged'].columns:
             self.assertTrue('_x' not in column)
 
+    @tag("smasher")
+    def test_no_smash_dupe_two(self):
```
kurtwheeler: I'm just wondering, what's this testing that isn't already tested? I'm not quite sure what the test function's name means.

jaclyn-taroni:
👍

workers/data_refinery_workers/processors/test_smasher.py
Miserlou: The test reflects an actual situation encountered where the tiles are the same, it also uses a QN target and has qn_normalize `true` on the dset

workers/data_refinery_workers/processors/utils.py
Miserlou: 3 million titles.. just wanted to expand the space

workers/data_refinery_workers/processors/smasher.py
Miserlou: Had to add a [0] to this anyway

workers/data_refinery_workers/processors/smasher.py
Miserlou: Changed to base exception, which I think is necessary for two-channel samples

workers/data_refinery_workers/processors/test_smasher.py
kurtwheeler: Ah okay. I don't think there's anything in the code which explains this. The code looks like it's operating on two different files. If you could add a comment or docstring that would be helpful.

PR #651, opened 2018-09-25
Prevent a very stupid deploy failure.

kurtwheeler:
## Issue Number

N/A Came up during master deploy.

This is a nice check to have, but it needed an `-f` flag.

## Types of changes

## Functional tests

N/A

## Checklist


.circleci/update_docker_img.sh
```diff
@@ -32,7 +32,7 @@ CCDL_WORKER_IMGS="smasher illumina affymetrix salmon transcriptome no_op downloa
 # Create ~/refinebio/common/dist/data-refinery-common-*.tar.gz, which is
 # required by the workers and data_refinery_foreman images.
 ## Remove old common distributions if they exist
-rm ~/refinebio/common/dist/*
+rm -f ~/refinebio/common/dist/*
```
dongbohu: Do you think it makes more sense to use `rm -rf ...` here? Otherwise the command will complain if there are any sub-directory in `dist/`?

Issue #683, opened 2018-10-01
Silent FastQC Failures


Miserlou:
```
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod 2018-09-29 23:03:24,101 i-094075f4b46b5ec64 data_refinery_workers.processors.salmon ERROR [processor_job: 9187]: Shell call to FastQC failed with error message: Started analysis of ERR1631915.fastq
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod 2018-09-29 23:03:24,279 i-094075f4b46b5ec64 data_refinery_workers.processors.utils ERROR [failure_reason: Started analysis of ERR1631915.fastq] [processor_job: 9187]: Processor function _run_fastqc failed. Terminating pipeline.
```

Not a whole lot to go with here - my guess is that this is a memory issue again.

Hopefully our strategy of increasingly large Salmon placements will address all of these issues together.

Issue #681, opened 2018-10-01
File Not Found During SHA1 Check


Miserlou:
### Context
```
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod 2018-09-30 09:53:19,387 i-03b2b1628573bd053 data_refinery_workers.processors.utils ERROR [processor_job: 6648] [no_retry: False]: Unhandled exception caught while running processor function _run_salmon in pipeline:
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod Traceback (most recent call last):
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod   File "/home/user/data_refinery_workers/processors/utils.py", line 231, in run_pipeline
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod     last_result = processor(last_result)
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod   File "/home/user/data_refinery_workers/processors/salmon.py", line 675, in _run_salmon
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod     quant_file.calculate_sha1()
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod   File "/usr/local/lib/python3.5/dist-packages/data_refinery_common/models/models.py", line 759, in calculate_sha1
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod     with open(self.absolute_file_path, mode='rb') as open_file:
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod FileNotFoundError: [Errno 2] No such file or directory: '/home/user/data_store/processor_job_6648/SRR6025586_output/quant.sf'
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod 2018-09-30 09:53:25,258 i-03b2b1628573bd053 data_refinery_workers.processors.utils ERROR [failure_reason: Unhandled exception caught while running processor function _run_salmon in pipeline: [Errno 2] No such file or directory: '/home/user/data_store/processor_job_6648/SRR6025586_output/quant.sf'] [pipeline_applied: SALMON] [processor_job: 6648] [no_retry: False]: Processor job failed!
```

### Problem or idea

This actually seems like a fairly common error, and I can't immediately explain it. 

My guesses are:

[...]

Issue #680, opened 2018-10-01
fasterq-dump: Not Enough Memory


Miserlou:
It seems like once we started processing larger files, we started to see some messages like this:

```
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod 2018-09-30 11:24:34,903 i-03b2b1628573bd053 data_refinery_workers.processors.salmon ERROR [file: /home/user/data_store/DRP003977/DRR051067/DRR051067.sra] [stdout: 2018-09-30T06:11:52 fasterq-dump.2.9.1 err: unknown while writing file within file system module - unknown system error errno='Cannot allocate memory(12)'
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod 2018-09-30T06:11:53 fasterq-dump.2.9.1 err: you have exhausted your space
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod 2018-09-30T06:11:53 fasterq-dump.2.9.1 err: copy_machine.c push2q().KQueuePush() -> RC(rcCont,rcQueue,rcInserting,rcQueue,rcReadonly)] [processor_job: 6044]: Shell call to fasterq-dump failed with error message: 2018-09-30T06:11:52 fasterq-dump.2.9.1 err: unknown while writing file within file system module - unknown system error errno='Cannot allocate memory(12)'
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod 2018-09-30T06:11:53 fasterq-dump.2.9.1 err: you have exhausted your space
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod 2018-09-30T06:11:53 fasterq-dump.2.9.1 err: copy_machine.c push2q().KQueuePush() -> RC(rcCont,rcQueue,rcInserting,rcQueue,rcReadonly)
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod 2018-09-30 11:24:35,076 i-03b2b1628573bd053 data_refinery_workers.processors.utils ERROR [failure_reason: 2018-09-30T06:11:52 fasterq-dump.2.9.1 err: unknown while writing file within file system module - unknown system error errno='Cannot allocate memory(12)'
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod 2018-09-30T06:11:53 fasterq-dump.2.9.1 err: you have exhausted your space
[...]

PR #686, opened 2018-10-01
Increase RAM Amount Upon Retrying Salmon Jobs

Miserlou:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/680
https://github.com/AlexsLemonade/refinebio/issues/681 (Hopefully)
https://github.com/AlexsLemonade/refinebio/issues/683 (Hopefully)

If a SALMON job gets retried, increase the amount of RAM if possible.

This PR also includes an API change necessary for https://github.com/AlexsLemonade/refinebio-frontend/issues/346

## Types of changes

## Functional tests

New test case.

### DIFF
--- a/api/data_refinery_api/views.py
+++ b/api/data_refinery_api/views.py
@@ -2,7 +2,7 @@ from django.conf import settings
 from django.db.models import Count
 from django.db.models.aggregates import Avg
 from django.db.models.expressions import F
-from django.http import Http400, Http404, HttpResponse, HttpResponseRedirect
+from django.http import Http404, HttpResponse, HttpResponseRedirect, HttpResponseBadRequest
 from django.shortcuts import get_object_or_404
 
 from django_filters.rest_framework import DjangoFilterBackend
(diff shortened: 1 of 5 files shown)


api/data_refinery_api/views.py
```diff
@@ -408,6 +408,12 @@ def get_object(self, pk):
             return Experiment.public_objects.get(pk=pk)
         except Experiment.DoesNotExist:
             raise Http404
+        except Exception:
```
kurtwheeler: Do we actually get here if an accession code is used instead of a ID? It seems like the initial lookup would raise a DoesNotExist which would trigger the previous except block and raise the Http404 before we got here.

Miserlou: There are other exceptions than Experiment.DoesNotExist, since we're allowing a more complex input string now.

kurtwheeler: So are you saying an accession code instead of an id will trigger a different exception than DoesNotExist or that you want to be more general about what you catch? I'm thinking what we want to do is remove the except block right above this so this one handles all exceptions, even the DoesNotExist.

Miserlou: I changed it from a 404 to a 400 error in the final block, as Bad Request is probably a better explanation for why you'd ever end up there.

PR #690, opened 2018-10-02
[WIP] The Dharma Initiative

Miserlou:
## Issue Number

n/a

Nomad has no forward capacity planning, so our immense backlog of smaller downloader jobs will block the allocation of larger processor jobs.

This is a hack to prevent a maximum number of downloader jobs from running on any single node at a time. The hack is in two parts:

1) Detection if the total number of jobs running on a node is greater than MAX
2) Delay of suicide until the next death-window, so that enough RAM can be simultaneously allocated for processor jobs.

## Types of changes

## Functional tests
New test added. Testing for real now.


workers/data_refinery_workers/downloaders/utils.py
```diff
@@ -25,6 +28,7 @@
 SYSTEM_VERSION = get_env_variable("SYSTEM_VERSION")
 # TODO: extend this list.
 BLACKLISTED_EXTENSIONS = ["xml", "chp", "exp"]
+MAX_JOBS_PER_NODE = 15
```
kurtwheeler: Two things about this. The first is that I think this is high if we're planning on scaling out to 20 instances. The second is that this should be an environment variable so that we can pass in whatever value we want to the docker container without having to rebuild it.

workers/data_refinery_workers/downloaders/test_array_express.py
```diff
@@ -103,3 +105,53 @@ def test_download_multiple_zips(self, mock_send_job):
...
+        sample.technology = "MICROARRAY"
+        sample.manufacturer = "AFFYMETRIX"
+        sample.has_raw = True
+        sample.platform_accession_code = "hgu133a"
+        sample.save()
+
+        OriginalFileSampleAssociation.objects.get_or_create(sample=sample, original_file=original_file)
+
+        try:
+            utils.start_job(dlj3.id)
```
kurtwheeler: How does this trigger dharma if there's only 2 tests running before this one is started and the cap is 15?

workers/data_refinery_workers/downloaders/utils.py
Miserlou: Envariablized.

workers/data_refinery_workers/downloaders/test_array_express.py
Miserlou: Bad test assertions. Added more robust assertions.

Miserlou:
Addressed comments, also added support for IOPS constraint on Salmon jobs.

Issue #701, opened 2018-10-05
Platform badge on search page always displays the mixed platform badge


dvenprasad:
### Context

 Noticed this walking through refine.bio

### Problem or idea
On the search page the platform badge is always the mixed platform badge.
![screen shot 2018-10-05 at 11 33 39 am](https://user-images.githubusercontent.com/15315514/46544825-9a0b1c80-c892-11e8-8cbc-67c232d1f713.png)

But on the experiment page it displays the correct platform badge:
![screen shot 2018-10-05 at 11 34 54 am](https://user-images.githubusercontent.com/15315514/46544861-b3ac6400-c892-11e8-98ed-0f6fb56cf341.png)

Could be related to #589 

### Solution or next step
Tagging @arielsvn and @Miserlou for input

Miserlou:
I thiiink there is already a fix for this that hasn't been deployed yet?

PR #589, opened 2018-09-06
Adds `get_sample_technologies` to model and `technologies` to API response

Miserlou:
## Issue Number

https://github.com/AlexsLemonade/refinebio-frontend/issues/268

## Purpose/Implementation Notes

Adds `get_sample_technologies` to model and `technologies` to API response

## Types of changes
- New feature (non-breaking change which adds functionality)

## Functional tests

New assertion

PR #702, opened 2018-10-05
Re-Adds Technologies

Miserlou:
Related:
https://github.com/AlexsLemonade/refinebio/pull/589
https://github.com/AlexsLemonade/refinebio/issues/701

This was already approved by got reverted by mistake.

### DIFF
--- a/common/data_refinery_common/models/models.py
+++ b/common/data_refinery_common/models/models.py
@@ -325,10 +325,12 @@ class Experiment(models.Model):
         return fields
 
     def get_sample_technologies(self):
-        """ Get all metadata fields that are non-empty for at least one sample in the experiment.
-        See https://github.com/AlexsLemonade/refinebio-frontend/issues/211 for why this is needed.
+        """ Get a list of unique technologies for all of the associated samples
         """
-        technologies = list(filter(None, list(set([t['technology'] for t in self.samples.all().values('technology')]))))
+        tech_values_qs = self.samples.all().values('technology')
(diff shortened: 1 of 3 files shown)


Miserlou:
Ariel confirmed that this works with frontend.

common/data_refinery_common/models/models.py
```diff
@@ -324,6 +324,13 @@ def get_sample_metadata_fields(self):
                 fields.append(field)
         return fields
 
+    def get_sample_technologies(self):
+        """ Get all metadata fields that are non-empty for at least one sample in the experiment.
+        See https://github.com/AlexsLemonade/refinebio-frontend/issues/211 for why this is needed.
+        """
+        technologies = list(filter(None, list(set([t['technology'] for t in self.samples.all().values('technology')]))))
```
kurtwheeler: The Jone's Complexity on this line seems a bit high. Also I think it might be a bit simpler to just append `if t['technology'] is not None` onto that list comprehension than it is to filter and cast back to a list.

common/data_refinery_common/models/models.py
```diff
@@ -324,6 +324,13 @@ def get_sample_metadata_fields(self):
                 fields.append(field)
         return fields
 
+    def get_sample_technologies(self):
+        """ Get all metadata fields that are non-empty for at least one sample in the experiment.
+        See https://github.com/AlexsLemonade/refinebio-frontend/issues/211 for why this is needed.
```
kurtwheeler: This docstring is a copy-paste.

Issue #679, opened 2018-10-01
Foreman is Extremely Chatty When It Can't Reach Nomad


Miserlou:
When Nomad is down, foreman will constantly output 4 tracebacks for each failure, which is in an infinite loop by design. This makes the logs pretty much unusable:

```
data-refinery-log-group-circleci-prod log-stream-foreman-circleci-prod 2018-10-01 15:46:45,880 i-00ddefc69b1a5b643 data_refinery_foreman.foreman.main ERROR [processor_job: 147276]: Couldn't query Nomad about Processor Job.
data-refinery-log-group-circleci-prod log-stream-foreman-circleci-prod Traceback (most recent call last):
data-refinery-log-group-circleci-prod log-stream-foreman-circleci-prod   File "/usr/local/lib/python3.5/dist-packages/urllib3/connectionpool.py", line 387, in _make_request
data-refinery-log-group-circleci-prod log-stream-foreman-circleci-prod     six.raise_from(e, None)
data-refinery-log-group-circleci-prod log-stream-foreman-circleci-prod   File "<string>", line 2, in raise_from
data-refinery-log-group-circleci-prod log-stream-foreman-circleci-prod   File "/usr/local/lib/python3.5/dist-packages/urllib3/connectionpool.py", line 383, in _make_request
data-refinery-log-group-circleci-prod log-stream-foreman-circleci-prod     httplib_response = conn.getresponse()
data-refinery-log-group-circleci-prod log-stream-foreman-circleci-prod   File "/usr/local/lib/python3.5/dist-packages/raven/breadcrumbs.py", line 375, in getresponse
data-refinery-log-group-circleci-prod log-stream-foreman-circleci-prod     rv = real_getresponse(self, *args, **kwargs)
data-refinery-log-group-circleci-prod log-stream-foreman-circleci-prod   File "/usr/lib/python3.5/http/client.py", line 1197, in getresponse
data-refinery-log-group-circleci-prod log-stream-foreman-circleci-prod     response.begin()
data-refinery-log-group-circleci-prod log-stream-foreman-circleci-prod   File "/usr/lib/python3.5/http/client.py", line 297, in begin
data-refinery-log-group-circleci-prod log-stream-foreman-circleci-prod     version, status, reason = self._read_status()
data-refinery-log-group-circleci-prod log-stream-foreman-circleci-prod   File "/usr/lib/python3.5/http/client.py", line 258, in _read_status
[...]

PR #718, opened 2018-10-08
Adds Better Nomad Exception Handling in the Foreman

Miserlou:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/679

Avoid Nomad being down flooding our logs.

## Types of changes

### DIFF
--- a/foreman/data_refinery_foreman/foreman/main.py
+++ b/foreman/data_refinery_foreman/foreman/main.py
@@ -1,3 +1,4 @@
+import socket
 import time
 from nomad import Nomad
 from nomad.api.exceptions import URLNotFoundNomadException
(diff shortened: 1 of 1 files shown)


foreman/data_refinery_foreman/foreman/main.py
```diff
@@ -210,8 +212,14 @@ def retry_lost_downloader_jobs() -> None:
                 # If this job really should be restarted we'll get it in the next loop.
                 if timezone.now() - job.created_at > MIN_LOOP_TIME:
                     lost_jobs.append(job)
+        except socket.timeout:
+            logger.info("Timeout connecting to Nomad - is Nomad down?", job_id=job.id)
+            lost_jobs.append(job)
+        except nomad.api.exceptions.BaseNomadException:
+            logger.info("Problem connecting to Nomad - is Nomad down?", job_id=job.id)
+            lost_jobs.append(job)
```
kurtwheeler: I'm not sure that we should assume these jobs are lost for either of these except blocks. They could be happily chugging along while the only issue going on is that the Foreman is having network issues. I would think that better behavior would be to do nothing. If the jobs are actually in need of retrying, the Foreman will loop back over them in a couple minutes.

Issue #550, opened 2018-08-24
Remove Affymetrix control probes from SCAN.UPC processed data


jaclyn-taroni:
### Context

I noticed that Affymetrix control probes were in the delivered in gene expression matrices that included only samples run on Affymetrix

```
	YA_36hpf_rep3 (YA36_5-3)
AFFX-BioB-3	-0.964241616
AFFX-BioB-5	-1.101939413
AFFX-BioB-M	-1.033939384
AFFX-BioC-3	-0.931904971
AFFX-BioC-5	-0.949069717
AFFX-BioDn-3	-0.88990334
AFFX-BioDn-5	-0.914673485
AFFX-CreX-3	-1.386955387
AFFX-CreX-5	-1.242860642
AFFX-DapX-3	0.923737962
AFFX-DapX-5	-0.29128977
AFFX-DapX-M	1.06311269
```

### Problem or idea

Remove genes / rows that begin with `AFFX-` which designates that these are the values for the control probes.

### Solution or next step

I don't think this is a must have for the initial release and in fact users may not actually mind the inclusion of these probes. I'm filing this ticket in service of keeping track of it and we can see what folks say.

If we do end up removing the control probes, it may make sense to include this step in the Affymetrix processor itself. If we decide to implement this after we've processed a lot of Affymetrix data, then I can imagine it makes sense to include this step in the smasher. For that reason, I've tagged this with `smasher` and `processor` labels.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

Miserlou:
If you want it done in the processor, we can't punt in into the future as we'll have to re-run all of the samples, either from source or through an additional post processor. So, it should either be included now or done at smash-time in the future, I'm leaning towards the latter so we don't through information away.

jaclyn-taroni:
Yes, that is why I wrote the Solution or next step that way. I will remove the processor label.

jaclyn-taroni:
Was this closed by #719 @Miserlou ?

PR #719, opened 2018-10-08
Remove Affy Control Probes When Smashing 

Miserlou:
## Issue Number

#550

Trivial implementation

## Types of changes


jaclyn-taroni:
This has implications for the QN target that has already been generated, correct?

Miserlou:
Ooo, that's a good point. Should this happen after the QN step?

jaclyn-taroni:
I would ideally remove these when we generate the QN target (other platforms don't have the same control probes or don't have probes at all) _and_ then we should take these out prior to QN when we smash.

Miserlou:
Okay, cool. That will happen from this PR, but we will need to regenerate our QN targets on prod.

PR #723, opened 2018-10-09
[WIP] Remove our dependence upon MIT's keyserver being up.

kurtwheeler:
## Issue Number

N/A, but it blocked a deploy: https://circleci.com/gh/AlexsLemonade/refinebio/8120

## Purpose/Implementation Notes

MIT's keyserver can go down. Since our deploy process pulls keys from it that means it is a single point of failure for our deploys. This removes our dependence upon them by just storing the public keys in the repo directly.

This PR is a WIP because we're waiting on public keys from @cgreene and @dongbohu 

## Types of changes

- New feature (non-breaking change which adds functionality)

## Functional tests

I tested that running `gpg --import keys/rich.gpg` worked. The true test will come during a deploy though.

## Checklist

- [x] Lint and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)
- [x] Any dependent changes have been merged and published in downstream modules

PR #692, opened 2018-10-02
Add template for smasher error email

arielsvn:
## Issue Number

N/A

## Purpose/Implementation Notes

@dvenprasad created a template for the email we send when there's an error in the smashing process. This PR adds it, in the same way we use for the email when a dataset is ready for download.

## Types of changes

- Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Functional tests

**I wasn't able to test this locally**

## Checklist

- [ ] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [x] Any dependent changes have been merged and published in downstream modules

## Screenshots

image

PR #729, opened 2018-10-10
Deploy recent changes to master including the Dharma initiative

kurtwheeler:
Deploys #692, # 686, #693, #694, #695, #696, #697, #705, #717, #702, #719, #721, and #723.


kurtwheeler:
I'm not actually sure if we should deploy this yet or not because of the disk space issue with EBS.

kurtwheeler:
A lot more has been pushed to dev since I opened this PR, but I think we're ready to kick off this deploy now that the Janitor is working.

kurtwheeler:
I forgot to check the negative test (that the foreman doesn't queue Janitor jobs TOO OFTEN). It looks like it's doing them every minute so we actually should test in staging again.

PR #734, opened 2018-10-11
Fix running nomad locally, end to end tests, and log level environment variable

kurtwheeler:
## Issue Number

N/A Ariel reported that he couldn't run processor jobs locally

We didn't set the meta fields that are needed for processor jobs on the local nomad client config, so there were no clients which could pick up our constrained processor jobs. Fixing that was fairly easy, but then the question became why didn't that cause our end to end tests to fail.

The reason our end to end tests weren't failing is because we generated no processor jobs and we didn't check the number of processor jobs so we waited until 0/0 jobs were successful and passed the test. The reason we weren't generating any processor jobs was because we were running the test on an agilent accession, which we're currently skipping since we don't support them yet.

I made sure that the end to end tests check that there was at least one processor job that ran and changed the accession to be an Affy experiment that needs to be no_op'd.

Along the way I was trying to increase visibility into what was going wrong by raising the log level and it wasn't working, so I made the nomad jobs actually use that env var.

Finally I also noticed that we weren't setting the filename for original files generated by array express so I set that property as well.

## Types of changes

## Functional tests

N/A since this is a unit test fix.

## Checklist


kurtwheeler:
@Miserlou I made a few changes since you reviewed. Wanna re-confirm your approval?

Issue #727, opened 2018-10-10
Allow users to filter by platforms


arielsvn:
### Context

https://github.com/AlexsLemonade/refinebio-frontend/issues/371#issuecomment-428334618

### Problem or idea

Users want to filter by Platform.

### Solution or next step

Add `platform` to the list of filters returned by the search endpoint.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

PR #737, opened 2018-10-15
Move from samples__platform_accession_code to samples__platform_name

Miserlou:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/727
https://github.com/AlexsLemonade/refinebio-frontend/issues/371

Filter requests must include `samples__platform_name` to use this field. 

## Types of changes

## Screenshots

screen shot 2018-10-15 at 12 50 35 pm


arielsvn:
I tested this locally, and it works:

image

But, I think we need to find a way to format the platform names, which will also help with https://github.com/AlexsLemonade/refinebio-frontend/issues/362. I'm not sure what's the best way to do this, should we start using the `/platforms` endpoint for this? Or is there any other way we can use to format the platform names in the frontend?

kurtwheeler:
This looks pretty good to me. Has it been tested at all?

arielsvn:
Yes, I tested it in https://github.com/AlexsLemonade/refinebio-frontend/pull/390 but we need to find a way to format the names of the platforms correctly. I guess we can discuss it in a different issue.

PR #741, opened 2018-10-15
Kurtwheeler/fix no op iamge

kurtwheeler:
## Issue Number

N/A the no_op image decided to break on us due to R dependencies changing.... somehow.

https://circleci.com/gh/AlexsLemonade/refinebio/8438 failed because:
```
2018-10-15 20:20:27,109 local data_refinery_workers.processors.utils ERROR [processor_job: 4] [pipeline_applied: NO_OP] [failure_reason: Status code 1 from gene_convert.R: b"Error in loadNamespace(i, c(lib.loc, .libPaths()), versionCheck = vI[[i]]) : \n  namespace 'rlang' 0.2.1 is already loaded, but >= 0.2.2 is required\nCalls: :: ... tryCatch -> tryCatchList -> tryCatchOne -> <Anonymous>\nExecution halted\n"] [no_retry: True]: Processor job failed!
```

Not the most useful error output, but running the image locally I was able to reproduce and get the full error output:
```
Error: package or namespace load failed for 'dplyr' in loadNamespace(i, c(lib.loc, .libPaths()), versionCheck = vI[[i]]):
 namespace 'rlang' 0.2.1 is already loaded, but >= 0.2.2 is required
```

I've bumped up rlang from 0.2.1 to 0.2.2 to hopefully fix the current version conflict we're seeing with dplyr. I have also tried to make it so that R will break at build time rather than run time.

I had thought we already protected ourselves against those kinds of issues, but it turns out that just installing a package doesn't mean you can actually use it...

@jaclyn-taroni when you get back we should discuss an audit of our R code for other issues where dependencies are not actually verified to be working before run time.

## Types of changes

## Functional tests

Circle is running the unit tests right now which is sufficient for this change.

## Checklist

### DIFF
--- a/workers/affymetrix_dependencies.R
+++ b/workers/affymetrix_dependencies.R
@@ -226,3 +226,7 @@ pd_annotation_pkgs <- c(
   'pd.zebrafish_3.12.0.tar.gz'
 )
 install_with_url(annotation_url, pd_annotation_pkgs)
+
+# Load this libraries because apparently just installing it isn't
+# enough to verify that the correct versions of dependencies are installed.
+library('foreach')
--- a/workers/illumina_dependencies.R
+++ b/workers/illumina_dependencies.R
@@ -39,3 +39,15 @@ illumina_pkgs <- c(
   'illuminaRatv1.db_1.26.0.tar.gz'
 )
 install_with_url(release_url, illumina_pkgs)
+
+# Load these libraries because apparently just installing them isn't
+# enough to verify that they have complementary versions.
+library("optparse")
+library(data.table)
+library("dplyr")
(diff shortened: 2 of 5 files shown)


Miserlou:
```
Error in library("dplyr") : there is no package called 'dplyr'
Execution halted
The command '/bin/sh -c Rscript illumina_dependencies.R' returned a non-zero code: 1
Could not build ccdlstaging/dr_illumina after three attempts.
Exited with code 1
```

>:[

Miserlou:
Changes look good, I made the requested ticket here: https://github.com/AlexsLemonade/refinebio/issues/742

PR #743, opened 2018-10-17
Makes the Foreman reliably dispatch Janitor jobs every half hour.

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio/pull/738 wasn't a good enough fix

When I tested https://github.com/AlexsLemonade/refinebio/pull/738 I didn't want to wait a half hour so I changed the minutes to be every minute. Well that happened to be the only way this works since the loop send_janitor_jobs only runs every 10 minutes and send_janitor_jobs doesn't have a loop of its own.

Anyway I ended up going with the `do_forever` implementation because it was super easy, we know it works, and if that thread ever dies we'll know because there's a loop monitoring its health.

## Types of changes

## Functional tests

I pushed a foreman image up with these changes, pulled it onto the staging foreman instance, restarted the foreman, made sure it started all the threads, made sure it dispatched janitor jobs, made sure those janitor jobs cleaned up the directories they were supposed to, and then made sure no more janitor jobs were dispatched for a half hour.

## Checklist

### DIFF
--- a/foreman/data_refinery_foreman/foreman/main.py
+++ b/foreman/data_refinery_foreman/foreman/main.py
@@ -396,7 +396,7 @@ def retry_lost_processor_jobs() -> None:
 
 @do_forever(JANITOR_DISPATCH_TIME)
 def send_janitor_jobs():
-    """ If we're in the send_minutes window, dispatch some janitors """
+    """Dispatch a Janitor job for each instance in the cluster"""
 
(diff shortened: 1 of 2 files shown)


Miserlou:
Approach looks okay but the tests need to be updated.

kurtwheeler:
@Miserlou the tests are passing!

Issue #710, opened 2018-10-08
Improve deploy times.


kurtwheeler:
### Context

Right now our deploy times are ~4 hours.

### Problem or idea

The thing that slows our deploys down is rebuilding the docker images because CircleCI cannot cache them.

### Solution or next step

Create a dedicated AWS instance for running deploys. This way it can always have the docker cache so it will build the images as quickly as possible.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

PR #744, opened 2018-10-17
Changes deploys to use an AWS instance so we keep our cache. 

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/710

Rebuilding our docker images takes ~3.5 hours because the circleci boxes don't have the Docker cache on them. This slows our deploys down a ton. By always using the same instance to run the deploys, we'll have the docker cache so rebuilding images won't take so long except when there's major changes to the images.

The way we do this is to have CircleCI ssh onto our deploy box and run the command necessary to deploy.

I have also included instructions on how to stand up the deploy box should that ever be necessary.

## Types of changes

## Functional tests

I used this script to run a deploy on the deploy box from my local machine. I set the env vars appropriately to do so. The whole stack went up nicely.

## Checklist


.circleci/git_decrypt.sh
```diff
@@ -1,12 +1,5 @@
...
-cd
-git clone https://github.com/AGWA/git-crypt.git
-cd git-crypt
-make
-sudo make install
-
 # Unlock encrypted files
 cd ~/refinebio/.circleci
-openssl aes-256-cbc -d -in exported.key.enc -out $KEY_FILENAME -k $OPENSSL_KEY
+openssl aes-256-cbc -md md5 -d -in exported.key.enc -out $KEY_FILENAME -k $OPENSSL_KEY
```
Miserlou: md5 whyyyyyy

kurtwheeler: The AMI we're using has a higher version of openssl than circleCI's. In the version switch the default method changed so to stay consistent we have to specify it now. Specifying it doesn't break older versions though so we don't have to worry about backwards compatibility.

.circleci/run_terraform.sh
```diff
@@ -65,9 +65,14 @@ elif [[ $branch == "dev" ]]; then
     ENVIRONMENT=staging
 else
     echo "Why in the world was run_terraform.sh called from a branch other than dev or master?!?!?"
-    exit 1
+    # TODO: uncomment this
+    # exit 1
```
Miserlou: ??

.circleci/update_docker_img.sh
```diff
@@ -13,9 +13,13 @@ elif [[ $branch == "dev" ]]; then
     DOCKERHUB_REPO=ccdlstaging
 else
     echo "Why in the world was update_docker_img.sh called from a branch other than dev or master?!?!?"
-    exit 1
+    # TODO: uncomment this!
+    # exit 1
```
Miserlou: ??

infrastructure/deploy.sh
```diff
@@ -113,7 +113,7 @@ if [[ ! -f terraform.tfstate ]]; then
     # Until terraform plan supports -var-file the plan is wrong.
     # terraform plan
 
-    terraform apply -var-file=environments/$env.tfvars -auto-approve > /dev/null
+    terraform apply -var-file=environments/$env.tfvars -auto-approve
```
Miserlou: I assume this is changed because we don't need the "security" of devnull anymore?

Miserlou: Maybe we can check something in the environment to determine this instead? I don't want to accidentally bump into this again.

infrastructure/deploy_box_instance_data.sh
```diff
@@ -0,0 +1,56 @@
...
+
+touch /var/log/docker_update.log
+chown ubuntu:ubuntu /var/log/docker_update.log
+
+touch /var/log/deploy.log
+chown ubuntu:ubuntu /var/log/deploy.log
+
+# Checkout the repo onto the box.
+cd /home/ubuntu
+git clone https://github.com/AlexsLemonade/refinebio.git
```
Miserlou: Branch info? Or will this only ever be for master?

kurtwheeler: This is just to init, the branch will be checked out at deploy time.

.circleci/run_terraform.sh
kurtwheeler: These are just like this until we've tested this because if this isn't here then test deploys fail. (You can't just comment it out locally for testing since the actual branch is checked out onto the deploy box as part of the deploy process.)

infrastructure/deploy_box_instance_data.sh
```diff
@@ -0,0 +1,56 @@
+# This script can be used as the instance-user-data.sh script for AWS
+# instances to configure a new instance to be used as a deploy box.
+
+# To create a new deploy box (there only ever needs to be one at a
+# time.), there are few steps to follow:
+#  - Make sure there is a deploy VPC. If there isn't create one so the deploy box is isolated.
```
Miserlou: This can't be managed by TF as well? "Create one" isn't very descriptive.

infrastructure/deploy_box_instance_data.sh
```diff
@@ -0,0 +1,56 @@
...
+#  - Make sure the deploy VPC has a security group. If it doesn't create one and allow ingress
+#    TCP connections on port 22 from 0.0.0.0/0
+#  - Create a new EC2 instance with the following settings:
```
Miserlou: Create it where? Via console or TF? Include URL here. Maybe a picture. Or couldn't this be TF managed as well?

kurtwheeler:
I destroyed the EC2 instance I created for this so that someone could follow my instructions to do so in order to make sure they are clear enough for someone who's not me to do this.

@Miserlou we had discussed you doing this, do you still want to or do you just want me to stand it back up and merge this in?

Miserlou:
Seemed to work fine!

PR #745, opened 2018-10-17
Fix stupid mistake where a constant was used instead of a parameter.

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio/pull/729#issuecomment-430684737

The `do_forever` was using a constant from the file it is defined in rather than the parameter that is passed to it. This was fine before `send_janitor_jobs` was using it because the constant it was using was the only value that was passed in, but we want janitor jobs to be queued ever 30 minutes, not every 2.

## Types of changes

## Functional tests

I pushed a foreman image to the `wkurt` dockerhub repo, pulled it on the staging foreman instance, restarted the foreman, then checked the logs. So far it's only queued one janitor job per instance and has not queued more since. At 1:54 it should queue another one per instance and at that point we will know this is working correctly.

## Checklist

### DIFF
--- a/foreman/data_refinery_foreman/foreman/main.py
+++ b/foreman/data_refinery_foreman/foreman/main.py
@@ -126,7 +126,7 @@ def do_forever(min_loop_time: timedelta) -> Callable:
 
                 loop_time = timezone.now() - start_time
                 if loop_time < min_loop_time:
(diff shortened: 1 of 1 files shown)


kurtwheeler:
I can confirm that additional Janitor Jobs were fired off at 1:54 and not before!

PR #747, opened 2018-10-18
Handle no downloader work gracefully.

kurtwheeler:
## Issue Number

N/A Came up during staging test

I saw a lot of downloader jobs failing but it turns out it was just because the files they were supposed to download had already been downloaded so they passed an empty list to create_processor_jobs which wasn't ready for it.

## Types of changes

## Checklist


Miserlou:
I don't think these tests will run in CI unless given a tag that matches one in the config. The change itself looks good though.

kurtwheeler:
Oh snap good catch.

PR #750, opened 2018-10-19
[WIP] Use Spot Fleet Rather Than Auto Scaling Groups

Miserlou:
Needs more testing, tuning, and cost prediction, but seems like it's working so far! Spot fleets are cool!


kurtwheeler:
Something I thought of last night is that we need to dynamically set `MAX_DOWNLOADER_JOBS_PER_NODE` based on the amount of RAM the host instance has since we'll have different amounts. I've started messing around with `psutil` package as a way to do this, the tricky thing will be making sure we're using the RAM of the host and not the amount of RAM allocated to the docker instance.

infrastructure/instances.tf
```diff
@@ -243,98 +243,265 @@ data "template_file" "nomad_client_script_smasher_smusher" {
...
-    depends_on = [
-              "aws_internet_gateway.data_refinery",
-              "aws_instance.nomad_server_1",
-              "aws_ebs_volume.data_refinery_ebs",
-              "aws_instance.pg_bouncer"
-    ]
-    user_data = "${data.template_file.nomad_client_script_smusher.rendered}"
+resource "aws_spot_fleet_request" "cheap_ram" {
+  iam_fleet_role      = "${aws_iam_role.data_refinery_spot_fleet.arn}"
+  allocation_strategy = "diversified"
```
kurtwheeler: Are you sure we want to use `diversified` instead of `lowestPrice`? It seems like lowestPrice would be overall cheaper, but could have more volatility if all our instances are of a single type and then that type's capacity gets snapped up.

However I think that what might be most appropriate for our use case is to follow the directions of the `Configuring Spot Fleet for Cost Optimization and Diversification` section:

> To create a fleet of Spot Instances that is both cheap and diversified, use the lowestPrice allocation strategy in combination with InstancePoolsToUseCount. Spot Fleet automatically deploys the cheapest combination of instance types and Availability Zones based on the current Spot price across the number of Spot pools that you specify. This combination can be used to avoid the most expensive Spot Instances.

What I couldn't determine from the documentation is what happens if there's no spot capacity for some instance types in the `diversified` strategy. Do the spot requests just sit and wait to be fulfilled or do they give up and use what is actually available? It seems to insinuate that at least...

infrastructure/instances.tf
```diff
@@ -243,98 +243,265 @@ data "template_file" "nomad_client_script_smasher_smusher" {
...
+    tags {
+        Name = "Spot Fleet Launch Specification x1e.32xlarge ${var.user}-${var.stage}"
+        User = "${var.user}"
+        Stage = "${var.stage}"
+    }
+  
+  }
+
+  ##
+  # c5d.18xlarge
```
kurtwheeler: C5's are compute optimized, not RAM optimized. I would have thought R5's would be more appropriate here.

infrastructure/instances.tf
Miserlou: My understanding is that `lowestPrice` is for riding the variance in spot price for a single instance type across regions and AZs. Since we're only operating on a single AZ and our limiting factor is capacity, I think that `diversified` is the correct strategy here.

infrastructure/instances.tf
```diff
@@ -243,98 +243,265 @@ data "template_file" "nomad_client_script_smasher_smusher" {
...
+
+  ##
+  # x1.16xlarge
+  ##
+  launch_specification {
+
+    # Client Specific
+    instance_type             = "x1.16xlarge"
+    weighted_capacity         = 10 # via https://aws.amazon.com/ec2/instance-types/
+    spot_price                = "${var.spot_price}"
```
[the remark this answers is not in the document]
Miserlou: `spot_price` isn't the value that you actually pay, it's the value of the highest-that-you're-willing-to-go. For us, that's always our spot price for the biggest class - we're not really concerned about the small variances for smaller classes.

kurtwheeler:
So since r5d.24xlarge has 768 GB and 20000/768 = 26.04 I think we should also set `MAX_CLIENTS` to 26 in `infrastructure/environments/prod.tfvars`.

Also we had discussed a strategy for trying to rotate volumes semi-fairly. I forget what it was, but has that been included in this PR yet?

Miserlou:
It doesn't, I think that might be out of scope as we might want to try a few strategies.

kurtwheeler:
Is there an initial strategy we want to try first? Or is that just what it's already doing?

Miserlou:
This strategy is AWS-generated chaos. The question is do we need CCDL-generated chaos, or to try to statefully tame the chaos.

kurtwheeler:
Cool, sounds pretty good to me. Earlier you mentioned some dev stack testing. Have you done that or should this just be tested in staging?

Miserlou:
I've tested this once before myself and I'm doing another test now. Curious at how the allocation will come in with these adjustments.

README.md
```diff
@@ -523,6 +523,33 @@ how to change this to another project.
...
+terraform state list
+```
+
+If you want to change a single entity in the state, you can use
+
+```
+terraform taint <your-entity-from-state-list>
+```
+
+And then rerun `deploy.sh` with the same parameters you originally ran it with.
```
kurtwheeler: The merge conflict ended up causing this section to be repeated. In my branch I moved it down below the `Autoscaling and Setting Spot Prices` section.

infrastructure/instances.tf
```diff
@@ -243,98 +243,296 @@ data "template_file" "nomad_client_script_smasher_smusher" {
...
+
+# resource "aws_autoscaling_policy" "clients_scale_down" {
+#     name = "asg-clients-scale-down-${var.user}-${var.stage}"
+#     scaling_adjustment = -1
+#     adjustment_type = "ChangeInCapacity"
+#     cooldown = 60
+#     autoscaling_group_name = "${aws_autoscaling_group.clients.name}"
+#     depends_on = ["aws_instance.nomad_server_1"]
+# }
+
```
kurtwheeler: Can we clean up this dead code? If we ever need to go back to it it'll be in the git history.

infrastructure/logging.tf
```diff
@@ -69,34 +69,35 @@ resource "aws_cloudwatch_log_stream" "log_stream_api_nginx_error" {
...
+#     namespace = "${var.user}-${var.stage}"
+#     period = "120"
+#     statistic = "Average"
+#     threshold = "${var.scale_down_threshold}"
+#     alarm_description = "The queue is too short - we need less workers!"
+#     alarm_actions = [
+#         "${aws_autoscaling_policy.clients_scale_down.arn}"
+#     ]
+
+# }
```
kurtwheeler: Can we clean up this dead code? If we ever need to go back to it it'll be in the git history.

Issue #699, opened 2018-10-05
Improve developer experience: running locally


jaclyn-taroni:
### Context

The idea of "improving developer experience" has been coming up around the office consistently. We'll want to make sure folks external to the group can run things locally. 

### Problem or idea

Two ways (so far) of approaching this have come up:
1) Have team members pair (e.g., backend dev + data science team, cover Mac and Ubuntu) and set things up locally.  Take detailed notes about any sticking points -> improve documentation where needed.
2) Create a couple configurations of VMs that we can spin up to simulate fresh Mac and Ubuntu boxes. Then we can use the script created by https://github.com/AlexsLemonade/refinebio/issues/714 to install everything and run end_to_end tests to make sure that our instructions/script work smoothly.

### Solution or next step

Figure out which approach to take with this and improve things.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [ ] You have proposed a solution or next step.

Miserlou:
Probably related: https://github.com/AlexsLemonade/refinebio/issues/714

jaclyn-taroni:
I was able to set this up locally last sprint and there were documentation changes as part of #996 that cleared up some issues @cansav09 ran up while setting this up on her MBP. 

Also to document a conversation we had in person:
* It is not particularly helpful to the backend team for science team members to have this running locally when tracking down errors that arise in production. It is more helpful for us to document our "suspicions" based on our experience on any ticket in question.
* If science team members need a particular experiment or set of samples processed, they should tell the backend team to run it in production rather than run it locally.

cgreene:
> If science team members need a particular experiment or set of samples processed, they should tell the backend team to run it in production rather than run it locally.

That makes me very nervous. Does this mean that the only way for Jackie or Candace to test a processor is to debug it _through_ Kurt or Rich? Or can you run a processor locally - just not via the entire refine.bio infrastructure?

Also, production? Seems dangerous.

PR #753, opened 2018-10-25
[WIP] Improve readme and many scripts.

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/699

This improves the readme a lot and improves a lot of our scripts to make running them easier.

This is a WIP that I think was almost done, but then I got distracted so I need to return to it to finish it up.

## Types of changes

- Documentation

## Functional tests

I don't remember anymore, so I should make sure to test this again.

## Checklist


kurtwheeler:
@Miserlou I've reviewed this and it looks pretty good to me. I think there's still a lot more we can improve as far as documentation goes, but this gets a lot of the worst stuff in better order. I haven't re-tested all the script changes yet, but I'll do that post-review in case there's major changes that necessitate another round of testing anyway.

For broader documentation improvements I think that we should consider coming up with an Information Architecture or something as a team so we know what topics we want need to have covered, what structure we want to use (i.e. one giant readme, multiple markdown files, a readthedocs page, etc), and what level of detail we want to cover. I'm happy to write lots of documentation, but the hardest part for me is knowing what to write about, what level of detail to write about it, and how to organize it.

Issue #559, opened 2018-08-28
We only store a single accession code for an experiment which can live in multiple databases


kurtwheeler:
### Context

This means that if a user knows the GEO accession of an experiment they would like to download, but we got it from SRA, we will tell them we don't have it.

### Problem or idea

We need a way to convert from the accession they are looking for to the accession that we stored.

### Solution or next step

We're not sure what the right way to do this is yet...

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

jaclyn-taroni:
* Would we want to be able to display alternate accessions? cc @dvenprasad 
* Add as search keywords? aliases?

dvenprasad:
@jaclyn-taroni and I met to discuss this.

- When someone searches with a GEO accession but we've picked it up from SRA, it should show up in the search results. Also applicable to similar cases involving GEO and ArrayExpress. 

- We will not display alternate accession codes on the UI since a portion of our user base will be confused by it. 

- On the Experiment page, we should add a field called "Source" and show the link to the experiment we surveyed the data from.

kurtwheeler:
> When someone searches with a GEO accession but we've picked it up from SRA, it should show up in the search results. Also applicable to similar cases involving GEO and ArrayExpress.

> We will not display alternate accession codes on the UI since a portion of our user base will be confused by it.

These two seem like they could be confusing in concert.  If I search for `E-GEOD-XXX` but we have the accession as `GSMXXX` we want to display it as a search result, but not show the `E-GEOD-XXX` accession? If the user doesn't know that `E-GEOD-XXX == GSMXXX` then how would they know we are actually displaying the experiment they searched for? 

If I search for X and only one result comes up which is Y, I'm probably more likely to assume that its a bug with the search algorithm rather than I am to assume that Y is actually X.

jaclyn-taroni:
The subset of users who know what the accession codes are and are therefore likely to search by accession code will probably be able to "grok" `GSEXXX` == `E-GEOD-XXX`. Many users think the `GSEXXX` are refine.bio internal accessions so added _more_ unfamiliar accession numbers could be confusing.

kurtwheeler:
Ok, I'm comfortable with this assumption about our users, however we may want to verify this after a while using hotjar @devenprasad

dvenprasad:
[...]

PR #757, opened 2018-10-29
Kurtwheeler/alternate accession codes

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/559

Makes it so that searching on `GSM1234` will match experiments with accession code `GSM1234` OR `E-GEOD-1234`. This won't help if someone searches `/experiment/?accession_code=GSM1234` though.

It does so by adding an `alternate_accession_code` to the experiment model which is included in searches and is set via the Experiment.save() method.

## Types of changes

## Functional tests

Unit tests.

## Checklist


common/data_refinery_common/models/models.py
```diff
@@ -279,6 +280,13 @@ def save(self, *args, **kwargs):
         if not self.id:
             self.created_at = current_time
         self.last_modified = current_time
+
+        if self.accession_code and not self.alternate_accession_code:
+            if self.accession_code.startswith('GSM'):
+                self.alternate_accession_code = 'E-GEOD-' + self.accession_code[3:]
+            elif self.accession_code.startswith('E-GEOD-'):
+                self.alternate_accession_code = 'GSM' + self.accession_code[7:]
```
Miserlou: This file uses `GSM` and the migration uses `GSE`. What's going on with that? Should that be tested?

kurtwheeler:
In an automated way? Looks like we could use https://github.com/plumdog/django_migration_testcase which was inspired by https://www.caktusgroup.com/blog/2016/02/02/writing-unit-tests-django-migrations/. The project hasn't been touched in over a year. I can try it though.

kurtwheeler:
I do think that a test on the staging server will be sufficient. I'll see if I can get django-migration-testcase working before we wanna do a staging deploy, and if not we'll just let that test it.

kurtwheeler:
Ok, I tried to write a migration test using that library. I got this exception:
```
ERROR: test_migration (data_refinery_api.tests.AltAccessionMigrationTest)
----------------------------------------------------------------------
  File "/usr/local/lib/python3.6/site-packages/django_migration_testcase/django_migrations.py", line 16, in setUp
    self.migrate(app_name, version)
  File "/usr/local/lib/python3.6/site-packages/django_migration_testcase/base.py", line 89, in migrate
    call_command(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/django/core/management/__init__.py", line 140, in call_command
    ', '.join(sorted(valid_options)),
TypeError: Unknown option(s) for migrate command: no_initial_data. Valid options are: app_label, database, fake, fake_initial, help, interactive, migration_name, no_color, no_input, pythonpath, run_syncdb, settings, skip_checks, stderr, stdout, traceback, verbosity, version.
```
which seems to be caused by this option being passed through: https://github.com/plumdog/django_migration_testcase/blob/master/django_migration_testcase/base.py#L78

However I think that's as far as it's worth it to debug since we can test this on staging and this package doesn't seem to be maintained since I'm the first one running into that.

Miserlou:
Ok!

PR #771, opened 2018-11-01
Isolate affy image tests and make them not use DLC

kurtwheeler:
## Issue Number

N/A https://circleci.com/gh/AlexsLemonade/refinebio/9082 failed

We hit a:
> failed to register layer: Error processing tar file(exit status 1): write /usr/local/lib/R/site-library/pd.genomewidesnp.6/extdata/pd.genomewidesnp.6.sqlite: no space left on device

error in the above-mentioned build. This was something we had been seeing before turning Docker Layer Caching off because if the cache is warmed on the volume that gets used by CircleCI for docker, then there may not be enough space to fit all of our images. This manifests itself by us not having enough room to download the Affy image onto it.

The solution I've come up with is to just not use The docker layer caching feature for the Affy image. However whether to use that feature or not is job-specific setting so I had to break them into their own job. However this is actually pretty sweet because the job they were in before took the longest to run, even with docker layer caching speeding things up.

## Types of changes

## Functional tests

This is a CI change so the passing CI tests show it's working.

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -124,7 +124,7 @@ jobs:
           command: .circleci/filter_tests.sh -t salmon
           no_output_timeout: 1h
 
-  tx_illumina_affy_agilent_tests:
+  tx_illumina_tests:
     working_directory: ~/refinebio
     machine:
       docker_layer_caching: true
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -165,6 +165,26 @@ jobs:
...
+  affy_agilent_tests:
+    machine: true
+    working_directory: ~/refinebio
+    steps:
+      - checkout
+
+      # Clean up stuff circle provides that we don't need.
+      - run: sudo bash .circleci/cleanup_instance.sh
+
+      # Setup Postgres in a Container
```
dongbohu: Any special reason why `C` is capitalized?

kurtwheeler: I guess because the comment is like a title which is why postgres is also capitalized? I just copied this line from the other job's config.

PR #782, opened 2018-11-06
Delete original files after processing locally.

kurtwheeler:
## Issue Number

N/A but Ariel mentioned his hard drive got filled up while crunching some data locally.

I think we added this to `delete_local_file` for computed files because since we don't upload files while running locally, if you remove the computed file then after processing you end up with nothing. However I can't think of a good reason to have it on the original files, so I think we should remove it. However I may have forgotten the reason.

## Types of changes

## Functional tests

N/A very minor local-only change, unit tests should be sufficient. 

## Checklist

### DIFF
--- a/common/data_refinery_common/models/models.py
+++ b/common/data_refinery_common/models/models.py
@@ -641,9 +641,6 @@ class OriginalFile(models.Model):
 
     def delete_local_file(self):
         """ Deletes this file from the local file system."""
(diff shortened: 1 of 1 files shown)


kurtwheeler:
Well I figure sometimes you'll wanna compare and sometimes you'll just wanna run stuff without having it flood your disk. We can't support both use cases by default so it seems like the better default is the one that doesn't kill your hard drive.

Issue #785, opened 2018-11-07
Browsable API Is Slow - Implement Select Field Cutoffs


Miserlou:
### Context
The API is now pretty fast when using `?format=json`, as we do from the front-end, but HTML requests are still very slow, sometimes leading to nginx timeout.

I think this is a result of the foreign relations which are being used to populate some of the templates that DRF renders. 

These browsable HTML only values can be changed here: https://www.django-rest-framework.org/api-guide/relations/#select-field-cutoffs - I think we can just lower (or zero) those values and solve the problem.

It should be fairly simple to implement, but we'll want to test with a lot of fixture data to make sure the change works as intended.

Miserlou:
Fixed in latest deploy. Still slow-ish for requests with lots of samples, but that's really a result of the response size and the browser's ability to apply the syntax highlighting.

PR #788, opened 2018-11-07
Prevent Filter Rendering With NoMarkupDjangoFilterBackend

Miserlou:
Hopefully a fix for https://github.com/AlexsLemonade/refinebio/issues/785, disables rendering of the widget that wasn't performant. Running a scale test with Dilla now.


Miserlou:
Tested this with 10K samples, seems like the fix works as expected. Ready for merge!

Issue #789, opened 2018-11-07
Foreman can't re-queue any jobs because of the database connection closing


kurtwheeler:
### Context

The Foreman logs are full of this:
```
2018-11-07 20:02:40,247 i-0d4a6a561fd82091c data_refinery_foreman.foreman.main ERROR: Exception caught by Foreman while running retry_hung_survey_jobs
 Traceback (most recent call last):
   File "/usr/local/lib/python3.5/dist-packages/django/db/backends/base/base.py", line 234, in _cursor
     return self._prepare_cursor(self.create_cursor(name))
   File "/usr/local/lib/python3.5/dist-packages/django/db/backends/postgresql/base.py", line 222, in create_cursor
     cursor = self.connection.cursor()
 psycopg2.InterfaceError: connection already closed
 
 The above exception was the direct cause of the following exception:
 
 Traceback (most recent call last):
   File "/home/user/data_refinery_foreman/foreman/main.py", line 65, in wrapper
     function(*args, **kwargs)
   File "/home/user/data_refinery_foreman/foreman/main.py", line 508, in retry_hung_survey_jobs
     for job in potentially_hung_jobs:
   File "/usr/local/lib/python3.5/dist-packages/django/db/models/query.py", line 268, in __iter__
     self._fetch_all()
   File "/usr/local/lib/python3.5/dist-packages/django/db/models/query.py", line 1186, in _fetch_all
     self._result_cache = list(self._iterable_class(self))
   File "/usr/local/lib/python3.5/dist-packages/django/db/models/query.py", line 54, in __iter__
     results = compiler.execute_sql(chunked_fetch=self.chunked_fetch, chunk_size=self.chunk_size)
   File "/usr/local/lib/python3.5/dist-packages/django/db/models/sql/compiler.py", line 1063, in execute_sql
     cursor = self.connection.cursor()
   File "/usr/local/lib/python3.5/dist-packages/django/db/backends/base/base.py", line 255, in cursor
     return self._cursor()
   File "/usr/local/lib/python3.5/dist-packages/django/db/backends/base/base.py", line 234, in _cursor
     return self._prepare_cursor(self.create_cursor(name))
   File "/usr/local/lib/python3.5/dist-packages/django/db/utils.py", line 89, in __exit__
     raise dj_exc_value.with_traceback(traceback) from exc_value
   File "/usr/local/lib/python3.5/dist-packages/django/db/backends/base/base.py", line 234, in _cursor
[...]

PR #790, opened 2018-11-07
Increases the database timeout to try to fix the foreman.

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/789

I'm not sure what's causing the errors reported in that issue. However my going theory is that when making large requests Django uses a cursor to iterate through the results of a query rather than ingesting them all into memory at once. Because the Foreman's database timeout was so low, when we made a query involving >40k results we would hit the timeout before we got through them all.

This increases the database timeout to 300 seconds, which is the default Django chooses when creating a new Django projects. Perhaps this is why.

## Types of changes

## Functional tests

 I can't think of a way to test this other than spinning up a dev deploy, but that seems way too heavyweight for such a small change. We already had everything set up to support this env var, we just weren't using it for some reason. 

## Checklist


kurtwheeler:
Well I just restarted pgbouncer but it didn't help the Foreman out at all. I'm going to deploy this and see if this fixes it.

PR #834, opened 2018-11-13
Kurtwheeler/smasher instance

kurtwheeler:
## Issue Number

I can't find it I thought there was one

Our cluster's capacity is not constant or reliable since we use spot instances. However we should reliably be able to deliver datasets. This means we need a non-spot instance that is dedicated to running smasher jobs. This PR adds one capable of running two smasher jobs at once. It sets Nomad constraints on the instance and surveyor/downloader jobs so that it won't run jobs that aren't smasher jobs, and adds constraints to smasher jobs so they will only be run by the smasher instance.

## Types of changes

## Functional tests

WIP

## Checklist


kurtwheeler:
You mean like explicitly checking something like `instance_type == 't3.large'` or `instance_type == "${smasher_instance_type}". I'd be hesitant to do it like that because then that would mean the behavior would be dependent on configuration settings, i.e. if we ever had another client node use the same instance type then suddenly the behavior would change and by the time it actually happened it might not be so obvious why. Or if we used smaller instances in dev or staging it could be a very subtle environment difference that changes behavior meaningfully.

PR #835, opened 2018-11-13
Interactive filtering

### LINKED ISSUE #780: Interactive filtering for search results
### Context

The filters are based off the search term only.

https://github.com/AlexsLemonade/refinebio-frontend/issues/374#issuecomment-436373470

### Problem or idea

We want to implement interactive filtering, the order in which filters are applied will matter when deciding what filters to display.

### Solution or next step

Change how filters are calculated as described in https://github.com/AlexsLemonade/refinebio-frontend/issues/374#issuecomment-436373470

I'm going to give this a try to see if I can get it working.

### New Issue Checklist

Blocks https://github.com/AlexsLemonade/refinebio-frontend/issues/374

arielsvn:
## Issue Number

close #780 
https://github.com/AlexsLemonade/refinebio-frontend/issues/374#issuecomment-436373470

This modifies the API to make the filters more interactive, when a filter in a category is applied the remaining categories will be updated. More details on https://github.com/AlexsLemonade/refinebio-frontend/issues/374#issuecomment-436373470

## Types of changes

## Functional tests

Tested locally, although it would be good to see this in staging since I don't have much data.

## Checklist

### DIFF
--- a/api/data_refinery_api/views.py
+++ b/api/data_refinery_api/views.py
@@ -204,10 +204,43 @@ class SearchAndFilter(generics.ListAPIView):
     def list(self, request, *args, **kwargs):
         """ Adds counts on certain filter fields to result JSON."""
         response = super(SearchAndFilter, self).list(request, args, kwargs)
-        qs = self.search_queryset(self.get_queryset())
-        response.data['filters'] = self.get_filters(qs)
+
+        # Calculate queryset with all filters applied
+        queryset = ExperimentFilter(request.GET, queryset=self.get_queryset()).qs
+        response.data['filters'] = self.get_filters(queryset)
+
+        last_category = self.last_category(request.get_full_path())
+        if last_category:
+            params_without_last_category = request.GET.copy()
+            params_without_last_category.pop(last_category)
+            
+            queryset_without_last_category = ExperimentFilter(params_without_last_category, queryset=self.get_queryset()).qs
+
+            # mapping between parameter names and category names
+            filter_name_map = {
+                'technology': 'technology',
+                'has_publication': 'publication',
+                'organisms__name': 'organism',
+                'platform': 'platforms'
+            }
+
+            # The filters of the last category are calculated differently, since they should stay the same
+            # we use a queryset with all filters applied except the ones in the last category
+            # ref https://github.com/AlexsLemonade/refinebio-frontend/issues/374#issuecomment-436373470
+            last_category_filters = self.get_filters(queryset_without_last_category)[filter_name_map[last_category]]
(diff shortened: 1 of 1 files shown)


arielsvn:
Discussed this with @Miserlou: the current solution is relying on the order of the parameters in the URL to determine the last filter category that was applied. This isn't a good idea... will update the code to use a new parameter instead.

api/data_refinery_api/views.py
```diff
@@ -152,21 +152,19 @@ class SearchAndFilter(generics.ListAPIView):
...
     pagination_class = LimitOffsetPagination
 
     filter_backends = (NoMarkupDjangoFilterBackend, filters.SearchFilter, filters.OrderingFilter)
     filter_class = ExperimentFilter
 
+    # Ordering
+    ordering_fields = ('total_samples_count', 'id', 'created_at',
+                       'source_first_published', 'accession_code',)
+    ordering = ('-total_samples_count',)
+
```
Miserlou: The ordering section is now repeated, so this code doesn't do anything

api/data_refinery_api/views.py
```diff
@@ -152,21 +152,19 @@ class SearchAndFilter(generics.ListAPIView):
...
     # Ordering
     ordering_fields = ('total_samples_count', 'id', 'created_at', 'source_first_published', 'accession_code',)
-    samples_count = django_filters.NumberFilter(method='filter_samples_count')
```
Miserlou: Why was this removed without comment?

api/data_refinery_api/views.py
arielsvn: good catch, not sure why this happened. I'll remove it

api/data_refinery_api/views.py
arielsvn: I thought it wasn't needed, since we now use the field `total_samples_count`. Sorting by number of samples works locally for me.

I should have added a comment though.

arielsvn:
> Does this break the ability to filter/order by number of samples? 

It doesn't, at least locally I'm able to sort by the total number of processed samples. I'll double check.

> What does this do to performance?

Good question, it would be good to try it with more data. Although I was thinking, we can reduce the number of DB queries per group of filters, right now each group of filters is calculated twice.

arielsvn:
@Miserlou I addressed the comments and added a new parameter `filter_order` as we discussed. Interactive filtering will only be enabled when this parameter is provided.

In terms of performance still isn't great, but I reduced it to making a single db query per filter group as it was before.

arielsvn:
https://github.com/AlexsLemonade/refinebio-frontend/pull/432 updates the frontend to send the `filter_order` parameter introduced in this request.

PR #838, opened 2018-11-14
Set Limit To Maximum Concurrent Survey Jobs

Miserlou:
## Issue Number

https://github.com/hashicorp/nomad/issues/4864

Nomad is garbage, part two.

We keep all of the job needs in the database, so we don't need to have all of the stuff dispatched to a queue anyway. This sets an arbitrary limit on the number of dispatched survey jobs at a time. It looks like our problems mostly started at >100,000 jobs, so the new limit of 1000 should be safer without limiting our throughput.

## Types of changes

## Functional tests

New test added. Currently running, might fail due to some unrelated R garbage.

### DIFF
--- a/foreman/data_refinery_foreman/foreman/main.py
+++ b/foreman/data_refinery_foreman/foreman/main.py
@@ -32,6 +32,9 @@ RUNNING_IN_CLOUD = get_env_variable_gracefully("RUNNING_IN_CLOUD", False)
 # greater than this because of the first attempt
 MAX_NUM_RETRIES = 2
 
+# This can be overritten by the env var "MAX_TOTAL_JOBS"
+DEFAULT_MAX_JOBS = 10000
+
 # The fastest each thread will repeat its checks.
 # Could be slower if the thread takes longer than this to check its jobs.
 MIN_LOOP_TIME = timedelta(minutes=2)
@@ -145,9 +148,9 @@ def handle_downloader_jobs(jobs: List[DownloaderJob]) -> None:
     nomad_client = Nomad(nomad_host, port=int(nomad_port), timeout=30)
     # Maximum number of total jobs running at a time.
     # We do this now rather than import time for testing purposes.
-    MAX_TOTAL_JOBS = int(get_env_variable_gracefully("MAX_TOTAL_JOBS", 1000))
-    all_jobs = nomad_client.jobs.get_jobs()
-    if len(all_jobs) >= MAX_TOTAL_JOBS:
+    MAX_TOTAL_JOBS = int(get_env_variable_gracefully("MAX_TOTAL_JOBS", DEFAULT_MAX_JOBS))
+    len_all_jobs = len(nomad_client.jobs.get_jobs())
+    if len_all_jobs >= MAX_TOTAL_JOBS:
         logger.info("Not requeuing job until we're running fewer jobs.")
         return False
 
@@ -159,7 +162,7 @@ def handle_downloader_jobs(jobs: List[DownloaderJob]) -> None:
         else:
             handle_repeated_failure(job)
 
-        if jobs_dispatched >= MAX_TOTAL_JOBS:
+        if (jobs_dispatched + len_all_jobs) >= MAX_TOTAL_JOBS:
             logger.info("We hit the maximum total jobs ceiling, so we're not handling any more downloader jobs now.")
             return False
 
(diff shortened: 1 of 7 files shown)


Miserlou:
I think bumping `rlang` fixed that R problem.

foreman/data_refinery_foreman/foreman/main.py
```diff
@@ -425,6 +425,20 @@ def requeue_survey_job(last_job: SurveyJob) -> None:
...
     """
+    nomad_host = get_env_variable("NOMAD_HOST")
+    nomad_port = get_env_variable("NOMAD_PORT", "4646")
+    nomad_client = Nomad(nomad_host, port=int(nomad_port), timeout=30)
+    # Maximum number of total jobs running at a time.
+    # We do this now rather than import time for testing purposes.
+    MAX_TOTAL_JOBS = int(get_env_variable_gracefully("MAX_TOTAL_JOBS", 1000))
+
+    lost_jobs = []
+    all_jobs = nomad_client.jobs.get_jobs()
```
kurtwheeler: This could be called in `handle_survey_jobs` and then the length of it could be used to determine how many times to call this function, rather than calling for each individual job.

Miserlou:
Like that?

foreman/data_refinery_foreman/foreman/main.py
```diff
@@ -139,12 +139,31 @@ def requeue_downloader_job(last_job: DownloaderJob) -> None:
...
+    nomad_client = Nomad(nomad_host, port=int(nomad_port), timeout=30)
+    # Maximum number of total jobs running at a time.
+    # We do this now rather than import time for testing purposes.
+    MAX_TOTAL_JOBS = int(get_env_variable_gracefully("MAX_TOTAL_JOBS", 1000))
+    all_jobs = nomad_client.jobs.get_jobs()
+    if len(all_jobs) >= MAX_TOTAL_JOBS:
+        logger.info("Not requeuing job until we're running fewer jobs.")
+        return False
+
+    jobs_dispatched = 0
```
kurtwheeler: This should start at `len(all_jobs)` not 0.

foreman/data_refinery_foreman/foreman/main.py
```diff
@@ -306,12 +325,31 @@ def requeue_processor_job(last_job: ProcessorJob) -> None:
...
```
kurtwheeler: Same.

foreman/data_refinery_foreman/foreman/main.py
```diff
@@ -484,14 +510,34 @@ def requeue_survey_job(last_job: SurveyJob) -> None:
...
```
kurtwheeler: Same.

foreman/data_refinery_foreman/foreman/main.py
```diff
@@ -139,12 +139,31 @@ def requeue_downloader_job(last_job: DownloaderJob) -> None:
...
 def handle_downloader_jobs(jobs: List[DownloaderJob]) -> None:
     """For each job in jobs, either retry it or log it."""
+
+    nomad_host = get_env_variable("NOMAD_HOST")
+    nomad_port = get_env_variable("NOMAD_PORT", "4646")
+    nomad_client = Nomad(nomad_host, port=int(nomad_port), timeout=30)
+    # Maximum number of total jobs running at a time.
+    # We do this now rather than import time for testing purposes.
+    MAX_TOTAL_JOBS = int(get_env_variable_gracefully("MAX_TOTAL_JOBS", 1000))
+    all_jobs = nomad_client.jobs.get_jobs()
```
kurtwheeler: I just realized that this will include dead jobs, we only want to consider pending and running jobs.

Miserlou:
The scalability problem isn't from the number of jobs the system can place, it's from how many Nomad can keep track of. I don't see any benefit to having it be higher 10000 (20TB/2GB), and everything more than that is just going to eat performance and stability. Even 10K seems high to me. Similarly, the status of jobs doesn't matter since they're still going to duplicated across the cluster (which is where the memory problems come from), so we'll want to wait for the dead ones to get GC'd anyway.

The threading issue was the original reason I had it check the length on a per-job basis, which we could still do (and is ultimately really the only way to be accurate, since processor jobs can dispatch jobs directly as well). If you want to put the whole foreman back into a single thread.. why did we bother with a thread pool at all? Do we still want to do that? Serious question.

kurtwheeler:
> If you want to put the whole foreman back into a single thread.. why did we bother with a thread pool at all? Do we still want to do that? Serious question.

I think it was a case of overzealous premature optimization on my part. Thinking about it now there really isn't reason for the Foreman to be multithreaded and I think now there's a good reason to revert that: by only having one place the Foreman queues jobs, we can give it better control over the queue and potentially later add logic over time to prioritize jobs.

I wasn't suggesting we go higher than 10k and it does make sense that we need to be concerned about dead jobs as well. I don't know what Nomad's GC cylce is like I just hope it won't let thousands of jobs sit in the dead state. As long as we always have pending jobs we should be good though.

You're right about querying Nomad before queuing each job being the only way to be perfectly accurate about the job queue depth, however Nomad's API always seemed somewhat slow to me and I was worried about the impact hammering it with HTTP requests as fast as possible would be on performance and stability as well.

Issue #796, opened 2018-11-09
We need a bigger deploy box, or to do it smarter


kurtwheeler:
### Context

The latest deployment got hung up a couple times because the deploy box got overwhelmed:
![image](https://user-images.githubusercontent.com/4026833/48235834-8b31f100-e38d-11e8-9b0a-0dba40b63f37.png)

I figured out that it had died by checking out the number of nomad jobs in the queue:
```
ubuntu@ip-10-0-233-82:~$ nomad status | wc -l
43110
ubuntu@ip-10-0-233-82:~$ nomad status | wc -l
43110
ubuntu@ip-10-0-233-82:~$ nomad status | wc -l
43110
ubuntu@ip-10-0-233-82:~$ nomad status | wc -l
43110
ubuntu@ip-10-0-233-82:~$ nomad status | wc -l
43110
ubuntu@ip-10-0-233-82:~$ nomad status | wc -l
43110
ubuntu@ip-10-0-233-82:~$ nomad status | wc -l
22798
ubuntu@ip-10-0-233-82:~$ nomad status | wc -l
22663
ubuntu@ip-10-0-233-82:~$ nomad status | wc -l
22525
ubuntu@ip-10-0-233-82:~$ nomad status | wc -l
22140
ubuntu@ip-10-0-233-82:~$ nomad status | wc -l
21693
ubuntu@ip-10-0-233-82:~$ nomad status | wc -l
16515
ubuntu@ip-10-0-233-82:~$ nomad status | wc -l
14447
ubuntu@ip-10-0-233-82:~$ nomad status | wc -l
9602
ubuntu@ip-10-0-233-82:~$ nomad status | wc -l
1347
ubuntu@ip-10-0-233-82:~$ nomad status | wc -l
1347
ubuntu@ip-10-0-233-82:~$ nomad status | wc -l
1347
ubuntu@ip-10-0-233-82:~$ nomad status | wc -l
1347
ubuntu@ip-10-0-233-82:~$ nomad status | wc -l
1347
ubuntu@ip-10-0-233-82:~$ nomad status | wc -l
1347
```

### Problem or idea

I think that what is happening is that when there's 10's of thousands of jobs the loop we kill the jobs in ends up getting too big and overwhelms the instance, causing the process to get killed.

This is the loop in question: https://github.com/AlexsLemonade/refinebio/blob/dev/infrastructure/deploy.sh#L210

### Solution or next step

There's probably two ways to fix this:
[...]

PR #842, opened 2018-11-15
Kill nomad jobs in groups of 500 during deploys

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/796

The old strategy for killing nomad jobs was to loop through them all with bash, kill each job in a background process, and then wait for all the processes to complete. However this ended up blowing out our RAM usage very heavily causing the issue reported in #796. (At least this what I think crashed it.)

This changes it so we only fire off 500 background processes at a time before waiting for them to complete so we won't have so many running at once.

## Types of changes

## Functional tests

Testing this was harder than implementing it, but what I did was:
  * Modified `deploy.sh` so only the nomad killing part ran and that the Nomad address was set to my local instance of Nomad.
  * Modified `run_nomad.sh` so it ran in server mode instead of dev mode. This meant that dispatched jobs wouldn't run so I could build up a reliable amount of them and not blow out my laptop running actual work.
  * Queued up 10k surveyor jobs using `./foreman/run_surveyor.sh surveyor_dispatcher --file s3://data-refinery-test-assets/ONE_PERCENT_CRUNCH.txt`.
  * Ran the modified `deploy.sh` script to kill them.
  * Monitored my RAM usage in `htop`.

I did this with and without my change. The baseline htop stats after I queued all the jobs with my other processes was 11.5GB with a limit of 15.4GB.

Without this change the RAM got completely saturated and had to use 1.5GB of swap.

With this change RAM never exceeded 13.4 GB and swap was unutilized. I also saw a pattern of the RAM going up a GB or two and then back down once the script hit the periodic `wait`.

I call it a win!

## Checklist

### DIFF
--- a/infrastructure/deploy.sh
+++ b/infrastructure/deploy.sh
@@ -219,9 +219,9 @@ if [[ $(nomad status) != "No running jobs" ]]; then
             counter=$((counter+1))
         fi
 
-        # Wait for all the jobs to stop every 500 so we don't knock
+        # Wait for all the jobs to stop every 100 so we don't knock
         # over the deploy box if there are 1000's.
-        if [[ "$counter" -gt 500 ]]; then
+        if [[ "$counter" -gt 100 ]]; then
             wait $(jobs -p)
             counter=0
         fi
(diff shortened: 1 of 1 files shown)


kurtwheeler:
It seems like in the deploy that couldn't get all the way through we were able to do 10's of thousands, but it may have been using swap and slowing things down. I've dropped it down to 100, I think we don't wanna go lower than that because then the time we spend waiting could slow us down. We originally put the jobs in the background because waiting for each one individually was too slow. This seems like a happy medium now.

Miserlou:
Sounds good!

Issue #844, opened 2018-11-16
Deploys are Borked


Miserlou:
### Context

There appear to be multiple problems in the deploy process.

The logs for each tag are all going to the same file, so it's very difficult to see what's actually happening. The end of the file looks like a successful deployment for `v1.1.3`, but there also references for everything up to `1.1.7-dev`. There are no timestamps, so it's impossible to tell what order things happened in. Staging and deploy logs are also sent to the same place, so it's impossible to distinguish which stage is happening.

It also appears that `nomad` and `terraform` are redownloaded constantly during the deploy process:

```
$ ls
git-crypt                  nomad_0.8.3_SHA256SUMS.7       nomad_0.8.3_SHA256SUMS.sig.5    nomad_0.8.3_linux_amd64.zip.3   terraform_0.11.8_SHA256SUMS.16      terraform_0.11.8_SHA256SUMS.sig.14   terraform_0.11.8_linux_amd64.zip.12
nomad_0.8.3_SHA256SUMS     nomad_0.8.3_SHA256SUMS.8       nomad_0.8.3_SHA256SUMS.sig.6    nomad_0.8.3_linux_amd64.zip.4   terraform_0.11.8_SHA256SUMS.17      terraform_0.11.8_SHA256SUMS.sig.15   terraform_0.11.8_linux_amd64.zip.13
nomad_0.8.3_SHA256SUMS.1   nomad_0.8.3_SHA256SUMS.9       nomad_0.8.3_SHA256SUMS.sig.7    nomad_0.8.3_linux_amd64.zip.5   terraform_0.11.8_SHA256SUMS.18      terraform_0.11.8_SHA256SUMS.sig.16   terraform_0.11.8_linux_amd64.zip.14
nomad_0.8.3_SHA256SUMS.10  nomad_0.8.3_SHA256SUMS.sig     nomad_0.8.3_SHA256SUMS.sig.8    nomad_0.8.3_linux_amd64.zip.6   terraform_0.11.8_SHA256SUMS.2       terraform_0.11.8_SHA256SUMS.sig.17   terraform_0.11.8_linux_amd64.zip.15
[...]

PR #845, opened 2018-11-16
Unbork Deploys

Miserlou:
## Issue Number
https://github.com/AlexsLemonade/refinebio/issues/844

Attempts to fix the issues with the current deploy set up.

- Sets local env vars required for checking remote tag existence.
- Uses different log file paths for different deploys
- Uses `wget -N` to avoid downloading duplicate files but only checking the last one

## Types of changes

## Tests

Haven't tested yet since it requires a fairly complicated set up that we don't have a handy way to reproduce yet.

### DIFF
--- a/.circleci/run_terraform.sh
+++ b/.circleci/run_terraform.sh
@@ -5,9 +5,9 @@ curl https://keybase.io/hashicorp/pgp_keys.asc | gpg --import
 # Install terraform and nomad
 cd
 TERRAFORM_VERSION=0.11.8
-wget https://releases.hashicorp.com/terraform/$TERRAFORM_VERSION/terraform_${TERRAFORM_VERSION}_linux_amd64.zip
-wget https://releases.hashicorp.com/terraform/$TERRAFORM_VERSION/terraform_${TERRAFORM_VERSION}_SHA256SUMS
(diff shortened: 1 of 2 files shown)


.circleci/run_terraform.sh
```diff
@@ -5,9 +5,9 @@ curl https://keybase.io/hashicorp/pgp_keys.asc | gpg --import
 # Install terraform and nomad
 cd
 TERRAFORM_VERSION=0.11.8
-wget https://releases.hashicorp.com/terraform/$TERRAFORM_VERSION/terraform_${TERRAFORM_VERSION}_linux_amd64.zip
-wget https://releases.hashicorp.com/terraform/$TERRAFORM_VERSION/terraform_${TERRAFORM_VERSION}_SHA256SUMS
-wget https://releases.hashicorp.com/terraform/$TERRAFORM_VERSION/terraform_${TERRAFORM_VERSION}_SHA256SUMS.sig
+wget -N https://releases.hashicorp.com/terraform/$TERRAFORM_VERSION/terraform_${TERRAFORM_VERSION}_linux_amd64.zip
```
cgreene: I am a fan of timestamping 👍

PR #852, opened 2018-11-19
Make it so that additional commits to dev won't break deploys that ar…

kurtwheeler:
…e running. Also makes it so that get_master_or_dev will allow logging.

## Issue Number

N/A Came up during a deploy: https://circleci.com/gh/AlexsLemonade/refinebio/9793

I merged a PR after I tagged v1.1.12-dev. This meant that the latest commit to the `dev` branch wasn't the same commit that got tagged, so our deploy stopped because it thought it was a bad tag. This makes it so we look through the list of tags to find the tag we're trying to deploy and then also see if that commit was on dev.

I also made that function not exit with status of 1 because it made it so the scripts exited before they could log why they were failing.

## Types of changes

## Functional tests

I've tested running the function and it works well, although it cannot differentiate between master and dev if a commit is part of both branches.

## Checklist


kurtwheeler:
So I'm not actually sure we want to fix things in this way. This implementation has a limitation in the scenario where if a commit has been merged into master and then tagged from the dev branch, we won't actually be able to tell that it was tagged on the dev branch. All we know is what commit got tagged, but not what branch that commit is supposed to be part of since that's not really how git branches work.

However, I don't think that there is a way to be able to tell what branch a tag happens on. The three options I can think of are:
1. Doing it this way and making it clear that a commit which has landed in master should not be used for a staging deploy
2. Formalize the `vX.X.X-dev` tag pattern actually check for that in the tag string.
3. Keep it how it is and just be careful with our merges while deploying.

I think I prefer option 1 because option 2 doesn't prevent accidentally tagging a branch other dev or master and option 3 means that everyone has to be sure there's no deploys queued up before they click the merge button.

Miserlou:
We settled on using a combination of 1) and 2).

kurtwheeler:
Ok, I've implemented the 1) and 2) combo and tested it. Seems to be good!

PR #859, opened 2018-11-21
Lower Nomad Garbage Collection Interval

Miserlou:
## Issue Number
n/a

Lowers the Nomad GC interval from the default 4h to 5m. This uses less resources and gives us much more time to allocate jobs before hitting our self-impose ceiling.

### DIFF
--- a/infrastructure/nomad-configuration/lead_server.hcl
+++ b/infrastructure/nomad-configuration/lead_server.hcl
@@ -10,6 +10,9 @@ data_dir = "/tmp/nomad_server1"
 server {
     enabled = true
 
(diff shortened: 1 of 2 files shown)


Miserlou:
Also raises disk size to avoid 

```
2018-11-21T16:10:10Z  Driver Failure  failed to initialize task "no_op" for alloc "6c62abfc-c06b-8e53-12e4-30691bda206d": Failed to pull `ccdl/dr_no_op:v1.2.0`: write /var/lib/docker/tmp/GetImageBlob461560485: no space left on device
```

and doubles the max jobs limit.

PR #872, opened 2018-11-27
Gives Clients a Cronjob to Kill Hung Salmon (fasterq-dump) Jobs

Miserlou:
## Issue Number
n/a

Lots of our Salmon processors were getting stuck, but their containers weren't dying. This seemed to be occuring on the `fasterq-dump` step.

Nomad provides no native timeout functionality, so we have to do it ourselves.

This PR installs a new python script during client init to kill all jobs 3 hours or older, and installs it with a crontab.

## Methods

Getting machine-readable status out of docker is a PITA.

## Types of changes

Before:

screen shot 2018-11-27 at 2 45 08 pm

After:

screen shot 2018-11-27 at 2 45 17 pm


Miserlou:
The problem was fasterq-dump, not Salmon:

Related:
https://github.com/ncbi/sra-tools/issues/159
https://github.com/ncbi/sra-tools/issues/161

PR #874, opened 2018-11-28
Adds timestamps to the S3 keys for transcriptome indices.

kurtwheeler:
## Issue Number

N/A came up because we think we need to recreate a few transcriptome indices but don't want to lose the ones we have.

We don't want to lose data, but every time a transcriptome index is created it replaces the last one for its organism/length. This adds a timestamp so we don't overwrite S3 objects.

## Types of changes

## Functional tests

None yet, this is only applicable in the cloud so I either need to test it in a dev stack or in the staging stack.

## Checklist

### DIFF
--- a/common/data_refinery_common/logging.py
+++ b/common/data_refinery_common/logging.py
@@ -4,16 +4,16 @@ import logging
 import sys
 
 from data_refinery_common.utils import get_env_variable_gracefully
-from data_refinery_common.utils import get_instance_id
+from data_refinery_common.utils import get_instance_id, get_volume_index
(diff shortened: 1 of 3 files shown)


kurtwheeler:
@Miserlou I didn't realize you had already approved this. I just pushed some additional changes to address #875, the log lines look like:
```
2018-11-29 15:40:20,662 local [volume: 0] data_refinery_foreman.surveyor.array_express INFO [survey_job: 20]: Surveying experiment with accession code: E-MTAB-3050.
```

PR #876, opened 2018-11-29
Prioritize queuing of downloader and processor jobs in foreman

kurtwheeler:
## Issue Number

N/A We never made an issue for this.

This PR makes the Foreman prioritize downloader and processor jobs. The prioritization is explained in a comment:
```
    # We want zebrafish data first, then hgu133plus2, then data
    # related to pediatric cancer, then to finish salmon experiments
    # that are close to completion.
```

The limitation of this PR is that it does it on each different monitoring function rather than prioritizing all jobs at once. However I think that this will be good enough.

## Types of changes

## Functional tests

The only way to really functionally test this is at scale, but the unit tests I added are rather comprehensive.

## Checklist


foreman/data_refinery_foreman/foreman/main.py
```diff
@@ -119,6 +137,123 @@ def get_active_volumes() -> Set[str]:
...
+
+    # Remove all the jobs we're moving to the front of the list
+    for job in sorted_jobs:
+        jobs.remove(job)
+
+    return sorted_jobs + jobs
+
+
+def prioritize_zebrafish_jobs(jobs: List) -> List:
+    zebrafish_jobs = []
```
Miserlou: docstring to explain what this does - is it sorting a list? pruning a list? etc. Same for function below

foreman/data_refinery_foreman/foreman/main.py
```diff
@@ -119,6 +137,123 @@ def get_active_volumes() -> Set[str]:
...
+
+        # We cannot simply filter on is_processed because that field
+        # doesn't get set until every sample in an experiment is processed.
+        # Instead we are looking for one successful processor job.
+        processed_samples = 0
+        for related_sample in related_samples:
+            sample_is_processed = False
+            for processor_job in related_sample.original_files.first().processor_jobs.all():
+                if processor_job.success == True:
+                    sample_is_processed = True
```
Miserlou: break here?

foreman/data_refinery_foreman/foreman/main.py
```diff
@@ -119,6 +137,123 @@ def get_active_volumes() -> Set[str]:
...
+        if type(job) is ProcessorJob and job.pipeline_applied != ProcessorPipeline.SALMON.value:
+            continue
+        elif type(job) is DownloaderJob:
+            is_salmon_sample = False
+            for original_file in sample.original_files.all():
+                if original_file.filename[-5:].upper() == "FASTQ" \
+                or original_file.filename[-8:].upper() == "FASTQ.GZ" \
+                or original_file.filename[-2:].upper() == "FQ" \
+                or original_file.filename[-3:].upper() == "SRA" \
+                or original_file.filename[-5:].upper() == "FQ.GZ":
```
Miserlou: Are these based on real data, or just memory? I feel like some (compressed?) file formats might be missing here.

foreman/data_refinery_foreman/foreman/main.py
```diff
@@ -119,6 +137,123 @@ def get_active_volumes() -> Set[str]:
...
+            for related_sample in experiment.samples.all():
+                related_samples.add(related_sample)
+
+        # We cannot simply filter on is_processed because that field
+        # doesn't get set until every sample in an experiment is processed.
+        # Instead we are looking for one successful processor job.
+        processed_samples = 0
+        for related_sample in related_samples:
+            sample_is_processed = False
+            for processor_job in related_sample.original_files.first().processor_jobs.all():
```
Miserlou: I initially thought that this might be wrong long since it's an `all`. Is there any advantage to doing it this way rather than using a `.filter(success=True)` rather than `.all` and then manually checking?

What happens if there are samples with no OriginalFiles? I think first() will return None, which doesn't have a `processor_jobs` so this will crash?

foreman/data_refinery_foreman/foreman/main.py
kurtwheeler: I got these out of `common/job_lookup.py` because that's what we use to decide if a file is RNA-seq there.

Miserlou: Roger roger. In that case, can it be functionalized?

PR #877, opened 2018-11-29
Better Janitor Error Handling // Index Integrity Verification

Miserlou:
## Issue Number
n/a

Small fixes, one to make the Janitor slightly more robust but slightly less aggressive, and one more to ensure that indexes aren't corrupt after download.

## Types of changes

### DIFF
--- a/foreman/data_refinery_foreman/foreman/main.py
+++ b/foreman/data_refinery_foreman/foreman/main.py
@@ -447,6 +447,16 @@ def retry_hung_processor_jobs() -> None:
                     hung_jobs.append(job)
         except URLNotFoundNomadException:
             hung_jobs.append(job)
+        except TypeError:
+            # Almost certainly a python-nomad issue:
+            # File "/usr/local/lib/python3.5/dist-packages/nomad/api/job.py", line 63, in get_job
+            #   return self.request(id, method="get").json()
+            # File "/usr/local/lib/python3.5/dist-packages/nomad/api/base.py", line 74, in request
+            #   endpoint = self._endpoint_builder(self.ENDPOINT, *args)
+            # File "/usr/local/lib/python3.5/dist-packages/nomad/api/base.py", line 28, in _endpoint_builder
+            #   u = "/".join(args)
+            # TypeError: sequence item 1: expected str instance, NoneType found
+            logger.info("Couldn't query Nomad about Processor Job.", processor_job=job.id)
         except Exception:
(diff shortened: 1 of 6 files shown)


foreman/data_refinery_foreman/foreman/main.py
```diff
@@ -447,6 +447,16 @@ def retry_hung_processor_jobs() -> None:
...
+        except TypeError:
+            # Almost certainly a python-nomad issue:
+            # File "/usr/local/lib/python3.5/dist-packages/nomad/api/job.py", line 63, in get_job
+            #   return self.request(id, method="get").json()
+            # File "/usr/local/lib/python3.5/dist-packages/nomad/api/base.py", line 74, in request
+            #   endpoint = self._endpoint_builder(self.ENDPOINT, *args)
+            # File "/usr/local/lib/python3.5/dist-packages/nomad/api/base.py", line 28, in _endpoint_builder
+            #   u = "/".join(args)
+            # TypeError: sequence item 1: expected str instance, NoneType found
+            logger.info("Couldn't query Nomad about Processor Job.", processor_job=job.id)
```
kurtwheeler: I've seen this before. I think it happens when you pass None to get_job()

workers/data_refinery_workers/processors/janitor.py
```diff
@@ -59,21 +59,32 @@ def _find_and_remove_expired_jobs(job_context):
...
+            except ProcessorJob.DoesNotExist:
+                # This job has vanished from the DB - clean it up!
+                logger.exception("Janitor found no record of " + item + " - why?")
+                pass
             except Exception:
                 # If we don't have a record of the job, we don't need the directory.
-                pass
-
+                logger.exception("Problem finding job record for " + item + " - why?")
+                continue
```
kurtwheeler: The comment above this now seems out of date.

Miserlou: There are two separate cases here, that we are handling in different ways. The first (object DNE) - then we say we found no record of it, and we `pass` so it gets deleted. If there is another type of Exception, that means that we are likely currently unable to access the database (PgBouncer is down, for example), so we don't want to perform any destructive action yet, hence the continue and "problem finding record".

kurtwheeler: Right, but I'm talking about the `# If we don't have a record of the job, we don't need the directory.` comment which used to precede the `pass`, but now it's before a continue so I don't think it makes sense.

PR #885, opened 2018-12-03
Hush Up the Foreman

Miserlou:
## Issue Number

Foreman is polluting the prod logs.

Also includes an explicit death check for missing `'internal_accession'` in AFFY, a logger error message for zerobyte t-indexes and sets a missing `job_context["job"].failure_reason`

### DIFF
--- a/workers/data_refinery_workers/processors/no_op.py
+++ b/workers/data_refinery_workers/processors/no_op.py
@@ -138,6 +138,7 @@ def _prepare_files(job_context: Dict) -> Dict:
         job_context["internal_accession"] = get_internal_microarray_accession(job_context["platform"])
         if not job_context["internal_accession"]:
             logger.error("Failed to find internal accession for code", code=job_context["platform"])
+            job_context["job"].failure_reason = "Failed to find internal accession for code" + job_context["platform"]
             job_context['success'] = False
             job_context["job"].no_retry = True
     except Exception as e:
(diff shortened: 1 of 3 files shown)


workers/data_refinery_workers/processors/no_op.py
```diff
@@ -160,6 +161,14 @@ def _convert_genes(job_context: Dict) -> Dict:
 def _convert_affy_genes(job_context: Dict) -> Dict:
     """ Convert to Ensembl genes if we can"""
 
+    if 'internal_accession' not in job_context:
+        error_msg = "Told to convert AFFY genes without an internal_accession - how did this happen?"
```
kurtwheeler: I'm not a fan of this "how did this happen" because it implies this is unexpected or that we don't know why it's happening, neither of which is the case.

Miserlou: This is unexpected, since it should have been taken care of in the prepare files function.

kurtwheeler: Ah okay, I thought that this was just about us not recognizing a platform, which is an expected error. However this is about `internal_accession` not being on `job_context`, which is weird and I cannot figure out how in the world it would ever happen.

kurtwheeler: So then the only suggestion I have is that this should probably be checking for `if 'internal_accession' not in job_context or not job_context['internal_accession']:

Miserlou: Addressed

PR #887, opened 2018-12-04
Default Volume Index Is Inaccurate

Miserlou:
## Issue Number
n/a

The logger is using '0' default volume index value, even though 0 is a valid volume index, so it looks like the Foreman is dispatching jobs from a client instance:

```
data-refinery-log-group-circleci-prod log-stream-foreman-circleci-prod 2018-12-04 15:33:08,156 i-0916d6941054aef49 [volume: 0] data_refinery_foreman.foreman.main INFO [index: 18] [job_id: 561731]: Sending Janitor with index:
data-refinery-log-group-circleci-prod log-stream-foreman-circleci-prod 2018-12-04 15:33:08,157 i-0916d6941054aef49 [volume: 0] data_refinery_common.message_queue INFO: Queuing JANITOR nomad job to run job JANITOR with id 561731.
``` 

## Methods

This PR changes the behavior to use -1 as the default index.

### DIFF
--- a/common/data_refinery_common/test_utils.py
+++ b/common/data_refinery_common/test_utils.py
@@ -94,7 +94,7 @@ class UtilsTestCase(TestCase):
 
     def test_volume_index(self):
         """Test that supported RNASeq platforms setting is set correctly."""
-        self.assertEqual(utils.get_volume_index(), "-1")
+        self.assertEqual(utils.get_volume_index(), "0")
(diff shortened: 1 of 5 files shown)


Miserlou:
Also fixes a pretty bad bug in `@do_forever` that caused `cleanup_the_queue` and `janitor` to be run every two minutes rather than every 30 minutes.

Issue #508, opened 2018-08-09
Processing species-level compendia


jaclyn-taroni:
### Context

I have been thinking quite a bit about integrating platforms recently as a result of the QN discussion (#488).

### Problem or idea

Because we're using an inner join in the smasher, we'll drop genes that are not included in every platform. If we were to include [`HGU133B`](https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GPL97) and [`HGU95E`](https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GPL95) in the human compendium in this manner, we would presumably be missing many highly expressed and important genes!

### Solution or next step

Tagging @cgreene for discussion and @Miserlou to comment on implementation. There are a few routes I can think of at this time, though I can not comment on feasibility:

*  Change smasher behavior in the compendium case such that we fill in missing genes with NAs. This would mean changing the arguments to `pandas.merge`. This would also probably change how we do QN. (I believe `preprocessCore::normalize.quantiles.use.target`, if we go that route, assumes genes are missing at random. Not a good assumption here.)
* We curate a list of platforms that we don't include in the compendium. This may be trickier for model organisms where I have less foundational knowledge, but I am sure I could figure it out. 

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

cgreene:
For the compendium of all the data for an organism, we'll need to give it some TLC. This likely means that we'll need to include the NAs (so we should have this be configurable for the smasher at least via the API, even if we use the inner join as the default for the web UI). Then we will likely want to prune things that are too missing to impute (>30% missing is the usual threshold). After pruning, we'll likely want to use KNN to impute the rest.

jaclyn-taroni:
A quick google of "knn impute python" gets me: https://github.com/iskandr/knnimpute

cgreene:
There are a lot of ways to do knn impute poorly, but that one looks good at a first glance.

jaclyn-taroni:
To summarize, for species-level compendia:
* We'd want to do a full outer join (union of all genes). Missing values should be NAs (I'm assuming that this is the default behavior, but we don't want it to be zero or some other value).
* Drop genes with >30% missing values
* Impute values (KNN impute)
* Quantile normalize

jaclyn-taroni:
[...]

PR #888, opened 2018-12-04
[WIP] Species Compendia

Miserlou:
## Issue Number

#508

See discussion on the related ticket. I stopped working on this PR to work on ops stuff, filing the PR now to not forget about about it.

## Types of changes

## Functional tests

## Checklist


workers/data_refinery_workers/processors/create_compendia.py
```diff
@@ -0,0 +1,218 @@
...
+    rows_to_filter = []
+    for (x, sum_val) in rnaseq_row_sums.items():
+        if sum_val < rnaseq_tenth_percentile:
+            rows_to_filter.append(x)
+
+    filtered_rnaseq_matrix = rnaseq_expression_matrix.drop(rows_to_filter)
+
+    # log2(x + 1) transform filtered_rnaseq_matrix; this is now log2_rnaseq_matrix
+    filtered_rnaseq_matrix_plus_one = filtered_rnaseq_matrix + 1
+    log2_rnaseq_matrix = np.log2(filtered_rnaseq_matrix)
```
jaclyn-taroni: Should this be `np.log2(filtered_rnaseq_matrix_plus_one)` or is there something OOP-y I'm missing? Based on the fact that you need to remove `-Inf` and `Inf` values here: https://github.com/AlexsLemonade/refinebio/pull/888/files#diff-5b6c34e495c6bb6ddf4320b09359c528R141, I'm guessing the add 1 step was not successful.

workers/data_refinery_workers/processors/create_compendia.py
```diff
@@ -0,0 +1,218 @@
...
+
+    # Set all zero values in log2_rnaseq_matrix to NA, but make sure to keep track of where these zeroes are
+    log2_rnaseq_matrix[log2_rnaseq_matrix==0]=np.nan
+
+    # Perform a full outer join of microarray_expression_matrix and log2_rnaseq_matrix; combined_matrix
+    combined_matrix = pd.merge(microarray_expression_matrix, log2_rnaseq_matrix, how='outer', left_index=True, right_index=True)
+
+    # Remove genes (rows) with >30% missing values in combined_matrix
+    thresh = len(combined_matrix.columns) * .6
```
jaclyn-taroni:
```suggestion
    thresh = len(combined_matrix.columns) * .7
```
Would be 30%?

workers/data_refinery_workers/processors/create_compendia.py
```diff
@@ -0,0 +1,225 @@
...
+    row_filtered_combined_matrix = combined_matrix.dropna(axis=0, thresh=thresh, how='any')
+
+    # Remove samples (columns) with >50% missing values in combined_matrix
+    # XXX: Find better test data for this!
+    thresh = len(combined_matrix.columns) * .5
+    row_col_filtered_combined_matrix_samples = row_filtered_combined_matrix.dropna(axis=1, thresh=thresh, how='any')
+
+    # "Reset" zero values that were set to NA in RNA-seq samples (i.e., make these zero again) in combined_matrix
+    # XXX: IterativeSVD doesn't see 0.000 as something to impute, only NA.
+    # So, we skip this step!
```
jaclyn-taroni: We don't want to skip this step. As we talked about yesterday, zeroes in RNA-seq data could be technical or biological (we have no easy way of knowing, here). We're setting the zeroes to NAs above to make it easier to perform the row and column filtering in lines 125-132. 

We do not want to impute the values that were originally zero in RNA-seq data, so we want to set them to `0.000` instead of leaving them as `NA`. 

Because the indices are changing based on the filtering, I don't think it will be sufficient to keep track of the indices where the zeros _were_. Instead, it makes sense to me to keep track of the gene labels and sample labels for values that were originally zero in the RNA-seq data and then set (any remaining post-filtering) values to zero.

You can also do this _after_ imputation but before quantile normalization, but I can see an argument where it's more efficient to do it _before_ imputation (= less values to impute) but I have no idea whether this actually makes a difference performance-wise.

Happy to chat in person or poke around at this step.

workers/data_refinery_workers/processors/create_compendia.py
```diff
@@ -0,0 +1,293 @@
...
+    organism_key = list(job_context['samples'].keys())[0]
+    annotation = ComputationalResultAnnotation()
+    annotation.result = result
+    annotation.data = {
+        "organism_id": job_context['samples'][organism_key][0].organism_id,
+        "organism_name": job_context['samples'][organism_key][0].organism.name,
+        "is_qn": False,
+        "is_compendia": True,
+        "samples": [sample.accession_code for sample in job_context["samples"][organism_key]],
+        "num_samples": len(job_context["samples"][organism_key])
```
kurtwheeler: Maybe it would also be good to have the experiments that were used listed in the annotation?

workers/data_refinery_workers/processors/test_compendia.py
```diff
@@ -0,0 +1,268 @@
...
+        result = ComputationalResult()
+        result.save()
+
+        danio_rerio = Organism.get_object_for_name("DANIO_RERIO")
+
+        micros = [] 
+        for file in os.listdir('/home/user/data_store/raw/TEST/MICROARRAY/'):
+
+            if 'microarray.txt' in file:
+                continue
```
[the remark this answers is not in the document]
kurtwheeler: Nvm, I just read the test script better and I can see that it gets copied over as well.

workers/data_refinery_workers/processors/test_compendia.py
```diff
@@ -0,0 +1,268 @@
...
+        experiment.accession_code = "GSE5678"
+        experiment.save()
+
+        result = ComputationalResult()
+        result.save()
+        rnas = []
+        for file in os.listdir('/home/user/data_store/raw/TEST/RNASEQ/'):
+
+            if 'rnaseq.txt' in file:
+                continue
```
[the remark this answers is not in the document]
kurtwheeler: Nvm, I just read the test script better and I can see that it gets copied over as well.

api/data_refinery_api/views.py
```diff
@@ -1057,3 +1058,19 @@ def get(self, request, format=None):
...
+###
+# Compendia
+###
+
+class CompendiaDetail(APIView):
+    """
+    Compendi
+    """
+
+    """List all processors."""
```
kurtwheeler: These docstrings are wonky.

common/data_refinery_common/migrations/0009_auto_20190118_1759.py
```diff
@@ -0,0 +1,29 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('data_refinery_common', '0008_auto_20190104_2135'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='computedfile',
+            name='compendia_organism',
```
kurtwheeler: Any reason to have this be `compendia_organism` instead of just `organism`?

workers/data_refinery_workers/processors/create_compendia.py
```diff
@@ -0,0 +1,293 @@
...
+    metadata_computed_file.filename = job_context['metadata_tsv_paths'][0].split('/')[-1]
+    metadata_computed_file.calculate_sha1()
+    metadata_computed_file.calculate_size()
+    metadata_computed_file.is_smashable = False
+    metadata_computed_file.is_qn_target = False
+    metadata_computed_file.result = result
+    metadata_computed_file.save()
+
+    # Create the resulting archive
+    final_zip_base = "/home/user/data_store/smashed/" + str(job_context["dataset"].pk) + "_compendia"
```
kurtwheeler: Could we put the organism and compendia_version in the name of the archive file so that it's easy to tell while looking through s3?

Miserlou: We do, it's below this code though

Miserlou:
From discussion:

If not cost prohibitive (shouldn't be at all), let's just get this out to prod so we can run a ZF compendia with everything we have so far as it is, which will help us investigate what's going on. I suggested it may also be interesting to create 3 compendia for comparison: ALL, MICROARRAY and RNASEQ.

Miserlou:
We really need to get this out if you could unblock this please.

PR #891, opened 2018-12-05
Fixes Bug In Affy Processor When Files Are Missing  // Blacklist Unsupported Files // Improve Debuggability

Miserlou:
## Issue Number
n/a

Fixes our largest source of error (although these jobs aren't actually processable) and improves debuggability elsewhere.

## Types of changes


kurtwheeler:
:+1:

PR #895, opened 2018-12-05
Prevent surveyor jobs from requeuing work we've already done.

kurtwheeler:
## Issue Number

N/A came up while I was considering what to queue next.

 Turns out we haven't surveyed all the experiments we wanted to thus far, so we should requeue them, but this means we wanna make we don't requeue the work we've already done.

The other function that queues Downloader jobs was already making this check, but this one wasn't.

## Types of changes

## Functional tests

I think the unit test covers this well.

## Checklist

### DIFF
--- a/foreman/data_refinery_foreman/surveyor/external_source.py
+++ b/foreman/data_refinery_foreman/surveyor/external_source.py
@@ -111,6 +111,15 @@ class ExternalSourceSurveyor:
                                                 ):
         """Creates a single DownloaderJob with multiple files to download.
         """
+        source_urls = [original_file.source_url for original_file in original_files]
+        # There is already a downloader job associated with this file.
+        old_assocs = DownloaderJobOriginalFileAssociation.objects.filter(
+            original_file__source_url__in=source_urls)
+        if len(old_assocs) > 0:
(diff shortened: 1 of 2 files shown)


foreman/data_refinery_foreman/surveyor/external_source.py
```diff
@@ -111,6 +111,15 @@ def queue_downloader_job_for_original_files(self,
                                                 ):
         """Creates a single DownloaderJob with multiple files to download.
         """
+        source_urls = [original_file.source_url for original_file in original_files]
+        # There is already a downloader job associated with this file.
+        old_assocs = DownloaderJobOriginalFileAssociation.objects.filter(
+            original_file__source_url__in=source_urls)
+        if len(old_assocs) > 0:
+            logger.debug("We found an existing DownloaderJob for these urls.",
```
Miserlou: Is this bad enough to warrant a logger.error?

kurtwheeler: Definitely not. My plan is to requeue everything that we have already queued up to make sure that nothing got skipped because of surveyor jobs getting interruptted, so I expect this error to happen once for every sample we've already downloaded, which should be 100's of thousands.

PR #904, opened 2018-12-07
Adds a Nomad Job Specification for Creating QN Targets

Miserlou:
## Issue Number

n/a

It's a pain in the butt to copy the prod env to a client to create a prod QN target, so this creates a nomad job spec for creating QN targets.

It'd be great if you could test this locally since our Nomad job register setup doesn't work on OSX, so I haven't tested this yet.

## Types of changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -450,7 +450,6 @@ Alternately, this can be done as a normal Nomad dispatch jobs:
 nomad job dispatch -meta ORGANISM=DANIO_RERIO CREATE_QN_TARGET
 ```
 
-
 ### Checking on Local Jobs
 
 _Note:_ The following instructions assume you have set the environment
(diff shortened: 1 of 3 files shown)


kurtwheeler:
I tested this on my local machine. I was able to queue a nomad job with 
```
nomad job dispatch -meta ORGANISM=DANIO_RERIO CREATE_QN_TARGET
```

and it's log output was:
```
nlo CREATE_QN_TARGET/dispatch-1544207334-8b778220
2018-12-07 18:38:52,245 local [volume: 0] data_refinery_workers.processors.management.commands.create_qn_target ERROR: No processed samples for that organism.
```

which is to be expected since I don't have processed data on my local machine yet. However I think the best test will be to do this with a non-trivial amount of data, and if it doesn't work it's not too big of a deal since the feature already is missing.

Issue #832, opened 2018-11-09
API is Serializing Far Too Much Information


Miserlou:
### Context

Database queries are much faster, but now it's slow because of how much information we're serializing. 

### Problem or idea

I suggest we stop serializing complete sample lists in the Search and Experiment and just return accession codes so that the individual samples can be queried by the separate Samples API.

arielsvn:
Merging https://github.com/AlexsLemonade/refinebio/pull/861 caused performance problems with the API

cgreene:
Tests got commented out - should probably be uncommented (or at least looked at) as this is getting resolved:
https://github.com/AlexsLemonade/refinebio/pull/901/commits/319fe446244d451a096baa083248645152625df1#diff-e841c1c2392f296f719c2151aa8167aeR29
https://github.com/AlexsLemonade/refinebio/pull/901/commits/319fe446244d451a096baa083248645152625df1#diff-e841c1c2392f296f719c2151aa8167aeR64

arielsvn:
Closing this to continue the discussion in https://github.com/AlexsLemonade/refinebio/issues/928. There're other recommendations there to improve performance.

PR #927, opened 2018-12-12
Disable full text search

arielsvn:
## Issue Number

#832 

Changes the behavior of the search filters, disabling Full-text search on several properties. Locally this lowers the response time from 14secs to 3secs, with the data from refine.bio. I guess it should be related to how the search term is translated into the SQL query with the full text search.

```sql
WHERE to_tsvector(COALESCE("experiments"."description", '')) @@ (plainto_tsquery('bio')) = true 
```

vs

```sql
WHERE UPPER("experiments"."description"::text) LIKE UPPER('%bio%') 
```

## Types of changes


Miserlou:
DDT clocked this as going from 7s to 2s for my test query. I'm not sure what this will do for accuracy, but we had discussed needing something like ElasticSearch to solve this properly in the future so I think this is a good change for now. Waiting for tests to pass then approval.

PR #932, opened 2018-12-14
[WIP] Create and Use tmpfs With fasterq-dump

Miserlou:
WIP, still testing..


Miserlou:
This PR is blocked by the release of Nomad 0.9. It should work as is then, but it'd be work exploring if -e 64 can also be enabled to drastically increased performance.

Miserlou:
Tracked here: https://github.com/hashicorp/nomad/issues/5014

Issue #930, opened 2018-12-12
Back Up Files Before Running fasterq-dump 


Miserlou:
### Context
`fasterq-dump` is garbage, and will corrupt local source files if it encounters a network error, which happens all the time, likely because we are constantly pummeling them with our magnificent high altitude orbital spam cannon.

A quick fix would be to extract the `.sra` file from a copy of the original file, so it can't be corrupted, which as forced a re-downloaded. I suspect this is a source of many of our blockages. 

### Problem or idea

Copy the file to a working directory and extract it from there.

kurtwheeler:
This was closed by #935

PR #935, opened 2018-12-14
Backup .sra Files Before fasterq-dump

Miserlou:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/930

## Methods

This change is isolated from this branch: https://github.com/AlexsLemonade/refinebio/pull/932/files

but that changeset can't land until Nomad 0.8.7+ is released, which won't be for another ~2(+?) weeks.

### DIFF
--- a/workers/data_refinery_workers/processors/salmon.py
+++ b/workers/data_refinery_workers/processors/salmon.py
@@ -75,9 +75,6 @@ def _prepare_files(job_context: Dict) -> Dict:
 
     original_files = job_context["original_files"]
     job_context["input_file_path"] = original_files[0].absolute_file_path
-    if len(original_files) == 2:
-        job_context["input_file_path_2"] = original_files[1].absolute_file_path
-
     if not os.path.exists(job_context["input_file_path"]):
         logger.error("Was told to process a non-existent file - why did this happen?",
             input_file_path=job_context["input_file_path"],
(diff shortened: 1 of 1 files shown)


workers/data_refinery_workers/processors/salmon.py
```diff
@@ -60,11 +60,39 @@ def _prepare_files(job_context: Dict) -> Dict:
...
+        logger.error("Was told to process a non-existent file - why did this happen?",
+            input_file_path=job_context["input_file_path"],
+            processor_job=job_context["job_id"]
+        )
+        job_context["job"].failure_reason = "Missing input file: " + str(job_context["input_file_path"])
+        job_context["success"] = False
+        return job_context
+
+    # Copy the .sra file so fasterq-dump can't corrupt it.
+    if job_context["input_file_path"][-4:] == ".sra":
```
kurtwheeler: Can these ever be `.SRA`?

workers/data_refinery_workers/processors/salmon.py
```diff
@@ -60,11 +60,39 @@ def _prepare_files(job_context: Dict) -> Dict:
...
+        new_input_file_path = os.path.join(job_context["work_dir"], original_files[0].filename)
+        shutil.copyfile(job_context["input_file_path"], new_input_file_path)
+        job_context['input_file_path'] = new_input_file_path
```
kurtwheeler: You also need to copy `input_file_path_2` over for the same reasons right?

workers/data_refinery_workers/processors/salmon.py
Miserlou: In practice no but we can check

workers/data_refinery_workers/processors/salmon.py
Miserlou: fasterq-dump is only ever going to take one file at a time, but we might as well copy both

kurtwheeler: It's  not a might as well thing is it, fasterq-dump will be called on both `input_file_path` and `input_file_path_2` so we need to back them both up right?

workers/data_refinery_workers/processors/salmon.py
```diff
@@ -87,12 +84,28 @@ def _prepare_files(job_context: Dict) -> Dict:
         job_context["success"] = False
         return job_context
 
+    if len(original_files) == 2:
+        job_context["input_file_path_2"] = original_files[1].absolute_file_path
+    if not os.path.exists(job_context["input_file_path_2"]):
```
kurtwheeler: This if statement is assuming that the job_context has had the `input_file_path_2` key set, but it is valid for that to not be populated. This if should just start with a `job_context.get("input_file_path_2", False) and` like you have below.

kurtwheeler: Or it could just be nested within the `if len(original_files) == 2` conditional block.

PR #962, opened 2018-12-27
Improve downloads API

arielsvn:
## Issue Number

https://github.com/AlexsLemonade/refinebio-frontend/issues/472

Right now, to render all the information in the downloads page we have to make two requests for information about a dataset:

- `/dataset/${dataset_id}`
- `/dataset/${dataset_id}/?details=true`

This PR changes it so that all information can be returned when making a single request to the second one. This urls avoids returning details for each samples and instead returns the accession codes of the samples grouped by organisms. This is needed for this section:

image

Also added some optimizations to the queries with `prefetch_related`

## Types of changes

## Functional tests

Api

## Checklist

### DIFF
--- a/common/data_refinery_common/models/models.py
+++ b/common/data_refinery_common/models/models.py
@@ -914,16 +914,6 @@ class Dataset(models.Model):
         self.last_modified = current_time
         return super(Dataset, self).save(*args, **kwargs)
 
-    def get_samples(self):
-        """ Retuns all of the Sample objects in this Dataset """
-
-        all_samples = []
-        for sample_list in self.data.values():
-            all_samples = all_samples + sample_list
-        all_samples = list(set(all_samples))
-
-        return Sample.objects.filter(accession_code__in=all_samples)
-
     def get_experiments(self):
         """ Retuns all of the Experiments objects in this Dataset """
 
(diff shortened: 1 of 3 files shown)


common/data_refinery_common/models/models.py
```diff
@@ -926,12 +925,7 @@ def get_samples(self):
 
     def get_experiments(self):
         """ Retuns all of the Experiments objects in this Dataset """
-
-        all_experiments = []
-        for experiment in self.data.keys():
-            all_experiments.append(experiment)
-        all_experiments = list(set(all_experiments))
-
+        all_experiments = self.data.keys()
```
Miserlou: I'm not sure why this was changed?

arielsvn: I'm not sure what was the initial reason, seems that it was added in https://github.com/AlexsLemonade/refinebio/pull/299 although it's not used there. I guess it's a good helper method to have, since the dataset is stored in a `JSONField`. 

Here it's being used in a serializer when the `/?details` parameter was introduced.
https://github.com/AlexsLemonade/refinebio/pull/517/files#diff-415cd6002ae0888c9bee613a31905f87R508

arielsvn: @Miserlou is there anything you don't like about these helpers? Should we move them to a different place?

Miserlou: Since the data is in a JSONField, I dont' know how valid that JSON has to be - is it still "valid" if it has two keys of the same value? etc. I don't think it's hugely important in normal usage.

arielsvn: but the `JSONField` is deserialized with a `dict` right? it shouldn't have two keys with the same value.

Issue #909, opened 2018-12-07
Run tximport before full RNA-SEQ experiment is ready


kurtwheeler:
### Context

We're having trouble making all the RNA-Seq data we have (mostly) processed available because we don't run tximport until the entire experiment is processed. So if an experiment has 100 samples and 99 of them have been processed but one failed, we can't let anyone get any of the samples because tximport hasn't been run.

### Problem or idea

We could find some kind of heuristic for a cutoff point and run tximport after that's been reached. If additional samples are processed after that, we could just rerun tximport and replace the previously generated data with the new stuff.

### Solution or next step

Is this something we want to do? If so what would our heuristic be? @jaclyn-taroni said in the demo day meeting that she would start thinking about these questions.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

jaclyn-taroni:
I'm going to note some ideas about how we would figure this out.

Let's consider 50% of samples processed with Salmon as our starting point, which we can adjust once we've done an initial pass and have more information.

* Identify RNA-seq datasets of various scales/sample sizes that have been fully processed (e.g., through `tximport` step) in refine.bio. I expect that the 50% cut off is a bad idea for `n = 8` but may be perfectly fine in the 100s/1000s of experiments space.
* Obtain the `quant.sf` files for these experiments.
* Perform subsampling, going as low as 50% of the samples (to start), and process with `tximport`.
* Calculate the error between the `lengthScaledTPM` from the full dataset and the subsampled `lengthScaledTPM`

This should all be done with the `ccdl/dr_salmon` image.

cgreene:
As a first step, could we get a count of how many samples are in each experiment and how many have been processed (for the ones that are either success / fail - not still in processing)? This way we know where to draw those thresholds for testing.

Miserlou:
I originally suggested this idea as a fallback for [this change](https://github.com/AlexsLemonade/refinebio/commit/0e3298af14dc057493c65ba4308fbf65b6686f3b), which may render this particularly endeavor pointless anyway. We're expecting this to land today, so we should have the result by tomorrow. I wouldn't spend too much more time thinking about this issue until we see the result of that change.

kurtwheeler:
[...]

Issue #297, opened 2018-05-31
Associate tximport's computational result with sample or experiment?


dongbohu:
### Context

This question was raised when PR #278 was reviewed. 

### Problem or idea

Right now the computational result of `tximport` is associated with all samples in a certain experiment. But since `tximport` is actually done at experiment level, it may make more sense to create a new `ExperimentResult` model and associate this result with the experiment.

Unless we have such a new experiment-sample association, it will be awkward when we want to know whether a certain experiment has been processed by `tximport` step. The association between result and samples doesn't really help, because each sample may be associated with multiple experiments.
  
### Solution or next step

If `tximport` is the only experiment-level data analysis, then we can keep it as a special case and do nothing. If we have other experiment-level analysis in the future, a new `ExperimentResult` seems to be a reasonable solution.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

PR #970, opened 2019-01-04
Kurtwheeler/experiment result associations

kurtwheeler:
## Issue Number

#297 
#909, part 1 of 2

In order to be able to run tximport before an experiment is fully processed, we also need to be able to run it more than once for the same experiment so that as additional samples are processed we can rerun it and include them. In order to be able to run tximport more than once, we need to be able to destroy the results created by previous runs so there's only one per experiment. However at the moment this is extremely difficult to do because tximport runs at the experiment level but is associated with samples.

This makes that much easier because it adds an ExperimentResultAssociation model and a management command to back-populate all the relationships between tximport results and the experiments they were run on.

Once this PR has landed and the management command has been run, a second PR can go out that actually uses these relationships.

## Types of changes

## Functional tests

The unit test is a good test for this.

## Checklist


kurtwheeler:
This depennds on #964

foreman/data_refinery_foreman/foreman/management/commands/assoc_experiment_results.py
```diff
@@ -0,0 +1,70 @@
+"""This command will go through ComputationalResult objects that are a
```
Miserlou: Will we ever need to actually run this again, or can it just be done automatically as part of the migration?

kurtwheeler: We won't need to run this again. However we can't run it at the same time as the migration because it needs the migration to have run so it can use the new data model. I could put it into a new migration, but that seemed like it'd be slightly more complicated to implement, it's not a migration we'd be able to rollback anyway, and I'm not actually sure how long this will take so I thought it might be nice to be able to do it outside of the deploy process (which is when the migration would run).

PR #976, opened 2019-01-10
Run tximport early for experiments that are large enough and have enough samples processed

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/909

The results from https://github.com/AlexsLemonade/tximport_partial_run_tests were that we should be okay to run tximport on experiments with >=20 samples if at least 80% of the samples have been processed. This will run tximport on each sample past that cutoff, which does in fact have the potential to be a lot of extra compute, however while implementing it I thought that was a price we were willing to pay to get at our RNA-Seq samples quicker. @Miserlou has raised some concerns that may mean the approach for this PR has to be reconsidered and then it be heavily modified:
  * This approach necessitates the need to delete old computed files/results, which means that the exact same data that was once downloaded from us may end up not being able to be retrieved again. (Otherwise you would end up with a lot of duplicated data files)
  * This can be a lot of extra compute power.

The alternative approach he suggested is to break tximport out into it's own job and run it if an experiment is completed or if an experiment has been past the cutoff without additional progress for some amount of time. This could be dispatched by the Foreman.

This may need a bit more discussion, but I am not going to be able to re-implement this before I go on vacation one way or the other. @cgreene and @jaclyn-taroni do either of you have strong opinions one way or the other?

## Types of changes

## Functional tests

I ran tximport on `SRP155220` in a dev stack in order to be able to generate all the quant files used and the entire experiment was able to be processed successfully and everything looked good, which means that it ran tximport multiple times but only ended up with one result.

The unit tests I added are also pretty comprehensive.

## Checklist


cgreene:
I am a little bit concerned about the "might not be able to download again" bit, but that is going to be par for the course with something like this. On the other hand, running tximport for every sample past that threshold sounds like it could get expensive quickly, as the cost will be highest in the largest experiments (where the condition also gets hit the most times).

One way to go would be to make this something that we could manually trigger if desired for specific experiments. If we really want a dataset, we can force it. But it doesn't get automatically forced.

kurtwheeler:
I broke this PR by messing up the git history, so I've recreated it here: https://github.com/AlexsLemonade/refinebio/pull/977

PR #977, opened 2019-01-10
Run tximport early for experiments that are large enough and have enough samples processed

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/909

This is a remake of https://github.com/AlexsLemonade/refinebio/pull/976 because I messed up my git history by adding a file that was too large, and then ran a thing from https://stackoverflow.com/questions/19573031/cant-push-to-github-because-of-large-file-which-i-already-deleted without reading the fact that this changes every commit hash in the history... So sorry about that, re-pinging @Miserlou, @cgreene, @jaclyn-taroni, even though it does look like this PR will need to be re-implemented to not run by default for all experiments.

The results from https://github.com/AlexsLemonade/tximport_partial_run_tests were that we should be okay to run tximport on experiments with >=20 samples if at least 80% of the samples have been processed. This will run tximport on each sample past that cutoff, which does in fact have the potential to be a lot of extra compute, however while implementing it I thought that was a price we were willing to pay to get at our RNA-Seq samples quicker. @Miserlou has raised some concerns that may mean the approach for this PR has to be reconsidered and then it be heavily modified:
  * This approach necessitates the need to delete old computed files/results, which means that the exact same data that was once downloaded from us may end up not being able to be retrieved again. (Otherwise you would end up with a lot of duplicated data files)
  * This can be a lot of extra compute power.

The alternative approach he suggested is to break tximport out into it's own job and run it if an experiment is completed or if an experiment has been past the cutoff without additional progress for some amount of time. This could be dispatched by the Foreman.

This may need a bit more discussion, but I am not going to be able to re-implement this before I go on vacation one way or the other. @cgreene and @jaclyn-taroni do either of you have strong opinions one way or the other?

## Types of changes

## Functional tests

I ran tximport on `SRP155220` in a dev stack in order to be able to generate all the quant files used and the entire experiment was able to be processed successfully and everything looked good, which means that it ran tximport multiple times but only ended up with one result.

The unit tests I added are also pretty comprehensive.

## Checklist

### DIFF
--- a/workers/data_refinery_workers/processors/salmon.py
+++ b/workers/data_refinery_workers/processors/salmon.py
@@ -24,6 +24,7 @@ from data_refinery_common.models import (
     ComputationalResultAnnotation,
     ComputedFile,
     Experiment,
+    ExperimentResultAssociation,
     ExperimentSampleAssociation,
     OrganismIndex,
     Pipeline,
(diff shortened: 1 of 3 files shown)


jaclyn-taroni:
> The alternative approach he suggested is to break tximport out into it's own job and run it if an experiment is completed or **if an experiment has been past the cutoff without additional progress for some amount of time.** This could be dispatched by the Foreman.

I like this idea or what @cgreene suggested on #976 (manually forcing this tximport to run "early").

cgreene:
This PR was opened but was replaced by a subsequent solution.

PR #986, opened 2019-01-22
Kurtwheeler/try new cran mirror

kurtwheeler:
## Issue Number

N/A related to github alert

I switched to the cran mirror: 

This is run by [Revolution Analytics](https://en.wikipedia.org/wiki/Revolution_Analytics), which is owned my Microsoft so they may be more invested in having a reliable service which goes down less.

The illumina deps also apparently had a bad rlang version so I fixed that too.

## Types of changes

## Functional tests

Unit tests will cover this.

## Checklist

### DIFF
--- a/workers/dockerfiles/Dockerfile.illumina
+++ b/workers/dockerfiles/Dockerfile.illumina
@@ -54,7 +54,7 @@ COPY .boto .boto
 
 COPY workers/data_refinery_workers/processors/requirements.txt .
 
-RUN  pip3 --no-cache-dir install -r requirements.txt
+RUN  pip3 install -r requirements.txt
 
 COPY common/dist/data-refinery-common-* common/
 
--- a/workers/dockerfiles/Dockerfile.illumina
+++ b/workers/dockerfiles/Dockerfile.illumina
@@ -9,7 +9,7 @@ RUN apt-get -y install apt-fast
 COPY workers/CRAN.gpg .
 RUN \
   apt-fast update -qq && \
-  apt-get install apt-transport-https  -y&& \
+  apt-get install -y apt-transport-https && \
   apt-fast install -y lsb-release && \
   echo "deb http://archive.linux.duke.edu/cran/bin/linux/ubuntu $(lsb_release -sc)/" \
       >> /etc/apt/sources.list.d/added_repos.list && \
--- a/workers/dockerfiles/Dockerfile.illumina
+++ b/workers/dockerfiles/Dockerfile.illumina
@@ -7,10 +7,14 @@ RUN apt-get update -qq
 RUN apt-get -y install apt-fast
 
 COPY workers/CRAN.gpg .
+RUN apt-key adv --keyserver keyserver.ubuntu.com --recv-keys E084DAB9
 RUN \
   apt-fast update -qq && \
+  apt-get install apt-transport-https  -y&& \
   apt-fast install -y lsb-release && \
   echo "deb http://archive.linux.duke.edu/cran/bin/linux/ubuntu $(lsb_release -sc)/" \
+      >> /etc/apt/sources.list.d/added_repos.list && \
+  echo "deb https://cloud.r-project.org/bin/linux/ubuntu $(lsb_release -sc)/" \
       >> /etc/apt/sources.list.d/added_repos.list && \
   apt-key add CRAN.gpg && \
   apt-fast update -qq && \
--- a/workers/dockerfiles/Dockerfile.illumina
+++ b/workers/dockerfiles/Dockerfile.illumina
@@ -7,7 +7,6 @@ RUN apt-get update -qq
 RUN apt-get -y install apt-fast
 
 COPY workers/CRAN.gpg .
-RUN apt-key adv --keyserver keyserver.ubuntu.com --recv-keys E084DAB9
 RUN \
   apt-fast update -qq && \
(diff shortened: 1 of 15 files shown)


Miserlou:
There are still R errors even after fixing the Python thing

```
* installing *source* package 'rlang' ...
** package 'rlang' successfully unpacked and MD5 sums checked
** libs
gcc -std=gnu99 -I/usr/share/R/include -DNDEBUG -I./lib/     -fpic  -g -O2 -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -g  -c capture.c -o capture.o
gcc -std=gnu99 -I/usr/share/R/include -DNDEBUG -I./lib/     -fpic  -g -O2 -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -g  -c export.c -o export.o
gcc -std=gnu99 -I/usr/share/R/include -DNDEBUG -I./lib/     -fpic  -g -O2 -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -g  -c internal.c -o internal.o
gcc -std=gnu99 -I/usr/share/R/include -DNDEBUG -I./lib/     -fpic  -g -O2 -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -g  -c lib.c -o lib.o
gcc -std=gnu99 -shared -L/usr/lib/R/lib -Wl,-Bsymbolic-functions -Wl,-z,relro -o rlang.so capture.o export.o internal.o lib.o -L/usr/lib/R/lib -lR
installing to /usr/local/lib/R/site-library/rlang/libs
** R
** byte-compile and prepare package for lazy loading
** help
*** installing help indices
*** copying figures
** building package indices
** testing if installed package can be loaded
* DONE (rlang)

The downloaded source packages are in
	'/tmp/RtmpjF3gP2/downloaded_packages'
Installing package into '/usr/local/lib/R/site-library'
(as 'lib' is unspecified)
inferring 'repos = NULL' from 'pkgs'
trying URL 'https://bioconductor.org/packages/3.6/bioc/src/contrib/BiocInstaller_1.28.0.tar.gz'
Content type 'application/x-gzip' length 18624 bytes (18 KB)
==================================================
downloaded 18 KB

ERROR: this R is version 3.2.3, package 'BiocInstaller' requires R >=  3.4.0
Error in install.packages("https://bioconductor.org/packages/3.6/bioc/src/contrib/BiocInstaller_1.28.0.tar.gz") : 
  (converted from warning) installation of package '/tmp/RtmpjF3gP2/downloaded_packages/BiocInstaller_1.28.0.tar.gz' had non-zero exit status
Execution halted
The command '/bin/sh -c Rscript illumina_dependencies.R' returned a non-zero code: 1
Could not build ccdlstaging/dr_illumina after three attempts.
Exited with code 1
```

workers/dockerfiles/Dockerfile.illumina
```diff
@@ -7,11 +7,15 @@ RUN apt-get update -qq
 RUN apt-get -y install apt-fast
 
 COPY workers/CRAN.gpg .
+RUN apt-key adv --keyserver keyserver.ubuntu.com --recv-keys E084DAB9
```
kurtwheeler: Wait why is this being done this way. Aren't we supposed to do this key file like `apt-key add CRAN.gpg`?

kurtwheeler:
@cgreene we aren't actually bumping the version of R, but the version of `rlang` which is a package... although it does look like the version of R somehow changed?

I really really don't understand how these issues keep popping up. I thought we had every R dep pinned to a specific version but something keeps changing and causing these version incompatibilities.

cgreene:
Actually it looks like you're specifically using the bioconductor 3.6 repos:
> https://bioconductor.org/packages/3.6/bioc/src/contrib/

I'm a bit surprised that this would have worked with R 3.2. The [release notes](https://bioconductor.org/news/bioc_3_6_release/) say:
> We are pleased to announce Bioconductor 3.6, consisting of 1473 software packages, 326 experiment data packages, and 911 annotation packages.
> 
> There are 100 new software packages, and many updates and improvements to existing packages; Bioconductor 3.6 is compatible with R 3.4.2, and is supported on Linux, 32- and 64-bit Windows, and Mac OS X. This release will include an updated Bioconductor Amazon Machine Image and Docker containers.

jaclyn-taroni:
Agree with @kurtwheeler that the R version got rolled back, because we did pin the Bioconductor version a bit ago and if I look at the API the R versions are `3.4.4` for all the samples I took a quick look at.

```
                            "os_pkg": {
                                "r-base": "3.4.4-1xenial0",
                                "python3": "3.5.1-3",
                                "python3-pip": "8.1.1-2ubuntu0.4"
                            }
```

It's not obvious to me where we are pinning the R version or _if_ we are pinning the R version. (My guess is this might have something to do with the Ubuntu base image?) If we're not pinning the R version, I bet doing so would resolve the Bioconductor issues at the very least.

kurtwheeler:
Woah...

@jaclyn-taroni I think you're right, I can't see us pinning a specific version of R anywhere... That very well could explain why we've had version incompatibility issues  changing despite thinking we had everything pinned. I'll make a PR to do that today.

PR #987, opened 2019-01-24
Kurtwheeler/pin r version

kurtwheeler:
## Issue Number

N/A Came up cause R deps broke our builds again

In the course of #986 we realized that we had all our R packages pinned except the R language itself (probably due to confusion with the rlang package). This pins R to the version that Bioconductor 3.6 specifies as its target.

## Types of changes

## Functional tests

Unit tests work, it means R is installed

## Checklist

### DIFF
--- a/workers/dockerfiles/Dockerfile.smasher
+++ b/workers/dockerfiles/Dockerfile.smasher
@@ -6,7 +6,7 @@ RUN add-apt-repository ppa:apt-fast/stable
 RUN apt-get update -qq
 RUN apt-get -y install apt-fast
 
-# The packages related to R are somewhat weird, see the README for more details.w
+# The packages related to R are somewhat weird, see the README for more details.
 
 COPY workers/CRAN.gpg .
 RUN \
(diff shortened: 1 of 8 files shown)


kurtwheeler:
A common base image is actually a pretty good idea. It was super obnoxious to have to edit 7 files as I figured out small tweaks to them all.

workers/dockerfiles/Dockerfile.smasher
```diff
@@ -6,17 +6,7 @@ RUN add-apt-repository ppa:apt-fast/stable
...
-# https://cran.revolutionanalytics.com/bin/linux/ubuntu/xenial/
-# If it needs to be updated then a version should be selected from
-# that list.
-
-# I am not entirely sure what the difference between r-base and r-base
-# core is, except that r-base is supposed to have some extra utilities
-# which increases the size of the package. However we don't get a ton
-# of choice because trying to use r-base causes a "you have held
-# broken pacakges" error.
+# The packages related to R are somewhat weird, see the README for more details.w
```
Miserlou: .w

Issue #990, opened 2019-01-24
Add CORS headers to failed api responses 


arielsvn:
### Context

In order to consume the API we need some headers in the response, such as `Access-Control-Allow-Origin: *`

### Problem or idea

When the API fails with a 500 error code these headers are missing. Eg:

![image](https://user-images.githubusercontent.com/1882507/51702631-4b6cc400-1fe2-11e9-95d8-cebdd439d647.png)
![image](https://user-images.githubusercontent.com/1882507/51702784-a9011080-1fe2-11e9-85d0-e6f8a8423e8e.png)

A successful request with the headers:

![image](https://user-images.githubusercontent.com/1882507/51702652-57f11c80-1fe2-11e9-915a-d9b616a4c902.png)

### Solution or next step

I'm not sure how we would add these headers to the failed responses, tagging @Miserlou 

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

Miserlou:
Is this the result of the error code or the HTTP method?

arielsvn:
closed with https://github.com/AlexsLemonade/refinebio/issues/990

PR #991, opened 2019-01-24
Raise CORSMiddleware Higher In Order

Miserlou:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/990

## Methods

Very small change, please confirm this is fixes your problem


Miserlou:
Never mind that, we can just set `CORS_ORIGIN_ALLOW_ALL` and be done with it since this is an API anyway.

PR #995, opened 2019-01-28
Add blurb for Nomad and Terraform about getting them on the PATH.

kurtwheeler:
## Issue Number

#699 

Getting Nomad and Terraform on her PATH were the only things that gave Candace any trouble... so far. She was actually familiar with using her PATH, she just ran into trouble because `sudo` didn't work for her. I'm not experienced with Macs, so I'm not entirely sure how to fix that, and we already link to https://www.nomadproject.io/guides/operations/install/index.html#precompiled-binaries in the readme,  but I do know that moving the binary into `/usr/local/bin` worked for her and is probably the best thing to recommend generally so I've mentioned it in the README.

## Types of changes

## Functional tests

N/A

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,9 +121,11 @@ The following services will need to be installed:
 - [jq](https://stedolan.github.io/jq/)
 - [Nomad](https://www.nomadproject.io/docs/install/index.html#precompiled-binaries) This may not work for Mac. [Citation Needed]
 
-Instructions for installing Docker, Nomad, and Homebrew can be found by
-following the link for those services. The others on that list can
-be installed by running: `brew install iproute2mac git-crypt terraform jq`.
+Instructions for installing Docker and Homebrew can be found by
+following the link for those services.
+Terraform and Nomad are both installed by downloading the macOS binary and placing it somewhere on your PATH.
+If you are unfamiliar with managing your PATH, a fairly reliable way to do this is to move the binary file to `/usr/local/bin`.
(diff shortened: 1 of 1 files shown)


Miserlou:
I don't think this is the right way to install Nomad on OSX, mine is handled by `homebrew` which makes a symlink from `/usr/local/bin/nomad` to the one in Cellar.

cgreene:
It does look like we have `brew` later, so perhaps the instructions should note that users should only use the brew installation method.

kurtwheeler:
@Miserlou did you look at this link: https://www.nomadproject.io/guides/operations/install/index.html#precompiled-binaries?

It only recommends two ways to do it and that's not one of them. Is that better? I don't know what the difference is and I don't know what Cellar is.

kurtwheeler:
Ohhh Cellar, like a beer cellar? Gotta be a homebrew concept :P

kurtwheeler:
It might make sense to have a Rich write the documentation for Macs since he actually uses one... I know brew is the popular way to install things in Macland, but is it even better than the officially supported methods?

Miserlou:
It's literally just `brew install nomad`

kurtwheeler:
You addressed none of my other comments, such as the fact that the source for these files does not recommend that method. I can put that, but I have literally no context for why that is or isn't a good idea other than you saying that's how you've done it. I am generally uncomfortable writing documentation that I cannot explain the rationale for.

Miserlou:
Then I don't know why this PR was opened at all, since that's exactly what you're doing.

Issue #928, opened 2018-12-12
[Notes] Make the API Faster


Miserlou:
Notes from today's investigation:

* We have so much data
* Full text search is bad
* Sample counts must be moved to an IntegerField on the Experiment object
* Platforms must be moved to a (StringField? ArrayField?) on the Experiment object
* experimentannotation_data search needs to be removed until a better solution can be found
* The four filter queries should be combined into a single one

arielsvn:
ref https://github.com/AlexsLemonade/refinebio/issues/832

cgreene:
We have had pretty good experience speeding up search w/ Elastic Search in the past. If that's the only part that's struggling, it's not horrible to do.

arielsvn:
> Sample counts must be moved to an IntegerField on the Experiment object

Adding the sample counts might not do much, tested locally and it doesn't seem to be a significant difference between ordering by `total_samples_count` or `created_at`.

arielsvn:
Another thing that might be good to move into the `Experiment` object is `samples__is_processed` and `samples__is_public`.

https://github.com/AlexsLemonade/refinebio/blob/86c22743a0db4e59fbe13eeff685621302ce02c7/common/data_refinery_common/models/models.py#L233-L237

Also removing that `.distinct()` call seems to reduce the response time slightly, not sure why do we need it there.

Miserlou:
It's friggin fast now

PR #996, opened 2019-01-28
Open ES Permissions, Fix [CORS, Platforms, README]

Miserlou:
Fixes ES deployment permissions and a few other small changes - double CORS headers, platforms not appearing and an outdated README.

#699
#928


Miserlou:
Oh, it also changes the default target capacity to 0 for dev and staging as it's very annoying to do devops and have this constantly reset itself to the expensive setting without warning.

common/data_refinery_common/migrations/0009_es_cache.py
```diff
@@ -31,10 +31,10 @@ def update_cached_values(apps, schema_editor):
 
         # via experiment.update_organism_names
         experiment.organism_names = list(set([organism.name for organism in experiment.organisms.all()]))
-        experiment.save()
 
         # experiment.update_platform_names
         experiment.platform_names = list(set([sample.platform_name for sample in experiment.samples.all()]))
+        experiment.save()
```
arielsvn: nice catch!

PR #998, opened 2019-01-30
Use Ngrams for More Robust ES Search

Miserlou:
## Issue Number

n/a

It was bothering me that we could only search for "medulloblastoma" but not for "medullo"

This adds ngram based indexing to the ES search results such that substrings between 4 and 12 characters will also hit.

## Types of changes

## Functional tests

Tested manually.


Miserlou:
They're adjustable parameters I chose after some mild trial and error. Less than 4 seemed to over match, more than 12, I don't know, it just doesn't seem like I'd type 12 letters and not just finish the word.

Issue #1000, opened 2019-01-30
Accession Search is Broken


Miserlou:
https://www.refine.bio/results?q=GSE85218

PR #1002, opened 2019-01-30
Search more fields via ES

Miserlou:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/1000

Some of the fields aren't being searched. This searches those too.

### DIFF
--- a/api/data_refinery_api/views.py
+++ b/api/data_refinery_api/views.py
@@ -165,9 +165,16 @@ class ExperimentDocumentView(DocumentViewSet):
     # Is this exhaustive enough?
     search_fields = (
         'title',
+        'publication_title',
         'description',
         'publication_authors',
         'submitter_institution',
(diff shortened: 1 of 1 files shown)


api/data_refinery_api/views.py
```diff
@@ -165,9 +165,16 @@ class ExperimentDocumentView(DocumentViewSet):
...
         'title',
+        'publication_title',
         'description',
         'publication_authors',
         'submitter_institution',
+        'accession_code',
+        'alternate_accession_code',
+        'publication_doi',
+        'pubmed_id',
+        'sample_metadata_fields',
```
kurtwheeler: Are you sure we want to be able to search on `sample_metadata_fields`? I thought this was mostly a cached field so we didn't have to calculate it on the fly for every experiment. I think it might end up matching a lot of extra results.

PR #1003, opened 2019-01-30
Allow supplying ALL as an organism to create_qn_target

Miserlou:
We have support for creating all QN targets at once, although this isn't usable in production since the nomad job spec doesn't support it, and the templating we use couldn't handle it anyway. This is a hack to allow using `ALL` as an organism - we already use `ALL` elsewhere in the app, so at least that's consistent.


Miserlou:
Also removes the SMASHER_CONSTRAINT from the COMPENDIA job specification.

Miserlou:
Well that sounds great for the future but we have a practical constraint of needing to get this done ASAP

Issue #1004, opened 2019-01-30
Send platforms accession codes in the search endpoint


arielsvn:
### Context

We need to show platform names correctly formatted in the frontend.

### Problem or idea

The search endpoint, is sending platform names that we display in the filters:

![image](https://user-images.githubusercontent.com/1882507/52011002-e2cf8c80-24a4-11e9-905a-f4d9f5568612.png)

These are not formatted correctly.

Also in order to filter the search results by platform we use the platform name in the url:  `&platform=[HG-U133A_2] Affymetrix Human Genome U133A 2.0 Array`. This makes the url look ugly when those values are encoded.

### Solution or next step

It would be ideal if we could make the filters work only with the platform accession code: `&platform=HG-U133A_2`. 

The API could use use platform accession codes everywhere. Now it's sending both the name and the formatted name for all objects. This info could be accessed with the `/platforms` endpoint.

For example, right now, the experiments endpoint sends the platform name and the formatted platform for each one of it's samples:
![image](https://user-images.githubusercontent.com/1882507/52011517-35f60f00-24a6-11e9-9e19-75671d10754f.png)

This is implemented in https://github.com/AlexsLemonade/refinebio/pull/769

Also, we should consider merging https://github.com/AlexsLemonade/refinebio/pull/783 which fixes the platform names of `Illumina*`

Tagging @Miserlou for additional comments.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

arielsvn:
We were able to make the platforms filter work with the platform names and also moved the formatting of platform names to the frontend.

I'm closing this and also #769, although we should consider removing `pretty_platforms` from everywhere in the API.

PR #1005, opened 2019-01-30
Adds Cached platform_accession_codes in ES 

Miserlou:
## Issue Number
https://github.com/AlexsLemonade/refinebio/issues/1004

Also makes `platform_accession_codes` the thing that `platform` filters on.


common/data_refinery_common/migrations/0012_experiment_platform_accession_codes.py
```diff
@@ -0,0 +1,29 @@
...
+from django.db import migrations, models
+
+def update_cached_values(apps, schema_editor):
+    """ """
+
+    Experiment = apps.get_model('data_refinery_common', 'Experiment')
+    for experiment in Experiment.objects.all():
+        # Model methods can't be used during migrations. :(
+        # https://stackoverflow.com/a/37685925/1135467
+        experiment.platform_accession_codes = list(set([sample.platform_accession_code for sample in experiment.samples.all()]))
```
kurtwheeler: I think this should be:
```
experiment.platform_accession_codes = get_platform_accession_codes()
```

Miserlou: You can't use model methods during migrations. There's already a comment and link to an explanation directly above this line.

kurtwheeler: Hah good call, I didn't notice that.

Miserlou:
@arielsvn 

example

```
        {
            "id": 1,
            "title": "Microarray analysis of in vitro differentiation of adult human pancreatic progenitor cells",
            "publication_title": "",
            "description": "We have developed an in vitro culture system that allows us to expand progenitor cells from human islet preparations and differentiate them into insulin-producing cells. We noticed however, that cultures from individual islet preparations had very heterogeneous outcomes, from good differentiation to almost none. We therefore speculated that our progenitor cell cultures contained different kinds of cells and that the true endocrine progenitor cells are present in the successful cultures, but not in the unsuccessful ones. To address this issue and to begin to identify markers for the true endocrine progenitor cells we compared global gene expression between a very successful culture (final insulin-expression 10% of islets) and an unsuccessful one (final insulin-expression 0%). We also included RNA from freshly isolated islets for control purposes. The cultures from donor A yielded substantial differentiation, while the cultures from donor B showed no successful differentiation.\n",
            "technology": "MICROARRAY",
            "accession_code": "E-MTAB-3050",
            "alternate_accession_code": null,
            "submitter_institution": "Massachusetts General Hospital",
            "has_publication": false,
            "publication_doi": "",
            "publication_authors": [],
            "sample_metadata_fields": [
                "sex",
                "age",
                "specimen_part",
                "subject"
            ],
            "platform_names": [
                "[HG_U95Av2] Affymetrix Human Genome U95 Version 2 Array"
            ],
            "platform_accession_codes": [
                "hgu95av2"
            ],
            "organism_names": [
                "HOMO_SAPIENS"
            ],
            "pubmed_id": "",
            "num_total_samples": 5,
            "num_processed_samples": 0
        }
```

common/data_refinery_common/migrations/0012_experiment_platform_accession_codes.py
Miserlou: and even then it would be `experiment.get_platform_accession_codes()`

Miserlou: :-P

Issue #978, opened 2019-01-10
We need a mechanism for creating downloader jobs for missing files.


kurtwheeler:
### Context

https://github.com/AlexsLemonade/refinebio/issues/930 means that at least some files that were downloaded are now missing and need to be re-downloaded. Additionally and probably more importantly, we killed our EBS volumes so all of the data we've downloaded so far is now missing. We don't currently have any mechanism to re-download data that has had a successful downloader job.

### Problem or idea

We need to have processor jobs re-create the downloader jobs that created them if the processor job detects that one or more files it needs is missing.

### Solution or next step

I think that basically the way to do this is to:
  - Find all the places where processor jobs check to see if their original_files exist. I think this is probably each processor's `prepare_files` function, but it could potentially be broken out to be part of `start_job` so that it's just one place. Here's an example of what I mean: https://github.com/AlexsLemonade/refinebio/blob/dev/workers/data_refinery_workers/processors/salmon.py#L78
  - Anywhere that check is, instead of logging an error and failing the job, call a util function which will recreate the downloader job that originally downloaded the original files associated with the processor job.
  - This recreated downloader job should have the same associations to original_files as the processor job.

Unfortunately I don't think this will work for Array Express and GEO because of extracted files. This is tricky because processor jobs are only associated with files they're supposed to process, not the compressed files that got extracted to produce those files. Therefore if you only know the original file the processor job was for, it's not immediately obvious how to find the compressed original_file that should be associated with the downloader job. HOWEVER, both the compressed original_file and the extracted original_file have the same `source_url` field. Therefore, for an array express or GEO job I would expect the flow to be something like:
  - Get the original file for the processor job.
  - Query original_files to find the original file where `original_file.filename == source_url.split("/")[-1]` is true
  - Create a new downloader job associated with that original file and let the foreman queue it.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

Miserlou:
Sounds good, I think the processor job should self-delete after going through this to avoid marking complete or marking as a retry.

PR #1006, opened 2019-01-30
Recreate needed Downloader jobs

kurtwheeler:
## Issue Number

#978 

Detects when ProcessorJobs are missing files that should be downloaded, recreates the DownloaderJob for those files, and deletes the ProcessorJob.

The really tricky thing about this PR is that ArrayExpress and GEO DownloaderJobs often download an archive file which is extracted and then different processor jobs are created for each of them. This makes recreating the DownloaderJob harder, but I tried to explain it well via comments.

The tests I've added are what the WIP on this PR is about. Ensembl and GEO are being very unreliable (well Ensembl is unreliable, GEO may or may not just be down) so I can't run them successfully enough. I think they are very close to done though, I just can't quite finish them because they can't download data. It's possible I can DRY up those tests, but it's hard to try to do so when I can't even run them all successfully.

## Types of changes

## Functional tests

I think the end to end tests are pretty dang comprehensive so I think maybe just a thorough test in staging might be sufficient.

## Checklist


kurtwheeler:
@Miserlou this should no longer be a WIP since the only test that is failing is caused by Ensembl's FTP server.

.circleci/config.yml
```diff
@@ -50,7 +50,7 @@ jobs:
 
       # Run Foreman Tests
       - run: mkdir -p test_volume && chmod -R a+rw test_volume
-      - run: ./foreman/run_tests.sh
+      - run: ./foreman/run_tests.sh --exclude-tag=salmon --exclude-tag=transcriptome
```
Miserlou: Document why exclude

common/data_refinery_common/models/models.py
```diff
@@ -698,6 +698,48 @@ def delete_local_file(self):
...
+            successful_processor_jobs = self.processor_jobs.filter(
+                success=True,
+                pipeline_applied=pipeline_applied
+            )
+        else:
+            successful_processor_jobs = self.processor_jobs.filter(success=True)
+
+        # Finally, if there is a successful processor job, then the file
+        # has been processed and doesn't need to be processed again.
+        # XXX: if we ever want to reprocess everything, this will be problematic.
```
Miserlou: Can we come up with a better solution than this, I'd rather not be introducing `XXX`es at this stage of development if they don't have a plan.

foreman/data_refinery_foreman/surveyor/test_end_to_end.py
```diff
@@ -126,3 +129,491 @@ def test_no_op(self):
...
+                try:
+                    processor_job = wait_for_job(processor_job, ProcessorJob, start_time)
+                    if processor_job.success:
+                        successful_processor_jobs.append(processor_job)
+                except:
+                    pass
+
+            self.assertEqual(len(successful_processor_jobs), 12)
+
+    # Commented out because GEO isn't giving files.
```
Miserlou: No longer true?

foreman/data_refinery_foreman/surveyor/test_end_to_end.py
```diff
@@ -126,3 +129,491 @@ def test_no_op(self):
...
+                try:
+                    processor_job = wait_for_job(processor_job, ProcessorJob, start_time)
+                    if processor_job.success:
+                        successful_processor_jobs.append(processor_job)
+                except:
+                    pass
+
+            self.assertEqual(len(successful_processor_jobs), target_job_count)
+
+    # Commented out because Ensembl's FTP isn't working reliably.
```
Miserlou: Not true anymore?

Miserlou:
I was expecting to see a migration file, how are we keeping track of state around this? I'd like to be able to know where and when this behavior occurs, and also to make sure than no undiscovered infinite-looping could occur. Maybe I just didn't see it.

Miserlou:
Also good work this looks like it was a nightmare

common/data_refinery_common/models/models.py
kurtwheeler: I'm just gonna take this out then. I think that if/when we want to reprocess everything then we're gonna have to do some planning and come up with a way to keep the first and subsequent runs separate. Maybe just like a field denoting which version jobs were run on.

Miserlou:
Is this ready for rereview

kurtwheeler:
It is now!

.circleci/config.yml
```diff
@@ -105,9 +92,32 @@ jobs:
...
+      - run: sudo chown -R circleci:circleci workers/test_volume/
+      - run:
+          command: .circleci/filter_tests.sh -t compendia
+          no_output_timeout: 1h
+
+
   # This tests workers tests tagged as 'salmon'
   salmon_and_api_tests:
     working_directory: ~/refinebio
+    # resource_class: xlarge
```
Miserlou: What is this?

.circleci/config.yml
```diff
@@ -140,6 +172,7 @@ jobs:
 
   tx_illumina_tests:
     working_directory: ~/refinebio
+    # resource_class: xlarge
```
Miserlou: ?

common/data_refinery_common/models/jobs.py
```diff
@@ -196,6 +196,11 @@ class Meta:
     # If the job is retried, this is the id of the new job
     retried_job = models.ForeignKey('self', on_delete=models.PROTECT, null=True)
 
+    # If the job was recreated because the data it downloaded got
+    # lost, deleted, or corrupted then this field will be true.
+    # This helps prevent an infinite loop of DownloaderJob recreation.
+    was_recreated = = models.BooleanField(default=False)
```
Miserlou: nope

.circleci/config.yml
kurtwheeler: Trying to get boxes big enough to actually run these tests. Circle told me this should work, but I got an error saying xlarge was an invalid setting. I'm in contact with them, but you're right: I should sort this out first.

run_es.sh
```diff
@@ -1,2 +1,5 @@
 #! /bin/bash
+
+docker rm -f dres 2> /dev/null
```
Miserlou: I don't love this at the start of a run script

workers/data_refinery_workers/downloaders/array_express.py
```diff
@@ -123,14 +123,25 @@ def download_array_express(job_id: int) -> None:
...
+            # actually needed to be downloaded before marking them as
+            # downloaded and queuing processor jobs.
+            if original_file.needs_downloading():
+                original_file.is_downloaded = True
+                original_file.is_archive = False
+                original_file.absolute_file_path = og_file['absolute_path']
+                original_file.filename = og_file['absolute_path'].split('/')[-1]
+                original_file.calculate_size()
+                original_file.save()
+                original_file.calculate_sha1()
```
Miserlou: Saving before sha1 ? I think this will error, is this tested?

workers/data_refinery_workers/processors/utils.py
```diff
@@ -48,6 +50,99 @@ def signal_handler(sig, frame):
...
         sys.exit(0)
 
+
+def create_downloader_job(undownloaded_files: OriginalFile) -> bool:
+    """Creates a downloader job to download `undownloaded_files`."""
+    original_downloader_job = None
+    archive_file = None
+    for undownloaded_file in undownloaded_files:
+        try:
+            original_downloader_job = DownloaderJobOriginalFileAssociation.objects.filter(
```
Miserlou: Getting association objects directly is weird

common/data_refinery_common/models/jobs.py
```diff
@@ -127,7 +127,7 @@ class Meta:
     failure_reason = models.TextField(null=True)
 
     # If the job is retried, this is the id of the new job
-    retried_job = models.ForeignKey('self', on_delete=models.PROTECT, null=True)
+    retried_job = models.ForeignKey('self', on_delete=models.SET_NULL, null=True)
```
Miserlou: If there are new fields on the models, I'd also expect new fields on the serializers

common/data_refinery_common/models/jobs.py
Miserlou: ( double =)

workers/data_refinery_workers/downloaders/array_express.py
kurtwheeler: Good catch. It didn't error out, but it definitely was getting covered by the end-to-end test for the array express downloader (I know cause I stuck a logging statement in there to make sure).

kurtwheeler:
Ok, @Miserlou I think I've addressed all your concerns. I'm not sure why my PR feedback commit made tests fail, but I think they should be passing.

PR #1012, opened 2019-02-01
[Master] Merge #1011

Miserlou:
If we're doing an FAPD today, I'd love to get this feature in since it makes the front end a lot more powerful.


Miserlou:
My man with the fast approval

PR #1013, opened 2019-02-01
[WIP] Way The Heck Faster Outer Joining

Miserlou:
`pd.merge` apparently gets quadratically slower as the number of columns increases (particularly in Py36/7 because of an implementation detail?), which we didn't really notice until creating QN targets from hundreds of thousands of samples. Since the QN inputs should all be from the same platform and therefore have the same number of gene identifiers, we can use an outer join rather than an inner join as they are functionally equivalent, but allows us to use a much faster technique of `pd.concat`.

WIP as I'm in the middle of a new test that should give us some metrics.


Miserlou:
This approach eats a lot of memory in a way that will cause docker to OOM kill if all sample files > Docker container memory limit.

Miserlou:
We might also want to seriously consider an approach where we can generate a QN target by choosing a % subsample of total available samples.

cgreene:
I'm also sure we could write code to make QN targets that uses less than 1MB of RAM and is linear with respect to the number of samples. I expect this is 10-20 lines of python.

What does the format of each sample look like and what does the output need to look like?

jaclyn-taroni:
Here's how we expect one would do this calculation:
* Create a set of genes from the first sample under consideration. If we come across and sample that is not using the same set of genes (either additional genes or missing genes), throw an error. 
  * We did have some issues where mislabeled samples caused this to fail. I forget exactly how we handle that case currently. We could potentially just drop the sample(s) that don't have the same sets of genes.
* Make a list with a number of elements equal to the number of genes in the set with all values initialized to zero.
* For each sample: sort it based on expression values, add the value to the matched spot in the index list.
  * If `NA` values are encountered (I think this is _highly unlikely_ given that all samples will be refine.bio processed with `SCAN`), set them to zero before sorting and adding.
* Keep track of the number of samples that were used and divide the list by that number of samples -> that's your QN target.
* Please also keep track of the _genes_ that were used (first noted here: https://github.com/AlexsLemonade/refinebio/issues/488#issuecomment-411741103 and if we're going to make changes, we should do this as well.)

jaclyn-taroni:
We can also include a target generated with `preprocessCore::normalize.quantiles.determine.target` as part of testing. We'll have to pay attention to floating-point imprecision, though.

cgreene:
@jaclyn-taroni's plan/implementation sounds like the way to go. I have some minor suggestions with regards to the error handling, but the plan/description of the implementation sounds good. With the affy processors from CEL files in particular, we really should have quite a lot of control over what we're seeing. If it's wrong, we really should figure it out and fix it instead of perpetuating bad data.

> We did have some issues where mislabeled samples caused this to fail. 

Yikes! Would it be better to start with a known good sample to define the genes list? Also, this raises whether or not our platform annotations should be revised based on the content of the data if it's really happening. How many samples that we processed from raw have mismatched numbers/names of genes?

> If NA values are encountered (I think this is highly unlikely given that all samples will be refine.bio processed with SCAN), set them to zero before sorting and adding.

This seems like an error. It seems as if we'd want to stop processing if this happens and figure out how the NAs crept in, fix samples that have fallen prey to these errors, and then re-run the QN.

Miserlou:
Implemented here: https://github.com/AlexsLemonade/refinebio/pull/1015

infrastructure/api-configuration/api-server-instance-user-data.tpl.sh
```diff
@@ -45,7 +45,8 @@ if [[ ${stage} == "staging" || ${stage} == "prod" ]]; then
     elif [[ ${stage} == "prod" ]]; then
         # The certbot challenge cannot be completed until the aws_lb_target_group_attachment resources are created.
         sleep 180
-        certbot --nginx -d api.refine.bio -n --agree-tos --redirect -m g3w4k4t5n3s7p7v8@alexslemonade.slack.com
+        RANDOM_API=$(( ( RANDOM % 8 ) + 2 )) # 2 to 9
+        certbot --nginx -d api.refine.bio -d api$RANDOM_API.refine.bio -n --agree-tos --redirect -m g3w4k4t5n3s7p7v8@alexslemonade.slack.com
```
kurtwheeler: :joy:

workers/data_refinery_workers/processors/test_smasher.py
```diff
@@ -1198,6 +1200,46 @@ def test_all_samples(self):
         self.assertEqual(row_num, 1)  # only two data rows in tsv file
         os.remove(tsv_filename)
 
+    @tag("smasher")
+    def test_all_precomendia(self):
```
kurtwheeler: What's precomendia?

PR #1015, opened 2019-02-04
Iteratively Create QN Target

Miserlou:
## Issue Number
https://github.com/AlexsLemonade/refinebio/pull/1013

Avoids any joins at all by adding 

We lose the ability to quality check since we don't have a smashed frame anymore, but the values for the test cases ended up being the same (we actually gained a point of precision, I don't know if that's good or bad).

## Methods
https://github.com/AlexsLemonade/refinebio/pull/1013

## Types of changes

## Functional tests
Same tests, slightly updated.

### DIFF
--- a/workers/data_refinery_workers/processors/qn_reference.py
+++ b/workers/data_refinery_workers/processors/qn_reference.py
@@ -35,23 +35,62 @@ def _prepare_input(job_context: Dict) -> Dict:
     # I'm not crazy about this yet. Maybe refactor later,
     # but I need the data now.
     job_context = smasher._prepare_files(job_context)
-    job_context = smasher._smash(job_context)
-
-    if not 'final_frame' in job_context.keys():
-        logger.error("Unable to prepare files for creating QN target.",
-            job_id=job_context['job'].id)
-        job_context["job"].failure_reason = "Couldn't prepare files creating QN target (no final_frame)."
-        job_context['success'] = False
-        return job_context
 
     # work_dir is already created by smasher._prepare_files
     outfile_base = job_context['work_dir'] + str(time.time()).split('.')[0]
-    outfile = outfile_base + '.tsv'
-    job_context['final_frame'].to_csv(outfile, sep='\t', encoding='utf-8')
-    job_context['smashed_file'] = outfile
     job_context['target_file'] = outfile_base + '_target.tsv'
 
     return job_context
+
+def _build_qn_target(job_context: Dict) -> Dict:
+
+    job_context['time_start'] = timezone.now()
+
+    # Get the gene list from the first input
+    geneset_target_frame = pd.read_csv(job_context['input_files']['ALL'][0].get_synced_file_path(), sep='\t', header=0,
+                                          index_col=0, error_bad_lines=False)
+
+    # Get the geneset
+    geneset = geneset_target_frame.index.values
+
+    # Sequentially build the target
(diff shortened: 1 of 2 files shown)


workers/data_refinery_workers/processors/qn_reference.py
```diff
@@ -35,24 +35,68 @@ def _prepare_input(job_context: Dict) -> Dict:
...
+    """ Iteratively creates a QN target file, method described here: https://github.com/AlexsLemonade/refinebio/pull/1013"""
+
+    job_context['time_start'] = timezone.now()
+
+    # Get the gene list from the first input
+    geneset_target_frame = pd.read_csv(job_context['input_files']['ALL'][0].get_synced_file_path(), sep='\t', header=0,
+                                          index_col=0, error_bad_lines=False)
+
+    # Get the geneset
+    geneset = geneset_target_frame.index.values
```
cgreene: May want to make this a set. Faster comparison.

workers/data_refinery_workers/processors/qn_reference.py
```diff
@@ -35,24 +35,68 @@ def _prepare_input(job_context: Dict) -> Dict:
...
+    sum_frame = sum_frame.set_index('index')
+
+    # Read and sum all of the inputs
+    num_valid_inputs = 0
+    for file in job_context['input_files']['ALL']:
+        input_frame = pd.read_csv(file.get_synced_file_path(), sep='\t', header=0,
+                                          index_col=0, error_bad_lines=False)
+
+        # If this input doesn't have the same geneset, we don't want it!
+        if list(input_frame.index.values) != list(geneset):
```
cgreene: Also I expect doing this as a set to be a faster operation. Perhaps substantially faster.

workers/data_refinery_workers/processors/qn_reference.py
```diff
@@ -35,24 +35,68 @@ def _prepare_input(job_context: Dict) -> Dict:
...
+                target_geneset=geneset,
+                bad_geneset=input_frame.index.values)
+            continue
+
+        # Sort the input
+        sample_name = list(input_frame.columns.values)[0]
+        expressions = input_frame.sort_values(sample_name)
+
+        # Add the sorted input to the sum frame
+        sum_frame['sum'] = sum_frame['sum'] + expressions[sample_name].values
```
cgreene: Vector summation! 🎉

workers/data_refinery_workers/processors/qn_reference.py
```diff
@@ -35,24 +35,68 @@ def _prepare_input(job_context: Dict) -> Dict:
...
+    for file in job_context['input_files']['ALL']:
+        input_frame = pd.read_csv(file.get_synced_file_path(), sep='\t', header=0,
+                                          index_col=0, error_bad_lines=False)
+
+        # If this input doesn't have the same geneset, we don't want it!
+        if list(input_frame.index.values) != list(geneset):
+            logger.error("Input frame doesn't match target geneset, skipping!",
+                bad_file=file,
+                target_geneset=geneset,
+                bad_geneset=input_frame.index.values)
```
cgreene: These should be a huge red flag. If we have these, our SCAN processor isn't producing what we expect in all cases (could happen if CEL files != what's annotated and we trust what's annotated). These will be very important to track down.

Issue #912, opened 2018-12-10
[Investigation] Track down unprocessed zebrafish RNA-Seq samples


kurtwheeler:
### Context

The last time I was trying to figure out what's holding up our RNA-Seq samples from being available on the front end I determined that there were ~12k samples that haven't had a successful processor job. However discovering their existence was a much simpler query than identifying them. 

### Problem or idea

Since zebrafish is the first organism we want to make a compendium for, we should try to get it as close to fully processed as possible. Really we should only be missing samples that are too poor quality to process.

### Solution or next step

Figure out what zebrafish samples haven't been processed and why. Open additional issues to handle the various reasons their processing failed and/or requeue samples that just need more retries or a higher RAM amount.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

kurtwheeler:
I have created a CSV with the sample id, sample accession, experiment id, and experiment accession!
[unprocessed_zebrafish_samples.txt](https://github.com/AlexsLemonade/refinebio/files/2665152/unprocessed_zebrafish_samples.txt)

kurtwheeler:
This is the code I used to generate that file:
```
from data_refinery_common.models import *

multiqc_result_ids = ComputedFile.objects.filter(filename='multiqc_report.html').values('result_id')

unprocessed_sample_ids = SampleResultAssociation.objects.exclude(result__in=multiqc_result_ids).values('sample_id')

unprocessed_zebra_samples = Sample.objects.filter(organism_id=46, id__in=unprocessed_sample_ids).prefetch_related("experiments")

with open('/tmp/unprocessed_zebrafish_samples.txt', 'w', newline='') as csvfile:
    csvfile.write("sample_id,sample_accession_code,experiment_id,experiment_accession_code\n")
    for sample in unprocessed_zebra_samples:
        csvfile.write(
            "{s_id},{s_ac},{e_id},{e_ac}\n".format(
                s_id=sample.id,
                s_ac=sample.accession_code,
                e_id=sample.experiments.first().id,
                e_ac=sample.experiments.first().accession_code))
```

kurtwheeler:
Ok I must have messed something up while running that before because it has way too many accessions and now that I'm trying to rerun it the queries are timing out. I'll follow up on this on Wednesday.

kurtwheeler:
[...]

PR #1032, opened 2019-02-06
Make Foreman target zebrafish samples

kurtwheeler:
## Issue Number

#227 
#912 

This PR makes the Foreman a fish shepherd, or a fishherd if you will.

The way it does this is by first preventing the Foreman from looking at any jobs that were created before yesterday. This is so that the Foreman won't be queuing up lots of the jobs we have in our system waiting to be run, so that instead it can just hunt down specific experiments/samples and queue jobs for those.

The second thing it does is to hunt down specific experiments/samples and requeue the jobs for those. It does this in a prioritized way, where the experiments with the greatest percentage of samples completed will have their samples queued up first.

My overall game plan is to get this thing running and hunting down unprocessed zebrafish samples. While it does this, I'm going to work on a tximport-only processor job so I can trigger it for experiments that get processed as much as possible. Looking through the experiments for such candidates will be a manual process until I can figure out a pattern I can automate.

## Types of changes

## Functional tests

The test I wrote for this command is pretty good and I added a couple foreman.main tests. Also I'd like to do a staging test.

## Checklist


common/data_refinery_common/migrations/0013_organism_experiments.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.1.5 on 2019-02-06 22:23
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('data_refinery_common', '0012_experiment_platform_accession_codes'),
```
Miserlou: this conflicts with https://github.com/AlexsLemonade/refinebio/pull/1006

foreman/data_refinery_foreman/foreman/management/commands/organism_shepherd.py
```diff
@@ -0,0 +1,240 @@
...
+    want, because we explicitly want to requeue these jobs regardless
+    of how many times they've been retried.
+    """
+    # All new jobs are going to be set at 2 retries so they only get
+    # tried once. Presumably all of these have failed at least once
+    # already, so there may be a good reason. Not immediately having
+    # them retried will give me a chance to actually take a look at
+    # what is happening to to them.
+    num_retries = 2
+    if type(job).__name__ == "ProcessorJob":
```
Miserlou: Not pretty

foreman/data_refinery_foreman/foreman/management/commands/organism_shepherd.py
```diff
@@ -0,0 +1,240 @@
...
+
+        for original_file in job.original_files.all():
+            DownloaderJobOriginalFileAssociation.objects.get_or_create(
+                downloader_job=new_job,
+                original_file=original_file
+            )
+
+        job_type = Downloaders[job.downloader_task]
+    else:
+        raise(ValueError("Told to requeue a job that's not a ProcessorJob nor DownloaderJob!"))
```
Miserlou: Raise is a statement not a function, I'm also guessing this is an untested branch

kurtwheeler: It is an untested branch because it _should never happen_. I don't actually know how I would get to this branch without changing the code to do so, it's just here as a just-in-case kinda thing.

kurtwheeler: However I will use `raise` correctly.

common/data_refinery_common/migrations/0013_organism_experiments.py
kurtwheeler: I think #1006 is ready to go out, I'll resolve this once that gets merged.

PR #1036, opened 2019-02-07
Fix shepherd command options, handle samples who don't have jobs.

kurtwheeler:
## Issue Number

N/A came up during staging test of #1032 

I forgot to add the add_arguments method to the management command.

I also forgot that sometimes it's valid for samples to not have any jobs associated with them because sometimes we don't want to process them. So this handles that case as well.

## Types of changes

## Functional tests

I've run this on the staging box by pushing a new image to my own Docker repo and then running the command using that image.

## Checklist

### DIFF
--- a/infrastructure/foreman-configuration/foreman-server-instance-user-data.tpl.sh
+++ b/infrastructure/foreman-configuration/foreman-server-instance-user-data.tpl.sh
@@ -29,7 +29,7 @@ docker run \\
        -e DATABASE_USER=${database_user} \\
        -e DATABASE_PASSWORD=${database_password} \\
        -e ELASTICSEARCH_HOST=${elasticsearch_host} \\
-       -e ELASTICSEARCH_PORT=80 \\
+       -e ELASTICSEARCH_PORT=${elasticsearch_port} \\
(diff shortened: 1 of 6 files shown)


kurtwheeler:
I also had to fix how I made it sleep and how many jobs it would try to queue at once. It works though!!!

infrastructure/foreman-configuration/foreman-server-instance-user-data.tpl.sh
```diff
@@ -28,6 +28,8 @@ docker run \\
        -e DATABASE_NAME=${database_name} \\
        -e DATABASE_USER=${database_user} \\
        -e DATABASE_PASSWORD=${database_password} \\
+       -e ELASTICSEARCH_HOST=${elasticsearch_host} \\
+       -e ELASTICSEARCH_PORT=80 \\
```
Miserlou: Why hardcoded?

kurtwheeler: Because it's not available as a parameter from terraform, but I just changed that.

PR #1048, opened 2019-02-11
Prevent hotfix deploys from running the deploy step twice.

kurtwheeler:
## Issue Number

N/A came up because I saw a deploy happen twice for one tag.

We made the hotfix deploy job run for tags that match /v.*-hotfix/, but we didn't stop it from running for tags that match /v.*/, which includes all the tags that would match /v.*-hotfix/. This makes it so the normal deploy ignores hotfix tags.

## Types of changes

## Functional tests

N/A this needs a hotfix tag to be tested.

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -286,7 +286,10 @@ workflows:
             branches:
               ignore: /.*/
             tags:
+              # From CirlceCI Docs:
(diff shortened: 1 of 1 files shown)


Miserlou:
Good

PR #1055, opened 2019-02-13
Filter Wrong Organism Samples from Compendia // Guard Against Unicode Metadata

Miserlou:
Two bugfixes in here: 

 - explicitly out an experiment's samples that are of the wrong organism type,
 - Properly handle unicode metadata, guard against against TSV writing fail


workers/data_refinery_workers/processors/test_smasher.py
```diff
@@ -1223,6 +1225,70 @@ def test_experiment(self):
         self.assertEqual(row_num, 0) # only one data row in tsv file
         os.remove(tsv_filename)
 
+    @tag("smasher")
+    def test_unicode_writer(self):
+        self.unicode_metadata = {
+            'experiments': {
+                "E-GEOD-😎": {
```
kurtwheeler: :sunglasses:

PR #1057, opened 2019-02-13
Reserve RAM on client instances so not more than 4 SALMON jobs can run at once.

kurtwheeler:
## Issue Number

#227 

I've observed 5 concurrent SALMON jobs as being enough to make fasterq-dump timeout. Therefore this uses Nomad's RAM reservation to make it so that 4 SALMON jobs will fit on a single instance, but not 5.

This can be undone once Nomad releases its tmpfs feature.

## Types of changes

## Functional tests

From one of our client instances:
```
ubuntu@ip-10-0-81-222:~$ TARGET_RAM=141072
ubuntu@ip-10-0-81-222:~$ TOTAL_RAM=$(free -m | grep Mem | tr -s ' ' | cut -d\  -f2)
ubuntu@ip-10-0-81-222:~$ RESERVED_RAM=$((TOTAL_RAM - TARGET_RAM))
ubuntu@ip-10-0-81-222:~$ echo $RESERVED_RAM
350733
ubuntu@ip-10-0-81-222:~$ echo $TOTAL_RAM
491805
```

I also tested the `sed` command by running it against client.tpl.hcl and making sure it replaced `CHANGE_ME` properly.

## Checklist

### DIFF
--- a/infrastructure/nomad-configuration/client-instance-user-data.tpl.sh
+++ b/infrastructure/nomad-configuration/client-instance-user-data.tpl.sh
@@ -115,6 +115,17 @@ EOF
 # Make the client.meta.volume_id is set to waht we just mounted
 sed -i "s/REPLACE_ME/$EBS_VOLUME_INDEX/" client.hcl
 
+# We want to leave enough RAM to run 4 SALMON jobs at once but not
+# enough to run 5.  Each SALMON job takes 32768MB of RAM, 32768 * 4 =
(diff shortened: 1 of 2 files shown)


kurtwheeler:
@Miserlou I'm not sure how else to test this. Do you think this is small enough for a hotfix or should we do a dev deploy? (Either way I will wait until the compendia job is done)

PR #1059, opened 2019-02-14
Hacks/optimizations to try to get fasterq-dump to actually succeed for Zebrafish

kurtwheeler:
## Issue Number

#227 

I want to run 3 of the smallest instance type, so if we get a bigger one it throws that off, so therefore I've temporarily disabled them.

Also fasterq-dump is still failing ~80% of the time. I'm trying to address this two ways:
  1. Decreaase the number of concurrent Salmon jobs from 4 to 3
  2. Increase the fasterq-dump timeout to allow it to complete.

I think 2 is justifiable based on the average file sizes for successful Salmon jobs

```
data_refinery=> select AVG(size_in_bytes) from original_files where id in (select original_file_id from processorjob_originalfile_associations where processor_job_id in (select id from processor_jobs where pipeline_applied='SALMON' and success ='t' order by last_modified desc limit 20));
         avg         
---------------------
 4902683849.50000000
```

 vs. Salmon jobs that fail because of this timeout:
```
data_refinery=> select AVG(size_in_bytes) from original_files where id in (select original_file_id from processorjob_originalfile_associations where processor_job_id in (select id from processor_jobs where failure_reason like '%timed out after %' and success ='f' order by last_modified desc limit 20));
         avg         
---------------------
 8265004603.15000000
```

As you can see the average file size for jobs that timeout is almost more than 1.5x as big as those that succeed.

## Types of changes

## Functional tests

N/A, this is mostly configuration/infrastructure.

## Checklist

### DIFF
--- a/infrastructure/instances.tf
+++ b/infrastructure/instances.tf
@@ -315,151 +315,151 @@ resource "aws_spot_fleet_request" "cheap_ram" {
   ##
   # x1.16xlarge
   ##
-  launch_specification {
+  # launch_specification {
 
-    # Client Specific
-    instance_type             = "x1.16xlarge"
-    weighted_capacity         = 10 # via https://aws.amazon.com/ec2/instance-types/
-    spot_price                = "${var.spot_price}"
-    ami                       = "${data.aws_ami.ubuntu.id}"
-    iam_instance_profile_arn  = "${aws_iam_instance_profile.data_refinery_instance_profile.arn}"
-    user_data                 = "${data.template_file.nomad_client_script_smusher.rendered}"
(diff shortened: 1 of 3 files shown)


Miserlou:
What's with the launch specs out?

kurtwheeler:
From the pull request description:

> I want to run 3 of the smallest instance type, so if we get a bigger one it throws that off, so therefore I've temporarily disabled them.

Miserlou:
Okay, I guess so. I hope you're keeping track of all the stuff we'll need to put back so we can do it all at once.

kurtwheeler:
Yeah I know.... I'm not stoked on it either but we have a nice record of all the PR's related to this nonsense.

Issue #1071, opened 2019-02-18
Detailed samples endpoint returning a huge JSON response


arielsvn:
### Context

We use the `/samples` endpoint to get detailed information about each sample.

### Problem or idea

I noticed this was really slow, and when inspected the response I saw there are a lot of files that are being added to the response. For sample [ERR657518](https://api.refine.bio/samples/?experiment_accession_code=ERP006132&offset=0&limit=1) we're returning 2827 files.

Here're two for reference.

```
{  
   "id":1079489,
   "filename":"ERR657546_output_gene_lengthScaledTPM.tsv",
   "size_in_bytes":1026229,
   "is_smashable":true,
   "is_qc":false,
   "sha1":"7d328cae179d00d98b6227262ac20d37c4507c4f",
   "s3_bucket":"data-refinery-s3-circleci-prod",
   "s3_key":"ya0uy4rr7peiwc49o1tq37ye_ERR657546_output_gene_lengthScaledTPM.tsv",
   "created_at":"2019-02-13T04:49:54.632412Z",
   "last_modified":"2019-02-13T05:24:56.316789Z"
},
{  
   "id":1079486,
   "filename":"ERR657549_output_gene_lengthScaledTPM.tsv",
   "size_in_bytes":1037777,
   "is_smashable":true,
   "is_qc":false,
   "sha1":"12109dfa0687dfac88dd906b509c9fe2a2567433",
   "s3_bucket":"data-refinery-s3-circleci-prod",
   "s3_key":"rojrnifyrcaymo9wo9jf1nf0_ERR657549_output_gene_lengthScaledTPM.tsv",
   "created_at":"2019-02-13T04:49:54.371617Z",
   "last_modified":"2019-02-13T05:24:55.966066Z"
},
```

### Solution or next step

A simple solution would be to modify the serializers to avoid sending that field.

Tagging @kurtwheeler in case there might be a bug behind this.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

kurtwheeler:
This looks to me like we're associated tximport files badly, like really badly. A single sample should not have multiple accession's worth of files associated with it.

PR #1073, opened 2019-02-19
Remove `files` from the sample details serializer

arielsvn:
## Issue Number

#1071 

Remove `files` from the response of the API, on the detailed samples endpoint (`/samples`).

### DIFF
--- a/api/data_refinery_api/serializers.py
+++ b/api/data_refinery_api/serializers.py
@@ -95,25 +95,8 @@ class ComputationalResultAnnotationSerializer(serializers.ModelSerializer):
                     'last_modified'
                 )
 
-class ComputedFileSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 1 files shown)


Miserlou:
I think this should be done with a new serializer, there are cases where we'll want to see the files assoc'd with a result outside of sample context.

arielsvn:
But if we add a new serializer, then the `ComputationalResultSerializer` won't be used anywhere else, since it was added specifically for the detailed samples endpoint. 

Do you think we should still keep it in case we need it in the future?

Miserlou:
We should have a results endpoint

arielsvn:
I see, I had missed that endpoint.

Added now a new serializer only with the fields we use in the frontend. Let me know what do you think @Miserlou

Issue #911, opened 2018-12-10
Break tximport out into its own job type


kurtwheeler:
### Context

Last week @Miserlou had the idea that the tximport step of the salmon pipeline may be using a lot more more RAM than the rest of the job. It does make sense that we may be having trouble finishing salmon experiments because once we get to the very final step, tximport, we get OOM-killed.

This is something that could be checked by looking in our database. If this is really what is going on then we would expect to see a lot of salmon experiments where there is only one missing sample because that was the sample whose job ran tximport.

### Problem or idea

We can break tximport out of the salmon job type and into its own. Salmon jobs which would run tximport can then instead create a new processor job to do so and queue it up. This way we can tweak the RAM requirements for the tximport jobs separately. It also would give us more clarity as to how frequently the tximport step specifically is failing.

### Solution or next step

It may be worth trying to determine if we have a lot of salmon experiments that only are missing one sample from being processed before actually implementing this. However this is the easiest answer to get out of the database given that we can't use the `is_processed` flag on the samples table because it's not set to true until tximport is run for the whole experiment.

However regardless of whether or not tximport getting OOM-killed is really the main culprit of what's blocking experiments from finishing, we may want to implement this anyway because it could work nicely with #909 because I expect there are a number of experiments where some number of samples are unprocessable and tximport will never be triggered from within Salmon. Even once we implement #909, the remaining samples won't be able to be processed so they won't run tximport even once the threshold has been lowered. (For example, consider an experiment that is 75% processed but whose remaining samples we cannot process. Even if we set the cutoff at 60%, the remaining 25% of that experiment's samples won't finish the salmon steps of the salmon pipeline so they won't make it to the tximport part. We may need to make a one-off management command to queue tximport jobs for all the experiments which have processed all the samples which are processable.)

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

cgreene:
Oo - getting the info on what's missing here:
>  If this is really what is going on then we would expect to see a lot of salmon experiments where there is only one missing sample because that was the sample whose job ran tximport.

will also inform #909

jaclyn-taroni:
Agreed!

jaclyn-taroni:
Let's plan to at least include the "checking the database"/information gathering part of this for this sprint.

cgreene:
Done via #1074

PR #1074, opened 2019-02-19
Tximport Only Job

kurtwheeler:
## Issue Number

#911 

This breaks tximport out into it's own job so we can queue it for experiments that are close enough.

## Methods

I used the 25 sample/80% cutoff that was determined by @cansav09 in https://github.com/AlexsLemonade/tximport_partial_run_tests/pull/3, however if any experiment gets to 100% tximport will still be run no matter how few samples it has.

## Types of changes

## Functional tests

Unit tests so far, but this needs more.

## Checklist


kurtwheeler:
Ok tests are passing, this should be good for review. My game plan for testing is to do it in staging because otherwise I have to run at least 20 salmon samples through. I've found three experiments in staging that I'll queue up a tximport job for:
* experiment with id 256 has 75/88 samples processed -- should work
* experiment with id 22 has 22/23 samples processed -- should not have enough saples
* experiment with id 50 has 39/106 samples processed -- should not have enough percentage

workers/nomad-job-specs/tximport.nomad.tpl
```diff
@@ -0,0 +1,94 @@
...
+
+        LOG_LEVEL = "${{LOG_LEVEL}}"
+      }
+
+      # The resources the job will require.
+      resources {
+        # CPU is in AWS's CPU units.
+        cpu = 1024
+        # Memory is in MB of RAM.
+        memory = 3276 #8
```
Miserlou: 8?

Miserlou: Also why 3276? Is this enough for large experiments? Should this have an increasing back off?

kurtwheeler: I think what happened here is that it was supposed to be 327678. That's already the highest tier for salmon, but it seemed like it's a good starting point for tximport. Based on how these jobs do we can add a increasing back off to it.

Or I can change this right now to start lower and go up to 32678, but I'm thinking tximport will need more RAM than the rest of the salmon pipeline. I'll admit though, I'm not sure how sure I am about that.

What do you think?

Miserlou: These are rare enough jobs that we can just start with max RAM. I'd give it more CPU as well.

Issue #1076, opened 2019-02-20
Enforce Deterministism/Uniqueness Of Smashed Column Titles


Miserlou:
### Context

Linking up samples and metadata in the compendia isn't as easy as it should be because we've had to apply some ~~hacks~~ creative engineering to ensure columnar uniqueness.

If we don't want to immediately trip up users, we should enforce and record uniqueness deterministically.

PR #1079, opened 2019-02-21
[Discussion] Dont Smash Multi-Channel Samples and Samples With Indeterminable Accession Codes

Miserlou:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/1076

What should we do with samples where the files have multiple channels or an indeterminable accession codes? Currently we smash them anyway, but this ends up breaking up code which joins sample data with sample metadata.

Fortunately, these samples are pretty rare overall but some of them might be important, I don't know enough about 2-channel samples. (That also might not be the primary reason, it might be something that came out of the processing step itself.)

Anyway, this PR logs and skips those samples rather than trying to brute force them into the smash/compendia result.


jaclyn-taroni:
@miserlou - can you post an example sample accession code of one of these multi-channel unsmashable files please? There are no 2 channel platforms on the supported microarray list to the best of my knowledge, so having an example will probably help get to the bottom of things. Thanks!

Miserlou:
I've been assuming that's the problem since that's what's triggering the condition in the test data, I have no idea if that's actually whats causing the issue in production. Here's some: `'E-MEXP-405-HS04', 'E-MEXP-405-HS03', 'E-MEXP-405-HS02', 'E-MEXP-405-HS01', 'GSM1707858', 'GSM1707864', 'GSM1707855'`

jaclyn-taroni:
These guys `E-MEXP-405-HSXX` don't surprise me, this experiment is strange in that it has a bunch of array design files and it is a 2 channel experiment. 

This set `'GSM1707858', 'GSM1707864', 'GSM1707855'` are from [`GSE69719`](https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GSE69719), which is a zebrafish Affymetrix ST array. I would not expect us to have trouble on these if we are using the sample accession codes as the column name unless something strange is happening with the CEL files. I counted 12 CEL files for 12 samples.

cgreene:
Todo: Confirm that problem samples are in production and not just test data (for our own sanity to understand why this happens).

We agree that this solution is good, but we just want to check the data source.

jaclyn-taroni:
For some reason, I thought `E-MEXP-405-HSXX` samples were from a strange 2-color experiment that was ChIP-chip. Searching ArrayExpress with "E-MEXP-405-HS01" doesn't return any results. I'm _fairly_ sure that's how l looked up that information to begin with but not absolutely sure.

If we look at [`E-MEXP-405` at ArrayExpress](https://www.ebi.ac.uk/arrayexpress/experiments/E-MEXP-405/), we can see that this is an Affymetrix experiment on the `zebrafish` platform. This is consistent with the information at `https://api.refine.bio/samples/?accession_code=E-MEXP-405-HS04`. It does not appear to be processed, so I imagine this particular sample is not a problem in production. I will note that I looked at the "raw data" from ArrayExpress and those files are `.txt`, not CEL files so I'm wondering if that contributes to any weirdness (possibly via our NOOP pipeline) in the test data.

workers/data_refinery_workers/processors/test_smasher.py
```diff
@@ -290,7 +294,9 @@ def test_smasher(self):
             print('###')
 
             final_context = smasher.smash(job.pk, upload=False)
-            final_frame = final_context['final_frame']
+            final_frame = final_context.get('final_frame')
+            if not final_frame:
+                continue
```
kurtwheeler: These three code blocks concern me because if there's a bug that means `final_frame` never gets set then we won't run any assertions and so this will just quietly succeed. We're only looping over two elements in each of these loops. For each element we should just either expect it to be in there or not.

Miserlou: I agree, and that's actually what was happening. Looking into it, it was Agilent data that shouldn't have been there.

Miserlou:
Addressed

PR #1081, opened 2019-02-22
Enable cloudfront compression by default

arielsvn:
I enabled it manually on aws. But adding it here so that it gets activated on the next deploy.

https://github.com/AlexsLemonade/refinebio-frontend/issues/543


arielsvn:
Pagespeed was estimating a 4.35 secs improvement, and the score went from 20 (slow) to 68 (average)

Issue #1068, opened 2019-02-18
Downloader job recreation seems to be broken for GEO jobs


kurtwheeler:
### Context

#1006 

### Problem or idea

GEO downloader jobs that were recreated don't get any original files associated with them.

### Solution or next step

Figure out why not and fix it

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

arielsvn:
@kurtwheeler can we close this?

kurtwheeler:
Yes! Good call! We definitely created a lot of successful GEO jobs!!!

PR #1082, opened 2019-02-22
Prevent samples from GEO from having their downloader jobs recreated.

kurtwheeler:
## Issue Number

#1068 

This is a bandaid for #1068 because:
  * We've processed the microarray data that we want to for now so we don't need to recreate these jobs at the moment.
  * We want to get RNA-Seq data back up and processing, we don't process RNA-Seq data from GEO anyway.
  * A proper fix for #1068 might be a bit tricky to get right because GEO metadata is so inconsistent.

## Types of changes

## Functional tests

This is straightforward, should be testable in our staging test today.

## Checklist

### DIFF
--- a/workers/data_refinery_workers/processors/utils.py
+++ b/workers/data_refinery_workers/processors/utils.py
@@ -134,6 +134,16 @@ def create_downloader_job(undownloaded_files: OriginalFile) -> bool:
         accession_code = original_downloader_job.accession_code
         original_files = original_downloader_job.original_files.all()
 
+        sample_object = original_files[0].samples.first()
+
+    # Disable GEO downloader job recreation until we resolve:
+    # https://github.com/AlexsLemonade/refinebio/issues/1068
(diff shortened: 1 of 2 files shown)


workers/data_refinery_workers/processors/utils.py
```diff
@@ -134,6 +134,16 @@ def create_downloader_job(undownloaded_files: OriginalFile) -> bool:
         accession_code = original_downloader_job.accession_code
         original_files = original_downloader_job.original_files.all()
 
+        sample_object = original_files[0].samples.first()
+
+    # Disable GEO downloader job recreation until we resolve:
+    # https://github.com/AlexsLemonade/refinebio/issues/1068
+    if sample_object.source_database == "GEO":
+        logger.info("Not recreating a downloader job for a sample coming from GEO.",
```
Miserlou: Change this if you want to see it in awslogs

Issue #644, opened 2018-09-24
Search results should be limited to organisms with QN targets


kurtwheeler:
### Context

> At this point, we'll only want to deliver quantile normalized data to users

https://github.com/AlexsLemonade/refinebio/issues/642

### Problem or idea

Since we only want to deliver quantile normalized data to users, and we don't want to show users data we won't give them, we shouldn't show them data that we cannot quantile normalize. We can only quantile normalize data that we have already built a QN target for. Therefore the front end should not expose samples which are from organisms we don't have QN targets for.

### Solution or next step

I'm not 100% sure what the best way to implement this is, so the next step is figuring out a good way to do so..

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

cansav09:
There are some species that we don't have QN targets for like Yeast that show up as downloadable, but are not actually downloadable. 

Example: [GSE26169](https://www.refine.bio/experiments/GSE26169/expression-data-for-saccharomyces-cerevisiae-oxidative-stress-response)

cansav09:
Further information on this phenomenon: apparently though I clicked the "Add to Dataset" button seen above, and I didn't think it added to my Downloadable dataset (because it still had a `0` badge like above), apparently it DID (sort of) add to my cart. 

I attempted to add this to my cart yesterday. I didn't think it added, but then today I downloaded a separate dataset and ended up with an empty folder and empty metadata file for this dataset. Presumably because it was still added to my cart. 
Note: GSE26169 is the dataset above that is Yeast, that doesn't actually have data that still got added even though it shouldn't have. GSE37418 is a different dataset that I was actually trying to download for another purpose. 

![Screen Shot 2019-04-17 at 10 34 26 AM](https://user-images.githubusercontent.com/23458084/56296392-6724c980-60fc-11e9-904e-12f3bd865a82.png)

cansav09:
@arielsvn It is unclear to me how much of this is a front-end issue, so you may want to read through my comments to see what may or may not need to be fixed on front-end.

arielsvn:
This should also be solved with https://github.com/AlexsLemonade/refinebio/issues/1162, without it we don't have information on the samples that are downloadable in the FE. 

I'm gonna close this to keep track of the issue in #1162

PR #1090, opened 2019-02-26
Adds qn_targets_available Endpoint

Miserlou:
## Issue Number
https://github.com/AlexsLemonade/refinebio/issues/644

One possible solution to the 644 problem, allows us to search/return all samples but gives frontend a method to filter/hide un-QN'able samples.


arielsvn:
LGTM. I'm gonna try to implement the frontend changes before we merge this to ensure we cover all cases.

arielsvn:
Sample response: 

```
[
  {'name': 'HOMO_SAPIENS', 'taxonomy_id': 9606}, 
  {'name': 'DANIO_RERIO', 'taxonomy_id': 7955}
]
```

arielsvn:
@Miserlou added https://github.com/AlexsLemonade/refinebio-frontend/pull/559 that ensures samples without QN targets are displayed as not processed in the samples table.

There're two details that I think we should add to the API:

1. Ensure that only samples with QN targets can be added to the dataset. This is especially important when we add entire experiments with the key `ALL`

2. Send the number of downloadable samples: samples that are processed and have QN targets. We need to display this number both in the search page and the experiments page.

image

As we discussed, we need to think what would be the best way to add this.

Miserlou:
I strongly agree with (1), although the result object should be the same at smashtime, they will get the "unsmashable files" warning for items they didn't add, I think (2) might be harder and that might not be the correct way to solve it, my gut says that should be turned into something cachable.

Miserlou:
The ['ALL'] issue is addressed now, (2) isn't but might be out of scope for this PR as it requires some additional architecture/API design consideration.

api/data_refinery_api/views.py
```diff
@@ -540,7 +541,12 @@ def perform_update(self, serializer):
             accessions = new_data['data'][key]
             if accessions == ["ALL"]:
                 experiment = get_object_or_404(Experiment, accession_code=key)
-                sample_codes = list(experiment.samples.filter(is_processed=True).values_list('accession_code', flat=True))
+
+                # QN organisms
+                organism_ids = list(ComputationalResultAnnotation.objects.filter(data__is_qn=True).values_list('data__organism_id', flat=True).order_by('data__organism_id'))
+                organisms = Organism.objects.filter(id__in=organism_ids)
```
arielsvn: What about calculating this outside of the `for...loop` above? In some cases we send multiple `ALL` keys:

image

api/data_refinery_api/views.py
```diff
@@ -540,7 +541,12 @@ def perform_update(self, serializer):
...
+
+                sample_codes = list(experiment.samples.filter(is_processed=True, organism__in=organisms).values_list('accession_code', flat=True))
```
arielsvn: Should we do the same check for when someone wants to add a sample with an accession code? And throw an error if the sample doesn't have any QN targets associated?

api/data_refinery_api/views.py
```diff
@@ -1336,6 +1342,19 @@ def get(self, request, format=None):
...
 
+class QNTargetsAvailable(APIView):
+    """
+    This is a list of all of the organisms which have available QN Targets
+    """
+    def get(self, request, format=None):
+        
+
+        organism_ids = list(ComputationalResultAnnotation.objects.filter(data__is_qn=True).values_list('data__organism_id', flat=True).order_by('data__organism_id'))
+        organisms = Organism.objects.filter(id__in=organism_ids)
```
arielsvn: Seems that we use this in multiple places, is it worthwhile to move it into `Organism`? Similarly to `Experiments.public_objects`

api/data_refinery_api/views.py
Miserlou: That's an interesting question but I'm not sure it's the API's duty to do that at that stage if we're returning them those objects only to have using them be a source of failure..

Miserlou:
I'd like to get this reviewed and merged as it will fix some of the sporadic failures in other PRs.

api/data_refinery_api/views.py
```diff
@@ -1357,6 +1361,19 @@ def get(self, request, format=None):
...
```
arielsvn: `Organism.get_objects_with_qn_targets()` ?

Issue #968, opened 2019-01-03
There are microarray samples which are mislabeled as RNA-Seq


kurtwheeler:
### Context

https://github.com/AlexsLemonade/refinebio/issues/963
https://github.com/AlexsLemonade/refinebio/issues/967

### Problem or idea

Some of our microarray samples have been mislabeled as RNA-Seq. This seems to be related to surveying super-series in GEO. Here is one example I came across: https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GSM1692993

### Solution or next step

We need to find all of these samples and rectify the data issues. The easiest solution may be to purge the affected experiments and re-survey them.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

kurtwheeler:
Here's an interesting one I found:

The sample with id 465479 two successful AFFY_TO_PCL jobs and one failed NO_OP job, despite being labeled as an RNA-seq sample.

So it looks like we're actually detecting the correct technology at some point in order to try and process it like that. When we do we should update the sample and/or experiment records, related #967

This might also be helpful when writing a data migration to correct these samples, looking at the jobs a sample had run on it looks like it could indicate that a sample is mislabeled.

kurtwheeler:
I've been tracking down where we are actually detecting the correct technology and setting it. It's during the processor job when we actually have the file and we run `affyio:read.celfile.header` on the file to see what's actually there.

What I think this means is that it won't be too hard to migrate these mis-labeled samples to have the correct technology/manufacturer, since we can just find all samples that had a AFFY_TO_PCL job run for it, but it won't be so easy to set the correct package information for them because we won't have the files any more. It still could be done with a slightly larger data-correction-job if we wanted to re-download the samples so we could figure it out.

Issue #967, opened 2019-01-03
GEO superseries cause incorrect platform/technology detection


kurtwheeler:
### Context

https://github.com/AlexsLemonade/refinebio/issues/963#issuecomment-451249805

### Problem or idea

We should detect the correct technology and platform for all samples, regardless of whether they are surveyed as part of a superseries or not.

### Solution or next step

Two things should be done here:
  1. Fix the code so that the samples in GSE69108 have the same platform/accession code whether surveyed via GSE69111 or GSE69108.
  2. Try to figure out if there's any other potential edge cases that could be broken and address those.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

Issue #966, opened 2019-01-03
We need to ignore RNA-Seq samples coming from GEO


kurtwheeler:
### Context

https://github.com/AlexsLemonade/refinebio/issues/963#issuecomment-451241279

### Problem or idea

We have nothing in our GEO surveyor which is attempting to ignore RNA-Seq samples.

### Solution or next step

Implement logic into the GEO surveyor that either prevents samples from being created for RNA-Seq samples or that prevents those samples from getting queued.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

Issue #963, opened 2019-01-02
Which files should be used when smashing by species?


kurtwheeler:
### Context

I'm currently working on #909, which is forcing me to review/rethink some of the ways we create, associate, and use files output by the salmon processor. One thing I've realized is that since a sample can belong to more than one experiment, it may have multiple smashable files which are each a slice of a different output from tximport. Currently our smash logic just uses the most recently created file. After discussing with @cansav09 we've identified that for experiment-level smash jobs we should be making sure to use files that came from the same tximport run. However we weren't able to figure out what to do when the smash job is at the species-level.

### Problem or idea

We need to figure out a good rule for picking a computed file to use for a given sample when smashing at the species level.

### Solution or next step

Unsure, but I imagine figuring out the answer to this question will involve @jaclyn-taroni 

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

cansav09:
A potentially oversimplified answer for the species level question is to use the result file from the larger sample size. As we discussed, in the "ideal world" we'd probably want to run tximport with all the samples together, but that isn't practical.

jaclyn-taroni:
If I’m understanding correctly: use the files from the same tximport run and the most recent set of files for species level smashing.

Do I have that right @cansav09? This is also what I was thinking.

cansav09:
@jaclyn-taroni  Sort of. 

My thought is that for tximport experiments, you should use the tximport result from the same result. So if sample 1 is in both experiments A and B, (and thus has two tximport results) and the user is trying to get experiment A, definitely use the tximport result for sample 1 from Experiment A. Because we know from the tximport tests I did that the result will be quite drastically changed depending on the samples it's processed with. 

However, if the use is trying to get the samples from Experiment A and Experiment B, and thus you have to choose between two different results for sample 1, then my thought is that you should choose the tximport result from the bigger experiment. So say Experiment A is 500 sample and Experiment B is 100 samples, for sample 1, choose the Experiment A result. (my thought is that you would like all the samples to be comparable as possible. So pick the result that would help sample 1 relate to more samples?). 

In an ideal world, we'd want to re-run Experiment A and B's tximport to get the best results and make them the most comparable. But it sounds like this is not a practical thing. And I don't think the benefits of doing this outweigh the impracticality of it. 

[...]

PR #1091, opened 2019-02-26
Don't download/process RNA-Seq samples from GEO

kurtwheeler:
## Issue Number

#966 
#967 
#968 

The goal of this PR is to prevent samples from making it through our system and still being mislabeled. Additionally this will prevent salmon from being run on samples that came from GEO.

## Types of changes

## Functional tests

I reran the GEO surveyor for GSE6911 because as I mentioned here, https://github.com/AlexsLemonade/refinebio/issues/963#issuecomment-451249805, it didn't get labelled properly. With these changes, it works now!

I think everything else is well covered by unit tests.

## Checklist


common/data_refinery_common/models/models.py
```diff
@@ -749,6 +749,12 @@ def needs_downloading(self, pipeline_applied=None) -> bool:
         # has been processed and doesn't need to be processed again.
         return successful_processor_jobs.count() == 0
 
+    def is_CEL_file(self) -> bool:
+        """Return true if original_file is a CEL file."""
+        upper_name = self.source_filename.upper()
+        return (len(upper_name) > 4 and upper_name[-4:] == ".CEL") \
+            or (len(upper_name) > 7 and upper_name[-7:] == ".CEL.GZ")
```
Miserlou: This I'm not sure about - is a CEL.GZ really a CEL file? Isn't it a gzip file?

foreman/data_refinery_foreman/surveyor/geo.py
```diff
@@ -359,27 +348,57 @@ def create_experiment_and_samples_from_api(self, experiment_accession_code) -> (
...
+                # It's okay to survey RNA-Seq samples from GEO, but we
+                # don't actually want to download/process any RNA-Seq
+                # data unless it comes from SRA.
+                if sample_object.technology != 'RNA-SEQ':
+                    created_samples.append(sample_object)
+
+                # Now that we've determined the technology at the
+                # sample level, we can set it at the experiment level,
+                # just gotta make sure to only do it once.
+                # XXX: Do we want to remove this field? It's not going
```
Miserlou: If we're touching this code, can we address this concern now or remove the triple X? :vin_diesel:

workers/data_refinery_workers/downloaders/utils.py
```diff
@@ -165,7 +165,7 @@ def delete_if_blacklisted(original_file: OriginalFile) -> OriginalFile:
         original_file.save()
         return None
 
-    return OriginalFile
+    return original_file
```
Miserlou: Wow, good catch :fish:

workers/run_tests.sh
```diff
@@ -479,6 +479,7 @@ for image in ${worker_images[*]}; do
                --env AWS_ACCESS_KEY_ID \
                --env AWS_SECRET_ACCESS_KEY \
                --volume $volume_directory:/home/user/data_store \
+               --memory=5G \
```
Miserlou: TIL 💡

kurtwheeler:
Ok, I think I addressed your concern @Miserlou

PR #1098, opened 2019-03-01
Correct Affy samples mislabelled as RNA-Seq

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/968
https://github.com/AlexsLemonade/refinebio/issues/963#issuecomment-451249805
https://github.com/AlexsLemonade/refinebio/issues/967

This is related to #1091 which corrects the way we survey GEO. That PR prevents us from mislabelling more Microarray data as RNA-Seq. This PR corrects the data that was already mislabelled.

## Types of changes

- Data migration

## Functional tests

I surveyed GSE69111 to get all of its samples mislabelled as RNA-Seq. I then ran this command and it corrected ONLY the microarray samples back to microarray.

I've retested this since fixing the platform detection stuff:
```
10 | GSM1692995     | postnatal day 1 (p1), biological rep3                    | GEO             |                    |                 |                           | t       | Illumina Genome Analyzer II | Illumina Genome Analyzer II | RNA-SEQ    | ILLUMINA     |
```
got corrected to:
```
10 | GSM1692995     | postnatal day 1 (p1), biological rep3                    | GEO             |                    |                 |                           | t       | mogene10st                  | [MoGene-1_0-st] Affymetrix Mouse Gene 1.0 ST Array | MICROARRAY | AFFYMETRTIX  |
```

(Sorry for the dump of partial rows straight from the DB, but I think it's good enough to show the technology, manufacturer, platform_name, and platform_accession_codes are changing correctly.)

## Checklist

### DIFF
--- a/common/data_refinery_common/models/models.py
+++ b/common/data_refinery_common/models/models.py
@@ -749,8 +749,8 @@ class OriginalFile(models.Model):
         # has been processed and doesn't need to be processed again.
         return successful_processor_jobs.count() == 0
 
-    def is_CEL_file(self) -> bool:
-        """Return true if original_file is a CEL file."""
+    def is_affy_data(self) -> bool:
+        """Return true if original_file is a CEL file or a gzipped CEL file."""
         upper_name = self.source_filename.upper()
         return (len(upper_name) > 4 and upper_name[-4:] == ".CEL") \
             or (len(upper_name) > 7 and upper_name[-7:] == ".CEL.GZ")
--- a/common/data_refinery_common/models/models.py
+++ b/common/data_refinery_common/models/models.py
@@ -749,6 +749,12 @@ class OriginalFile(models.Model):
         # has been processed and doesn't need to be processed again.
         return successful_processor_jobs.count() == 0
 
+    def is_CEL_file(self) -> bool:
+        """Return true if original_file is a CEL file."""
+        upper_name = self.source_filename.upper()
+        return (len(upper_name) > 4 and upper_name[-4:] == ".CEL") \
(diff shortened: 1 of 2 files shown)


kurtwheeler:
Actually, just realized that it didn't properly handle the platform information, this is still a WIP.

common/data_refinery_common/models/models.py
```diff
@@ -749,6 +749,12 @@ def needs_downloading(self, pipeline_applied=None) -> bool:
         # has been processed and doesn't need to be processed again.
         return successful_processor_jobs.count() == 0
 
+    def is_CEL_file(self) -> bool:
```
Miserlou: Discussion around this function name happened in the sister PR.

workers/data_refinery_workers/processors/management/commands/correct_affy_samples.py
```diff
@@ -0,0 +1,145 @@
...
+
+        target_file = open(file_path, "wb")
+        with urllib.request.urlopen(download_url) as request:
+            shutil.copyfileobj(request, target_file, CHUNK_SIZE)
+
+        urllib.request.urlcleanup()
+    except Exception:
+        logger.exception("Exception caught while downloading file %s from %s",
+                         file_path, download_url)
+        raise
```
Miserlou: In general I find this pattern to be a bit weird (I'd rather it be do a thing vs fail or return True/False rather than True/raise, this seems to be both) but it looks like you're handling it okay and this function is only used here so it's not a big deal. I'd refactor if you've got spare cycles but if you don't it's NBD.

cgreene:
This is no longer WIP

Issue #873, opened 2018-11-28
Route jobs away from `fasterq-dump` ?


jaclyn-taroni:
### Context

Related to https://github.com/AlexsLemonade/refinebio/pull/872, from that PR:

> Lots of our Salmon processors were getting stuck, but their containers weren't dying. This seemed to be occuring on the `fasterq-dump` step.

Some related issues for SRA tools: https://github.com/ncbi/sra-tools/issues/159, https://github.com/ncbi/sra-tools/issues/161

### Problem or idea

These hanging jobs will get killed as a result of #872, but will they run and hang again?

### Solution or next step

Some thoughts on what we might do next:
- Detect situations where `fasterq-dump` isn't working
- For these cases: use a different tool like `fastq-dump`, [`parallel-fastq-dump`](https://github.com/rvalieris/parallel-fastq-dump) (see: 
https://github.com/ncbi/sra-tools/issues/161#issuecomment-434729952) OR get these reads from European Nucleotide Archive 
- We could strip trailing Ns if that is the issue (https://github.com/ncbi/sra-tools/issues/159)

Tagging @Miserlou and @kurtwheeler to start the discussion

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

Miserlou:
I think I'd prefer to try to get them from ENA since the SRA file format was a stupid idea to begin with. We could also try updating the sra-tools version and trying again, I'm not sure if this bug is fixed already or not.

jaclyn-taroni:
@Miserlou about what % of Salmon jobs hang? Is it a low enough number such that if we fall back to ENA, we won't have to be concerned about knocking things over?

Miserlou:
Yes. This issue isn't actually our top priority right now, the corrupt (?) Salmon index issue is. Compared to that, it's only a small percentage of jobs that are getting stuck here.

jaclyn-taroni:
May be related: https://github.com/ncbi/sra-tools/issues/168

kurtwheeler:
I'm going to record an excerpt from a slack conversation between @Miserlou and I this morning that is relevant to this issue:

```
Rich [11:02 AM]
That would make sense, I don’t know what the under the hood differences are
We could also try retrying those samples from ena rather than ncbi

kurt-al [11:05 AM]
you mean like just the ones that fail?
[...]

PR #1106, opened 2019-03-07
Replace `fasterq-dump` With `fastq-dump -Z` and Named Pipes

Miserlou:

Using `fasterq-dump` at scale causes disk thrashing that kills our overall throughput. We tried solving this using ramdisks so that `fasterq-dump` just uses RAM rather than disk, but Nomad isn't ready for that as ramdisks aren't supported in `0.8.0` and don't work properly in `0.9.0-beta3`.

This solution uses named pipes created with `mkfifo` to pass the output from `fastq-dump -Z` from the `stdout` directly into Salmon. This doesn't use any disk IO as long as there is enough RAM available so that no paging is necessary.

I think this is probably a short-term fix but I think it'll get the job done. 

## Methods
Because we don't have direct access to FASTQ files directly any more, we can't use `fastqc`, so that step gets skipped entirely. Similarly, we can't use our old technique for calculating the length of the transcriptome index to use, so we parse the output of `sra-stats --xml --statistics` using the `untangle`.

There is a difference between in single and paired end reads, so we use `tee` and `grep` to split the streams in place, thanks so a solution posted online by Graham LaRue from UC Merced.

Since I think this fix will be temporary, I left the previous functionality in place and just commented it out of the pipeline.

## Types of changes

## Functional tests
Existing tests pass without modification, which is a good sign. May need one more test for assuring that both single and paired end reads work as expected (although they did in my manual testing).

### DIFF
--- a/workers/data_refinery_workers/processors/requirements.txt
+++ b/workers/data_refinery_workers/processors/requirements.txt
@@ -47,6 +47,7 @@ six==1.11.0               # via cycler, python-dateutil, retrying
 spectra==0.0.11           # via multiqc
 sympy==1.3
 urllib3==1.23             # via botocore, requests
+untangle==1.1.1
 
 # For ElasticSearch support:
 django-elasticsearch-dsl==0.5.1
--- a/workers/data_refinery_workers/processors/salmon.py
+++ b/workers/data_refinery_workers/processors/salmon.py
@@ -734,6 +734,7 @@ def _run_salmon(job_context: Dict) -> Dict:
     # them to disk.
     if job_context.get('sra_input_file_path', None):
 
+        # Single reads
         if job_context['sra_num_reads'] == 1:
 
             fifo = job_context["work_dir"] + "barney"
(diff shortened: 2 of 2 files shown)


cgreene:
Would close #1000

jaclyn-taroni:
☝️ Fixed that issue number, also closes #873

Miserlou:
@arielsvn can you review this? I'd like to try it on staging

luizirber:
Hi! I saw this PR and the previous discussion about SRA files, and I'm curious about a few things:

- Do you need the Named Pipes because salmon doesn't read files from stdin? I use the `fastq-dump -Z` output in [wort][0] to feed it into sourmash, but are there other benefits in making a named pipe (better buffering or more control?)
- I also saw [this post][1] from some time ago, do you have any other tricks to make downloading faster? =]

[0]: https://github.com/dib-lab/wort/blob/fbeeeb78a669dec8049e8d78bf8b6e2d7c38d488/wort/blueprints/compute/tasks.py#L41
[1]: https://www.ccdatalab.org/blog/a-desperate-plea-for-a-free-software-alternative-to-aspera

Miserlou:
Yes, Salmon doesn't read directly from `stdin`, there is an open ticket somewhere around Salmon related to that. There the author suggests using `/dev/fd0` but here we use FIFOs for the same purpose. Particularly for paired end reads, Salmon expects inputs as two separate files and can't handle interleaved pairs.

Aspera sucks. I'm praying that QUIC+HTTP/3 will emerge victorious from the hell pit of IETF standardizations as a suitable alternative, everybody will make the switch quickly (that should be the easy part given the absurdly high cost of Aspera server licenses), and Aspera will be one of those bad memories of internet past, like Java Applets and Adobe Shockwave.

Miserlou:
Seems to be working, `sra-stat` is occasionally quite slow but Salmon seems fast.

PR #1108, opened 2019-03-08
Set PG engine version in TF to match engine version in production

Miserlou:
Welp, this one is terrifying.

For some reason, all of of our PG servers are at 9.6.11. Our terraform value sets it to 9.6.6. 

It has been 9.6.6 for a year:

Screen Shot 2019-03-08 at 4 25 00 PM

And yet the prod instances are now at 9.6.11:
Screen Shot 2019-03-08 at 4 23 45 PM

So updating a stack with TF now causes a downgrade error:
Screen Shot 2019-03-08 at 4 27 36 PM

This updates the TF stack, but doesn't explain what the heck actually happened.

My guess is that AWS upgraded us and didn't tell us? I have no idea how that's possible. That seems like dangerous behavior.

## Types of changes

- Spooky ghosts (ghosts have made changes that can't be explaiend)

### DIFF
--- a/infrastructure/instances.tf
+++ b/infrastructure/instances.tf
@@ -549,6 +549,7 @@ resource "aws_db_instance" "postgres_db" {
   storage_type = "gp2"
   engine = "postgres"
   engine_version = "9.6.11"
+  auto_minor_version_upgrade = false
   instance_class = "db.${var.database_instance_type}"
   name = "data_refinery"
   port = "${var.database_hidden_port}"
--- a/infrastructure/instances.tf
+++ b/infrastructure/instances.tf
@@ -548,7 +548,7 @@ resource "aws_db_instance" "postgres_db" {
   allocated_storage = 100
   storage_type = "gp2"
   engine = "postgres"
-  engine_version = "9.6.6"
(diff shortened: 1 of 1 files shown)


Miserlou:
@kurtwheeler can you think of anything that would explain this

Miserlou:
Okay apparently RDS has something called "auto minor version upgrades" that are on by default, but related notifications aren't: https://aws.amazon.com/about-aws/whats-new/2018/12/amazon-rds-enhances-auto-minor-version-upgrades/

I don't even know if I'm mad or not

Miserlou:
Okay, I have changed the TF auto_minor_version_upgrade setting, although this means that there are more things being changed by TF during the upgrade, which adds more to the anxiety.

kurtwheeler:
Ok, we can do this in two steps if it would make you less anxious. I don't think it needs to change RIGHT NOW, but just before AWS does it to us again.

Miserlou:
Nah, I want to see what will happen on staging. I think it'll be okay we just need to keep an eye on it.

Miserlou:
This went okay.

Issue #731, opened 2018-10-11
Lock down S3 buckets


kurtwheeler:
### Context

It's possible to download objects from our S3 buckets if you the URL to the object, despite the permissions to list the contents of S3 buckets being locked down.

### Problem or idea

We should make it so that you need to have permissions to the data to download it, for legal reasons.

### Solution or next step

However will want people to be able to pull data from those buckets eventually, a prime example of this is transcriptome indices. I think this is the strategy @Miserlou mentioned for enabling that: https://docs.aws.amazon.com/STS/latest/APIReference/Welcome.html

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

kurtwheeler:
@cgreene I've been digging into this one a bit, what I'm thinking is setting all our buckets to private and then using https://boto3.amazonaws.com/v1/documentation/api/latest/guide/s3.html#generating-presigned-urls to generate a URL that will be valid for only a few hours to allow users to download data. This way they can get whatever we want to let them at, but only after they've agreed to our T&C. Will this satisfy the lawyers?

However, there's a question about what data needs to be locked down in this manner:
```
kurt-al [4:24 PM]
OK got it. Unsmashed processed data, downloadable datasets, and compendia should be private, while QN targets and tx-indices are public but only discoverable via API

jackie-al [4:25 PM]
yea I suspect those that would want those things are sufficiently computationally savvy
we may want a page that's like "here's some other stuff to make things work" eventually

kurt-al [4:26 PM]
there won't be legal concerns around distributing qn targets/tx-indices?

jackie-al [4:26 PM]
shouldn't be
I don't think one could figure out the original values of the samples used to derive the QN targets
and the transcriptome index doesn't have anything to do with individuals

kurt-al [4:27 PM]
What about in the "we used your data to make a decision and it was wrong" category?

jackie-al [4:28 PM]
oh hm
```

I can't remember if those are actual concerns or if I just overheard them as something that might be an issue. Which types of S3 objects do we need to make private and/or are there any that can be public?

This change will mean anything we put in a private bucket won't be accessible unless there's a webpage for it which allows you to agree to T&C and get a URL. At least not until we implement a way to agree via API.

cgreene:
[...]

PR #1109, opened 2019-03-11
Provide access to resources that should still be accessible even once we lock down our buckets

kurtwheeler:
## Issue Number

#731 

In order to be able to lock down our S3 buckets we need to provide a mechanism by which people can still get the data that they're supposed to. The mechanism this adds is that ComputedFile resources will have a `download_url` field if the request included an `API_KEY` header (or if the key was present in the POST-data for datasets). This `download_url` is a temporary presigned URL from AWS which allows access to the resource using that URL for a few hours.

## Types of changes

## Functional tests

In addition to the unit tests, I also tested this by making requests against the API using curl/the browser.

## Checklist

### DIFF
--- a/common/data_refinery_common/models/models.py
+++ b/common/data_refinery_common/models/models.py
@@ -956,7 +956,7 @@ class ComputedFile(models.Model):
 
     def create_download_url(self):
         """ Create a temporary URL from which the file can be downloaded."""
-        if self.s3_bucket and self.s3_key:
+        if settings.RUNNING_IN_CLOUD and self.s3_bucket and self.s3_key:
             return S3.generate_presigned_url(
                 ClientMethod='get_object',
                 Params={
--- a/common/data_refinery_common/models/models.py
+++ b/common/data_refinery_common/models/models.py
@@ -956,13 +956,14 @@ class ComputedFile(models.Model):
 
     def create_download_url(self):
         """ Create a temporary URL from which the file can be downloaded."""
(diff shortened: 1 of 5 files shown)


common/data_refinery_common/models/models.py
```diff
@@ -939,16 +939,34 @@ def get_synced_file_path(self, force=False, path=None):
...
+        return self.create_download_url()
+
+    def create_download_url(self):
+        """ Create a temporary URL from which the file can be downloaded."""
+        if settings.RUNNING_IN_CLOUD and self.s3_bucket and self.s3_key:
+            return s3.generate_presigned_url(
+                ClientMethod='get_object',
+                Params={
+                    'Bucket': self.s3_bucket,
+                    'Key': self.s3_key
```
Miserlou: What is ExpireIn going to be if we don't set it, and do we want the default?

kurtwheeler: OK, looks like I figured this guy out finally. I went with a day because that's how long datasets last. I have a link that shouldn't work this time tomorrow. I'll test that tomorrow but when I switched `ExpiresIn` to a parameter instead of an entry in the `Params` parameter it complained because I used milliseconds and that was more than a week worth of seconds (which is the limit).  Anyway I took that to mean I was finally setting that correctly.

Miserlou:
Compendia tests failing, probably needs AWS creds supplied

PR #1115, opened 2019-03-12
Actually lock down S3 buckets

kurtwheeler:
## Issue Number

#731 

In #1109 I set the API up so that it can grant access to non-public objects. Once that goes out the frontend will have to change so it sends api keys on request. Once that change is made we can actually release this so that the buckets actually get locked down.

QN targets are currently in the same bucket as all the other results, but we don't want to lock down QN Targets. Therefore I've made a new bucket that can be public to store these in and a migration to move their metadata. The S3 objects themselves will also have to be moved, but there's only 12 of them so far.

## Types of changes

## Functional tests

None yet. I figure that the migration is going to get out of date and then I'll want to retest this anyway so I'll wait until we're actually ready to release this to test it. It's pretty straightforward so I don't think the testing will take long once it's actually time.

## Checklist


arielsvn:
@kurtwheeler the FE is now using `download_url` to access the files, we should be able to lock down the buckets.

kurtwheeler:
@Miserlou this is now actually ready for review when you get a chance

infrastructure/disk.tf
```diff
@@ -92,9 +92,20 @@ resource "aws_s3_bucket" "data_refinery_transcriptome_index_bucket" {
...
+  tags {
+    Name        = "data-refinery-s3-qn-target-${var.user}-${var.stage}"
+    Environment = "${var.stage}"
+  }
+}
+
 resource "aws_s3_bucket" "data_refinery_compendia_bucket" {
   bucket = "data-refinery-s3-compendia-${var.user}-${var.stage}"
-  acl    = "public-read"
+  acl    = "private"
```
Miserlou: Why do compendia results need to be private?

common/data_refinery_common/migrations/0019_auto_20190404_1427.py
```diff
@@ -0,0 +1,36 @@
+# Generated by Django 2.1.5 on 2019-04-04 14:27
+
+from django.db import migrations
+from data_refinery_common.utils import get_env_variable
+
+
+# We want this to throw if it can't access this, no point in running a
+# migration to set everything to a bad value.
+S3_BUCKET_NAME = get_env_variable("S3_QN_TARGET_BUCKET_NAME")
```
Miserlou: Would prefer this to match `S3_QN_TARGET_BUCKET_NAME`

common/data_refinery_common/migrations/0019_auto_20190404_1427.py
```diff
@@ -0,0 +1,36 @@
...
+    Based off of:
+    https://simpleisbetterthancomplex.com/tutorial/2017/09/26/how-to-create-django-data-migrations.html
+
+    We can't import the ComputedFile model directly as it may be a newer
+    version than this migration expects. We use the historical version.
+    '''
+    # Pagination isn't necessary here because we have very few QN targets.
+    ComputedFile = apps.get_model('data_refinery_common', 'ComputedFile')
+    for computed_file in ComputedFile.objects.filter(is_qn_target=True):
+        computed_file.s3_bucket = S3_BUCKET_NAME
```
Miserlou: Why does this not automatically move the result from one bucket to the other? How will this process occur?

Miserlou:
The TF part of this is okay, I was expecting something that would actually move the files though.

common/data_refinery_common/migrations/0019_auto_20190404_1427.py
kurtwheeler: Yeah this is a good point. Looking back at this now I also think I should have included code to move the data too. Not sure why I didn't before, but I'll add that in.

common/data_refinery_common/migrations/0019_auto_20190404_1427.py
kurtwheeler: This is a good point!

infrastructure/disk.tf
kurtwheeler: Well my thinking was that they contain the data from specific samples, as per this comment: https://github.com/AlexsLemonade/refinebio/issues/731#issuecomment-469443058. However I guess they could be smashed up enough that it's not a risk? @cgreene could you clarify your previous comment to indicate if compendia should be treated more like dataset downloads or more like QN targets/tx-indices?

Miserlou: Maybe it's that we're expecting to distribute those via Zenodo?

cgreene: Conferred with @lemonadejay - we should require the checkbox for compendia as well 👍

common/data_refinery_common/models/models.py
```diff
@@ -902,6 +902,52 @@ def sync_from_s3(self, force=False, path=None):
             logger.exception(e, computed_file_id=self.pk)
             return None
 
+    def change_s3_location(self, new_bucket: str, new_key: str) -> bool:
+        old_bucket = self.s3_bucket
```
Miserlou: docstring

common/data_refinery_common/migrations/0020_update_qn_bucket.py
```diff
@@ -18,17 +19,25 @@ def update_qn_bucket(apps, schema_editor):
     We can't import the ComputedFile model directly as it may be a newer
     version than this migration expects. We use the historical version.
     '''
+
+    if not settings.RUNNING_IN_CLOUD:
```
Miserlou: Can we confirm this is set correctly, even it's run on the deploy box? Does that count as "the cloud"?

common/data_refinery_common/migrations/0020_update_qn_bucket.py
```diff
@@ -18,17 +19,25 @@ def update_qn_bucket(apps, schema_editor):
...
     ComputedFile = apps.get_model('data_refinery_common', 'ComputedFile')
     for computed_file in ComputedFile.objects.filter(is_qn_target=True):
-        computed_file.s3_bucket = S3_BUCKET_NAME
-        computed_file.save()
+        if not computed_file.s3_bucket or not computed_file.s3_key:
+            continue
+
+        if not computed_file.change_s3_location(S3_QN_TARGET_BUCKET_NAME, computed_file.s3_key):
+            # What the hell am I supposed to do if errors start ocurring during a migration?!?
+            break
```
Miserlou: Needs more information than just a break

common/data_refinery_common/migrations/0020_update_qn_bucket.py
kurtwheeler: Yes, when this is run on the deploy box it's done so using whatever the environment that is being deployed is: https://github.com/AlexsLemonade/refinebio/blob/dev/infrastructure/deploy.sh#L252

Issue #833, opened 2018-11-09
Missing Organism for GSE54716


Miserlou:
https://api.refine.bio/experiments/GSE54716/

`organisms` is `[]`, which shouldn't happen. Not a problem for GSE54717.

kurtwheeler:
I just tested this accession locally and it was properly associated with the correct organism. I wonder if this is a bug that has already been fixed?

As far as fixing this data goes though... I propose two steps:
  1. Do a query to figure out if there's any other experiments that had the same problem.
  2. Unsurvey those experiments and re-survey them in order to re-collect the metadata and fix the data.

@Miserlou does that sound like a good plan to you?

kurtwheeler:
I've found 129 total experiments without organisms:
```
  id  | accession_code 
------+----------------
  183 | E-MEXP-2381
 1560 | GSE106981
 2380 | GSE33539
  340 | GSE37164
 2799 | GSE49675
 1329 | GSE22337
 2075 | GSE42373
 1469 | GSE13033
 2448 | GSE44597
  497 | GSE34567
 1883 | GSE53537
  696 | GSE38123
 2707 | GSE59795
 2826 | GSE41498
 1680 | GSE107417
 2463 | GSE37044
 2465 | GSE23973
 1716 | GSE12999
 1772 | GSE12538
 1785 | GSE103171
 1805 | GSE22317
   72 | E-MEXP-3893
  136 | E-TABM-748
  200 | E-MEXP-2215
  285 | E-MTAB-1476
  279 | GSE31186
  300 | E-MEXP-2288
  430 | GSE37163
  415 | GSE31185
  613 | GSE55881
  472 | GSE13157
  531 | GSE45635
  510 | GSE23203
  539 | GSE36743
  560 | GSE34886
  706 | GSE17818
  661 | GSE17816
  646 | GSE54716
  676 | GSE32298
  675 | GSE35417
  693 | GSE27382
  738 | GSE64888
 2007 | GSE44808
 2484 | GSE32666
 2061 | GSE31229
 2138 | GSE37045
 2168 | GSE33112
 2498 | GSE51835
 2201 | GSE57750
 2259 | GSE43114
 2277 | GSE41914
 2293 | GSE40624
 2299 | GSE34776
 2312 | GSE4218
 2414 | GSE37371
 2578 | GSE54378
 2611 | GSE54762
 2631 | GSE48501
 2633 | GSE60124
 3019 | GSE49695
 3110 | GSE49586
[...]

PR #1119, opened 2019-03-12
Prevent and correct missing experiment/organism associations

kurtwheeler:
## Issue Number

#833 

When a GEO experiment was surveyed that had samples which had previously been discovered, these samples would get properly associated with the experiment, however we were not properly creating links between the experiment and the organism of those samples. This caused some experiments to have no organism links and probably others to be missing some (it's rather hard to find those so I didn't bother, but the migration corrects them anyway.)

This PR both corrects the GEO surveyor to prevent this issue in the future and a data migration to correct the already affected experiments.

## Types of changes

## Functional tests

I looked at the prod database and found experiment GSE54716 didn't have any organism_names. I also found that it shared some samples with GSE54717. I was then able to reproduce this locally by first surveying GSE54717 and then GSE54716.

Once I implemented my solution I cleared out the database and resurveyed these experiments. They both had the correct `organism_names`.

I then cleared out the database again and resurveyed without out my surveyor changes. This caused GSE54716 to be missing `organism_names` again, which I then was able to correct with my data migration.

## Checklist


Miserlou:
Good job, I think the migration conflicts with other branches

kurtwheeler:
Thanks. I can correct the migration once other branches go out. Basically as each branch that has a migration goes out I'll rebuild them. I _think_ that's easier to deal with than the git-history mess that can get made by pulling branches into other branches.

Miserlou:
It basically comes down to what order do you want these merged in

kurtwheeler:
I don't know that I care. Basically whatever order they get approved in :P

PR #1127, opened 2019-03-14
Use Paginator During Migration 0017

Miserlou:
We are too big to use `Sample.objects.all()`. This uses a paginator instead.


Miserlou:
This is blocking deployment: https://circleci.com/gh/AlexsLemonade/refinebio/14706

Miserlou:
All tests are failing due to unrelated stupidness in our testing setup that means that even accepted fixes don't actually get applied to CI until a deploy happens, and deploys don't happen unless tests pass. /giphy hellscape

PR #1130, opened 2019-03-14
Increase Max Clients

Miserlou:
Sets max_clients to 20 so we can crunch crunch crunch.


kurtwheeler:
Hotfix?

Miserlou:
Sure, why not

PR #1133, opened 2019-03-15
Detect SRA based on source, not file extention

Miserlou:
Sometimes files downloaded have a '.sra' extension, but apparently they don't.

This uses the source url rather than the file extension.


Miserlou:
Tested before and after locally with `ERR519481`

Issue #1131, opened 2019-03-14
Missing quant files during tximport run


Miserlou:
```
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod Traceback (most recent call last):
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod   File "/home/user/data_refinery_workers/processors/utils.py", line 480, in run_pipeline
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod     last_result = processor(last_result)
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod   File "/home/user/data_refinery_workers/processors/salmon.py", line 740, in tximport
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod     tximport_inputs = _get_tximport_inputs(job_context)
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod   File "/home/user/data_refinery_workers/processors/salmon.py", line 729, in _get_tximport_inputs
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod     quant_files.append(ComputedFile.objects.filter(result=result, filename="quant.sf")[0])
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod   File "/usr/local/lib/python3.5/dist-packages/django/db/models/query.py", line 303, in __getitem__
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod     return qs._result_cache[0]
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod IndexError: list index out of range
```

Miserlou:
@kurtwheeler if you have a db connection open can you see how often this one is occuring

kurtwheeler:
```
data_refinery=> select count(*) from processor_jobs where pipeline_applied='SALMON' and failure_reason like '%list index out of range%' and end_time < now() - interval '1 days';
 count 
-------
   100
(1 row)
```

kurtwheeler:
I suspect this may have been related to #2932. After we rerun the samples in that and cleanup the dead ones I'd be interested to see if these errors still persist.

kurtwheeler:
This hasn't happened since `2019-03-14 21:32:08.214767+00`. I think this may be resolved by #2932.

PR #1135, opened 2019-03-15
Add better logging if tximport cannot find a quant file it is expecting to.

kurtwheeler:
## Issue Number

#1131 

We know this is where things are going wrong. A little better logging will help dig into why.

## Types of changes

## Functional tests

Unit tests should cover this.

## Checklist


Miserlou:
This will be logged, but that doesn't mean we'll ever to be able to actually catch this error.

a) Should this result in a failure of the job?
b) If so, should can't we store this information in failure_reason?

workers/data_refinery_workers/processors/salmon.py
```diff
@@ -726,18 +726,34 @@ def _get_tximport_inputs(job_context: Dict) -> Dict[Experiment, List[ComputedFil
...
+                    logger.exception(
+                        "Salmon quant result found without quant.sf ComputedFile!",
+                        processor_job=job_context["job_id"],
+                        quant_result=result.id,
+                        experiment=experiment.id
+                    )
+                    job.failure_reason = (
+                        "Salmon quant result {} found without quant.sf"
+                        " ComputedFile in experiment {}!"
+                    ).format(str(result.id), str(experiment.id))
```
Miserlou: Is there anyway we can get the sample accession here? This is going to be harder to investigate this way.

Issue #1146, opened 2019-03-18
Force New SRA Endpoints


Miserlou:
Experiments surveyed before the new SRA endpoints were swapped in are still being downloaded from the old endpoints. This appears to be killing NCBI. We should try to intercept these URLs and replace them with the new ones.

PR #1148, opened 2019-03-19
[Discussion] Force New Endpoints

Miserlou:
## Issue Number
https://github.com/AlexsLemonade/refinebio/issues/1146

Forces new SRA endpoints. Detailed testing actually reveals that these endpoints are way way slower (20 vs 150 Mb/s) than the old ones, but hopefully they are less prone to failure. Not sure how we want to proceed, but here it is.

### DIFF
--- a/workers/data_refinery_workers/downloaders/sra.py
+++ b/workers/data_refinery_workers/downloaders/sra.py
@@ -1,6 +1,6 @@
 from contextlib import closing
 from typing import List
-import django.utils.timezone
+from django.utils import timezone
 import os
 import shutil
 import subprocess
--- a/workers/data_refinery_workers/downloaders/sra.py
+++ b/workers/data_refinery_workers/downloaders/sra.py
@@ -1,5 +1,6 @@
 from contextlib import closing
 from typing import List
+import django.utils.timezone
 import os
 import shutil
 import subprocess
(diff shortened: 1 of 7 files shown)


workers/data_refinery_workers/downloaders/sra.py
```diff
@@ -35,6 +36,15 @@ def _download_file(download_url: str,
         download_url = download_url.replace('.uk/', '.uk:/')
         return _download_file_aspera(download_url, downloader_job, target_file_path, source="ENA")
     elif "ncbi.nlm.nih.gov" in download_url and not force_ftp:
+        # Try to convert old-style endpoints into new-style endpoints if possible
+        try:
+            if 'anonftp' in download_url:
+                accession = download_url.split('/')[-1].split('.sra')[0]
+                new_url = get_fasp_sra_download(accession)
+                if new_url:
+                    download_url = new_url
```
kurtwheeler: I think that we probably also need to update the OriginalFile object's `source_url` field because otherwise we won't actually know exactly where the file came from.

Miserlou: Yeah I was thinking about that but I'm not sure if it's the right play as we destroy source information, for instance, if we want to fall back and use the old URL on retries, etc.

kurtwheeler: Good point, and also I'm not convinced the URL we get back from NCBI is even persistent so storing it could be completely useless.

PR #1152, opened 2019-03-20
Kurtwheeler/better expected failures

kurtwheeler:
## Issue Number

N/A Came up based on running in prod

There's a couple places that we purposefully fail jobs:
  1. When there's too many downloader jobs and the only way we have to reduce that number is by having them kill themselves.
  2. When a job gets a signal from Nomad or Docker.

When this happens we should set a failure reason so we can tell the difference between these expected failures and failures that don't have a failure reason for an unexpected reason.

Also this stops us logging about queuing jobs in prod because we have 100's of thousands of jobs and that's just a lot of logging.

## Types of changes

## Functional tests

WIP.

## Checklist

### DIFF
--- a/common/data_refinery_common/message_queue.py
+++ b/common/data_refinery_common/message_queue.py
@@ -79,10 +79,10 @@ def send_job(job_type: Enum, job, is_dispatch=False) -> bool:
     else:
         raise ValueError("Invalid job_type: {}".format(job_type.value))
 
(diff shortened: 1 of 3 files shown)


Miserlou:
Good, but I actually came up with the same thing in https://github.com/AlexsLemonade/refinebio/pull/1153/files and it requires a specific string.. =/

Miserlou:
Closing because this has been reimplemented in 1153

Issue #1080, opened 2019-02-21
Some experiments have an incorrect title


arielsvn:
### Context

We added slugs to the experiment url, and while generating a sitemap, google started complaining that a few experiments had invalid urls.

### Problem or idea

Problem was that the urls were too long (will fix in the frontend). But the experiments in question have an incorrect title.

![image](https://user-images.githubusercontent.com/1882507/53202240-00030100-35f4-11e9-9bf4-f0d0fcab08b1.png)

I think all of them can be found by searching [title:AEExperimentType](https://www.refine.bio/search?q=title:AEExperimentType). Some of them are:

- https://www.refine.bio/experiments/E-MTAB-392
- https://www.refine.bio/experiments/E-MEXP-76
- https://www.refine.bio/experiments/E-MEXP-1513

### Solution or next step

Not sure how to address this. Tagging @Miserlou and @kurtwheeler for next step.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

kurtwheeler:
These seem to have all been fixed. The linked experiments look good and searching title:AEExperimentType has no results.

PR #1154, opened 2019-03-20
Fix for #1080

Miserlou:
Source was bad data in the metadata files, not the API response.


Miserlou:
The existing experiments will have to be re-surveyed.

PR #1160, opened 2019-03-21
Log stderr too when populating elasticsearch index.

kurtwheeler:
## Issue Number

N/A trying to get elasticsearch index working automatically again.

I should have included this in #1157 but I didn't think about it. /var/log/api_cron.log is empty so I guess all the output is happening on stderr.

## Types of changes

## Functional tests

I tested that a command like this works as expected.

## Checklist


kurtwheeler:
I tested this by actually changing the crontab on the API instance to what I most recently committed here. It worked, this should be G2G!

Miserlou:
Nice! Merge it!

Issue #1155, opened 2019-03-20
Foreman Needs Pagination


Miserlou:
We saw a foreman OOM death, likely needs a paginator.

Miserlou:
This looks like it actually will mean a larger refactor for the foreman.

Miserlou:
:taco: to kurt for this fix

Miserlou:
I think this actually introduced a more serious DB connection issue, possibly illustrated here:

![Screen Shot 2019-03-25 at 4 06 54 PM](https://user-images.githubusercontent.com/139987/54950874-afdfcd00-4f18-11e9-82c9-0cfaca03ff6d.png)

![Screen Shot 2019-03-25 at 4 11 58 PM](https://user-images.githubusercontent.com/139987/54950907-c1c17000-4f18-11e9-8266-e1c142ec172e.png)

Let's joint code review this section tomorrow AM and see if we can something less CPU hungry. Clues here: https://github.com/ross/performant-pagination ?

PR #1164, opened 2019-03-21
Make foreman use pagination, remove some outdated comments from tests.

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/1155

The Foreman got OOM-killed because it was pulling too many objects from the DB at once. This uses pagination so we only pull 200 at a time.

## Types of changes

## Functional tests

The unit tests do a good job of testing this because functionally nothing should have changed other than our memory usage. I'll make sure this actually does well on staging before pushing to prod though.

## Checklist

### DIFF
--- a/foreman/data_refinery_foreman/foreman/main.py
+++ b/foreman/data_refinery_foreman/foreman/main.py
@@ -1,26 +1,18 @@
+import datetime
 import nomad
 import socket
 import time
 import traceback
 
+from django.conf import settings
+from django.core.paginator import Paginator
+from django.db import transaction
+from django.utils import timezone
+from functools import wraps
 from nomad import Nomad
 from nomad.api.exceptions import URLNotFoundNomadException
 from typing import List, Set
-from functools import wraps
-import datetime
-from django.conf import settings
-from django.utils import timezone
-from django.db import transaction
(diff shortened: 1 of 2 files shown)


Miserlou:
Huge!! I think EBI is down which is causing test failure right now, don't think it's related to these changes but let's make sure things are passing before this gets merged.

foreman/data_refinery_foreman/foreman/main.py
```diff
@@ -880,44 +946,50 @@ def retry_lost_survey_jobs() -> None:
...
-            if job_status != "pending" and job_status != "running":
-                logger.debug(("Determined that a survey job needs to be requeued because its"
-                             " Nomad Job's status is: %s."),
-                            job_status,
-                            job_id=job.id
+    if potentially_lost_jobs.count() == 0:
+        return
+
+    paginator = Paginator(potentially_lost_jobs, 200)
+    for page_idx in range(1, paginator.num_pages + 1):
```
Miserlou: This looks funny because pages are 0 indexed and range isn't inclusive of the final value, but I think it's right. Maybe needs a comment? Probably not.

Issue #1167, opened 2019-03-22
Bad Strand Mapping Stat


Miserlou:
```
ta-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod 2019-03-21 21:22:21,103 i-04f994c9abb658f9e [volume: 9] data_refinery_workers.processors.utils ERROR [processor_job: 2101316] [no_retry: False]: Unhandled exception caught while running processor function _run_salmon in pipeline:
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod Traceback (most recent call last):
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod   File "/usr/local/lib/python3.5/dist-packages/django/db/backends/utils.py", line 85, in _execute
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod     return self.cursor.execute(sql, params)
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod psycopg2.DataError: invalid input syntax for type json
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod LINE 1: ..., "created_at", "last_modified") VALUES (1044248, '{"compati...
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod                                                              ^
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod DETAIL:  Token "NaN" is invalid.
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod CONTEXT:  JSON data, line 1: ..._fragment_ratio": 1.0, "strand_mapping_bias": NaN...
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod The above exception was the direct cause of the following exception:
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod Traceback (most recent call last):
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod   File "/home/user/data_refinery_workers/processors/utils.py", line 482, in run_pipeline
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod     last_result = processor(last_result)
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod   File "/home/user/data_refinery_workers/processors/salmon.py", line 982, in _run_salmon
data-refinery-log-group-circleci-prod log-stream-processor-docker-circleci-prod     kv.save()
[...]

PR #1172, opened 2019-03-25
Guard against invalid Salmon JSON output

Miserlou:
## Issue Number
https://github.com/AlexsLemonade/refinebio/issues/1167

There is a bug in our version of Salmon. It's already fixed in later versions. This guards against catastrophe in that failure case, since Postgres' JSONField won't accept invalid JSON.


Miserlou:
I think these test failures are unrelated as they appear to be foreman tests, and this is only touching a single processor.

Issue #735, opened 2018-10-12
Salmon results have too many associations with tximport computed files


kurtwheeler:
### Context

https://api.refine.bio/samples/?accession_code=SRR3098580 has the following files associated with it:
```
SRR3098579_output_gene_lengthScaledTPM.tsv
SRR3098580_output_gene_lengthScaledTPM.tsv
SRR3098581_output_gene_lengthScaledTPM.tsv
SRR3098582_output_gene_lengthScaledTPM.tsv
txi_out.RDS
```
When it should only have 
```
SRR3098580_output_gene_lengthScaledTPM.tsv
txi_out.RDS
```

### Problem or idea

Looks like when we split the txi_out.RDS files into multiple files, one for each sample, we associate them all with all of the samples, rather than matching them up correctly.

### Solution or next step

Make sure that each `XXXXXX_output_gene_lengthScaledTPM.tsv` file is only associated with a sample that has that accession code.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

kurtwheeler:
That sounds good. I've tagged it now let's just make sure to investigate after a deploy.

kurtwheeler:
I just looked into this again to see if it was in fact still happening and it looks like it is: https://api.refine.bio/samples/?accession_code=SRR1981231

PR #1173, opened 2019-03-25
Kurtwheeler/no more multiqc

kurtwheeler:
## Issue Number

This will help with https://github.com/AlexsLemonade/refinebio/issues/735

We don't actually have `.fastq` files any more since we dump them using named pipes, so we can't run fasterqc so this removes that. However, now multiqc only uses the salmon output, which is already in JSON, to produce other JSON. Therefore @jaclyn-taroni has approved the removal of multiqc from our salmon pipeline.

This takes them out and removes a bit of other dead code.

## Methods

MultiQC and FastQC are officially gonezo.

## Types of changes

## Functional tests

The unit tests for these cover this functionality pretty well.

## Checklist


Miserlou:
Excellent, this solves a number of problems at the same time.

workers/data_refinery_workers/processors/salmon_quant.yml
```diff
@@ -11,6 +11,7 @@ os_pkg:
 # Command lines (commands whose outputs will be kept track of)
 cmd_line:
   - salmon --version
+  - fasterq-dump --version
```
Miserlou: We no longer use fasterq-dump, I guess this sholuld be fastq-dump?

I also suggest that sra-stat should go in here

PR #1186, opened 2019-03-28
Break out of foreman pagination early.

kurtwheeler:
## Issue Number

#1179

The foreman can pull a lot of pages of jobs to requeue. There can be way more jobs to iterate through than we want to actually queue, so this makes it so that once we hit the queue capacity for each job type we stop looping over the pages of jobs.

## Types of changes

## Functional tests

They're hard to do for this without the volume of data we have in the cloud.

## Checklist

### DIFF
--- a/foreman/data_refinery_foreman/foreman/main.py
+++ b/foreman/data_refinery_foreman/foreman/main.py
@@ -895,40 +895,32 @@ def requeue_survey_job(last_job: SurveyJob) -> None:
 
     return True
 
-
-def handle_survey_jobs(jobs: List[SurveyJob]) -> None:
-    """For each job in jobs, either retry it or log it.
-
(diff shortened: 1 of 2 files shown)


kurtwheeler:
I agree with that, it seems like a good refactor. I can move where we check the queue's capacity out of the `handle_*_jobs` functions, into their own functions, and then call those from the `retry_*` functions now though!

kurtwheeler:
Ok, `foreman.main` tests are passing locally for me, I don't think the end-to-end tests will time out, however there was significant work since you approved it so this should probs be re-reviewed.

Miserlou:
Changes look good, but I'm still confused about the behavior on CI. I think it's outside of our control. Let's do this

Issue #1184, opened 2019-03-28
Dynamically Compute Wheeler Limit


Miserlou:
The downloader job limit is hardcoded, this should be made dynamic based on cluster capacity.

PR #1187, opened 2019-03-28
Dynamically Calculate Max Downloader Jobs Limit Based On Nomad Cluster Size

Miserlou:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/1184

Replaces a hardcoded value with a value based on the number of nodes in the cluster. For 10 nodes, sets a theoretical 10GiB/s max, though I don't think we'll ever get close to this.

Also includes a guard against a semi-rare `sra-stat` failure.

### DIFF
--- a/foreman/data_refinery_foreman/foreman/main.py
+++ b/foreman/data_refinery_foreman/foreman/main.py
@@ -48,9 +48,9 @@ MAX_NUM_RETRIES = 2
 # This can be overritten by the env var "MAX_TOTAL_JOBS"
 DEFAULT_MAX_JOBS = 20000
 
-
 # This is the maximum number of non-dead nomad jobs that can be in the queue.
-MAX_TOTAL_DOWNLOADER_JOBS = 400
+MAX_TOTAL_DOWNLOADER_JOBS = 0
+TIME_OF_LAST_SIZE_CHECK = timezone.now()
 
 # The minimum amount of time in between each iteration of the main
 # loop. We could loop much less frequently than every two minutes if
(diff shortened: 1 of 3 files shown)


foreman/data_refinery_foreman/foreman/main.py
```diff
@@ -112,6 +111,25 @@ def handle_repeated_failure(job) -> None:
...
+    global TIME_OF_LAST_SIZE_CHECK
+
+    if (timezone.now() - TIME_OF_LAST_SIZE_CHECK > window):
+        # Assuming they're all similar to an X1, give 50 downloaders per node.
+        if not nomad_client:
+            nomad_host = get_env_variable("NOMAD_HOST")
+            nomad_port = get_env_variable("NOMAD_PORT", "4646")
+            nomad_client = Nomad(nomad_host, port=int(nomad_port), timeout=30)
+
+        MAX_TOTAL_DOWNLOADER_JOBS = len(nomad_client.nodes) * 50
```
kurtwheeler: Our cluster can have 10-26 nodes, which means this value will generally end up being 500-1300. Right now we have this value hardcoded to be 400 and our database has been sitting 90-98% CPU utilization. Before we deploy this, I think we should do a load test with the value at 1300 to make sure that our database isn't going to go down if we ever end up with 26 small nodes.

kurtwheeler: Ok based on our discussion, we're going to put a cap of 1000 no matter the cluster size. I guess the one last thing is that we should decrease the count of nodes by 1 because the smasher instance doesn't run downloader jobs.

PR #1188, opened 2019-03-28
Small Fixes

Miserlou:
Some small changes here, one fix of a broken TXI failure logger, the other making a very chatty janitor shut up.


workers/data_refinery_workers/processors/janitor.py
```diff
@@ -62,6 +62,8 @@ def _find_and_remove_expired_jobs(job_context):
                     # This job is running, don't delete  the working directory.
                     if job_status == "running":
                         continue
+                    if job_status == "pending":
+                        continue
```
kurtwheeler: Could this just be added to line 63 with an `or`?

PR #1190, opened 2019-03-29
Improves the way we count how many nodes are active. 

kurtwheeler:
## Issue Number

N/A some tweaks based off of prod performance.

I don't think the python nomad API was being used correctly. I'm not actually sure what our max downloader jobs limit was, but it seems like it somehow ended up being 1000 based off of how many jobs there were before any nodes came up, which was an issue caused by this bug.

This also makes it so that we start checking downloader jobs immediately rather than waiting two minutes.

This also decreases our main Foreman MIN_LOOP_TIME to 15 seconds from 2 minutes so we don't have downtime in between the foreman queuing work.

This also stops jobs operating on original_files that have been processed.

## Types of changes

## Functional tests

The unit tests cover this fairly well, I also looked at the actual data getting returned by the Nomad API calls to make sure my filtering was correct.

## Checklist

### DIFF
--- a/workers/data_refinery_workers/downloaders/utils.py
+++ b/workers/data_refinery_workers/downloaders/utils.py
@@ -109,6 +109,18 @@ def start_job(job_id: int, max_downloader_jobs_per_node=MAX_DOWNLOADER_JOBS_PER_
         logger.error("This downloader job has already been started!!!", downloader_job=job.id)
         raise Exception("downloaders.start_job called on a job that has already been started!")
 
+    for original_file in job.original_files.all():
+        if original_file.has_ever_been_processed():
+            logger.error(("Original file has a successful processor job, it doesn't need"
+                          " to be downloaded! Aborting!"),
+                         job_id=job.id,
+                         original_file_id=original_file.id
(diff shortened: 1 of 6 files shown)


workers/data_refinery_workers/downloaders/utils.py
```diff
@@ -109,17 +109,22 @@ def start_job(job_id: int, max_downloader_jobs_per_node=MAX_DOWNLOADER_JOBS_PER_
...
+        # SRA jobs only have one sample.
+        sample = job.orignal_files.first().samples().first():
+        for computed_file in sample.computed_files.all():
+            if compted_file.s3_bucket and compted_file.s3_key:
+                logger.error(("Sample has a good computed file, it must have been processed, "
+                              "so it doesn't need to be downloaded! Aborting!"),
+                             job_id=job.id,
+                             original_file_id=original_file.id
+                )
+                job_context["original_files"] = []
```
Miserlou: start_job does not have a job_context. The code above uses `sys.exit(0)` after setting job.success explicitly.

PR #1208, opened 2019-04-03
Force TXI downloads to local working directory

Miserlou:
## Issue Number
https://github.com/AlexsLemonade/refinebio/issues/1203

My theory is that these were being cleaned up early and causing TXI failures for larger jobs as the inputs were being cleaned up too early.

### DIFF
--- a/common/data_refinery_common/utils.py
+++ b/common/data_refinery_common/utils.py
@@ -293,29 +293,3 @@ def get_fasp_sra_download(run_accession: str):
             logger = logging.getLogger(__name__)
             logger.exception("Bad FASP CGI response", data=data, text=resp.text)
             return None
-
-def has_original_file_been_processed(original_file) -> bool:
-    """Returns True if original_file has been completely processed, returns False otherwise.
-    """
-    if not original_file:
-        return False
-
-    sample = original_file.samples.first()
-    if not sample:
-        return False
-
-    if sample.source_database == "SRA":
-        for computed_file in sample.computed_files.all():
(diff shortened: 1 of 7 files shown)


Miserlou:
Test failures in Illumina are unrelated.

workers/data_refinery_workers/processors/salmon.py
```diff
@@ -451,7 +451,8 @@ def _run_tximport_for_experiment(
     tximport_path_list_file = job_context["work_dir"] + "tximport_inputs.txt"
     with open(tximport_path_list_file, "w") as input_list:
         for quant_file in quant_files:
-            input_list.write(quant_file.get_synced_file_path() + "\n")
+            quant_work_path = job_context["work_dir"] + quant_file.filename
+            input_list.write(quant_file.get_synced_file_path(quant_work_path) + "\n")
```
kurtwheeler: For this and the other place you made an analogous change, I think it should be:
```
input_list.write(quant_file.get_synced_file_path(path=quant_work_path) + "\n")
```
since that function definition is:
```
def get_synced_file_path(self, force=False, path=None):
```

common/data_refinery_common/utils.py
```diff
@@ -295,12 +295,9 @@ def get_fasp_sra_download(run_accession: str):
             return None
 
 def has_original_file_been_processed(original_file) -> bool:
-    """Returns True if original_file is from SRA and has been processed
+    """Returns True if original_filehas been processed, returns False otherwise.
```
kurtwheeler: original_filehas

Issue #944, opened 2018-12-18
Lost Smasher Job // Smasher OOM


Miserlou:
### Context

While investigating other issues, I found that one of my Smasher jobs did not process, and it seems never got picked up.

https://api.refine.bio/dataset/b5e9e8c3-432f-4b17-8983-86be1dea338d/

### Problem or idea

It says `is_processing: true`, but I can find no record of it ever actually starting or running on the Smasher instance. I got no email. I think it was simply lost.

Should the Foreman be able to handle Smasher retries/lost jobs?

Miserlou:
Actually, smashing with a similar dataset, I was able to capture this:

```
ubuntu@ip-10-0-197-11:~$ nomad status 062b633b
ID                  = 062b633b
Eval ID             = d56b884e
Name                = SMASHER/dispatch-1545152062-d0a6dd60.jobs[0]
Node ID             = d20e2736
Job ID              = SMASHER/dispatch-1545152062-d0a6dd60
Job Version         = 0
Client Status       = failed
Client Description  = <none>
Desired Status      = run
Desired Description = <none>
Created             = 9m12s ago
Modified            = 16s ago

Task "smasher" is "dead"
Task Resources
CPU       Memory   Disk    IOPS  Addresses
1024 MHz  4.0 GiB  10 MiB  0

Task Events:
Started At     = 2018-12-18T16:55:34Z
Finished At    = 2018-12-18T17:03:18Z
Total Restarts = 0
Last Restart   = N/A

Recent Events:
Time                  Type            Description
2018-12-18T17:03:18Z  Not Restarting  Policy allows no restarts
2018-12-18T17:03:18Z  Terminated      Exit Code: 137, Exit Message: "OOM Killed"
2018-12-18T16:55:34Z  Started         Task started by client
2018-12-18T16:54:22Z  Driver          Downloading image ccdl/dr_smasher:v1.4.5
2018-12-18T16:54:22Z  Task Setup      Building Task Directory
2018-12-18T16:54:22Z  Received        Task received by client
```

Miserlou:
Appear to have a similar problem with `7816271c-eb32-4bbd-9b38-eb1cc8eb9775` - is `E-TABM-105` the problem?

Miserlou:
Semi related: https://github.com/AlexsLemonade/refinebio/pull/950

This one is an absolute bastard to track down.

Miserlou:
[...]

PR #1210, opened 2019-04-03
Add Smasher Support To The Foreman

Miserlou:
## Issue Number
https://github.com/AlexsLemonade/refinebio/issues/944

The foreman currently ignores lost smasher jobs. Somehow, Nomad can drop dispatched smasher jobs that get dispatched but never started. This hopefully adds support for those Smashers which get dropped.

### DIFF
--- a/foreman/data_refinery_foreman/foreman/main.py
+++ b/foreman/data_refinery_foreman/foreman/main.py
@@ -693,8 +693,7 @@ def handle_processor_jobs(jobs: List[ProcessorJob],
     jobs_dispatched = 0
     for count, job in enumerate(jobs):
 
-        if not ignore_ceiling:
-            if jobs_dispatched >= queue_capacity:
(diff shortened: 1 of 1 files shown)


foreman/data_refinery_foreman/foreman/main.py
```diff
@@ -691,9 +692,11 @@ def handle_processor_jobs(jobs: List[ProcessorJob],
 
     jobs_dispatched = 0
     for count, job in enumerate(jobs):
-        if jobs_dispatched >= queue_capacity:
-            logger.info("We hit the maximum total jobs ceiling, so we're not handling any more processor jobs now.")
-            return
+
+        if not ignore_ceiling:
+            if jobs_dispatched >= queue_capacity:
```
kurtwheeler: and we combine these two lines with `and`?

PR #1211, opened 2019-04-03
Missing library on timedelta call

Miserlou:
Trivial


Miserlou:
I thought main test would have caught this

PR #1225, opened 2019-04-09
Lower Smasher Instance Class, Raise Smasher RAM Usage

Miserlou:
## Issue Number
https://github.com/AlexsLemonade/refinebio/issues/944

@cansav09 got a smasher OOM yesterday, 944 wasn't fully resolved for very large datasets.

I'm not sure how it happened (must have been during compendia processing time), but the smasher instance was way way too powerful, this lowers it down to something more sane but raises the RAM usage of each individual job.

### DIFF
--- a/workers/nomad-job-specs/smasher.nomad.tpl
+++ b/workers/nomad-job-specs/smasher.nomad.tpl
@@ -65,9 +65,9 @@ job "SMASHER" {
       # The resources the job will require.
       resources {
         # CPU is in AWS's CPU units.
-        cpu = 1024
+        cpu = 2048
         # Memory is in MB of RAM.
-        memory = 12288
+        memory = 24576
       }
 
       logs {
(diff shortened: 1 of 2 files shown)


workers/nomad-job-specs/smasher.nomad.tpl
```diff
@@ -65,9 +65,9 @@ job "SMASHER" {
       # The resources the job will require.
       resources {
         # CPU is in AWS's CPU units.
-        cpu = 1024
+        cpu = 2048
         # Memory is in MB of RAM.
-        memory = 12288
+        memory = 24576
```
kurtwheeler: This box has 32GB of RAM and it only runs these jobs. Is there any reason to not put this number closer to the total RAM for the instance? With these changes we can't run more than one job at a time anyway.

Miserlou: We can raise it a little bit, but there should be some overhead for the nomad client

PR #1230, opened 2019-04-10
Better handle og_files without associated samples 

Miserlou:
Suppresses some sporadic nastiness for the requeueing original files without associated samples - but why do we have those? This will help us diagnose.

### DIFF
--- a/foreman/data_refinery_foreman/foreman/main.py
+++ b/foreman/data_refinery_foreman/foreman/main.py
@@ -329,6 +329,7 @@ def requeue_downloader_job(last_job: DownloaderJob) -> bool:
         elif ram_amount == 4096:
             ram_amount = 8192
 
+
     original_file = last_job.original_files.first()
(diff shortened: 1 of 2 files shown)


foreman/data_refinery_foreman/foreman/main.py
```diff
@@ -329,20 +329,28 @@ def requeue_downloader_job(last_job: DownloaderJob) -> bool:
...
         return False
 
-    first_sample = original_file.samples.first()
-    if first_sample and first_sample.is_blacklisted:
-        last_job.no_retry = True
-        last_job.failure_reason = "Sample run accession has been blacklisted by SRA."
-        last_job.save()
-        logger.info("Avoiding requeuing for DownloaderJob for blacklisted run accession: " + str(first_sample.accession_code))
-        return False
+    if original_file.samples:
```
kurtwheeler: I'm pretty sure this returns a Manager, so it will never be None.

PR #1233, opened 2019-04-11
Reduce jobs/node

Miserlou:
SRA capacity seems greatly reduced. We need to re-lower our DLJ/Node until it's capable again.


kurtwheeler:
It seems more appropriate to reduce our capacity rather than make this change. With only 5 downloader jobs running on a node at a time, I don't think we'll get enough data to properly utilize our resources.

Miserlou:
Yeaaahhhh, that's probably true. Hm.

Miserlou:
We confirmed with SRA that limits are on a per machine basis and are this low anyway. The number of jobs/node is kinda of independent than our total computing resources.

PR #1242, opened 2019-04-15
Cut EBS volume costs

kurtwheeler:
## Issue Number

N/A came up while thinking about what to do to make prod run at maximum efficiency.

This drops the two smallest instance types so we don't need so many volumes. Our smallest weight was 4 before which meant we could have 25 instances at once (with max capacity 100).

By dropping instances smaller than weight 8, we can't have more than 12 instances as long as we keep capacity at 96 or lower. Therefore the env var for prod MAX_CLIENTS has been changed to 12.

Also we had 10 staging volumes, which may have been necessary before, but now we're definitely fine with just 2. Therefore the env var for prod MAX_CLIENTS has been changed to 2.

Overall, this means that we're going from 30 9TB volumes to only 14, which means we should have about a 53% reduction in costs for EBS volumes. The downside to this is that we could more frequently have periods of downtime where we cannot get spot instance capacity.

This PR also drops the max downloader jobs per node value down to 10 because we're seeing our volumes starting to get full.

## Types of changes

- Operational tweak

## Functional tests

N/A this is infrastructure related

## Checklist


Miserlou:
The idea is good, this needs to be tested from a high number of staging volumes to the new low number as I'm not sure how Terraform will handle the destruction of data.

kurtwheeler:
I agree, I very much hope only some are destroyed, not that all are destroyed and the new number created. Hard to say what will happen before staging test though.

Issue #1256, opened 2019-04-16
No * IAM Permissions


cgreene:
### Context

Some of our IAM permissions still have wildcards.

### Problem or idea

We shouldn't have permissions with wildcards. These should be fully specified.

### Solution or next step

Revise permissions.

PR #1259, opened 2019-04-16
IAM Lockdown

Miserlou:
## Issue Number
https://github.com/AlexsLemonade/refinebio/issues/1256

AWS resource permissions lockdown.

### DIFF
--- a/infrastructure/permissions.tf
+++ b/infrastructure/permissions.tf
@@ -159,21 +159,7 @@ resource "aws_iam_policy" "cloudwatch_policy" {
                 "logs:CreateLogGroup",
                 "logs:CreateLogStream",
                 "logs:PutLogEvents",
-                "logs:DescribeLogStreams"
-            ],
-            "Resource": [
-              "arn:aws:logs:${var.region}:${aws_cloudwatch_log_group.data_refinery_log_group.name}:${aws_cloudwatch_log_stream.log_stream_surveyor.name}",
-              "arn:aws:logs:${var.region}:${aws_cloudwatch_log_group.data_refinery_log_group.name}:${aws_cloudwatch_log_stream.log_stream_processor.name}",
-              "arn:aws:logs:${var.region}:${aws_cloudwatch_log_group.data_refinery_log_group.name}:${aws_cloudwatch_log_stream.log_stream_downloader.name}",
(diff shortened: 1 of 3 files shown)


Miserlou:
This is pretty much ready for first review. I don't like the manual listing of the EBS volumes, as it means that it has to be kept in sync with `max_clients`, but TF won't have loops until 0.12 and they're still on the 0.11 release, despite it being a year overdue.

infrastructure/permissions.tf
```diff
@@ -156,8 +180,15 @@ resource "aws_iam_policy" "cloudwatch_policy" {
...
-            "Resource":
-                "*"
+            "Resource": [
+              "arn:aws:logs:${var.region}:${aws_cloudwatch_log_group.data_refinery_log_group.name}:${aws_cloudwatch_log_stream.log_stream_surveyor.name}",
+              "arn:aws:logs:${var.region}:${aws_cloudwatch_log_group.data_refinery_log_group.name}:${aws_cloudwatch_log_stream.log_stream_processor.name}",
+              "arn:aws:logs:${var.region}:${aws_cloudwatch_log_group.data_refinery_log_group.name}:${aws_cloudwatch_log_stream.log_stream_downloader.name}",
+              "arn:aws:logs:${var.region}:${aws_cloudwatch_log_group.data_refinery_log_group.name}:${aws_cloudwatch_log_stream.log_stream_foreman.name}",
+              "arn:aws:logs:${var.region}:${aws_cloudwatch_log_group.data_refinery_log_group.name}:${aws_cloudwatch_log_stream.log_stream_api.name}",
+              "arn:aws:logs:${var.region}:${aws_cloudwatch_log_group.data_refinery_log_group.name}:${aws_cloudwatch_log_stream.log_stream_api_nginx_access.name}",
+              "arn:aws:logs:${var.region}:${aws_cloudwatch_log_group.data_refinery_log_group.name}:${aws_cloudwatch_log_stream.log_stream_api_nginx_error.name}"
```
kurtwheeler: Could this block be combined with the one above it? It looks like they both affect the same resources with the same effect.

Miserlou: Good eye! Clears up some bytes too.

PR #1261, opened 2019-04-17
Bash variables should be in double quotes, now they all are! 

kurtwheeler:
## Issue Number

N/A came up while @jaclyn-taroni  was trying to test things for me

This doesn't touch anything in `infrastructure/` because it's not worth messing with that stuff over something like this. Also it shouldn't impact outside developers' experiences because they won't be deploying.

## Types of changes

## Functional tests

Besides unit tests and a staging test, @jaclyn-taroni will be running some Salmon jobs locally with this branch before it gets merged.

## Checklist


kurtwheeler:
The perl template magic happens when running locally and for end to end tests so I think it should be well covered.

jaclyn-taroni:
This works on my computer.

PR #1265, opened 2019-04-19
Parallel QN Dispatcher

Miserlou:
Provides a more convenient way to re-gen QN targets in parallel. Testing on a dev stack now.


workers/data_refinery_workers/processors/management/commands/create_qn_target.py
```diff
@@ -50,71 +50,80 @@ def add_arguments(self, parser):
...
+
+                dataset = Dataset()
+                dataset.data = {organism.name + '_(' + biggest_platform + ')': sample_codes}
+                dataset.aggregate_by = "ALL"
+                dataset.scale_by = "NONE"
+                dataset.quantile_normalize = False
+                dataset.save()
+
+                job = ProcessorJob()
+                job.pipeline_applied = "QN_REFERENCE"
```
kurtwheeler: Should this not be `RUN_QN_JOB` or ProcessorPipeline.QN_REFERENCE.value?

workers/nomad-job-specs/qn_dispatcher.nomad.tpl
```diff
@@ -0,0 +1,95 @@
...
+        max_files = 1
+        max_file_size = 1
+      }
+
+      # Don't run on the smasher instance, it's too small and should be running smasher jobs.
+      constraint {
+        attribute = "${meta.is_smasher}"
+        operator = "!="
+        value = "true"
+      }
```
kurtwheeler: This constraint should be on the the other job spec as well right?

workers/data_refinery_workers/processors/management/commands/qn_dispatcher.py
```diff
@@ -0,0 +1,86 @@
+"""This command will create and run survey jobs for each experiment
+in the experiment_list. experiment list should be a file containing
+one experiment accession code per line.
+"""
```
kurtwheeler: This docstring isn't accurate.

workers/data_refinery_workers/processors/management/commands/qn_dispatcher.py
```diff
@@ -0,0 +1,86 @@
...
+
+class Command(BaseCommand):
+
+    def handle(self, *args, **options):
+        """ Handle it! """
+
+        organisms = Organism.objects.all()
+
+        for organism in organisms:
+            samples = Sample.processed_objects.filter(organism=organism, has_raw=True, technology="MICROARRAY", is_processed=True)
```
kurtwheeler: We should probably also filter on `is_public` since we use that as a poor man's `is_deleted` right?

workers/data_refinery_workers/processors/management/commands/qn_dispatcher.py
```diff
@@ -0,0 +1,86 @@
...
+            if samples.count() < MIN:
+                logger.error("Proccessed samples don't meet minimum threshhold",
```
kurtwheeler: I don't think this is an error. There's going to be plenty of organisms that don't meet the threshold and that's expected. I don't mind logging about it because it could be handle to have it spit out Yea or Nay for every organism, but it's not an error.

workers/data_refinery_workers/processors/management/commands/qn_dispatcher.py
```diff
@@ -0,0 +1,86 @@
...
+                    organism=organism,
+                    count=samples.count(),
+                    min=MIN
+                )
+                continue
+
+            platform_counts = samples.values('platform_accession_code').annotate(dcount=Count('platform_accession_code')).order_by('-dcount')
+            biggest_platform = platform_counts[0]['platform_accession_code']
```
kurtwheeler: This biggest_platform may have less than 100 samples. We should decide whether or not to make a qn target based on each organism's largest platform, not their full sample count.

workers/data_refinery_workers/processors/management/commands/qn_dispatcher.py
Miserlou: This is already taken care of by the `processed_objects` manager

PR #1234, opened 2019-04-11
Kurtwheeler/fix geo recreation

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/1068

## Purpose/Implementation Notes

This fixes recreation of GEO downloader jobs along with fixing some other issues identified along the way.

## Types of changes

- Bugfix (non-breaking change which fixes an issue)

## Functional tests

The end to end tests are covering this well. If they are not sufficient I should add more.

## Checklist

- [x] Lint and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)
- [x] Any dependent changes have been merged and published in downstream modules

PR #1266, opened 2019-04-19
Disable AWS logging for workers and set log level to DEBUG.

kurtwheeler:
## Issue Number

#1234 

The staging test isn't working at all and I have no idea why. All the jobs that I think should recreate their downloader jobs instead exit with status code 1 and I have no idea why. There's no logs about it, there doesn't seem to be any way we could exit with status code 1 unless we gave an invalid job name or id, and I am out of non-nuclear options to try to debug this.

Therefore I'm setting the log level to DEBUG and keeping the logs in nomad rather than having them go off into awslogs where they'll get  so intermingled it'll be hard to decipher given the huge volume I'm expecting. This will also help us not generate too too many cloudwatch log entries.

## Types of changes

- Debugging


format_nomad_with_env.sh
```diff
@@ -171,6 +171,12 @@ else
     environment_file="$script_directory/infrastructure/prod_env"
 fi
 
+# Temporarily set the logging config to nothing so we can see the logs
+# through nomad.
+if [[ $env == "staging" ]]; then
+    export LOGGING_CONFIG=""
```
Miserlou: I assume blank defaults to DEBUG

kurtwheeler: No, this is about where to send the logs. It defaults to using the docker driver to send them to Nomad, which we normally overwrite to send them to cloudwatch.

kurtwheeler:
One last note about the problem this is to diagnose, I tried surveying a completely new experiment to see if freshly generated processor jobs would run and they all failed with status code 1 just the same.

PR #1270, opened 2019-04-23
No more direct policy attachments

Miserlou:
The IAM role policy limit is not any individual policy, but the _total_ bytes. Annoying!

The limit is higher for attached policies, so let's try it that way.


Miserlou:
Took a few tries and found another bug along the way, but it's working solidly on a dev stack now. Log streams need to be dynamically created.

PR #1278, opened 2019-04-25
Fix and improve logging for the cleanup_the_queue function in the Foreman

kurtwheeler:
## Issue Number

N/A came up in prod

We didn't have any capacity all night but our nomad queue had a depth of >9k this morning. This fixes that and gives us a little more logging so we know what's going on with it in the future.

## Types of changes

## Functional tests

I should add some for this but right now I want to get MA running again.

### DIFF
--- a/foreman/data_refinery_foreman/foreman/main.py
+++ b/foreman/data_refinery_foreman/foreman/main.py
@@ -1304,6 +1304,8 @@ def cleanup_the_queue():
     Therefore we clear out jobs of that type every once in a while so
     our queue is dedicated to jobs that can actually be placed.
     """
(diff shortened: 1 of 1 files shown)


foreman/data_refinery_foreman/foreman/main.py
```diff
@@ -1349,10 +1354,13 @@ def cleanup_the_queue():
                     job_record = ProcessorJob(nomad_job_id=job["ID"])
                     job_record.num_retries = job_record.num_retries - 1
                     job_record.save()
+                    num_jobs_killed += 1
                 except:
                     # If we can't do this for some reason, we'll get it next loop.
```
Miserlou: Logging here plz

PR #1281, opened 2019-04-25
Fixes the retry_lost_downloader_jobs function so it can't queue too many

kurtwheeler:
## Issue Number

N/A Discovered in prod

We were seeing a lot of downloader jobs getting queued... way past our cap. I think this is why and this should fix that!

## Types of changes

### DIFF
--- a/foreman/data_refinery_foreman/foreman/main.py
+++ b/foreman/data_refinery_foreman/foreman/main.py
@@ -599,6 +599,7 @@ def retry_lost_downloader_jobs() -> None:
     page_count = 0
     while queue_capacity > 0:
         lost_jobs = []
+        jobs_queued_from_this_page = 0
         for job in page.object_list:
             try:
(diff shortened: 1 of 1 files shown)


foreman/data_refinery_foreman/foreman/main.py
```diff
@@ -613,7 +614,7 @@ def retry_lost_downloader_jobs() -> None:
                                      job_id=job.id
                         )
                         lost_jobs.append(job)
-                else:
+                elif jobs_queued_from_this_page < queue_capacity:
                     # The job never got put in the Nomad queue, no
                     # need to recreate it, we just gotta queue it up!
                     send_job(Downloaders[job.downloader_task], job=job, is_dispatch=True)
```
Miserlou: Am I reading this wrong or should their be an incrementor here?

PR #1318, opened 2019-05-28
Fix aws permissions

arielsvn:
Aws policies were failing when `max_clients=0`.

### DIFF
--- a/infrastructure/deploy.sh
+++ b/infrastructure/deploy.sh
@@ -181,7 +181,7 @@ echo "Confirming Nomad cluster.."
 start_time=$(date +%s)
 diff=0
 nomad_status=$(check_nomad_status)
-while [ "$diff" -lt "900" -a $nomad_status != "200" ]; do
+while [ "$diff" -lt "1200" -a $nomad_status != "200" ]; do
     sleep 1
     nomad_status=$(check_nomad_status)
     let "diff = $(date +%s) - $start_time"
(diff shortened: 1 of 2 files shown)


infrastructure/deploy.sh
```diff
@@ -181,7 +181,7 @@ echo "Confirming Nomad cluster.."
 start_time=$(date +%s)
 diff=0
 nomad_status=$(check_nomad_status)
-while [ "$diff" -lt "900" -a $nomad_status != "200" ]; do
+while [ "$diff" -lt "1200" -a $nomad_status != "200" ]; do
```
kurtwheeler: Can you revert this change since it ended up being something different?

infrastructure/permissions.tf
```diff
@@ -50,12 +50,12 @@ resource "aws_iam_policy_attachment" "fleet_role" {
...
 
 resource "aws_iam_instance_profile" "data_refinery_instance_profile" {
-  name  = "data-refinery-instance-profile-${var.user}-${var.stage}"
+  name = "data-refinery-instance-profile-${var.user}-${var.stage}"
   role = "${aws_iam_role.data_refinery_instance.name}"
 }
 
 resource "aws_iam_policy" "s3_access_policy" {
-  name = "data-refinery-s3-access-policy-${var.user}-${var.stage}"
+  name        = "data-refinery-s3-access-policy-${var.user}-${var.stage}"
```
kurtwheeler: Waiiiit noo, I hate this style and want to try to get more standard around doing it without unnecessary spaces in the middle of our lines. There's a few places where we do this cause the code got copied from elsewhere without getting fixed but we should fix that stuff not change everything else.

arielsvn:
will send changes in another PR

Issue #1317, opened 2019-05-28
RNA-Seq samples with unmated reads cause salmon to hang forever


kurtwheeler:
### Context

This issue is mostly solved but I thought it'd be good to have something to track it anyway.

### Problem or idea

Use the --split-3 option to filter out the unmated reads.

### Solution or next step

Make sure to test samples that have and do not have unmated reads.

PR #1321, opened 2019-05-29
Kurtwheeler/fastq dump split 3

kurtwheeler:
## Issue Number

#1317 

This PR attempts to use the `--split-3` option of fastq-dump to filter out unmated reads so that salmon doesn't hang. I think it does it fairly well, however there is an issue in that sometimes one pipe fills up, causing fastq-dump and then accordingly salmon to hang. One of the samples used in the salmon tests triggers this error. I think the thing that needs to happen to make this work generally is to have much larger buffers in between fastq-dump and salmon.

I tried to use `fcntl` to do this but examples such as https://www.golinuxhub.com/2018/05/how-to-view-and-increase-default-pipe-size-buffer.html ended up hanging because opening a named pipe causes the program to hang until something is doing the opposite (for example if you open it for reading the program will hang until something starts writing to it). I'm not sure if this is a python issue or not, trying to do this in C is probably a decent next step.

Other next steps that could be pursued would be to use a message queue or an in-memory database like rabbitmq or redis. Additionally there could be a hack that would be simpler where we link a whole bunch of smaller pipes together to combine the size of their buffers together.

## Methods

The scientific implications of this PR are that for samples with unmated reads, we will throw those unmated reads away.

## Types of changes

## Functional tests

I've chosen samples for the unit tests that cover the cases that we need to: a sample with unmated reads and one without.

## Checklist

### DIFF
--- a/foreman/data_refinery_foreman/surveyor/test_end_to_end.py
+++ b/foreman/data_refinery_foreman/surveyor/test_end_to_end.py
@@ -575,7 +575,10 @@ class SraRedownloadingTestCase(TransactionTestCase):
             organism = Organism(name="HOMO_SAPIENS", taxonomy_id=9606, is_scientific_name=True)
             organism.save()
 
-            survey_job = surveyor.survey_experiment("SRP040623", "SRA")
+            # survey_job = surveyor.survey_experiment("SRP040623", "SRA")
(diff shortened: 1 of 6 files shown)


kurtwheeler:
I'm going to close this because we're pursuing the tmpfs option since this didn't work for many samples (where the reads weren't interleaved)

PR #1327, opened 2019-05-31
[WIP] Switch the SRA downloader over to use HTTPS instead of aspera.

kurtwheeler:
## Issue Number

N/A came out of a discussion with SRA Chris

This is likely to not pass tests because I haven't touched the tests yet :(

However this does in fact use HTTP to download from SRA instead of Aspera. This is because Chris said that they had to tweak their aspera servers a lot to get the uploads to Google Cloud and AWS and so those servers are not currently working.

## Types of changes

## Functional tests

I downloaded some data with this switch.

## Checklist

- [] Lint and unit tests pass locally with my changes

### DIFF
--- a/common/data_refinery_common/utils.py
+++ b/common/data_refinery_common/utils.py
@@ -265,22 +265,22 @@ def calculate_sha1(absolute_file_path):
 
     return hash_object.hexdigest()
 
-def get_fasp_sra_download(run_accession: str):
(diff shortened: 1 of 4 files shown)


kurtwheeler:
@arielsvn I think this is very close. I hope that at this point I've fixed the tests but if not then it should only be minor changes to do so from here on out. I'd like to try to deploy this and #1326 today if possible.

PR #1329, opened 2019-06-04
Deploy S3 bucket lockdown, change SRA downloads to HTTP

kurtwheeler:
Brings master up to date with dev. The commit log has further detail.


kurtwheeler:
I'll note that the staging test for the QN buckets didn't quite work, see #1328. However, I don't think that needs to hold things up. We can redeploy once we retest that change, until then we just won't create more QN targets.

PR #1334, opened 2019-06-05
[WIP] Management command to retry all unprocessed samples from a source database

kurtwheeler:
## Issue Number

N/A came up while trying to get us back up and running. 

The organism shepherd is great and all, but now that we expect to have access to all of SRA it seems to make sense to me that we should just try queuing a fresh downloader job for every sample!

## Types of changes

## Functional tests

This is the thing I'm currently hung up on, hence the WIP. Recreating enough failed jobs in a dev or local stack won't be too easy.


kurtwheeler:
@arielsvn I've tested this locally, seems to work well. I think I'd like to try and hotfix this out so I can use this command to requeue work because there are a lot jobs failing with:
```
Missing file for processor job but unable to recreate downloader jobs!
```
because the job was recreated in the past.

I want to be able to queue up everything that hasn't been processed yet.

foreman/data_refinery_foreman/foreman/management/commands/retry_samples.py
```diff
@@ -0,0 +1,70 @@
...
+                    if create_downloader_job(sample.original_files.all(), force=True):
+                        creation_count += 1
+
+            logger.info(
+                "Created %d new downloader jobs because their samples lacked computed files.",
+                creation_count
+            )
+
+            if not page.has_next():
+                break
```
arielsvn: Does this only works on the first page? or I'm missing something?

kurtwheeler: Good catch! It's addressed now.

PR #1347, opened 2019-06-20
Fix a syntax error in the smasher.

kurtwheeler:
## Issue Number

N/A Came up while working on unrelated tests

This is a bad one. The smasher's code had a syntax error so all the smasher jobs failed. We didn't even get notified about it.

## Types of changes

## Functional tests

N/A but the tests are actually running now.

## Checklist


kurtwheeler:
Yes but the tests should have caught it for you. I made #1348 to try to prevent this from happening in the future.

Issue #542, opened 2018-08-21
Get SRA sample metadata from BioSample database


jaclyn-taroni:
### Context

Yesterday, we were taking a look at some zebrafish crunch RNA-seq samples.

### Problem or idea

To quote from the MetaSRA publication ([Bernstein, et al. _Bioinformatics._ 2017.](
https://doi.org/10.1093/bioinformatics/btx334))

> Unfortunately, [large-scale] studies are currently difficult due to the poor structure of the metadata associated with samples in the SRA. These metadata are centrally stored at the BioSample database ([Barrett et al., 2012](https://doi.org/10.1093/nar/gkr1163)), with each SRA sequencing experiment referencing a sample-specific metadata record within this database. The BioSample database organizes each sample’s metadata as sets of key-value pairs where the key is a property of the sample and the value is a property-value of the sample. 

We are not currently extracting metadata from the BioSample database ([documentation](https://www.ncbi.nlm.nih.gov/biosample/docs/); unclear how useful that will be). I've also found this: https://catalog.data.gov/dataset/biosample/resource/61cb64a9-14c1-4216-8af9-4a4d68286ef1 which seems a bit worrisome.

### Solution or next step

We should circle back to this following the initial release. [MetaSRA](http://metasra.biostat.wisc.edu) uses [SRAdb](https://github.com/seandavi/SRAdb), so I'm linking those here for reference. 

EMBL-EBI also has a [BioSamples database](https://www.ebi.ac.uk/biosamples/) which seems like it contains ENA data and has [API documentation](https://www.ebi.ac.uk/biosamples/docs/references/api), so if that covers all of SRA that may be an avenue to explore.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

cansav09:
### Examples of BioSample documentation vs what is in Refine.bio

Most RNA-seq datasets appear to only have sample titles. BioSample has more information. If we want this data to be readily useable by the users, we really should try to get the BioSample metadata for RNA-seq samples. 

- Example: SRP066834
[refinebio entry](https://www.refine.bio/experiments/SRP066834/human-cerebral-organoids-recapitulate-gene-expression-programs-of-fetal-neocortex-development)
[...]

PR #1350, opened 2019-06-25
Allow SRA metadata to be processed, and test the harmonization

wvauclain:
## Issue Number

#542 

A bug was preventing us from getting metadata from SRA samples. I fixed the bug and added a test for harmonized SRA metadata.

## Types of changes

## Functional tests

https://github.com/AlexsLemonade/refinebio/issues/542#issuecomment-505498535

I also wrote a new test to verify that SRA metadata is harmonized.

## Checklist

### DIFF
--- a/foreman/data_refinery_foreman/surveyor/sra.py
+++ b/foreman/data_refinery_foreman/surveyor/sra.py
@@ -474,7 +474,6 @@ class SraSurveyor(ExternalSourceSurveyor):
             else:
                 sample_object.manufacturer = "UNKNOWN"
 
-            # Directly apply the harmonized values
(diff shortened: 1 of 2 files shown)


foreman/data_refinery_foreman/surveyor/sra.py
```diff
@@ -467,10 +475,7 @@ def _generate_experiment_and_samples(self, run_accession: str, study_accession:
                 sample_object.manufacturer = "UNKNOWN"
 
             # Directly apply the harmonized values
```
kurtwheeler: I think we don't need this comment any more since it's essentially in the function name/docstring.

PR #1355, opened 2019-06-26
Created a management command to refresh SRA metadata

### LINKED ISSUE #542: Get SRA sample metadata from BioSample database

wvauclain:
## Issue Number

Fixes #542 

I wrote a management command to refresh the metadata for all RNA-Seq samples. This will allow us to get the missing metadata for all pre-existing RNA-Seq samples once #1350 lands.

## Types of changes

## Functional tests

I commented out the fix in #1350, then downloaded some samples and confirmed that they were missing metadata. After running the management command, the samples were no longer missing metadata.

## Checklist

## Screenshots


foreman/data_refinery_foreman/foreman/management/commands/update_sra_metadata.py
```diff
@@ -0,0 +1,39 @@
...
+            source_database = options["source_database"]
+            sra_samples = Sample.objects.filter(
+                source_database=source_database
+            )
+
+        for sample in sra_samples:
+            logger.debug("Refreshing metadata for a sample.",
+                         sample=sample.accession_code)
+            metadata = SraSurveyor.gather_all_metadata(sample.accession_code)
+            SraSurveyor._apply_harmonized_metadata_to_sample(sample, metadata)
```
kurtwheeler: I think we need to either make this management command specific to SRA or change this code to run other surveyors. As it is now, I could pass `ARRAY_EXPRESS` in as the `source_database` parameter and this would try to resurvey SRA for data about an object that SRA doesn't contain.

It could be useful one day to be able to resurvey other data sources without creating another command so maybe it would be best to switch the surveyor based on the `source-database` parameter.

Issue #1351, opened 2019-06-25
Increase POSIX-compliance of shell scripts


wvauclain:
### Context

I tried to run `source common.sh && export NOMAD_ADDR=http://$(get_ip_address):4646` on my laptop, which runs `zsh`, and I got an error that was indecipherable, but turned out to be a syntax error because a condition used `==` instead of `=`.

### Problem or idea

Some scripts (especially `common.sh`) are not technically POSIX-compliant. This causes issues when using a shell like `zsh`, because `bash` and `zsh` can have slightly different syntax additions for shell scripting.

### Solution or next step

I am going to edit all shell scripts so they execute using `/bin/sh`. I will then use `shellcheck` to check POSIX-compliance, and execute the scripts using `dash` to check for problems on Ubuntu, where `dash` is symlinked to `/bin/sh`.

wvauclain:
This was the error when running the command in `zsh`.

```
$ source common.sh && export NOMAD_ADDR=http://$(get_ip_address):4646
get_ip_address:1: = not found
```

wvauclain:
It looks like some scripts should stay in `bash`, especially those that deal with arrays, but I will still try to convert as many of them as possible to `sh`, especially if they are ever sourced.

Issue #714, opened 2018-10-08
Create install_all.sh script


kurtwheeler:
### Context

Right now we have a decently long getting started section. It's broken up like this so that we don't make a ton of changes to the user's computer without their knowledge.

### Problem or idea

Everything we instruct the user to do could be done by a single script. We should offer this as an alternative to walking through our guide. This also will give us the ability to test that our system runs after following our instructions exactly.

### Solution or next step

Create an `install_all.sh` script that will detect whether it's installing onto Mac or Ubuntu, install any missing dependencies, start the database, build local images, get Nomad running, and anything else that should be done before a developer can run the entire system locally.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

Issue #713, opened 2018-10-08
Reorganize and standardize scripts


kurtwheeler:
### Context

Right now we have scripts scattered throughout the project. I know there's scripts in the top level directory, `infrastructure/`, and `common/`. The arguments they accept are not as consistent as they could be and sometimes they use environment variables instead of arguments.

### Problem or idea

I think it's time we created a `scripts/` directory so all our scripts are in a predictable location. I also think we should make sure no scripts assume anything about the environment variables of the environment they are called in. 

### Solution or next step

If a script uses an environment variable, it should instead take that in as an argument. Also consolidate the scripts into a single directory.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

PR #1357, opened 2019-06-27
Clean Up All User/Developer-facing Scripts

wvauclain:
## Issue Number

#1351 #714 #713 

I refactored all user-facing shell scripts into either POSIX-compliant `.sh` scripts or Bash `.bash` scripts using `shellcheck` and `dash`. Then I created an `install_all.sh` script to automatically install all of the development dependencies. I also moved all of the scripts into a `scripts/` directory, and changed all the references to those scripts.

## Types of changes

## Functional tests

I `grep`ed locally to make sure that all references to each script was updated, then I ran each modified script locally to make sure that they work.

## Checklist

## Screenshots


wvauclain:
I still need to test this on macOS before we merge.

scripts/format_nomad_with_env.sh
```diff
@@ -231,97 +238,101 @@ export_log_conf (){
...
             done
         fi
     done
-elif [[ "$project" == "surveyor" ]]; then
+elif [ "$project" = "surveyor" ]; then
 
     # Iterate over all the template files in the directory.
-    for template in $(ls -1 nomad-job-specs | grep \.tpl); do
+    for template in nomad-job-specs/*.tpl; do
+	template="$(basename "$template")"
```
kurtwheeler: I have the same question as above about this.

scripts/install_all.sh
```diff
@@ -0,0 +1,193 @@
...
+
+# Config variables
+TERRAFORM_VERSION="0.12.3"
+
+# This script should always run as if it were being called from
+# the directory it lives in.
+script_directory="$(perl -e 'use File::Basename;
+ use Cwd "abs_path";
+ print dirname(abs_path(@ARGV[0]));' -- "$0")"
+cd "$script_directory" || exit
```
kurtwheeler: This is repeated.

scripts/install_all.sh
```diff
@@ -0,0 +1,193 @@
...
+	v)
+	    OUTPUT="/dev/stdout"
+	    ;;
+        \?)
+            echo "Invalid option: -$OPTARG" >&2
+            print_usage >&2
+            exit 1
+            ;;
+        :)
+            echo "Option -$OPTARG requires an argument." >&2
```
kurtwheeler: None of the options to this script require an argument do they?

scripts/install_all.sh
```diff
@@ -0,0 +1,193 @@
...
+echo "Starting postgres and installing the database..."
+./run_postgres.sh > $OUTPUT
+./install_db_docker.sh > $OUTPUT
+
+echo "Starting elasticsearch and building the ES Indexes..."
+./run_es.sh > $OUTPUT
+./run_manage.sh search_index --rebuild -f > $OUTPUT
+
+echo "Updating common dependencies..."
+./update_models.sh > $OUTPUT
```
kurtwheeler: Doesn't this script require the virtualenv to be active?

README.md
```diff
@@ -296,12 +306,17 @@ linter within the project's directory tree, it will enforce a line length limit
 of 100 instead of 80. This will also be true for editors which rely on either
 linter.
 
+All user-facing scripts have been linted with `shellcheck` for common
```
kurtwheeler: Can you describe here how to run this linter so that as people change scripts they can re-lint them?

workers/run_command.sh
```diff
@@ -45,13 +48,13 @@ if [ ! -d "$volume_directory" ]; then
...
 
-source common.sh
+. ./scripts/common.sh
 HOST_IP=$(get_ip_address)
 DB_HOST_IP=$(get_docker_db_ip_address)
 
 chmod -R a+rwX "$volume_directory"
 
-./prepare_image.sh -i "$image" -s workers
+./scripts/prepare_image.sh -i "$image" -s workers
```
kurtwheeler: I know it's not part of this PR but it looks like I accidentally committed some specific command stuff in the bottom of this script. Could you change it so that instead of running that `Rscript` command it just runs whatever was passed in as arguments to the script?

scripts/format_nomad_with_env.sh
```diff
@@ -231,97 +238,101 @@ export_log_conf (){
 
 # This actually performs the templating using Perl's regex engine.
 # Perl magic found here: https://stackoverflow.com/a/2916159/6095378
-if [[ "$project" == "workers" ]]; then
+if [ "$project" = "workers" ]; then
     # Iterate over all the template files in the directory.
-    for template in $(ls -1 nomad-job-specs | grep \.tpl); do
+    for template in nomad-job-specs/*.tpl; do
+	template="$(basename "$template")"
```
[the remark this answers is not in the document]
wvauclain: Shellcheck lints against the old for loop. Piping `ls` into `grep` can get fragile when filenames are not alphanumeric, especially if filenames contain spaces or newlines.

Source: https://github.com/koalaman/shellcheck/wiki/SC2010

scripts/format_nomad_with_env.sh
```diff
@@ -24,124 +24,130 @@ print_options() {
...
-export NOMAD_HOST_IP=$(get_ip_address)
 
-if [ "$env" == "test" ]; then
-    export VOLUME_DIR="$script_directory/test_volume"
+# shellcheck disable=SC1091
+. ../scripts/common.sh
+DB_HOST_IP=$(get_docker_db_ip_address)
+export DB_HOST_IP
+NOMAD_HOST_IP=$(get_ip_address)
+export NOMAD_HOST_IP
```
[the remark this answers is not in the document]
wvauclain: This is another shellcheck linting error. Apparently the reason for this is that when you assign and export in one line, the return code from the command execution is ignored. If we had `set -e` on and `$(get_ip_address)` failed for some reason, this script would exit properly but if export was on the same line it wouldn't.

Source: https://github.com/koalaman/shellcheck/wiki/SC2155

scripts/install_all.sh
wvauclain: I think these both run inside docker, so they don't need the virtualenv.

PR #1360, opened 2019-07-01
Remove num_downloadable migration and readd it as a management command

kurtwheeler:
## Issue Number

N/A came up during a deploy because the migration took too long

This migration takes a really long time. Our deploys wait for migrations to complete before continuing, so this was holding everything up. By moving this to a management command, I can just run it manually after the deploy completes successfully.

## Types of changes

## Functional tests

N/A, just moving code around.

## Checklist


foreman/data_refinery_foreman/foreman/management/commands/update_downloadable_samples.py
```diff
@@ -0,0 +1,9 @@
+import django.contrib.postgres.fields
+from data_refinery_common.models import Experiment, ComputationalResultAnnotation
+
+class Command(BaseCommand):
+    def handle(self, *args, **options):
+        for experiment in Experiment.objects.all():
+            organism_ids = list(ComputationalResultAnnotation.objects.filter(data__is_qn=True).values_list('data__organism_id', flat=True))
```
arielsvn: can we put `organism_ids` outside of the `for-loop`?

kurtwheeler: Oh snap good call!

PR #1373, opened 2019-07-05
Kurtwheeler/run tximport

kurtwheeler:
## Issue Number

#1074 

This PR adds a mangagement command which will create a tximport job for every experiment that is past the minimum thresholds and hasn't already had tximport run on it. This should enable us to unlock a lot more RNA-Seq samples in the frontend.

## Types of changes

## Functional tests

I am going to finish this functional test first thing Monday morning. Today I have been running a 21 sample experiment in a local environment. I killed one downloader jobs so only 20 of those samples will get processed. I will then run this command twice to run tximport and to test that it's not run more than once.

## Checklist


wvauclain:
As long as the tests pass

PR #1376, opened 2019-07-08
Add downloadable samples filter

arielsvn:
## Issue Number

https://github.com/alexslemonade/refinebio-frontend/issues/639

Adds a new filter on the search api that counts the number of results that have downloadable samples.

image

## Types of changes

## Functional tests

Tested locally

## Checklist

### DIFF
--- a/api/data_refinery_api/views.py
+++ b/api/data_refinery_api/views.py
@@ -309,7 +309,7 @@ class ExperimentDocumentView(DocumentViewSet):
             'facet': TermsFacet,
             'enabled': True # These are enabled by default, which is more expensive but more simple.
         },
-        'organism_names': {
+        'organism': {
             'field': 'organism_names',
             'facet': TermsFacet,
             'enabled': True,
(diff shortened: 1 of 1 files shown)


api/data_refinery_api/views.py
```diff
@@ -327,6 +332,16 @@ class ExperimentDocumentView(DocumentViewSet):
             'enabled': True,
             'global': False,
         },
+        'num_downloadable_samples__gt': {
+            'field': 'num_downloadable_samples',
+            'facet': RangeFacet,
+            'options': {
+                'ranges': [
+                    ("0", (0, None)),
```
wvauclain: I'm not sure how ranges work with elasticsearch, but does this range exclude 0? At a first glance it looks like it might include 0.

arielsvn: Good catch, just confirmed in the [docs](https://www.elastic.co/guide/en/elasticsearch/reference/7.1/search-aggregations-bucket-range-aggregation.html)

> this aggregation includes the from value and excludes the to value for each range.

arielsvn:
I realized that this will always return 0, since the filter was counting number of downloadable samples of the experiments with 0 downloadable samples. Will file a PR in the FE soon, closing this since it won't be needed.

PR #1384, opened 2019-07-11
Added a Syntax Test

### LINKED ISSUE #1348: Syntax errors cause our tests to silently not run
### Context

#1347 

### Problem or idea

If we have a syntax error in our code then our test runner fails silently instead of failing the tests. This is because we use tags, and if we say to only run tags the test runner for some reason doesn't even consider syntax errors a problem.

### Solution or next step

Maybe there's a way to tell the python unit test runner to fail if there's syntax errors. If not then we need to check to make sure that the test output never says "ran 0 tests".

wvauclain:
## Issue Number

Fixes #1348 

I added a CircleCI test that runs `flake8`, a python linter. Eventually, if we migrate to full-on PEP8 standards, this could check all of PEP8, but for now I only have it check for `E9` (syntax errors).

## Types of changes

## Functional tests

I undid the changes from #1347 to make sure that the syntax test would catch that error, and then I re-did the changes to make sure the tests passed.

## Checklist

## Screenshots

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -113,6 +113,19 @@ jobs:
           command: .circleci/filter_tests.sh -t compendia
           no_output_timeout: 1h
 
+  syntax_test:
+    working_directory: ~/refinebio
+    docker:
+      - image: circleci/python:3.6.4
+    steps:
+      - checkout
+
+      # Install flake8
+      - run: sudo pip install flake8
+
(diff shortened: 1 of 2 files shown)


cgreene:
Screen Shot 2019-07-11 at 3 20 38 PM

Also confirmed that it fails with the syntax error

PR #1386, opened 2019-07-11
Don't run the compendia generation on the smasher instance

### LINKED ISSUE #1385: Compendia Jobs Wait Indefinitely
### Context

I was trying to generate a new fish compendium for @jaclyn-taroni, but I couldn't get the job to start.

### Problem or idea

It looks like the compendia jobs are constrained to run on smasher nodes, but the RAM requirements of compendia jobs are 180GB and the smashers only have 32GB of RAM.

### Solution or next step

I think that if I remove that constraint we should be able to generate a compendium.

wvauclain:
## Issue Number

Closes #1385 

Don't run the compendia generation on the smasher. The compendia job specification requires more ram than is available on smasher instances, so the jobs hang indefinitely.

## Types of changes

- Tuning
- Bugfix

## Functional tests

N/A

## Checklist

## Screenshots

### DIFF
--- a/workers/nomad-job-specs/create_compendia.nomad.tpl
+++ b/workers/nomad-job-specs/create_compendia.nomad.tpl
@@ -77,7 +77,12 @@ job "CREATE_COMPENDIA" {
         max_file_size = 1
       }
 
-      ${{SMASHER_CONSTRAINT}}
+      # Don't run on the smasher instance, it's too small
+      constraint {
+        attribute = "${meta.is_smasher}"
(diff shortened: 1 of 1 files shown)


arielsvn:
Any idea why this constraint was there?

cgreene:
I think it was because the processing instances other than the smasher are not always up, so they may not be available for work.

cgreene:
However, if this is just for the compendia, we can make sure to always have an instance up when we run that job.

Issue #1378, opened 2019-07-09
Diagnose the large query that seems to fire about every 5 minutes


kurtwheeler:
### Context

![image](https://user-images.githubusercontent.com/4026833/60893956-cbed4d00-a22f-11e9-9adb-41cd94a86fb2.png)

There seems to be some kind of query that jumps our DB usage by like 80% about every 5 minutes.

### Problem or idea

This query is probably a significant drain on our database. As we prepare to scale up bigger and faster the database is likely to become a bottleneck for us, so reducing load on it could be very beneficial for going fast.

### Solution or next step

We should enable query logging for our RDS instance: https://aws.amazon.com/premiumsupport/knowledge-center/rds-postgresql-query-logging/

It looks like there's a setting we can have only queries that take too long get logged. We should use that with a high bar like a second or two, and then see if we net these queries. Once we find what query is responsible for that pattern, we should see if we can optimize it and/or see if we can run it less frequently.

wvauclain:
I added query logging to our staging server, and I found a bunch of long queries, like these ones:

```
2019-07-11 13:27:05 UTC:10.0.79.106(40914):drpostgresuser@data_refinery:[26626]:LOG: duration: 7500.662 ms statement: SELECT DISTINCT "samples"."id", "samples"."accession_code", "samples"."title", "samples"."organism_id", "samples"."source_database", "samples"."source_archive_url", "samples"."source_filename", "samples"."source_absolute_file_path", "samples"."has_raw", "samples"."platform_accession_code", "samples"."platform_name", "samples"."technology", "samples"."manufacturer", "samples"."protocol_info", "samples"."sex", "samples"."age", "samples"."specimen_part", "samples"."genotype", "samples"."disease", "samples"."disease_stage", "samples"."cell_line", "samples"."treatment", "samples"."race", "samples"."subject", "samples"."compound", "samples"."time", "samples"."is_processed", "samples"."is_blacklisted", "samples"."is_public", "samples"."created_at", "samples"."last_modified" FROM "samples" WHERE "samples"."is_public" = true ORDER BY "samples"."source_archive_url" ASC LIMIT 25 OFFSET 475
[...]

PR #1389, opened 2019-07-12
Cache the stats view

### LINKED ISSUE #1388: Cache dashboard data
### Context

We have been investigating what is causing our database's CPU to spike every five minutes, and some of the possible candidates for those queries are the executive dashboard and dashboard API requests.

### Problem or idea

Even if those requests aren't what's causing the database CPU usage to spike, they are hefty requests that are prone to timing out. Since they already refresh every five minutes on the frontend, it might be a good idea to have the backend run the queries once every five minutes and serve up the cached queries via the API instead of direct queries. We might even be able to cut down the refresh time if these queries are run once every N minutes instead of once for every M clients every N minutes.

### Solution or next step

Investigate if there is a way to cache queries with Django REST Framework with a certain timeout.

wvauclain:
## Issue Number

Related to #1378 
Closes #1388 

Cache the stats page so that it is only actually generated once every five minutes, then each subsequent request hits the cache.

## Types of changes

## Functional tests

```
$ time curl "localhost:8000/stats/?range=day"
curl "localhost:8000/stats/?range=day"  0.00s user 0.01s system 8% cpu 0.145 total

$ time curl "localhost:8000/stats/?range=day"
curl "localhost:8000/stats/?range=day"  0.00s user 0.01s system 32% cpu 0.027 total

$ time curl "localhost:8000/stats/?range=day"
curl "localhost:8000/stats/?range=day"  0.01s user 0.00s system 37% cpu 0.025 total

$ time curl "localhost:8000/stats/?range=day"
curl "localhost:8000/stats/?range=day"  0.01s user 0.00s system 39% cpu 0.020 total

$ time curl "localhost:8000/stats/?range=year"
curl "localhost:8000/stats/?range=year"  0.01s user 0.00s system 8% cpu 0.123 total
```

## Checklist

## Screenshots


api/data_refinery_api/views.py
```diff
@@ -888,6 +890,7 @@ class Stats(APIView):
         description="Specify a range from which to calculate the possible options",
         enum=('day', 'week', 'month', 'year',)
     )])
+    @method_decorator(cache_page(60*5))  # Cache the stats view for five minutes
```
arielsvn: Is it possible to use `@cache_page` directly here?

wvauclain: I tried that first, but I think it's `Python`'s fault. `@method_decorator` wraps a function decorator to work with methods too. Without it, `@cache_page` tries to look at `self` instead of the `request` parameter.

wvauclain:
I'm not completely happy with this because the first request every five minutes will still likely time out while the request regenerates. I'm trying to find out if there is a way to automatically refresh the cache.

cgreene:
Do we have a cache already configured? If not, does this work just using the memory of the worker? In that case, I don't know how effective it would be.

wvauclain:
Yeah, you're right:

"Local-memory caching
This is the default cache if another is not specified in your settings file. ...This cache is per-process (see below) and thread-safe."

api/data_refinery_api/views.py
```diff
@@ -1114,6 +1145,15 @@ def _get_intervals(self, objects, range_param, field = 'created_at'):
                       .values('start') \
                       .filter(start__gte=range_to_start_date.get(range_param))
 
+
+STAT_CACHE_REFRESH_INTERVAL_MINUTES = 5
+scheduler = BackgroundScheduler()
+for timeframe in ('day', 'week', 'month', 'year', None):
+    scheduler.add_job(Stats.update_cache, 'interval', [timeframe],
+                      minutes=STAT_CACHE_REFRESH_INTERVAL_MINUTES)
+scheduler.start()
```
arielsvn: Is this a good place to instantiate the `BackgroundScheduler`? I'm not sure about it, but I'd imagine these should be in a different file?

wvauclain: I'm not sure, since it has to be imported somewhere in order to load.

api/data_refinery_api/views.py
```diff
@@ -889,18 +892,32 @@ class Stats(APIView):
...
     )])
     def get(self, request, format=None):
+        cached_stats = cache.get("stats", None)
         range_param = request.query_params.dict().pop('range', None)
 
+        # This should only need to run once for each range at most. From then on,
+        # the cache is refreshed in the background using apscheduler.
+        # We index the dict by str(range_param) because None is the default range.
+        if cached_stats is None or cached_stats.get(str(range_param), None) is None:
+            cached_stats = Stats.update_cache(range_param)
```
arielsvn: Will the cache be invalidated if there's a new deploy?

wvauclain: It looks like it will, so if someone views the dashboard within `STAT_CACHE_REFRESH_INTERVAL_MIN` minutes of deploy they won't get a cached result.

PR #1403, opened 2019-07-19
Change our CRAN mirror to a Microsoft fixed CRAN snapshot

### LINKED ISSUE #1402: Consider fixed CRAN snapshots
### Context

Tests are failing on CircleCI recently because no_op is not being cached and it refuses to build.

### Problem or idea

Currently, builds are failing for some of our images because of a nested dependency three layers down upgrading into an incompatible version.

Blob depends on vctrs, RSQLite depends on blob, and AnnotationDbi depends on RSQLite, so a minor version bump for vctrs that depends on a new version of rlang meant that AnnotationDbi would no longer compile.

### Solution or next step

@jaclyn-taroni came up with the idea that we should consider fixing the entirety of CRAN to one point in time to avoid this kind of error:
https://mran.microsoft.com/documents/rro/reproducibility

wvauclain:
## Issue Number

Fixes #1402

Change our CRAN mirror to a Microsoft fixed CRAN snapshot taken a couple of days before two transitive dependencies of `AnnotationDbi` bumped minor versions, which caused some of our docker containers to stop building.

## Types of changes

## Functional tests

I built the `no_op` and `illumina` containers locally.

## Checklist

## Screenshots


wvauclain:
Updated pull request description after most recent commit

PR #1405, opened 2019-07-22
Associate downloader jobs with a volume

### LINKED ISSUE #1371: We need be smarter about running downloader jobs on a per-volume basis
### Context

We limit the number of downloader jobs that queue based on how many instances we currently have running, but this doesn't give us very much control how many run on each instance and it's also a very rough control. We have variable sizes of instances that we run on so a large instance may be able to process the data from a larger number of downloader jobs than a smaller instance. Currently we have the limit set to 50 jobs per instance. This is definitely quite a bit faster than the instances can keep up with and the volumes are all slowly filling up.

After downloading an additional 1000 human experiments the volume that has been in use consistently (4) is more than two thirds full:
```
/dev/nvme5n1    8.8T  6.4T  2.0T  77% /var/ebs
```

### Problem or idea

If we ever run for a long enough time without cycling our volumes then they will eventually fill up because we download data faster than we process it. Even if we were to set the downloader jobs limit at a point where we were processing faster than downloading... that would just mean we'd be under utilizing our resources.

### Solution or next step

I think we should handle this making the Foreman smarter about queuing downloader jobs. I think that if we give downloader jobs a volume index similar to the way that we do with processor jobs, then we can have the Foreman calculate the work depth for each volume and only queue more downloader jobs for each volume when the work depth drops below some threshold.

wvauclain:
## Issue Number

Closes #1371 

For each available volume, we now calculate how many running and pending jobs that volume has (its work depth). If that work depth is below a certain limit (300 for now), we queue up downloader jobs on that volume to bring the total number of jobs back up to 300.

## Types of changes

## Functional tests

I changed the test to make sure that this works as expected.

## Checklist

## Screenshots


common/data_refinery_common/message_queue.py
```diff
@@ -77,26 +77,28 @@ def send_job(job_type: Enum, job, is_dispatch=False) -> bool:
...
+            # have an attached volume index, so use that.
+            if job.volume_index is None:
+                job.volume_index = get_volume_index()
+                job.save()
+            nomad_job = nomad_job + "_" + job.volume_index + "_" + str(job.ram_amount)
+        elif isinstance(job, SurveyJob):
+            nomad_job = nomad_job + "_" + str(job.ram_amount)
+        elif isinstance(job, DownloaderJob):
+            volume_index = job.volume_index if settings.RUNNING_IN_CLOUD else "0"
+            nomad_job = nomad_job + "_" + volume_index + "_" + str(job.ram_amount)
```
kurtwheeler: I think moving this code block would introduce a bug because we'd start queuing downloader jobs immediately. In fact, I don't think this is where we should be setting the volume_index. I think we should have the foreman decide which volume to put downloader jobs on as it dispatches them.

kurtwheeler: Ok, as we discussed in person I now see why this is in fact correct: this will only fire when `is_dispatch=True` which will be the case when the Foreman is actually queuing downloader jobs.

foreman/data_refinery_foreman/foreman/main.py
```diff
@@ -50,14 +51,19 @@
...
+VOLUME_WORK_DEPTH = dict()
+TIME_OF_LAST_WORK_DEPTH_CHECK = timezone.now() - datetime.timedelta(minutes=10)
+
+# The number of downloader jobs currently in the queue
+DOWNLOADER_JOBS_IN_QUEUE = 0
+TIME_OF_LAST_DOWNLOADER_JOB_CHECK = timezone.now() - datetime.timedelta(minutes=10)
+
+# The desired number of active + pending jobs on a volume. Downloader jobs
+# will be assigned to instances until this limit is reached.
+DESIRED_WORK_DEPTH = 300
```
kurtwheeler: I think this number is too low. Right now with 2 8-weighted instances up we have 381 jobs running simultaneously. If we had a 20-weighted instance up it seems reasonable that it could probably run ~500 jobs itself. Therefore in order to have a bit of padding I'm thinking that we should set this to 1000. Even if we were at full capacity with 12 8-weighted instances we'd still only have a 12k job work queue depth, which is well under the 20k limit that the size of our nomad instances can handle.

wvauclain: That makes sense. I didn't pick 300 for any particular reason, figuring that we would tune it later.

foreman/data_refinery_foreman/foreman/main.py
```diff
@@ -391,60 +385,76 @@ def requeue_downloader_job(last_job: DownloaderJob) -> bool:
...
-            total = total + job['JobSummary']['Children']['Running']
+    nomad_host = get_env_variable("NOMAD_HOST")
+    nomad_port = get_env_variable("NOMAD_PORT", "4646")
+    nomad_client = Nomad(nomad_host, port=int(nomad_port), timeout=30)
+
+    global TIME_OF_LAST_DOWNLOADER_JOB_CHECK
+    global DOWNLOADER_JOBS_IN_QUEUE
+
+    if (timezone.now() - TIME_OF_LAST_DOWNLOADER_JOB_CHECK > window):
+        TIME_OF_LAST_DOWNLOADER_JOB_CHECK = timezone.now()
```
kurtwheeler: Can you move this down near the end of this function?

foreman/data_refinery_foreman/foreman/main.py
```diff
@@ -391,60 +385,76 @@ def requeue_downloader_job(last_job: DownloaderJob) -> bool:
...
+        try:
+            all_downloader_jobs = nomad_client.jobs.get_jobs(prefix="DOWNLOADER")
+        except:
+            # Nomad is down, return an impossibly high number to prevent
+            # additonal queuing from happening:
+            DOWNLOADER_JOBS_IN_QUEUE = sys.maxsize
+
+        total = 0
+        for job in all_downloader_jobs:
```
kurtwheeler: If nomad is down `all_downloader_jobs` will be undefined. I think this block should probably be inside the `try`.

PR #1409, opened 2019-07-23
Allow hotfixes from dev

### LINKED ISSUE #1404: Allow hotfix deploys on `dev`
### Context

We deploy prod when a new tag is created that ends in `-hotfix`

### Problem or idea

Even if the tag is created on `dev`, it will always deploy to production. I think we should make it so that creating the tag on `dev` deploys to staging. Just like all other tags. With the advantage would be that in both cases the tests aren't run before the deploy.

### Solution or next step

Removing these two lines should do the work:

https://github.com/AlexsLemonade/refinebio/blob/dev/.circleci/remote_deploy.sh#L74-L75

wvauclain:
## Issue Number

Closes #1404

I removed the lines that automatically put every hotfix onto master.

NOTE: In order for the dev hotfix deploys to work, the proper tag format is v$VERSION-dev-hotfix.

## Types of changes

## Functional tests

I tested the changes to `get_master_or_dev()` locally.

## Checklist

## Screenshots

### DIFF
--- a/.circleci/remote_deploy.sh
+++ b/.circleci/remote_deploy.sh
@@ -71,9 +71,7 @@ source ~/refinebio/scripts/common.sh
 # Circle won't set the branch name for us, so do it ourselves.
 branch=$(get_master_or_dev $CIRCLE_TAG)
 
-if [[ "$CIRCLE_TAG" == *"-hotfix" && "$branch" == "dev" ]]; then
-    DOCKERHUB_REPO=ccdl
-elif [[ "$branch" == "master" ]]; then
+if [[ "$branch" == "master" ]]; then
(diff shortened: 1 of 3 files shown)


arielsvn:
> NOTE: In order for the dev hotfix deploys to work, the proper tag format is `v$VERSION-dev-hotfix`.

Nice change. What would happen if someone pushes a tag with `v$VERSION-hotfix` to `dev` ?

wvauclain:
If dev is ahead of master, then `v$VERSION-hotfix` would be deployed to `ccdlstaging`, which isn't necessarily a problem but it would mess with our naming scheme.

PR #1414, opened 2019-07-24
Send output of scripts back to CircleCI

### LINKED ISSUE #1413: CircleCI times out while building images
### Context

When all of the docker images had to update so that we could pin their dependencies, the deploys timed out because CircleCI was not reading any output for four hours.

### Problem or idea

It looks like at some point something changed, so these lines no longer output to the CircleCI console. If these commands sent their STDOUT and STDERR to the CircleCI console, I don't think we would time out:

https://github.com/AlexsLemonade/refinebio/blob/f1bbf93f9de7ed38e61aa3bf6a198c92af08d0d7/.circleci/remote_deploy.sh#L59-L66

### Solution or next step

Figure out how to send the output from the docker commands back to CircleCI so that it detects output and doesn't kill our containers for timing out.

wvauclain:
## Issue Number

Fixes #1413 

Using `2>&1 | tee -a` instead of `&>>` sends the output both to the file and to STDOUT. CircleCI times out because there is no output on STDOUT, so if we send the output of the commands back to STDOUT CircleCI should no longer time out.

## Types of changes

## Functional tests

## Checklist

## Screenshots


.circleci/remote_deploy.sh
```diff
@@ -101,9 +101,9 @@ echo "Finished building new images, running run_terraform.sh."
 
 run_on_deploy_box "sudo touch /var/log/deploy_$CIRCLE_TAG.log"
 run_on_deploy_box "sudo chown ubuntu:ubuntu /var/log/deploy_$CIRCLE_TAG.log"
-run_on_deploy_box "source env_vars && echo -e '######\nStarting new deploy for $CIRCLE_TAG\n######' >> /var/log/deploy_$CIRCLE_TAG.log 2>&1"
+run_on_deploy_box "source env_vars && echo -e '######\nStarting new deploy for $CIRCLE_TAG\n######' 2>&1 | tee -a /var/log/deploy_$CIRCLE_TAG.log"
```
kurtwheeler: The output from this script contains secrets so I purposefully prevented it's output from making it CircleCI. However I agree that we can allow the output of building the docker images  to go to CircleCI.

PR #1416, opened 2019-07-25
Fix sample counts on search filters

### LINKED ISSUE #1145: Samples that belong to multiple experiments counted several times in search filters
### Context

https://github.com/AlexsLemonade/refinebio/pull/1124 Added sample counts to the filters.

### Problem or idea

Each experiment has a cached value `num_processed_samples`, for each filter value these values were added on Elastic Search. The problem is that some samples might belong to multiple experiments, and because of this, they are being counted several times in the filters.

image

### Solution or next step

I'm not sure how we could be able to calculate the number of samples accurately, and quickly. @Miserlou any ideas?

### New Issue Checklist

### LINKED ISSUE #1401: Some experiments are not being returned by the search endpoint
### Context

I was trying to find experiment `GSE75356`

### Problem or idea

We have in the DB and it has an [experiments page](https://www.refine.bio/experiments/GSE75356/a-zebrafish-melanoma-model-reveals-emergence-of-neural-crest-identity-during-melanoma-initiation) but searching for it's accession code or title doesn't return the experiment in the results.

On the API it should be here:

https://api.refine.bio/search/?search=accession_code:GSE75356

### Solution or next step

Not sure what the issue is, we need to investigate. Maybe it's not indexed on ES?

arielsvn:
## Issue Number

close #1145 
close #1401 

This indexes the accession codes of the downloadable samples for each experiment so that we are able to count their unique values for each filter.

The number of samples are calculated using an ES cardinality aggregation, which is [an approximation](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-metrics-cardinality-aggregation.html#_counts_are_approximate).

Indexing the experiments is a bit slower now. Searching for experiments also takes longer, although locally it was <500ms.

## Types of changes

## Functional tests

Tested the numbers returned on the filters with manual calculations of the actual results. In the case of Technology the differences were always <1000.

## Checklist


common/data_refinery_common/models/models.py
```diff
@@ -420,6 +422,27 @@ def pretty_platforms(self):
...
+        # This method will be called for each experiment when the ES index is being created
+        # This cache is to ensure that we don't repeat the query to calculate the organisms with 
+        # qn targets multiple times.
+        qn_organisms = cache.get('qn_organisms')
+        if qn_organisms == None:
+            qn_organisms = Organism.get_objects_with_qn_targets()
+            cache.set('qn_organisms', qn_organisms, 30*60) # save value for 1h
+
+        return list(self.samples.filter(is_processed=True, organism__in=qn_organisms)\
+                                      .values_list('accession_code', flat=True))
```
kurtwheeler: Rather than caching a list of organisms that have QN targets... what if we just added a `most_recent_qn_result` field to the organisms table? When we create a new QN target we can update that and then when we need a QN target for a given organism we won't have to look through all the computational results to find it, it'll just be on the organism object? This would of course involve modifying the QN target code to update that as they're generated and a migration to populate that field for all existing QN targets.

arielsvn:
@kurtwheeler thanks for the tip, great suggestion! I just added a reference to `ComputationalResult` on the organisms, let me know what do you think.

wvauclain:
Before you merge this, I just merged another PR with a migration, so your migration should be 0027.

PR #1417, opened 2019-07-26
Try setting the repos using a data structure instead of what the docs say

kurtwheeler:
## Issue Number

#1403

In #1403 we changed the way we specified repos from `options(repos=structure(c(CRAN="https://cran.revolutionanalytics.com")))` to `options(repos="https://cran.microsoft.com/snapshot/2019-07-03")`. I think that despite the docs saying:

> character vector, the base URL(s) of the repositories to use, e.g., the URL of a CRAN mirror such as "https://cloud.r-project.org"

that it does indeed need to have the special data structure to be used by devtools. I dunno why, but I built it on my local machine and it worked.

## Types of changes

## Functional tests

I built the image on my local machine and got a nice:

```
Successfully built 308e96f1abd8
Successfully tagged dr_affymetrix:test
```

## Checklist

### DIFF
--- a/workers/affymetrix_dependencies.R
+++ b/workers/affymetrix_dependencies.R
@@ -1,7 +1,7 @@
 # Turn warnings into errors because biocLite throws warnings instead
 # of error if it fails to install something.
 options(warn=2)
-options(repos="https://cran.microsoft.com/snapshot/2019-07-03")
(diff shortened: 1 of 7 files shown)


kurtwheeler:
I'm going to merge this so we can have the dev box build the new image to test it there

PR #1418, opened 2019-07-26
Add worker_version to dataset

arielsvn:
## Issue Number

https://github.com/alexslemonade/refinebio-frontend/issues/636

Adds `worker_version` field to the dataset endpoint, so that we can know in the frontend when something might have changed in the smashing process.

## Types of changes

## Functional tests

Tested locally.

## Checklist


api/data_refinery_api/serializers.py
```diff
@@ -722,6 +729,13 @@ def get_organism_samples(self, obj):
 
         return result
 
+    def get_worker_version(self, obj):
+        processor_jobs = obj.processor_jobs.order_by('-created_at').values_list('worker_version', flat=True)
```
kurtwheeler: I think we should probably filter for `success=True` right?

arielsvn: Talked in person, assuming that datasets can only be processed once this will return the correct value for processed datasets. For the datasets that failed to process, it will return the latest version that we used to try to process them, which still might be useful.

arielsvn:
Just updated the migration, in case you want to take another look @kurtwheeler

PR #1436, opened 2019-07-31
Add index to downloader and processor job tables

arielsvn:
## Issue Number

#1378 

Added a new index to the downloader and processor job tables on the fields `created_at`, `start_time` and `end_time`. These should improve the performance of queries like the following:

```
LOG: duration: 2072.185 ms statement: SELECT "downloader_jobs"."id", "downloader_jobs"."downloader_task", "downloader_jobs"."accession_code", "downloader_jobs"."no_retry", "downloader_jobs"."start_time", "downloader_jobs"."end_time", "downloader_jobs"."success", "downloader_jobs"."nomad_job_id", "downloader_jobs"."ram_amount", "downloader_jobs"."num_retries", "downloader_jobs"."retried", "downloader_jobs"."worker_id", "downloader_jobs"."worker_version", "downloader_jobs"."failure_reason", "downloader_jobs"."retried_job_id", "downloader_jobs"."was_recreated", "downloader_jobs"."created_at", "downloader_jobs"."last_modified" FROM "downloader_jobs" WHERE ("downloader_jobs"."created_at" > '2017-02-05T00:00:00+00:00'::timestamptz AND "downloader_jobs"."no_retry" = false AND "downloader_jobs"."retried" = false AND "downloader_jobs"."success" = false) ORDER BY "downloader_jobs"."id" ASC LIMIT 2001
LOG: duration: 1017.450 ms statement: SELECT "processor_jobs"."id", "processor_jobs"."pipeline_applied", "processor_jobs"."no_retry", "processor_jobs"."ram_amount", "processor_jobs"."volume_index", "processor_jobs"."start_time", "processor_jobs"."end_time", "processor_jobs"."success", "processor_jobs"."nomad_job_id", "processor_jobs"."num_retries", "processor_jobs"."retried", "processor_jobs"."worker_id", "processor_jobs"."worker_version", "processor_jobs"."failure_reason", "processor_jobs"."retried_job_id", "processor_jobs"."created_at", "processor_jobs"."last_modified" FROM "processor_jobs" WHERE ("processor_jobs"."created_at" > '2019-07-15T12:29:08.683003+00:00'::timestamptz AND "processor_jobs"."end_time" IS NULL AND "processor_jobs"."no_retry" = false AND "processor_jobs"."pipeline_applied" = 'SMASHER' AND "processor_jobs"."retried" = false AND "processor_jobs"."start_time" IS NULL AND "processor_jobs"."success" IS NULL)
```

These are performed by the foreman.

If we update to django 2.2 we can reduce the scope of the index and only index the rows relevant to those queries, with the [condition](https://docs.djangoproject.com/en/2.2/ref/models/indexes/#django.db.models.Index.condition) parameter.

Also moved `JOB_CREATED_AT_CUTOFF` to a more recent date.

## Types of changes

## Functional tests

Applied the migrations and checked the query plan locally to ensure it was using the index instead of a `Seq Scan`

```
data_refinery=# explain analyze verbose select id from downloader_jobs where success='f' and no_retry='f' and retried='f' and created_at > '2019-06-05 00:00:00'::timestamp;
                                                                    QUERY PLAN                                                                     
---------------------------------------------------------------------------------------------------------------------------------------------------
 Index Scan using downloader_jobs_created_at on public.downloader_jobs  (cost=0.43..7.37 rows=1 width=4) (actual time=0.011..0.011 rows=0 loops=1)
   Output: id
   Index Cond: (downloader_jobs.created_at > '2019-06-05 00:00:00'::timestamp without time zone)
   Filter: ((NOT downloader_jobs.success) AND (NOT downloader_jobs.no_retry) AND (NOT downloader_jobs.retried))
 Planning time: 2.990 ms
 Execution time: 0.058 ms
(6 rows)
```

vs **production** (ignore the times since prod has 4x more rows than my local db)

```
                                                                                              QUERY PLAN                                                                                              
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Seq Scan on public.downloader_jobs  (cost=0.00..149446.10 rows=111639 width=4) (actual time=719.426..719.426 rows=0 loops=1)
   Output: id
   Filter: ((NOT downloader_jobs.success) AND (NOT downloader_jobs.no_retry) AND (NOT downloader_jobs.retried) AND (downloader_jobs.created_at > '2019-06-05 00:00:00'::timestamp without time zone))
   Rows Removed by Filter: 4377583
 Planning time: 0.066 ms
 Execution time: 719.444 ms
```

## Checklist


arielsvn:
Also added another index on `worked_id` for downloader jobs since there are multiple queries that are taking >1sec

```
duration: 1193.056 ms  statement: SELECT COUNT(*) AS "__count" FROM "downloader_jobs" WHERE ("downloader_jobs"."end_time" IS NULL AND "downloader_jobs"."retried" = false AND "downloader_jobs"."start_time" IS NOT NULL AND "downloader_jobs"."success" IS NULL AND "downloader_jobs"."worker_id" = 'i-0ccf43149f8428a6b')
duration: 1198.485 ms  statement: SELECT COUNT(*) AS "__count" FROM "downloader_jobs" WHERE ("downloader_jobs"."end_time" IS NULL AND "downloader_jobs"."retried" = false AND "downloader_jobs"."start_time" IS NOT NULL AND "downloader_jobs"."success" IS NULL AND "downloader_jobs"."worker_id" = 'i-0ccf43149f8428a6b')
duration: 1274.428 ms  statement: SELECT COUNT(*) AS "__count" FROM "downloader_jobs" WHERE ("downloader_jobs"."end_time" IS NULL AND "downloader_jobs"."retried" = false AND "downloader_jobs"."start_time" IS NOT NULL AND "downloader_jobs"."success" IS NULL AND "downloader_jobs"."worker_id" = 'i-0ccf43149f8428a6b')
duration: 1378.323 ms  statement: SELECT COUNT(*) AS "__count" FROM "downloader_jobs" WHERE ("downloader_jobs"."end_time" IS NULL AND "downloader_jobs"."retried" = false AND "downloader_jobs"."start_time" IS NOT NULL AND "downloader_jobs"."success" IS NULL AND "downloader_jobs"."worker_id" = 'i-0ccf43149f8428a6b')
```

common/data_refinery_common/models/jobs.py
```diff
@@ -87,6 +87,17 @@ class ProcessorJob(models.Model):
...
         db_table = "processor_jobs"
 
+        indexes = [
+            models.Index(
+                fields=['created_at'], 
+                name='processor_jobs_created_at',
+                # A partial index might be better here, given our queries we don't
+                # need to index the whole table. We need to update to Django 2.2
+                # for this to be supported.
+                # condition=Q(success=None, retried=False, no_retry=False)
```
kurtwheeler: While this comment seems to have good information, it's likely to be forgotten about here. Do we have an issue open about this? Maybe we can make one and link to it here?

arielsvn:
> You mention in the PR description that you also added an index to start_time and end_time, but I guess you decided to take those out of this?

Yes, I think it only makes the index more complicated and it doesn't add much value since we query with a range on `created_at`.

> you took all the order_by('id') calls to improve performance as well?

The results should be sorted by `created_at` by default from the index, sorting by `id` would force Postgres to load the matching results in memory. If you want we can add `.order_by('created_at')` to make this more evident.

kurtwheeler:
> The results should be sorted by created_at by default from the index

Are you sure about this? Just cause otherwise the pagination won't work. I agree though that it should work like this. Adding `.order_by('created_at')` might be good just to make things explicit, but it's not strictly necessary.

common/data_refinery_common/models/jobs.py
arielsvn: Good idea, added here https://github.com/AlexsLemonade/refinebio/issues/1454

arielsvn: Good idea, added here #1454

arielsvn:
> Are you sure about this? Just cause otherwise the pagination won't work.

Yes, the index should be able to retrieve the results sorted by `created_at`, so adding `order_by('created_at')` shouldn't require an [additional step](https://www.postgresql.org/docs/current/indexes-ordering.html).

wvauclain:
I think it's worth pointing out that under the hood, `PerformantPaginator` applies a separate ordering (the default is `pk`, but you can customize it) to the queryset:

https://github.com/AlexsLemonade/refinebio/blob/40305a5ccabfda4af56b1207313e0a952df65439/foreman/data_refinery_foreman/foreman/performant_pagination/pagination.py#L155-L156

You might want to change the `Paginator` constructor calls to `Paginator(jobs, PAGE_SIZE, 'created_at')` to avoid this and instead order by something in the index, since it might undo your gains.

arielsvn:
@kurtwheeler added the calls to `order_by('created_at')` and also `Paginator(jobs, PAGE_SIZE, 'created_at')` as @wvauclain suggested. Let me know if you guys see anything else that can be improved.

arielsvn:
Tests passed, @kurtwheeler you might want to take another look. I added a special case on the `PerformantPaginator`. https://github.com/AlexsLemonade/refinebio/pull/1436/commits/caffdacc15a7ef177c183141872ac05f684ae221

common/data_refinery_common/models/jobs.py
kurtwheeler: Wait can you link to it here?

arielsvn: you mean on the comment? sure!

```suggestion
                # condition=Q(success=None, retried=False, no_retry=False)
                # https://github.com/AlexsLemonade/refinebio/issues/1454
```

PR #1438, opened 2019-07-31
Fix an issue with tximport getting the index length 

kurtwheeler:
## Issue Number

#1074 

Also try to prevent no_retry jobs from being recreated by checking for them in the `needs_processing` method of OriginalFile.

Salmon quant computational results have multiple annotations. Tximport was assuming there was  only one. This fixes that.

## Types of changes

## Functional tests

N/A tests cover this well and we're testing these things on staging ATM.

### DIFF
--- a/common/data_refinery_common/models/models.py
+++ b/common/data_refinery_common/models/models.py
@@ -297,7 +297,7 @@ class Experiment(models.Model):
     # Cached Computed Properties
     num_total_samples = models.IntegerField(default=0)
     num_processed_samples = models.IntegerField(default=0)
-    num_downloadable_samples = models.IntegerField(default=0)    
+    num_downloadable_samples = models.IntegerField(default=0)
     sample_metadata_fields = ArrayField(models.TextField(), default=list)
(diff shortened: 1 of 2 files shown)


common/data_refinery_common/models/models.py
```diff
@@ -737,9 +743,10 @@ def needs_processing(self, own_processor_id=None) -> bool:
         if own_processor_id:
             incomplete_processor_jobs = incomplete_processor_jobs.exclude(id=own_processor_id)
 
-        # If the file has a processor job that hasn't even started yet,
-        # then it doesn't need another.
-        if incomplete_processor_jobs.count() > 0:
+        # Check if there's any jobs which should block another
+        # processing attempt.
+        blocking_jobs = no_retry_processor_jobs | incomplete_processor_jobs
```
arielsvn: cool, didn't know you could combine querysets like this.

kurtwheeler: Yeah me neither! You can use the `&` operator too apparently!

PR #1440, opened 2019-08-01
Add a check for if a processor job is a tximport job before failing

wvauclain:
## Issue Number

N/A came up while testing the tximport management command

Tximport jobs fail because their original files were already processed. They should have been processed already by Salmon, so it doesn't make sense to fail for that reason.

## Types of changes

## Functional tests

## Checklist

## Screenshots

### DIFF
--- a/workers/data_refinery_workers/processors/utils.py
+++ b/workers/data_refinery_workers/processors/utils.py
@@ -197,7 +197,7 @@ def start_job(job_context: Dict):
         raise Exception("processors.start_job called on job %s that has already been started!" % str(job.id))
 
     original_file = job.original_files.first()
-    if not job.pipeline_applied == ProcessorPipeline.TXIMPORT and original_file\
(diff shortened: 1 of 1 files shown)


workers/data_refinery_workers/processors/utils.py
```diff
@@ -197,7 +197,8 @@ def start_job(job_context: Dict):
         raise Exception("processors.start_job called on job %s that has already been started!" % str(job.id))
 
     original_file = job.original_files.first()
-    if original_file and not original_file.needs_processing(job_context["job_id"]):
+    if not job.pipeline_applied == ProcessorPipeline.TXIMPORT and original_file\
```
kurtwheeler: I'm fairly certain this needs to be `ProcessorPipeline.TXIMPORT.value` because job.pipeline_applied is a string and ProcessorPipeline.TXIMPORT is an Enum.

wvauclain: Good catch

PR #1452, opened 2019-08-05
Run compendia jobs on the smasher instance

kurtwheeler:
## Issue Number

Reverts https://github.com/AlexsLemonade/refinebio/pull/1386

It's best to run compendia jobs on the smasher instance because it won't get cycled by AWS and we don't have to worry about enough capacity being available to run the job. It does mean we have to scale our smasher instance up and then back down to run compendia, but it's better than running these jobs on spot instances.

## Types of changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -545,6 +545,13 @@ or for all organisms with sufficient data:
 nomad job dispatch -meta ORGANISM= CREATE_COMPENDIA
 ```
 
+Compendia jobs run on the smasher instance.
+However they require a very large amount of RAM to be able to complete.
+Our smasher instance does not generally have enough RAM to be able to run them, so if you need to run a smasher job you should temporarily increase the size of the smasher instance.
+This can be done by changing the terraform variable `smasher_instance_type` which can be found in qinfrastructure/variables.tf`.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -545,6 +545,13 @@ or for all organisms with sufficient data:
 nomad job dispatch -meta ORGANISM= CREATE_COMPENDIA
 ```
 
+Compendia jobs run on the smasher instance.
+However they require a very large amount of RAM to be able to complete.
+Our smasher instance does not generally have enough RAM to be able to run them, so if you need to run a smasher job you should temporarily increase the size of the smasher instance.
+This can be done by changing the terraform variable `smasher_instance_type` which can be found in qinfrastructure/variables.tf`.
```
wvauclain: I think you have a typo here

wvauclain:
The documentation changes are super helpful

README.md
wvauclain: I still think you want a second backtick at the start of infrastructure/variables.tf

kurtwheeler:
Sweet, I've addressed your comments.

Issue #1391, opened 2019-07-15
Missing samples from Affymetrix microarray experiments


jaclyn-taroni:
### Context

I was looking for a large microarray experiment on refine.bio. 

### Problem or idea

I've noticed a few experiments where I would expect we'd be able to process all the samples:

* [`GSE116143`](https://www.refine.bio/experiments/GSE116436/drug-induced-change-in-gene-expression-across-nci-60-cell-lines-after-exposure-to-15-anticancer-agents-for-2-6-and-24h) has 7623 total examples and 7208 downloadable samples.
* [`GSE37418`](https://www.refine.bio/experiments/GSE37418/novel-mutations-target-distinct-subgroups-of-medulloblastoma) has 76 total samples and 71 downloadable samples.
* [`GSE37382`](https://www.refine.bio/experiments/GSE37382/subgroup-specific-somatic-copy-number-aberrations-in-the-medulloblastoma-genome-mrna) has 285 total samples and 255 downloadable samples.
* [`GSE85033`](https://www.refine.bio/experiments/GSE85033/a-machine-learning-classifier-trained-on-cancer-transcriptomes-detects-nf1-inactivation-signal-in-glioblastoma) still has no downloadable samples.

### Solution or next step

We should investigate why we're failing to process Affymetrix samples.

jaclyn-taroni:
Note: above, I assume this is due to processing failures but this may not be the case.

cgreene:
It should be exceptionally rare that we fail to process an Affymetrix sample with a CEL file available.

arielsvn:
After some investigation, seems like they failed for different reasons.

* [`GSE116143`](https://www.refine.bio/experiments/GSE116436/drug-induced-change-in-gene-expression-across-nci-60-cell-lines-after-exposure-to-15-anticancer-agents-for-2-6-and-24h) has 7623 total examples and 7208 downloadable samples.
Seems that no processor jobs were created for the samples that aren't processed samples 



* [`GSE37418`](https://www.refine.bio/experiments/GSE37418/novel-mutations-target-distinct-subgroups-of-medulloblastoma) has 76 total samples and 71 downloadable samples.
Several errors here:
[...]

PR #1453, opened 2019-08-05
Ensure a downloader job can be created before trying to create one

arielsvn:
## Issue Number

#1391 

The `create_missing_downloader_jobs` takes a long time because it tries to create downloader jobs for samples with unsupported platforms. This checks that a downloader job can be created before actually trying to create a new downloader job.

## Types of changes

## Functional tests

Didn't tested locally.

## Checklist


arielsvn:
> I don't know that it's worth interrupting the job and starting it over again to deploy them.

Yes, I agree. 

Added another modification on the query to ensure next time we run we don't go over samples with unsupported platforms at all.

PR #1460, opened 2019-08-07
Brings master up to date with dev, deploying many changes including some tuning

kurtwheeler:
We're scaling up to 4 volumes to be able to go faster.

There's also a lot of stuff that's gotten merged in recently that's going out with this PR. For that reason I think I'd like to do a non-hotfix deploy. I was hoping to do this really quickly tonight so we could go faster now that the `create_missing_downloader_jobs` command has finished, but I don't think we should rush this deploy since there's so many changes going out now.


arielsvn:
`main_tests` is failing for me on other pr

Issue #1480, opened 2019-08-12
Run survey jobs in more of trickle rather than a deluge


kurtwheeler:
### Context

When we run 200 surveyor jobs all at the same time we kill the database:

![image](https://user-images.githubusercontent.com/4026833/62894516-d3b27c80-bd1a-11e9-8411-393969d9c9ef.png)

### Problem or idea

We should run surveyor jobs 1-5 at a time. This will probably generate data faster than we can process it, but without knocking over our own database so hard.

### Solution or next step

I'm not 100% sure on how we should implement this just yet. I know it needs to happen though. The simplest naive solution I can think of would be to de-parallelize survey jobs, let each one do a list of experiment accessions one at a time, and then we can just choose how many to run simultaneously.

arielsvn:
Looking at the logs, there're a lot of queries like these:

```
2019-08-12 20:00:04 UTC:10.0.73.172(37070):drpostgresuser@data_refinery:[21349]:LOG:  duration: 1866.064 ms  statement: SELECT "original_files"."id", "original_files"."filename", "original_files"."absolute_file_path", "original_files"."size_in_bytes", "original_files"."sha1", "original_files"."s3_bucket", "original_files"."s3_key", "original_files"."source_url", "original_files"."is_archive", "original_files"."source_filename", "original_files"."has_raw", "original_files"."is_downloaded", "original_files"."is_public", "original_files"."created_at", "original_files"."last_modified" FROM "original_files" WHERE ("original_files"."has_raw" = true AND "original_files"."source_filename" = 'SRR6652790' AND "original_files"."source_url" = 'dbtest@sra-download.ncbi.nlm.nih.gov:data/sracloud/traces/sra58/SRR/006496/SRR6652790')
[...]

PR #1500, opened 2019-08-16
Adds a management command for feeding surveyor jobs to the system.

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/1480

I've been having trouble feeding enough accessions into our system to keep it busy. This management command will go through every accession we know of and check if it's in our database. If it isn't it'll queue up a surveyor job for the experiment, sleep for 30 seconds, and continue. This should help our surveyor jobs not be all completely in sync so they hopefully use the database at different times. It also won't create more survey jobs if there are 15 or more surveyor jobs in the Nomad queue.

## Types of changes

## Functional tests

I've run this locally and verified that it runs, that it creates survey jobs, and that it won't keep creating surveyor jobs. 

## Checklist


foreman/data_refinery_foreman/foreman/management/commands/feed_the_beast.py
```diff
@@ -0,0 +1,78 @@
...
+                        # Probably having trouble communicating with Nomad, let's try again next loop.
+                        continue
+
+                    if num_surveyor_jobs < 15:
+                        try:
+                            queue_surveyor_for_accession(missing_accessions.pop())
+                            time.sleep(30)
+                        except:
+                            # We don't want to stop, gotta keep feeding the beast!!!!
+                            logger.exception("Exception caught while looping through all accessions!")
```
arielsvn: more context here would be helpful? maybe the experiment accession code?

foreman/data_refinery_foreman/foreman/management/commands/feed_the_beast.py
```diff
@@ -0,0 +1,78 @@
...
+                                num_surveyor_jobs = num_surveyor_jobs + job['JobSummary']['Children']['Pending']
+                                num_surveyor_jobs = num_surveyor_jobs + job['JobSummary']['Children']['Running']
+                    except:
+                        logger.exception("Exception caught counting surveyor jobs!")
+                        # Probably having trouble communicating with Nomad, let's try again next loop.
+                        continue
+
+                    if num_surveyor_jobs < 15:
+                        try:
+                            queue_surveyor_for_accession(missing_accessions.pop())
```
arielsvn: won't this affect the cycle above? `for accession_code in missing_accessions`

kurtwheeler:
@arielsvn I think I've addressed your comments! Good catch on the for loop!

arielsvn:
Great! This looks good! 🔥

PR #1505, opened 2019-08-17
Fix the feed_the_beast command

kurtwheeler:
## Issue Number

#1500 

I was a lil sleep deprived and rushing yesterday, I made a dumb looping mistake where I iterated over the list of accessions wrong. I made the end index BATCH_SIZE instead of `BATCH_START + BATCH_SIZE`.

This reorders the accessions it uses to put zebrafish, human, mouse, rat, fly, and worm in front in that order..

## Types of changes

## Functional tests

I changed BATCH_SIZE to  2 so I could actually see it loop in a few minutes.

## Checklist


foreman/data_refinery_foreman/foreman/management/commands/feed_the_beast.py
```diff
@@ -30,7 +30,7 @@ def handle(self, *args, **options):
 
         BATCH_SIZE = 1000
         batch_index = 0
-        batch_accessions = all_accessions[batch_index * BATCH_SIZE:BATCH_SIZE]
+        batch_accessions = all_accessions[0:BATCH_SIZE]
```
cgreene: This change won't actually do anything. You could start at batch_index if you want to avoid the 0. 🤷‍♂

foreman/data_refinery_foreman/foreman/management/commands/feed_the_beast.py
```diff
@@ -78,4 +78,6 @@ def handle(self, *args, **options):
             if batch_index * BATCH_SIZE >= len(all_accessions):
                 break
 
-            batch_accessions = all_accessions[batch_index * BATCH_SIZE:BATCH_SIZE]
+            batch_start = batch_index * BATCH_SIZE
+            bach_end = batch_start + BATCH_SIZE
+            batch_accessions = all_accessions[batch_start:batch_end]
```
cgreene: This change looks like it should fix things though 👍

foreman/data_refinery_foreman/foreman/management/commands/feed_the_beast.py
kurtwheeler: I don't, this was just for clarity!

PR #1513, opened 2019-08-19
Create command to rerun Salmon on experiments where multiple versions were applied

### LINKED ISSUE #1496: Tximport fails for experiments where samples were processed with different versions of Salmon
### Context

We got some Tximport failing with the following error message:

```sql
data_refinery=> select id, failure_reason from processor_jobs where pipeline_applied='TXIMPORT' and created_at > '2019-08-12 00:00:00'::timestamp and success='f' and failure_reason is not null order by end_time desc limit 1 OFFSET 4;
   id    |                                                         failure_reason
---------+---------------------------------------------------------------------------------------------------------------------------------
 3374939 | Found non-zero exit code from R code while running tximport.R: Read 1658 items                                                 +
         | Parsed with column specification:                                                                                              +
         | cols(                                                                                                                          +
         |   gene_id = col_character(),                                                                                                   +
         |   tx_name = col_character()                                                                                                    +
         | )                                                                                                                              +
         | reading in files with read_tsv                                                                                                 +
         | 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20 21 22 23 24 25 26 27 28 29 30 Error: all(txId == raw[[txIdCol]]) is not TRUE+
         | In addition: Warning message:                                                                                                  +
         | In txId == raw[[txIdCol]] :                                                                                                    +
         |   longer object length is not a multiple of shorter object length                                                              +
         | Execution halted
(1 row)
```

### Problem or idea

This was caused because we used different versions of salmon to process the samples in some experiments, which generates different quant files.

### Solution or next step

Talked with @kurtwheeler, we want to re-run the latest version of Salmon on the unprocessed experiments and then re-run tximport.

arielsvn:
## Issue Number

close #1496 

This modifies `get_quant_results_for_experiment` to use a queryset instead to calculate the latest `ComputationalResults` for each sample in a given experiment.

Updates `should_run_tximport` to return `False` if multiple salmon versions were applied to the samples.

Also adds a command to find experiments that haven't been tximported and where multiple versions of salmon were applied on their samples, and it creates Processor jobs for the old samples to re-run salmon on them.

## Types of changes

## Functional tests

Added tests.

## Checklist


workers/dockerfiles/Dockerfile.transcriptome
```diff
@@ -30,6 +30,10 @@ RUN cd RSEM && make install
 RUN rm -rf RSEM
 
 # Install Salmon
+
+# Tximport requires all experiments to be processed with the same version of Salmon to work https://github.com/AlexsLemonade/refinebio/issues/1496
+# This is something that should be considered when updating salmon, because
+# all unprocessed experiments will require to have salmon run on them again.
```
cgreene:
```suggestion
# all samples from incomplete experiments must have salmon run on them again.
```
I think this is true with the suggested revision. If an experiment is not yet tximported, but all the samples are done, I think it will be fine. However, the current wording suggests that will also fail. Just a minor change for those who find this in the future.

workers/dockerfiles/Dockerfile.salmon
```diff
@@ -40,7 +40,12 @@ RUN groupadd user && useradd --create-home --home-dir /home/user -g user user
 WORKDIR /home/user
 
 # Install Salmon
+
+# Tximport requires all experiments to be processed with the same version of Salmon to work https://github.com/AlexsLemonade/refinebio/issues/1496
+# This is something that should be considered when updating salmon, because
+# all unprocessed experiments will require to have salmon run on them again.
```
arielsvn:
```suggestion
# all samples from incomplete experiments must have salmon run on them again.
```

common/data_refinery_common/test_job_management.py
```diff
@@ -1,18 +1,16 @@
...
 from unittest.mock import patch, call
 from urllib.error import URLError
 
-from data_refinery_workers.downloaders import utils
+from data_refinery_common.job_management import create_processor_job_for_original_files
 
 
 class UtilsTestCase(TestCase):
-    @tag('downloaders')
+    @tag('processor')
```
kurtwheeler: This tag is important for where in the test workflow we run the test. We will filter tests for the tag `downloaders`, however we don't for `processor` because we have processor-specific tags for each of the processor images. Basically this tag is specifying which image should be used to process it, and there's no `processor` image. Honestly this can probably still be run in the `downloaders` tests because it doesn't use code that's specific to any processor image.

kurtwheeler: Alternatively the tag could probably just be removed and this could run in the common tests.

arielsvn: I like that better, I'm gonna remove it!

foreman/data_refinery_foreman/foreman/management/commands/rerun_salmon_old_samples.py
```diff
@@ -0,0 +1,108 @@
+""" This management command will search for experiments where samples haven't been processed and
+that multiple versions of salmon have been used on the samples.
+"""
+
+import random
+import sys
+import time
```
kurtwheeler: `random` and `time` aren't used by this command. Are any other of these imports unnecessary?

foreman/data_refinery_foreman/foreman/management/commands/rerun_salmon_old_samples.py
```diff
@@ -0,0 +1,108 @@
...
+    
+    for experiment in eligible_experiments:
+        update_salmon_versions(experiment)
+
+class Command(BaseCommand):
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--accession-code",
+            type=str,
+            help=("Optional parameter to allow running the command on a single experiment")
```
kurtwheeler: Oh good call, this'll be nice for making sure it works before we do it to everything!

foreman/data_refinery_foreman/foreman/management/commands/rerun_salmon_old_samples.py
```diff
@@ -0,0 +1,108 @@
...
+        )
+
+    def handle(self, *args, **options):
+        if options["accession_code"] is None:
+            update_salmon_all_experiments()
+        else:
+            experiment = Experiment.objects.all().filter(accession_code=options["accession_code"]).first()
```
kurtwheeler: `all().filter` is kinda redundant right? I think there's a lot of `all().filter`'s in this PR, they can probably just be `objects.filter`

foreman/data_refinery_foreman/foreman/management/commands/rerun_salmon_old_samples.py
```diff
@@ -0,0 +1,91 @@
...
+from data_refinery_common.utils import get_env_variable, get_active_volumes
+from data_refinery_common.job_management import create_processor_job_for_original_files
+
+logger = get_and_configure_logger(__name__)
+
+def update_salmon_versions(experiment: Experiment):
+    quant_results = get_quant_results_for_experiment(experiment)
+    salmon_versions = list(quant_results.order_by('-organism_index__created_at')\
+                                   .values_list('organism_index__salmon_version', flat=True)\
+                                   .distinct())
```
kurtwheeler: Why is this method of counting salmon versions different than the one used by `should_run_tximport` in rna_seq.py? I would think they're doing the same check so they should be the same.

foreman/data_refinery_foreman/foreman/management/commands/rerun_salmon_old_samples.py
```diff
@@ -0,0 +1,108 @@
...
+    # find the samples that were not processed with `latest_salmon_version` and trigger new processor jobs for them
+    newest_computational_results = ComputationalResult.objects.all()\
+        .filter(
+            samples=OuterRef('id'),
+            processor__name=ProcessorEnum.SALMON_QUANT.value['name']
+        )\
+        .order_by('-created_at')
+    
+    samples = experiment.samples.all().annotate(
+            salmon_version=Subquery(newest_computational_results.values('organism_index__salmon_version')[:1])
```
kurtwheeler: Can you test that this is set to None if the sample has no computational results? (i.e., not an error)

foreman/data_refinery_foreman/foreman/management/commands/test_rerun_salmon_old_samples.py
```diff
@@ -0,0 +1,288 @@
...
+    transcriptome_processor.save()
+    computational_result_short = ComputationalResult(processor=transcriptome_processor)
+    computational_result_short.save()
+
+    organism_index = OrganismIndex()
+    organism_index.index_type = "TRANSCRIPTOME_SHORT"
+    organism_index.organism = zebrafish
+    organism_index.result = computational_result_short
+    organism_index.absolute_directory_path = "/home/user/data_store/ZEBRAFISH_INDEX/SHORT"
+    organism_index.salmon_version='v0.9.0'
```
kurtwheeler: This version doesn't match the version of the tx processor which was "used to create" it

foreman/data_refinery_foreman/foreman/management/commands/test_rerun_salmon_old_samples.py
```diff
@@ -0,0 +1,288 @@
...
+
+        SampleResultAssociation.objects.get_or_create(
+            sample=sample,
+            result=quant_result
+        )
+
+    # Create another OrganismIndex with a newer version of
+    transcriptome_processor = Processor()
+    transcriptome_processor.name = "Transcriptome"
+    transcriptome_processor.version = "v9.9.10"
```
kurtwheeler: Should this be `v0.13.0`?

foreman/data_refinery_foreman/foreman/management/commands/test_rerun_salmon_old_samples.py
```diff
@@ -0,0 +1,288 @@
...
+    comp_file.sha1="ABC"
+    comp_file.s3_key = "key"
+    comp_file.s3_bucket = "bucket"
+    comp_file.save()
+
+
+    # Create the already processed samples along with their
+    # ComputationalResults and ComputedFiles. They don't need
+    # original files for this test because we aren't going to run
+    # salmon quant on them.
```
kurtwheeler: I think you copied this comment from the other test because you do create original files here.

foreman/data_refinery_foreman/foreman/management/commands/test_rerun_salmon_old_samples.py
```diff
@@ -0,0 +1,288 @@
...
+    OrganismIndex,
+    Processor,
+    ProcessorJob,
+    ProcessorJobOriginalFileAssociation,
+    Sample,
+    SampleResultAssociation,
+)
+from data_refinery_foreman.foreman.management.commands.rerun_salmon_old_samples import update_salmon_all_experiments
+
+def setup_experiment(complete_accessions: List[str], incomplete_accessions: List[str]) -> Dict:
```
kurtwheeler: Can you change the names of the parameters here to `new_version_accessions` and `old_version_accessions` or something like that?

kurtwheeler:
Ok, this job is only going to run once and there's only 632 experiments so I think that with that sleep you added the imperfection of this query is acceptable and we should just get it out!

Issue #1477, opened 2019-08-12
Avoid reindexing the full db in elastic search every 20 mins


arielsvn:
### Context

We reindex the entire db of experiments every 20 mins.

### Problem or idea

This impacts negatively the performance of the DB.

![image](https://user-images.githubusercontent.com/1882507/62874300-fdee4500-bcee-11e9-8b00-1ec569296ef4.png)

The biggest query that we do while reindexing calculates the accession codes of the downloadable samples for each experiment (we use this to calculate the totals in the search filters).

https://github.com/AlexsLemonade/refinebio/blob/dev/common/data_refinery_common/models/models.py#L427-L435

This has to be repeated once for each experiment.

### Solution or next step

Not sure if there's a way to improve that query.

Alternatively, we could only reindex the experiments that were modified in the 20 mins, or that have a sample that was modified in that interval. That way we would avoid throwing away the entire index.

cgreene:
This is done with #1518 - note #1555 arises because of this fix.

PR #1518, opened 2019-08-21
Update ES incrementally instead of rebuilding indices from scratch every time

kurtwheeler:
## Issue Number

Hmm, I don't know that we ever actually made an issue for it

We have about a 5-10% database spike every minute from rebuilding all the indices for elasticsearch. For whatever reason, it starts over and reindexes every model rather than just the ones that changed. It's honestly probably because it's hard to know what field to use for the cutoff. Fortunately all of our models have a `last_modified` field so we can just use that!

## Types of changes

## Functional tests

I ran the command with a bunch of old experiments and saw:

```
Indexing 0 'Experiment' objects
```

I then updated one of the experiments' last_modified field and ran it again and saw:
```
Indexing 1 'Experiment' objects
```

## Checklist

### DIFF
--- a/api/data_refinery_api/management/commands/update_es_index.py
+++ b/api/data_refinery_api/management/commands/update_es_index.py
@@ -20,7 +20,7 @@ class Command(BaseCommand):
         We have updated it so that it will do incremental updates
         rather than looping over the full queryset every time.
         """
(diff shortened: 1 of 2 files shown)


cgreene:
Issue is #1477

PR #1525, opened 2019-08-22
[HOTFIX] Deploy salmon rerun fix and scale up larger

kurtwheeler:
## Issue Number

#1524 
#1523


cgreene:
Also looks like it gets rid of our smallest instance type.

PR #1548, opened 2019-08-29
Small improvements to be able to queue up lots of work

kurtwheeler:
## Issue Number

N/A 3 small changes that we realized were necessary.

This makes the dashboard job stats use the same cutoff date as the Foreman.

This also fixes the logic in needs_downloading that caused the salmon rerun jobs to not get queued.

Finally, this bumps the downloader job max up to 1000 since the external sources seem to be able to handle the volume we're running at and we've been having trouble keeping a queue populated on all 11 instances.

## Types of changes

## Functional tests

I tested the stats endpoint and I made sure that the conditional I changed gave the expected result when run on prod data (I execed onto the foreman docker container and ran it against a sample that didn't get run for this reason).


kurtwheeler:
https://circleci.com/gh/AlexsLemonade/refinebio/21268?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link was failing because I made all salmon jobs unhappy that they weren't running tximport. That last commit should fix that.

Issue #1560, opened 2019-09-03
Don't fail when detecting column header for illumina sample


arielsvn:
### Context

We have `4287` samples that failed because we weren't able to find an id column in their headers. For example for sample [GSM2677584](https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GSM2677584) we failed with the error:

`Could not find any ID column in headers ['ID_REF ', 'SAMPLE 1', 'Detection Pval', 'SAMPLE 2', 'Detection Pval', 'SAMPLE 3', 'Detection Pval', 'SAMPLE 4', 'Detection Pval', 'SAMPLE 5', 'Detection Pval', 'SAMPLE 6', 'Detection Pval', 'SAMPLE 7', 'Detection Pval', 'SAMPLE 8', 'Detection Pval', 'SAMPLE 9', 'Detection Pval', 'SAMPLE 10', 'Detection Pval', 'SAMPLE 11', 'Detection Pval', 'SAMPLE 12', 'Detection Pval', 'SAMPLE 13', 'Detection Pval', 'SAMPLE 14', 'Detection Pval', 'SAMPLE 15', 'Detection Pval', 'SAMPLE 16', 'Detection Pval', 'SAMPLE 17', 'Detection Pval', 'SAMPLE 18', 'Detection Pval'] for file /home/user/data_store/processor_job_1310904/GSE100301_non-normalized.txt.sanitized`

https://github.com/AlexsLemonade/refinebio/issues/86#issuecomment-379308817

### Problem or idea

Seems like we are not parsing the first column correctly, there's trailing whitespace on `'ID_REF '`, and that's causing the job to fail in 

https://github.com/AlexsLemonade/refinebio/blob/dev/workers/data_refinery_workers/processors/illumina.py#L115-L121

There're more samples with the same error (~14k), however, the fix doesn't seem to be as easy for the others.

<details><summary>queries</summary>

```sql
select count(*)
from (
    select samples.accession_code, (
        select processor_jobs.failure_reason
        from original_file_sample_associations, processorjob_originalfile_associations, processor_jobs
        where samples.id=original_file_sample_associations.sample_id
            and original_file_sample_associations.original_file_id=processorjob_originalfile_associations.original_file_id
            and processorjob_originalfile_associations.processor_job_id=processor_jobs.id
        order by processor_jobs.start_time desc           
        limit 1
      ) as job_failure_reason
    from samples
    where samples.is_processed='f'
  ) tt 
[...]

PR #1561, opened 2019-09-03
Strip whitespaces when reading columns in illumina files

arielsvn:
## Issue Number

#1560 

Strips whitespaces when parsing tsv columns in ilumina files.

## Types of changes

## Functional tests

None

## Checklist


arielsvn:
I didn't actually tested running it, but looked into some of the original files and it seems the whitespace is only on the header.

image

The rows don't have any whitespaces between them.

arielsvn:
Will include this change in https://github.com/AlexsLemonade/refinebio/pull/1568

PR #1563, opened 2019-09-04
Update README.md

arielsvn:
Updates path to `common.sh` script

### DIFF
--- a/README.md
+++ b/README.md
@@ -573,7 +573,7 @@ variable NOMAD_ADDR to include the IP address of your development
 machine. This can be done with:
 
 ```bash
-source common.sh && export NOMAD_ADDR=http://$(get_ip_address):4646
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -573,7 +573,7 @@ variable NOMAD_ADDR to include the IP address of your development
 machine. This can be done with:
 
 ```bash
-source common.sh && export NOMAD_ADDR=http://$(get_ip_address):4646
+source scripts/common.sh && export NOMAD_ADDR=http://$(get_ip_address):4646
```
davidsmejia: Maybe it should be ```./scripts/common.sh```

arielsvn: Sure, let's change it. I'm not sure which one is more appropriate, the one above also worked.

```suggestion
source ./scripts/common.sh && export NOMAD_ADDR=http://$(get_ip_address):4646
```

Issue #1564, opened 2019-08-18
Total length of queues counts are off in the dashboard


kurtwheeler:
![image](https://user-images.githubusercontent.com/4026833/63226795-b82cf300-c1ac-11e9-8bff-70c981043e3b.png)

```
data_refinery=> select now(), count(*) from downloader_jobs where start_time is NULL and success is null and created_at > '2019-06-05 00:00:00'::timestamp;
              now              | count 
-------------------------------+-------
 2019-08-18 15:40:56.139861+00 | 38473
(1 row)

data_refinery=> select now(), count(*) from processor_jobs where start_time is NULL and success is null and created_at > '2019-06-05 00:00:00'::timestamp;
              now              | count 
-------------------------------+-------
 2019-08-18 15:41:36.622155+00 | 19325
(1 row)
```
The processor jobs number isn't really too far off, but you can see that the downloader jobs have an extra 22k or so in the dashboard. I used 2019-06-05 because that's the cutoff date the foreman uses. Perhaps we can turn that into an environment variable that can be used by the API as well?

Also the numbers overlap on the pie charts sometimes and it makes it really hard to read them.

PR #1565, opened 2019-09-04
filter jobs counts and averages if there is a range param 

davidsmejia:
## Issue Number

#1564 

Filter out the numbers of jobs and job aggregates so they are only from the range requested.

## Methods

Adds an empty filter to all requests and limits to the time range when there is a rang_param passed in. This is then used for getting the counts and the avg times.

## Types of changes

## Functional tests

None yet.

## Checklist


api/data_refinery_api/views.py
```diff
@@ -1001,8 +1013,12 @@ def _get_job_stats(cls, jobs, range_param):
                                       created_at__gt=JOB_CREATED_AT_CUTOFF)),
         )
         # via https://stackoverflow.com/questions/32520655/get-average-of-difference-of-datetime-fields-in-django
-        result['average_time'] = jobs.filter(start_time__isnull=False, end_time__isnull=False, success=True).aggregate(
-                average_time=Avg(F('end_time') - F('start_time')))['average_time']
+        result['average_time'] = jobs.filter(start_filter).filter(
+            start_time__isnull=False,
+            end_time__isnull=False,
+            success=True).aggregate(
```
kurtwheeler: This is my preferred style for method chaining across lines in python, not sure that's gospel or anything but I think we're moderately consistent about it: https://github.com/AlexsLemonade/refinebio/blob/dev/workers/data_refinery_workers/downloaders/geo.py#L320

davidsmejia: yup! 👍

Issue #1276, opened 2019-04-24
Some hgu133XXX samples may have been processed with wrong CDFs


kurtwheeler:
### Context

@cansav09 killing it again at detecting data integrity issues!!!

https://api.refine.bio/experiments/?accession_code=GSE12417 has 3 platforms, hgu133a, hgu133b, and hgu133plus2. We used to be much worse at detecting microarray platforms, we would just assume that the first one we found for the experiment was the platform for all the samples because of GEO's metadata practices.

### Problem or idea

 We detected that GSE12417 was hgu133a so we tried to run all the samples with that platform. It looks like the hgu133b samples failed because there wasn't enough probe overlap, but hgu133plus2 contains all the hgu133a probes so those seem to have succeeded enough to be marked successful.

### Solution or next step

We've since fixed our affy pipeline so that it will look at the .CEL file and determine the correct platform so erunning all the microarray data should cause the samples that didn't get processed to be processed with the correct platform. However this hgu133 overlap issue could have affected other experiments so those experiments and this one should be forced to be reprocessed so we can have them labelled and processed with the correct platform.

kurtwheeler:
I forgot to note this while originally posting, but samples with accessions between GSM316661 - GSM316664 had a platform of hgu133plus2 but still were successfully processed. This may help with debugging.

PR #1566, opened 2019-09-05
Adds management command to correct GEO samples with bad metadata

kurtwheeler:
## Issue Number

#1276 

It's rather impossible to tell by only looking at the database which samples are wrong and which are not since GEO has mixed-platform experiments. So the only way I could think to figure out which samples had the incorrect platform was to check GEO.

As #1276 mentions, some of these mis-labeled samples were successfully processed so in order to retrigger processing and the correction of metadata, I decided to just unsurvey the whole experiment and resurvey it. This may result in a decent number of reprocessing of samples that aren't incorrect, but these are all microarray samples so they'll go quick and I think this will be a cleaner solution than trying to both correct the metadata and then convincing the system to allow these samples to be reprocessed.

## Types of changes

## Functional tests

I can't really recreate the situation I'm trying to correct except by manually changing stuff in the database which isn't really much different than the unit tests I wrote.

My plan is to do a dry-run in staging, run it without dry-run in staging (assuming there's actually some samples there that have this issue), then do a dry run in prod, then finally assuming that it looks good do the actual run in prod.

## Checklist

### DIFF
--- a/foreman/data_refinery_foreman/foreman/management/commands/correct_affy_cdfs.py
+++ b/foreman/data_refinery_foreman/foreman/management/commands/correct_affy_cdfs.py
@@ -54,7 +54,6 @@ class Command(BaseCommand):
                         break
 
                 if wrong_platform:
-                    print(type(options["dry_run"]))
                     if options["dry_run"]:
                         logger.info("Would have re-surveyed experiment with accession code %s",
                                     experiment.accession_code)
--- a/foreman/data_refinery_foreman/foreman/management/commands/correct_affy_cdfs.py
+++ b/foreman/data_refinery_foreman/foreman/management/commands/correct_affy_cdfs.py
@@ -1,0 +1,73 @@
+"""
+This command checks if samples from GEO have incorrect platform
+information and re-surveys them.
+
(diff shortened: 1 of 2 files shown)


foreman/data_refinery_foreman/foreman/management/commands/correct_affy_cdfs.py
```diff
@@ -0,0 +1,73 @@
...
+                for sample in samples:
+                    gpl = gse.gsms[sample.accession_code].metadata['platform_id'][0]
+                    internal_accession = get_internal_microarray_accession(gpl)
+                    if internal_accession != sample.platform_accession_code:
+                        wrong_platform = True
+                        break
+
+                if wrong_platform:
+                    print(type(options["dry_run"]))
+                    if options["dry_run"]:
```
arielsvn: I like this option!

kurtwheeler:
Mmm good call, it should indeed have a sleep.

kurtwheeler:
Wait, I don't really think so. We don't do it in parallel. GEO should be able to handle one request at a time.

PR #1607, opened 2019-09-16
Clean up after GEOparse since they won't do it...

kurtwheeler:
## Issue Number

#1276 

GEO provides metadata in files... #BecauseBio

So GEOparse has to download the files and read them in to get the metadata. At this point GEOparse could clean up those files, but it won't. This probably isn't much of an issue for most people because they aren't downloading hundreds of thousands of these. We do.

Therefore all this PR does is cleanup the metadata files that GEOparse litters. I think this is what went wrong when we ran this command before because I reran the command as a dry run, exec'd onto the docker container, and it had lots of `_family.soft.gz` files. I saw the disk slowing filling up.

## Types of changes

## Functional tests

The unit tests run. I don't have affy samples that need correcting though so it's a tad tricky to test locally.

## Checklist

### DIFF
--- a/foreman/data_refinery_foreman/foreman/management/commands/correct_affy_cdfs.py
+++ b/foreman/data_refinery_foreman/foreman/management/commands/correct_affy_cdfs.py
@@ -11,6 +11,7 @@ the experiment and reprocess it correctly.
 
 from django.core.management.base import BaseCommand
 import GEOparse
+import shutil
 
 from data_refinery_common.models import Experiment, Sample
 from data_refinery_common.logging import get_and_configure_logger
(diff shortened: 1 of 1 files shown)


foreman/data_refinery_foreman/foreman/management/commands/correct_affy_cdfs.py
```diff
@@ -40,30 +44,36 @@ def handle(self, *args, **options):
...
+                        else:
+                            logger.info("Re-surveying experiment with accession code %s",
+                                        experiment.accession_code)
+
+                            purge_experiment(experiment.accession_code)
+
+                            queue_surveyor_for_accession(experiment.accession_code)
+                finally:
+                    # GEOparse downloads files here and never cleans them up! Grrrr!
+                    download_path = GEO_TEMP_DIR + experiment.accession_code + '_family.soft.gz'
```
arielsvn: Does GEO download this file for all experiments? I'm asking because in the GEO downloader we have to handle several types of files.

kurtwheeler: I think that's because the downloader is dealing with more than just metadata.

PR #1609, opened 2019-09-16
Add option to include quant files in dataset

### LINKED ISSUE #1602: API: Expose quant files
### Context

With #1587 it became clear that we should allow computational folks to access these.

### Problem or idea

There is no way to add `quant.sf` files to a collection for downloading. We probably want to expose these only via the API (perhaps, in the front-end, we could provide information for the samples with only quant files by linking to the newly written quant file documentation).

### Solution or next step

We should modify the API to allow folks to retrieve quant files if they verify that they really, really want them.

arielsvn:
## Issue Number

close #1602 

Adds an additional option to the dataset object: `quant_sf_only` to only include the `quant.sf` files in the downloaded dataset.

## Types of changes

## Functional tests

Added local test.

## Checklist


kurtwheeler:
I see this is still a draft, but I don't think `include_quant_files` is a good name for the field. One thing is quant files is apparently a kallisto term instead of salmon. I think @jaclyn-taroni prefers to call them quantSF files. The other thing about the field name is that it says "include" but really instead of just additionally including quantSF files you're instead requesting to ONLY receive quantSF files. @jaclyn-taroni do you think `quant_sf_only` would be an intuitive name for the field?

workers/data_refinery_workers/processors/smasher.py
```diff
@@ -502,10 +513,18 @@ def _smash(job_context: Dict, how="inner") -> Dict:
 
         # Once again, `key` is either a species name or an experiment accession
         for key, input_files in job_context['input_files'].items():
+            # Check if we need to copy the quant.sf files
+            if job_context['dataset'].quant_sf_only:
+                outfile_dir = smash_path + key + "/"
+                os.makedirs(outfile_dir, exist_ok=True)
+                sync_quant_files(outfile_dir, input_files, job_context)
+                # we ONLY want to give quant sf files to the user if that's what they requested
+                continue
```
kurtwheeler: Hmm, have you fully considered the implications of terminating this loop here? I can at least tell that `num_samples` will stay at 0, so we'll report to the user in the metadata that we've served 0 samples regardless of what we've actually given them. There's probably others, but even that one alone is enough to make me think the metadata might not look so good.

arielsvn:
I looked into the metadata, here's what results when generating a dataset with a single experiment and a single sample:

```
{
    "created_at": "2019-09-18T17:51:47",
    "experiments": {
        "GSE51088": { ... }
    },
    "num_experiments": 1,
    "num_samples": 1,
    "quant_sf_only": true,
    "samples": {
        "GSM1237818": { ... }
    }
}
```

I removed the other fields: `aggregate_by`, `scale_by`, `non_aggregated_files`, `ks_statistic`, `ks_pvalue`, `ks_warning`, `quantile_normalized` as they don't seem relevant when the user only requests the `quant.sf` files. Do you agree @kurtwheeler ?

jaclyn-taroni:
I agree FWIW @arielsvn 😄

workers/data_refinery_workers/processors/test_smasher.py
```diff
@@ -527,6 +527,75 @@ def test_no_smash_all_diff_species(self):
...
+        pjda.save()
+
+        final_context = smasher.smash(job.pk, upload=False)
+        
+        # Check that the sample was really generated
+        self.assertTrue(os.path.exists(final_context['output_dir'] + '/HOMO_SAPIENS/GSM1237818_quant.sf'))
+        self.assertTrue(final_context['metadata']['quant_sf_only'])
+        self.assertEqual(final_context['metadata']['num_samples'], 1)
+        self.assertEqual(final_context['metadata']['num_experiments'], 1)
+
```
kurtwheeler:
```suggestion
self.assertTrue('aggregate_by' not in final_context['metadata'])
```

workers/data_refinery_workers/processors/test_smasher.py
```diff
@@ -527,6 +527,75 @@ def test_no_smash_all_diff_species(self):
...
+        pjda.save()
+
+        final_context = smasher.smash(job.pk, upload=False)
+        
+        # Check that the sample was really generated
+        self.assertTrue(os.path.exists(final_context['output_dir'] + '/HOMO_SAPIENS/GSM1237818_quant.sf'))
+        self.assertTrue(final_context['metadata']['quant_sf_only'])
+        self.assertEqual(final_context['metadata']['num_samples'], 1)
+        self.assertEqual(final_context['metadata']['num_experiments'], 1)
+self.assertTrue('aggregate_by' not in final_context['metadata'])
```
arielsvn:
```suggestion
        self.assertTrue('aggregate_by' not in final_context['metadata'])
```

workers/data_refinery_workers/processors/test_smasher.py
```diff
@@ -528,6 +528,75 @@ def test_no_smash_all_diff_species(self):
...
+        pjda.save()
+
+        final_context = smasher.smash(job.pk, upload=False)
+        
+        # Check that the sample was really generated
+        self.assertTrue(os.path.exists(final_context['output_dir'] + '/HOMO_SAPIENS/GSM1237818_quant.sf'))
+        self.assertTrue(final_context['metadata']['quant_sf_only'])
+        self.assertEqual(final_context['metadata']['num_samples'], 1)
+        self.assertEqual(final_context['metadata']['num_experiments'], 1)
+        self.assertTrue('aggregate_by' not in final_context['metadata'])
```
kurtwheeler:
```suggestion
        self.assertTrue('aggregate_by' not in final_context['metadata'])
        
```

Issue #1485, opened 2019-08-13
Parallelize recreating compendia for all organisms


kurtwheeler:
### Context

Our compendia code either creates one compendium or all the compendia... sequentially... in the same job. So if one job fails they all fail and if we have to restart we do them in the same order rebuilding any that were already successful.

### Problem or idea

We should switch over to the dispatcher job pattern we use for survey jobs.

### Solution or next step

We should have a compendia job spawn other compendia jobs if it is instructed to run more than one of them. While we're at it, it might be nice to add an option to run a list of them rather than just one or all.

PR #1610, opened 2019-09-17
Parallelize recreating compendia for all organisms

kurtwheeler:
## Issue Number

#1485 

Makes create_compendia dispatch jobs if requested to create more than one. This also makes create_compendia accept a list of comma separated organism names and dispatch jobs for each of them.

## Types of changes

## Functional tests

 I've passed a list of organisms to a compendia job through nomad.

## Checklist

### DIFF
--- a/workers/data_refinery_workers/processors/management/commands/create_compendia.py
+++ b/workers/data_refinery_workers/processors/management/commands/create_compendia.py
@@ -26,7 +26,7 @@ def create_job_for_organism(organism=Organism):
     Fetch all of the experiments and compile large but normally formated Dataset.
     """
     data = {}
-    experiments = Experiment.objects.filter(id__in=(ExperimentOrganismAssociation.objects.filter(organism=organism)).values('experiment'))
+    experiments = Experiment.objects.filter(organisms=organism).prefetch_related('samples')
     for experiment in experiments:
         data[experiment.accession_code] = list(experiment.samples.filter(organism=organism).values_list('accession_code', flat=True))
 
--- a/workers/data_refinery_workers/processors/management/commands/create_compendia.py
+++ b/workers/data_refinery_workers/processors/management/commands/create_compendia.py
@@ -13,12 +13,40 @@ from data_refinery_common.models import (
     Dataset,
     ProcessorJobDatasetAssociation,
     ExperimentOrganismAssociation,
-    OrganismIndex,
     ExperimentSampleAssociation
 )
 from data_refinery_workers.processors import create_compendia
 
 logger = get_and_configure_logger(__name__)
+
+
+def create_job_for_organism(organism=Organism):
+    """Returns a compendia job for the provided organism.
+
+    Fetch all of the experiments and compile large but normally formated Dataset.
+    """
+    data = {}
+    experiments = Experiment.objects.filter(id__in=(ExperimentOrganismAssociation.objects.filter(organism=organism)).values('experiment'))
+    for experiment in experiments:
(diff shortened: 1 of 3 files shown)


workers/data_refinery_workers/processors/management/commands/create_compendia.py
```diff
@@ -13,57 +13,70 @@
...
-            dset.quantile_normalize = False
-            dset.save()
-
-            pjda = ProcessorJobDatasetAssociation()
-            pjda.processor_job = job
-            pjda.dataset = dset
-            pjda.save()
-
-            final_context = create_compendia.create_compendia(job.id)
+            organisms = options["organisms"].upper().replace(" ", "_").split(",")
```
arielsvn: we probably don't need to do `replace(" ", "_")`, I think the command parameters can't have spaces

workers/data_refinery_workers/processors/management/commands/create_compendia.py
```diff
@@ -13,57 +13,70 @@
...
 logger = get_and_configure_logger(__name__)
 
 
+def create_job_for_organism(organism=Organism):
+    """Returns a compendia job for the provided organism.
+
+    Fetch all of the experiments and compile large but normally formated Dataset.
+    """
+    data = {}
+    experiments = Experiment.objects.filter(id__in=(ExperimentOrganismAssociation.objects.filter(organism=organism)).values('experiment'))
```
arielsvn: Is it the same to do `Experiment.objects.filter(organisms=organism)` ?

workers/data_refinery_workers/processors/management/commands/create_compendia.py
```diff
@@ -13,57 +13,70 @@
...
+        If --organism is supplied will immediately create a compedium
+        for it. If not a new job will be dispatched for each organism
+        with enough microarray samples.
         """
-
-        dataset_ids = []
-
-        if options["organism"] is None:
+        if options["organisms"] is None:
             all_organisms = Organism.objects.all()
```
arielsvn: Where do we check that the organisms have enough microarray samples?

workers/data_refinery_workers/processors/management/commands/create_compendia.py
arielsvn: And since we're here, does it makes sense to `prefetch_related('samples')`?

workers/data_refinery_workers/processors/management/commands/create_compendia.py
kurtwheeler: In the compendia job.

workers/data_refinery_workers/processors/management/commands/create_compendia.py
kurtwheeler: Why not? `./workers/run_command.sh -i compendia "python3 manage.py create_compendia --organisms 'DANIO_RERIO,HOMO SAPIENS'"` works

workers/data_refinery_workers/processors/management/commands/create_compendia.py
arielsvn: Is it worth checking here to avoid dispatching these jobs?

workers/data_refinery_workers/processors/management/commands/create_compendia.py
arielsvn: nice! I think I was calling the commands differently when I had the one with the regular expression

workers/data_refinery_workers/processors/management/commands/create_compendia.py
kurtwheeler: Mmmm I think it might be more trouble than it's worth.

Issue #1603, opened 2019-09-16
Create quantpendia for all organisms


kurtwheeler:
### Context

Will build off of https://github.com/AlexsLemonade/refinebio/issues/1601

### Problem or idea

#1601 made it so we can run smasher jobs on just quant files. Now we want to run a smasher job for the full organism for all organisms.

### Solution or next step

This is a two parter:
  1. Make it so compendia jobs can run for just quant files.
  2. Run compendia jobs for all organism.

arielsvn:
We have quantpendias for all organisms now, except mouse, but we're building it for #1587.

PR #1619, opened 2019-09-18
Extend create compendium command to generate quantpendia

kurtwheeler:
## Issue Number

#1603

We want to make a quantpendia for every organism. This command will now do so if the `quant-sf-only` field is set to True.

## Types of changes

## Functional tests

I've run this command from the command line and through Nomad setting the flag to various values and making sure that it gets pass through to the dataset object correctly.

## Checklist


arielsvn:
Are you marking the generated compendia as a quantcompendia somehow? How can we differentiate this one from our regular compendia?

I see that it will get `is_compendia=True` here

https://github.com/AlexsLemonade/refinebio/blob/dev/workers/data_refinery_workers/processors/create_compendia.py#L320

Issue #1572, opened 2019-09-09
Some organisms are missing QN targets that should have them


kurtwheeler:
### Context

We don't have a rat QN target. There's a lot of rat samples so we should be able to make one.

Remember, we can create a QN target for any organism for which we have processed 100 samples from the same affy platform.

### Problem or idea

Jackie has provided the following list of organisms, their largest platform, and how many samples there are for that platform. We would expect most if not all of these organisms to have a QN target.

[affymetrix_platforms_for_QN_reference.txt](https://github.com/AlexsLemonade/refinebio/files/3591306/affymetrix_platforms_for_QN_reference.txt)

### Solution or next step

Look into why we don't have a QN target for rat or other organisms that we would expect to have one for and generate them.

cgreene:
🐀 ➡️ 📈

PR #1625, opened 2019-09-19
Make the foreman retry QN jobs

kurtwheeler:
## Issue Number

#1572 

The foreman wasn't retrying QN jobs because they didn't have a volume index. This fixes that by letting the foreman retry jobs without volume indices. To avoid this suddenly retrying lots of jobs that were previously filtered out I've moved the job_cutoff date up.

Also set the QN s3 env var in the other QN job. Yes we should have two because one gets dispatched to with a processor job already created and the other will just go for any supplied organism.

## Types of changes

## Functional tests

The unit tests cover this rather well I think.

## Checklist

### DIFF
--- a/foreman/data_refinery_foreman/foreman/main.py
+++ b/foreman/data_refinery_foreman/foreman/main.py
@@ -9,6 +9,7 @@ from django.conf import settings
 #from django.core.paginator import Paginator
 from data_refinery_foreman.foreman.performant_pagination.pagination import PerformantPaginator as Paginator
 from django.db import transaction
+from django.db.models.expressions import Q
 from django.utils import timezone
 from functools import wraps
 from nomad import Nomad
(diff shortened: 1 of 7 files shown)


foreman/data_refinery_foreman/foreman/main.py
```diff
@@ -799,11 +800,11 @@ def retry_failed_processor_jobs() -> None:
...
-        success=False,
-        retried=False,
-        no_retry=False,
-        volume_index__in=active_volumes,
-        created_at__gt=JOB_CREATED_AT_CUTOFF
+        Q(success=False,
+          retried=False,
+          no_retry=False,
+          created_at__gt=JOB_CREATED_AT_CUTOFF) \
+        and (Q(volume_index__isnull=True) | Q(volume_index__in=active_volumes))
```
arielsvn: Should we use `&` instead of `and`?

```suggestion
        & (Q(volume_index__isnull=True) | Q(volume_index__in=active_volumes))
```

PR #1629, opened 2019-09-20
Access quant files through computational results

arielsvn:
## Issue Number

#1602 

Seems that we don't associate `ComputedFiles` with `quant.sf` results with the samples, instead this changes the code to fetch them through the `ComputationalResults` relations.

## Types of changes

## Functional tests

Tested locally

## Checklist

### DIFF
--- a/workers/data_refinery_workers/processors/smasher.py
+++ b/workers/data_refinery_workers/processors/smasher.py
@@ -50,7 +50,6 @@ def log_state(message):
     cpu = psutil.cpu_percent()
     logger.debug("cpu:%s - ram:%s -  %s", cpu, ram, message)
 
-
 def _prepare_files(job_context: Dict) -> Dict:
     """
     Fetches and prepares the files to smash.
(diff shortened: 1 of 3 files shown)


workers/data_refinery_workers/processors/smasher.py
```diff
@@ -467,15 +466,15 @@ def sync_quant_files(output_path, files_sample_tuple, job_context: Dict):
     """ Takes a list of ComputedFiles and copies the ones that are quant files to the provided directory.
         Returns the total number of samples that were included """
     num_samples = 0
-    for (computed_file, sample) in files_sample_tuple:
+    for (_, sample) in files_sample_tuple:
+        latest_computed_file = sample.get_most_recent_quant_sf_file()
```
kurtwheeler: If I'm not mistaken, we made this loop over a tuple so that we could be more performant about querying the database by getting the samples and their computed files at the same time. It looks like this takes us back to one query per sample, when we could be looking at 100's of thousands for the human/mouse quantpendias.

arielsvn: Do you have any recommendations on how we can do it? Not sure there's a way around making a query per sample, I'm gonna try to think about it.

Each one of these queries should be straightforward, but we could be making a lot of them.

```sql
SELECT "computed_files".* 
FROM "computed_files" 
WHERE ("computed_files"."filename" = 'quant.sf' AND "computed_files"."result_id" IN (
    SELECT U0."id" 
    FROM "computational_results" U0 INNER JOIN "sample_result_associations" U1 ON (U0."id" = U1."result_id")
    WHERE U1."sample_id" = 1521073)
) ORDER BY "computed_files"."created_at" DESC;
```

I ran it in prod and got `Execution time: 3.775 ms`

kurtwheeler: Could it all be done in one query with something roughly like
```
Sample.objects.filter(
  accession_code__in=samples
).prefetch_related(
  'results__computed_files'
).filter(
  result__computed_file.name='quant.sf'
)
```

kurtwheeler: I mean that would have to be done in `_prepare_files()` I think.

arielsvn: hmm but isn't that more costly? we would be fetching in memory all samples, all their computational results and also all their computed files. Seems like a premature optimization.

Removing the part where we were making a query for each sample had a big impact on the compendia generation performance?

kurtwheeler: That's a good point. We should probably rework this to use pagination for that reason.

I don't think we can make 150k queries in a loop without hurting the DB. Each one has overhead that adds up when you make too many in rapid succession. Generally queries in a long loop without a sleep are a bad idea.

arielsvn: Talked with @kurtwheeler about this, we're going to try it on staging to see how it behaves. The db queries should not be executed successfully because the files need to be downloaded in between each one of them.

Issue #1604, opened 2019-09-16
Downloaded Data Numbers - API


cgreene:
### Context

Sprint planning meeting.

### Problem or idea

We do not currently have a way to get `1 PB -Over a petabyte of raw data will be processed` from the about page based on actual downloaded data.

### Solution or next step

We should add an API endpoint to calculate this.

arielsvn:
New api is still timing out for this https://api.refine.bio/v1/stats-about/

arielsvn:
Discussed with @kurtwheeler  and @davidsmejia, two ideas came up:

- Increase the timeout for that endpoint
- Take another look at the queries to see if we can optimize them further

cgreene:
Can we hard code for now? Thinking about what gets us there asap.

PR #1640, opened 2019-09-23
Arielsvn/1604 about page numbers

arielsvn:
## Issue Number

#1604 

Adds a new endpoint `/stats-about` with the stats we need in the about page. I tried to simplify the queries so that they don't time out.

Also added a new index on `ComputedFile.filename`.

## Types of changes

## Functional tests

Tested on staging.

## Checklist


api/data_refinery_api/views.py
```diff
@@ -845,6 +845,28 @@ class ProcessorJobList(generics.ListAPIView):
...
 ###
 
+class AboutStats(APIView):
+    """ Returns general stats for the site, used in the about page """
+    @method_decorator(cache_page(10 * 60))
+    def get(self, request, version, format=None):
+        result = {
+            # count the total number of samples that are processed or that have a quant.sf file associated with them
+            'samples_available': Sample.objects.filter(is_processed=True).count() \
+                + Sample.objects.filter(is_processed=False, technology='RNA-SEQ', results__computedfile__filename='quant.sf').distinct().count(),
```
kurtwheeler: Could you break this up a bit more? There's a lot going on in this one (but split into two) line of code. Maybe like init 2 variables and set `samples_available` to the sum in this line or something?

api/data_refinery_api/views.py
```diff
@@ -845,6 +845,28 @@ class ProcessorJobList(generics.ListAPIView):
...
+            'total_size_in_bytes': OriginalFile.objects.aggregate(total_size=Sum('size_in_bytes'))['total_size'],
+            # count organisms with qn targets or that have at least one sample with quant files
+            'supported_organisms': Organism.objects.filter(qn_target__isnull=False).count() \
+                +  Organism.objects.filter(
+                    qn_target__isnull=True,
+                    sample__is_processed=False,
+                    sample__technology='RNA-SEQ',
+                    sample__results__computedfile__filename='quant.sf'
+                  ).distinct().count(),
```
kurtwheeler: A similar comment about breaking this up. Combining the full calculation of these stats into a single expression makes them less readable I think.

arielsvn:
Just addressed the comments, let me know what do you think @kurtwheeler

Issue #1587, opened 2019-09-11
.quant file request


cgreene:
### Context

I was contacted by Maria Chikina ( @mchikina ), who would like access to a large number of .quant files for a project.

### Problem or idea

It is not currently possible to get quant files.

### Solution or next step

We could tackle this in a couple ways:
* The API could allow one to request quant files for RNA-seq samples. Most of our web-based users probably won't want this functionality, so we probably wouldn't initially expose it via the React UI.
* We could make "compendia" of quant files. We'd put them in a directory and zip them, along with the metadata.

mchikina:
Also, any metadata (like tissue, etc) would be useful.

jaclyn-taroni:
@mchikina are you looking specifically for human samples? If so, you can likely leverage [MetaSRA](http://metasra.biostat.wisc.edu/) for the tissue labels.

mchikina:
This is great. Hopefully the tissue information will come out automatically but it would be nice to have something to verify.

arielsvn:
@mchikina we now have the ability to generate datasets that include `quant.sf` files. Do you need one with human samples? or any other type of samples?

mchikina:
Yes. Human would be a top priority for me. I am also interested in mouse.

arielsvn:
@mchikina you can download all of our human `quant.sf` files [here](https://s3.amazonaws.com/data-refinery-s3-circleci-prod/HOMO_SAPIENS_1_1574170428.zip) (it's `256gb` compressed).

arielsvn:
All our mouse `quant.sf` can be downloaded [here](https://s3.amazonaws.com/data-refinery-s3-circleci-prod/MUS_MUSCULUS_1_1574233541.zip), this one is `185gb` compressed.

I'm closing this issue, feel free to re-open or file a new one if you need something else.

PR #1642, opened 2019-09-24
Increase smasher size

arielsvn:
## Issue Number

#1587 

One of the jobs to generate the quantcompendia failed with 50k human samples because it ran out of space.

The total size of the quantfiles that we have is `5000GB`. This increases the smasher to `3000GB` to be prepared for when we generate the datasets in parallel. 

```sql
data_refinery=> select sum(size_in_bytes) from computed_files where filename='quant.sf';
      sum      
---------------
 5603241992755
(1 row)
```

## Types of changes

## Functional tests

None

## Checklist

### DIFF
--- a/infrastructure/instances.tf
+++ b/infrastructure/instances.tf
@@ -274,7 +274,7 @@ resource "aws_instance" "smasher_instance" {
   # Should be more than enough to store 2 jobs worth of data at a time.
   root_block_device = {
     volume_type = "gp2"
-    volume_size = 100
+    volume_size = 3000
   }
 }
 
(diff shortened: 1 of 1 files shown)


infrastructure/instances.tf
```diff
@@ -274,7 +274,7 @@ resource "aws_instance" "smasher_instance" {
   # Should be more than enough to store 2 jobs worth of data at a time.
   root_block_device = {
     volume_type = "gp2"
-    volume_size = 100
+    volume_size = 3000
```
arielsvn:
```suggestion
    volume_size = 6000
```

@cgreene recommended making it bigger.

PR #1643, opened 2019-09-24
Filter un-uploaded quant files from tximport. Fix QN bucket perms.

kurtwheeler:
## Issue Number

N/A came up while running #1630 

QN jobs were still failing because of bucket permissions. This should address that.

Additionally we had problems because `needs_processing` and `get_quant_results_for_experiment` weren't disregarding computed files that hadn't been uploaded. This makes those functions filter out computed files that weren't uploaded so we either can reprocess them or realize that we aren't ready for tximport.

## Types of changes

## Functional tests

The unit tests are actually really good for this, especially since I don't have the test data for it.

## Checklist

### DIFF
--- a/common/data_refinery_common/models/models.py
+++ b/common/data_refinery_common/models/models.py
@@ -789,14 +789,24 @@ class OriginalFile(models.Model):
             return False
 
         if sample.source_database == "SRA":
-            for computed_file in sample.computed_files.prefetch_related('result__organism_index').all():
-                    if computed_file.s3_bucket and computed_file.s3_key \
-                       and computed_file.result.organism_index != None \
-                       and computed_file.result.organism_index.salmon_version == CURRENT_SALMON_VERSION:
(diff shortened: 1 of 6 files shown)


common/data_refinery_common/models/models.py
```diff
@@ -789,14 +789,24 @@ def needs_processing(self, own_processor_id=None) -> bool:
...
+            # If there's no smashable file then we should check the quant.sf file.
+            if not computed_file:
+                computed_file = sample.get_most_recent_quant_sf_file()
+
+            # If there's neither a quant.sf file nor a smashable file
+            # then we definitely need to process it.
+            if not computed_file:
+                return True
+
+            if computed_file.s3_bucket and computed_file.s3_key \
```
arielsvn: I feel tempted to suggest to add a new method to `ComputedFile` to check for this. Like, `is_uploaded_to_s3`. We can add it later if we think it's helpful

PR #1647, opened 2019-09-25
paginate svd migration

davidsmejia:
## Issue Number

#1645 

Paginate compendia svd algorithm assignments in migration as to not overload the DB

### DIFF
--- a/common/data_refinery_common/migrations/0038_set_existing_compendia_svd_algorithm.py
+++ b/common/data_refinery_common/migrations/0038_set_existing_compendia_svd_algorithm.py
@@ -1,9 +1,15 @@
 from django.db import migrations, models
+from django.core.paginator import Paginator
 
 def set_existing_compendia_svd_algorithm(apps, schema_editor):
     """ Fixes existing compendia svd algorithm - defaulted NONE corrects to ARPACK """ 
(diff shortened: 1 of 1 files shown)


common/data_refinery_common/migrations/0038_set_existing_compendia_svd_algorithm.py
```diff
@@ -1,9 +1,15 @@
 from django.db import migrations, models
+from django.core.paginator import Paginator
 
 def set_existing_compendia_svd_algorithm(apps, schema_editor):
     """ Fixes existing compendia svd algorithm - defaulted NONE corrects to ARPACK """ 
     ComputedFile = apps.get_model('data_refinery_common', 'ComputedFile')
-    ComputedFile.objects.all().filter(is_compendia=True).update(svd_algorithm="ARPACK")
+
+    paginator = Paginator(ComputedFile.objects.all().order_by('id'), 1000)
```
kurtwheeler: Why aren't you doing the filter in this query you're passing to the paginator? We're going to iterate through millions of files that are not compendia to get a few that are.

davidsmejia: 👍 wasn't sure if that was possible, glad to know it is.. updated accordingly

PR #1650, opened 2019-09-25
paginate svd algorithm default value in migration 

davidsmejia:
## Issue Number

#1645 deploy fail

Migrations were being hung up when creating a new column with a default value on a table that contains a substantial (>1MM?) rows. 

- First update your model, adding the new column, but set ```null=True```
- run ```./script/update_models.sh```
- then create a migration that can more responsibly assign the default values (i.e. using paginator)
- change model: remove ```null=True``` and add ```default="My Value"```
- run ```./scripts/update_models.sh```

This should allow django to migrate without attempting to write all rows at once.

## Types of changes

## Functional tests

Migrations run locally

## Checklist


common/data_refinery_common/migrations/0038_paginate_default_svd_algorithm.py
```diff
@@ -0,0 +1,40 @@
...
+    paginator = Paginator(ComputedFile.objects.all().order_by('id').filter(is_compendia=False), 1000)
+    for page_idx in range(1, paginator.num_pages):
+        for computedFile in paginator.page(page_idx).object_list:
+            computedFile.update(svd_algorithm="NONE")
+
+def set_computed_file_svd_algorithm_arpack(apps, schema_editor):
+    """ Set svd_algorithm to ARPACK for exiting compendia """
+    ### is_compendia=True   
+    ComputedFile = apps.get_model('data_refinery_common', 'ComputedFile')
+    paginator = Paginator(ComputedFile.objects.all().order_by('id').filter(is_compendia=True), 1000)
```
kurtwheeler: Can you either change this to `ComputedFile.objects.filter(is_compendia=True).order_by('id')` or just don't filter at all and go through the entire list with:
```
if computed_file.is_compendia:
  computedFile.update(svd_algorithm="ARPACK")
else:
  computedFile.update(svd_algorithm="NONE")
```

I think django might be smart enough to create that query in a non-horrible manner, but it looks risky. I worry that `ComputedFile.objects.all().order_by('id')` would order the entire table and then filter it. We could order the entire table and iterate over it all since we have to anyway or we could filter it and then order but I'd rather not order the whole table twice.

common/data_refinery_common/migrations/0038_paginate_default_svd_algorithm.py
```diff
@@ -0,0 +1,40 @@
...
+    for page_idx in range(1, paginator.num_pages):
+        for computedFile in paginator.page(page_idx).object_list:
+            computedFile.update(svd_algorithm="ARPACK")
```
kurtwheeler: Can this variable name be computed_file instead of computedFile?

davidsmejia: OYE haha too much javascript 👍

common/data_refinery_common/migrations/0038_paginate_default_svd_algorithm.py
davidsmejia: 👍

davidsmejia: I was curious about this as well.

PR #1654, opened 2019-09-25
batch pagination update

davidsmejia:
## Issue Number

N/A - fixing deploy - migration hangs on batch updates

This is a custom implementation of the pagination util that will allow for update calls on a queryset.

### DIFF
--- a/common/data_refinery_common/migrations/0038_paginate_default_svd_algorithm.py
+++ b/common/data_refinery_common/migrations/0038_paginate_default_svd_algorithm.py
@@ -1,26 +1,36 @@
 from django.db import migrations, models
-from django.core.paginator import Paginator
+
+def batch_update(queryset, batch=1000, **changes):
+    """ Update per batch """
+    is_done = False
+    current = 1
+
+    while not is_done:
+        start = (current - 1) * batch
+        end = start + batch
+        pks = queryset.values_list('pk', flat=True)[start:end][::1]
+        queryset.model.objects.filter(pk__in=pks).update(**changes)
+        is_done = len(pks) <= batch
+        if not is_done:
+            current = current + 1
 
 def set_computed_file_svd_algorithm_arpack(apps, schema_editor):
     """ Set svd_algorithm for computed files - ARPACK when is_compendia"""
     ComputedFile = apps.get_model('data_refinery_common', 'ComputedFile')
-    paginator = Paginator(ComputedFile.objects.filter(is_compendia=True).order_by('id'), 1000)
-    for page_idx in range(1, paginator.num_pages):
-        paginator.page(page_idx).object_list.update(svd_algorithm="ARPACK")
+    arpack_queryset = ComputedFile.objects.filter(is_compendia=True).order_by('id')
+    batch_update(arpack_queryset, svd_algorithm="ARPACK")
 
 def set_computed_file_svd_algorithm_none(apps, schema_editor):
     """ Set svd_algorithm for computed files - None when not is_compendia"""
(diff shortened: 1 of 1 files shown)


common/data_refinery_common/migrations/0038_paginate_default_svd_algorithm.py
```diff
@@ -1,26 +1,36 @@
...
+
+def batch_update(queryset, batch=1000, **changes):
+    """ Update per batch """
+    is_done = False
+    current = 1
+
+    while not is_done:
+        start = (current - 1) * batch
+        end = start + batch
+        pks = queryset.values_list('pk', flat=True)[start:end][::1]
```
arielsvn: Should this be: `queryset[start:end].values_list('pk', flat=True)`? `values_list` will return a list with every id.

kurtwheeler: What's `[::1]` do?

davidsmejia:
Should probably actually only call the update function if pks isnt empty

common/data_refinery_common/migrations/0038_paginate_default_svd_algorithm.py
davidsmejia: same as list(queryset)

davidsmejia: my understanding is that the slice is what executes the queryset but with a limit offset

davidsmejia: so it would be the same

davidsmejia: At the top of this page it explains which operations execute the query list.

https://docs.djangoproject.com/en/2.2/ref/models/querysets/#values-list

davidsmejia: thought it was easier to read but may have been mistaken

common/data_refinery_common/migrations/0038_paginate_default_svd_algorithm.py
```diff
@@ -1,26 +1,36 @@
...
+        queryset.model.objects.filter(pk__in=pks).update(**changes)
+        is_done = len(pks) <= batch
+        if not is_done:
```
kurtwheeler: I don't think this will terminate at the correct time. You're selecting `batch` ids each time, and then checking to see if they are less than or equal to `batch`. I would expect that condition to be true every loop. I think a better way might be to change the while condition to:
```
num_objects = queryset.count()
while end < num_objects:
```

davidsmejia: batch is a full page, so it wont be called again once fewer than 1000 is returned

davidsmejia: however i should update it to not call the update method on no objects.... since that is a waste

kurtwheeler: What if exactly 1000 is returned? Which I think will be every loop.

Basically I think you have an off by one error but I suggested an implementation that avoids the possibility of an off by one error instead of just fixing it. But either solution works for me. But right now I'm pretty sure this only loops once because `len(pks) == batch` will be true the first time.

davidsmejia: oh it should be less than! good catch.

PR #1667, opened 2019-09-26
Clean up disk space better in correct_affy_cdfs

kurtwheeler:
## Issue Number

#1276 

Ran out of disk space again. I think I didn't indent my previous "fix" enough. This should keep the disk space usage low for that command.

## Types of changes

## Functional tests

I ran it locally but I don't have much data to correct.


kurtwheeler:
> Seems like we could also remove each file after we're done with them.
That's what I'm doing! I just  don't trust GEOParse to not litter files around because it's really disrespectful of disk space.

Issue #1258, opened 2019-04-16
Generate New Compendia


cgreene:
### Context

We have enough data now to start generating compendia for many organisms.

### Problem or idea

Depends #1221 - once we get the updated array data we should generate compendia for the organisms with QN targets.

### Solution or next step

@Miserlou will use this as an opportunity to walk @kurtwheeler through compendia generation.

kurtwheeler:
I was just able to rebuild the fish compendium! I should kick off compendia jobs for all organisms that meet the criteria for it right @cgreene ? Just making sure nothing has changed in the last few months.

cgreene:
Let's kick this off for the top 3 or top 5 organisms or so to test our process. We will want to generate new compendia when we hit 1m samples to be distributed as the first compendia from the non-beta release.

cgreene:
This is done! 🎊

PR #1673, opened 2019-09-26
Fix uploading compendia by using/setting the correct bucket env var.

kurtwheeler:
…

## Issue Number

#1258 

Also fix GEOParse cleanup.


foreman/data_refinery_foreman/foreman/management/commands/correct_affy_cdfs.py
```diff
@@ -92,7 +92,13 @@ def handle(self, *args, **options):
                     # GEOparse downloads files here and never cleans them up! Grrrr!
                     download_path = GEO_TEMP_DIR + experiment.accession_code + '_family.soft.gz'
                     # It's not a directory, but ignore_errors is useful.
-                    shutil.rmtree(download_path, ignore_errors=True)
+                    try:
+                        os.remove(download_path, ignore_errors=True)
+                    except:
```
arielsvn: do we need the `try...except` if we have `ignore_errors=True`?

Also, `except...pass` makes me nervous.

kurtwheeler: Oh good catch! `ignore_errors=True` isn't an option on os.remove(), I just forgot to delete it from the other method. I've changed this so we log so we can tell why we're failing to delete stuff.

foreman/data_refinery_foreman/foreman/management/commands/correct_affy_cdfs.py
```diff
@@ -92,7 +92,13 @@ def handle(self, *args, **options):
...
                     download_path = GEO_TEMP_DIR + experiment.accession_code + '_family.soft.gz'
                     # It's not a directory, but ignore_errors is useful.
-                    shutil.rmtree(download_path, ignore_errors=True)
+                    try:
+                        os.remove(download_path)
+                    except:
+                        # Don't anything interrupt this, like say,
+                        # GEOParse downloading a directory instead of
+                        # a file...
+                        logger.exception("Failed to delete an archive.")
```
arielsvn: Does this include the exception that was raised?

kurtwheeler: Yup, that's what `logger.exception` does.

PR #1682, opened 2019-09-27
Add CREATE_COMPENDIA to job switch.

kurtwheeler:
## Issue Number

#1258 

We still can't dispatch multiple compendia jobs with a single command. This isn't urgent at all cause I can do them one by one in prod until we're ready for another deploy.


arielsvn:
I missed this when adding the field `ProcessorPipeline.CREATE_COMPENDIA`

Issue #1694, opened 2019-09-30
Finish setting up vscode to run refine.bio in a remote container


arielsvn:
### Context

VScode recently added [remote containers](https://code.visualstudio.com/docs/remote/containers), which allows the IDE to run inside a docker container and provide all modern ide features like IntelliSense, code competition, linting, debugging tests, etc

### Problem or idea

I have been playing with this and it's almost ready on the branch [arielsvn/vscode-remote-containers](https://github.com/AlexsLemonade/refinebio/tree/arielsvn/vscode-remote-containers)

![2019-09-20-vscode-refinebio-debugger](https://user-images.githubusercontent.com/1882507/65887994-d1f55480-e36c-11e9-817e-fcfde7aa36f0.png)

### Solution or next step

I need to add some notes in the readme and then I think it's ready for a PR.

arielsvn:
Closing this, we'll come up with a way to make it non vscode specific.

PR #1698, opened 2019-09-30
Arielsvn/vscode remote containers

arielsvn:
## Issue Number

#1694 

This adds some configuration files that help my IDE work better with refine.bio

## Types of changes

## Checklist


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,37 @@
+// If you want to run as a non-root user in the container, see .devcontainer/docker-compose.yml.
+{
+	"name": "Refinebio API",
+	"dockerComposeFile": "docker-compose.yml",
+	"service": "app",
+	"workspaceFolder": "/workspace",
+
+	// Use 'settings' to set *default* container specific settings.json values on container create.
+	// You can edit these settings after create using File > Preferences > Settings > Remote.
```
kurtwheeler: Well this is very ugly :(

arielsvn: yeah - seems like github doesn't like comments in json files

kurtwheeler:
Can you explain this PR a bit better?

> help my IDE work better with refine.bio

isn't really specific enough for me to tell whether it is accomplishing its goals or not.

api/data_refinery_api/views.py
```diff
@@ -46,7 +46,7 @@
 from rest_framework.settings import api_settings
 from rest_framework.views import APIView
 
-from data_refinery_api.serializers import (
+from .serializers import (
```
kurtwheeler: Why are you changing this to use relative imports?

arielsvn:
Sure, This creates a docker container to run the project and allows the IDE to run from inside that container. So I get IntelliSense, code navigation, debugging tests and all the features of the IDE. There are more details [here](https://code.visualstudio.com/docs/remote/containers).

This will only be useful to people using vscode - which is only me for now.

.dockerignore
```diff
@@ -3,7 +3,6 @@ volume
 test_volume
 */test_volume
 volume
-.git
```
kurtwheeler: My `.git` directory is 502MB. This is 502 MB that will be push from docker to docker every time I run any `docker build` command with this change. Why is this change necessary?

api/data_refinery_api/views.py
arielsvn: Because that way the IDE can locate all those types and provide code competition and type checking. Does that have any other drawbacks?

.devcontainer/devcontainer.json
kurtwheeler: Well that's just because JSON doesn't support comments. It's not valid JSON but I guess if you wanna use JSON for configs people write special parsers that support it?

api/data_refinery_api/views.py
kurtwheeler: Why is that necessary? I don't understand why that'd make a difference.

Pep8 has this to say:
```
Absolute imports are recommended, as they are usually more readable and tend to be better behaved (or at least give better error messages) if the import system is incorrectly configured (such as when a directory inside a package ends up on sys.path):
...
However, explicit relative imports are an acceptable alternative to absolute imports, especially when dealing with complex package layouts where using absolute imports would be unnecessarily verbose:
...

Standard library code should avoid complex package layouts and always use absolute imports.

Implicit relative imports should never be used and have been removed in Python 3.
```

api/environments/local
```diff
@@ -16,7 +16,7 @@ NOMAD_HOST=nomad
 LOCAL_ROOT_DIR=/home/user/data_store
 USE_S3=False
 S3_BUCKET_NAME=data-refinery
-S3_QN_TARGET_BUCKET_NAME=
+S3_QN_TARGET_BUCKET_NAME=data-refinery
```
kurtwheeler: Why are you setting this? Local should not upload files to S3.

setup.cfg
```diff
@@ -3,3 +3,9 @@ max-line-length: 99
 
 [flake8]
 max-line-length: 99
+
+[tool:pytest]
```
kurtwheeler: `pytest`? But we use python unittest...

setup.cfg
```diff
@@ -3,3 +3,9 @@ max-line-length: 99
...
+DJANGO_SETTINGS_MODULE = api.data_refinery_api.settings
+django_find_project = false
+python_paths = api
+python_files = api/data_refinery_api/tests.py common/**/test_*.py
```
kurtwheeler: Also this all seems to be duplicated in .devcontainer/pytest.cfg

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,37 @@
...
+	"postCreateCommand": "./.devcontainer/postcreate.sh",
+
+	// Add the IDs of extensions you want installed when the container is created in the array below.
+	"extensions": [
+		"ms-python.python",
+		"donjayamanne.python-extension-pack",
+		"VisualStudioExptTeam.vscodeintellicode",
+		"ms-azuretools.vscode-docker"
+	]
+}
```
kurtwheeler: This is all specific to VSCode? Are you able to rename the directory? We have a lot of development containers so devcontainer isn't very specific...

.devcontainer/docker-compose.yml
```diff
@@ -0,0 +1,42 @@
...
+  app:
+    build:
+      context: ..
+      dockerfile: .devcontainer/Dockerfile
+    ports:
+      - 8000:8000
+    volumes:
+      - ..:/workspace
+      - ~/.ssh:/root/.ssh-localhost:ro
+      - /var/run/docker.sock:/var/run/docker.sock
```
kurtwheeler: Why are you exposing the docker socket to the container?

arielsvn: This allowed me to access my local docker install from inside the container.

.devcontainer/devcontainer.json
arielsvn: hmm I need to check, not sure it's possible. The directory also has a `.` at the start, suggesting is a special folder

setup.cfg
arielsvn: good catch! forgot to remove this file.

kurtwheeler:
> Sure, This creates a docker container to run the project and allows the IDE to run from inside that container. So I get IntelliSense, code navigation, debugging tests and all the features of the IDE. There are more details here.

It does more than that though. It also adds docker compose. What do you mean by "creates a docker container to run the project"? Is it actually running the project or just vscode or not even vscode but actually just a remote server for a local vscode to connect to?

> This will only be useful to people using vscode

As it's written not it's definitely not useful to anyone not using vscode, and actually maybe not to people using vscode as well. Can you add some high level documentation about what this stuff is doing and how to use it? Maybe in the docker compose, maybe in the README, maybe whereever else makes sense. This very well may be rather useful outside of vscode as well but it's rather hard to tell so far.

api/environments/local
arielsvn: I just needed it to have some value, without it, some tests were failing because they couldn't find this variable. I'm gonna try to find another way to resolve it.

.dockerignore
arielsvn: does that happens every time you run `docker build` or just once? This allowed me to use git from inside the container. I'm gonna try to see if there's a way to add this file specifically for my image.

arielsvn:
Thanks for reviewing, I'm gonna address those comments later!

.dockerignore
kurtwheeler: Every time you run build Docker sends the whole context over a stream to the docker daemon and then looks at what is so it can tell where the cache needs to be busted at. You could probably mount the .git directory as a volume though.

.devcontainer/devcontainer.json
kurtwheeler: Hmm well that seems kind of unfriendly... maybe it's configurable? Go figure vscode would assume it's the only way anyone would develop on your project...

.devcontainer/docker-compose.yml
kurtwheeler: Righto, but why?

api/environments/local
kurtwheeler: Ah ok. Maybe just an issue with the way the env var was being retrieved then.

.devcontainer/docker-compose.yml
arielsvn: I wanted to be able to also run the commands we have from inside docker, to have the full development experience. From inside the container, I can execute `./api/manage.py makemigrations` which I like. However, I'd also be nice to be able to run `./scripts/makemigrations.sh`.

.devcontainer/devcontainer.json
arielsvn: It doesn't seem to be a problem for vscode, this is a configuration file for the IDE.

api/data_refinery_api/views.py
arielsvn: ok I'll try to find another way. I saw relative imports in other parts of the code, we might want to update those as well.

setup.cfg
arielsvn: yeah I wasn't able to setup `unittest`, I'm gonna give it another try. pytest seems to also work well when running the tests.

image

arielsvn:
I'm closing this, for now, will re-open after we get out of beta with these comments addressed.

Issue #1696, opened 2019-09-30
Some computed files seems to be in S3 and they aren't


arielsvn:
### Context

We found this while running the quantpendia.

### Problem or idea

For some reason, they have values on `s3_bucket` and `s3_key` but they aren't in S3

### Solution or next step

We need a command to ensure that all the files are in S3.

arielsvn:
We used to remove the `quant.sf` files from S3, this might explain this problem.

arielsvn:
Ran a command to check all computed files on staging and it found 58 that were not in S3.

<details>

```
2019-09-30 21:28:58,724 i-0a35b53f5566c9a51 [volume: -1] data_refinery_foreman.foreman.management.commands.check_computed_files INFO: Checked 2000 computed files to see if they are in S3, so far found 42
2019-09-30 21:29:37,953 i-0a35b53f5566c9a51 [volume: -1] data_refinery_foreman.foreman.management.commands.check_computed_files INFO: Checked 4000 computed files to see if they are in S3, so far found 56
2019-09-30 21:30:16,629 i-0a35b53f5566c9a51 [volume: -1] data_refinery_foreman.foreman.management.commands.check_computed_files INFO: Checked 6000 computed files to see if they are in S3, so far found 57
2019-09-30 21:30:55,696 i-0a35b53f5566c9a51 [volume: -1] data_refinery_foreman.foreman.management.commands.check_computed_files INFO: Checked 8000 computed files to see if they are in S3, so far found 58
2019-09-30 21:31:34,988 i-0a35b53f5566c9a51 [volume: -1] data_refinery_foreman.foreman.management.commands.check_computed_files INFO: Checked 10000 computed files to see if they are in S3, so far found 58
2019-09-30 21:32:14,038 i-0a35b53f5566c9a51 [volume: -1] data_refinery_foreman.foreman.management.commands.check_computed_files INFO: Checked 12000 computed files to see if they are in S3, so far found 58
2019-09-30 21:32:49,946 i-0a35b53f5566c9a51 [volume: -1] data_refinery_foreman.foreman.management.commands.check_computed_files INFO: Checked 14000 computed files to see if they are in S3, so far found 58
2019-09-30 21:33:26,916 i-0a35b53f5566c9a51 [volume: -1] data_refinery_foreman.foreman.management.commands.check_computed_files INFO: Checked 16000 computed files to see if they are in S3, so far found 58
[...]

PR #1699, opened 2019-09-30
Add command to check if computed files are in S3

arielsvn:
## Issue Number

#1696 

This adds a command that iterates over all computed files to check if they are in S3. If they aren't it clears their `s3_bucket` and `s3_key` fields.

I used [iterator()](https://docs.djangoproject.com/en/dev/ref/models/querysets/#iterator) to iterate over all computed objects, seems like this will do something similar to the `Paginator` we have used.

## Types of changes

## Functional tests

None, should we test this on staging?

## Checklist

### DIFF
--- a/foreman/data_refinery_foreman/foreman/management/commands/check_computed_files.py
+++ b/foreman/data_refinery_foreman/foreman/management/commands/check_computed_files.py
@@ -23,7 +23,7 @@ class Command(BaseCommand):
         """
         computed_files_queryset = ComputedFile.objects.filter(s3_key__isnull=False, s3_bucket__isnull=False)
         total_files = 0
-        modified_files = 0
+        missing_file_ids = []
 
         # iterate over all computed files without loading them in memory
         # https://docs.djangoproject.com/en/dev/ref/models/querysets/#iterator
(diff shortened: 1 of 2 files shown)


foreman/data_refinery_foreman/foreman/management/commands/check_computed_files.py
```diff
@@ -0,0 +1,47 @@
...
+            # https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.head_object
+            try:
+                S3.head_object(Bucket=computed_file.s3_bucket, Key=computed_file.s3_key)
+            except ClientError:
+                # Not found
+                modified_files += 1
+                logger.debug('Computed file not found on S3. Removing s3 fields.', computed_file=computed_file)
+                computed_file.s3_key = None
+                computed_file.s3_bucket = None
+                computed_file.save()
```
kurtwheeler: Hmm, this looks like it'll work, but it'll make a bunch of requests to update the computed files one by one. What if you instead kept a list of ids and then did something like:
```
ComputedFile.objects.filter(id__in=missing_file_ids).update(s3_key=None, s3_bucket=None)
```
?

arielsvn:
@kurtwheeler added your suggestion here. Let me know if you want to take another look.

PR #1701, opened 2019-09-30
Remove authentication section from api docs

### LINKED ISSUE #1608: Remove 'Basic Authentication' from API docs
### Context

We don't actually use HTTP's Basic Authentication, but the API docs say we do.

### Problem or idea

This confused the first external user of our API: @awm33

### Solution or next step

We should not say we use Basic Authentication on our API docs.

arielsvn:
## Issue Number

close #1608 

Remove authentication section from api docs

## Functional tests

Tested locally

## Checklist

## Screenshots

image


kurtwheeler:
BTW @arielsvn this PR says "review required" but the only change since I requested changes was a merge from dev.

arielsvn:
yes, I still have to address your comments here. Thanks for taking a look anyways!

arielsvn:
@kurtwheeler this is ready for review. I improved the documentation on `/dataset` and `/token` and added sample code.

api/data_refinery_api/views.py
```diff
@@ -540,16 +557,20 @@ class CreateApiTokenView(generics.CreateAPIView):
...
+    import requests
+    import json
+    ​
+    response = requests.post('https://api.refine.bio/v1/token/')
+    token_id = response.json()['id']
+    response = requests.put('https://api.refine.bio/v1/token/' + token_id + '/', json.dumps({'is_activated': True}), headers={'Content-Type': 'application/json'})
     ```
 
-    The token id needs to be sent on the `API_KEY` header on http requests.
+    The token id needs to be attached to the header of the requests with the key `API-KEY`
```
kurtwheeler: I would phrase this `The token id needs to be provided in the HTTP request in the `API-KEY` header.` Just a suggestion though, that might not be as clear to others as it seems to me.

We should definitely have a period at the end of the sentence though :)

api/data_refinery_api/views.py
```diff
@@ -377,9 +377,26 @@ class CreateDatasetView(generics.CreateAPIView):
 @method_decorator(name='put', decorator=swagger_auto_schema(operation_description="""
 Modify an existing Dataset.
 
-Set `start` to `true` along with a valid activated API token (from `/token/`) to begin smashing and delivery.
+In order to begin smashing, you have to provide an activated api token. The token can be
+requested with [/token](#tag/token), and has to be sent in the header of the request
+with the key: `API-TOKEN`. Example:
```
kurtwheeler: I think I realized what seems weird to me about this phrasing. It's the use of `key`. I don't think I've ever heard `key` used when talking about HTTP headers. I've generally heard each field referred to as a header, but I think that might also be slightly incorrect technically. I think the term we should use is "header field", as in "must be provided in the `API-KEY` header field on the request."

I also think that instead of saying `The token can be requested with [/token](#tag/token)...`, we should just say something like "An activated API key must be provided in the `API-KEY` header field on the request. To acquire and activate an API key see the documentation for the `/token` endpoint. The reason for this change is that if someone doesn't actually go look at the documentation for the `/token` endpoint then they might end up trying an inactivated key.

arielsvn:
Thanks for the comments @kurtwheeler. Writing documentation is not easy :) Let me know if it's more clear now. They will look like this with the code formatted:

image

image

PR #1706, opened 2019-10-01
Parallelize command to check computed files

arielsvn:
## Issue Number

#1696 

Parallelize command to check computed files. Hopefully, it will run faster like this.

## Types of changes

## Functional tests

None

## Checklist

### DIFF
--- a/common/data_refinery_common/utils.py
+++ b/common/data_refinery_common/utils.py
@@ -441,15 +441,18 @@ def _aggregate_nomad_jobs(aggregated_jobs):
 
     return nomad_pending_jobs, nomad_running_jobs
 
-def queryset_iterator(queryset, page_size = 2000):
-    """ use the performant paginator to iterate over a queryset """
+def queryset_page_iterator(queryset, page_size = 2000):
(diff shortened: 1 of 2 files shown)


common/data_refinery_common/utils.py
```diff
@@ -441,15 +441,18 @@ def _aggregate_nomad_jobs(aggregated_jobs):
...
+        yield page.object_list
 
         if not page.has_next():
             break
         else:
             page = paginator.page(page.next_page_number())
+
+def queryset_iterator(queryset, page_size = 2000):
+    """ use the performant paginator to iterate over a queryset """
+    return (page for page in queryset_page_iterator(queryset, page_size))
```
kurtwheeler: What exactly is the point of this function? It looks like it just pulls the whole queryset into memory and returns a list of pages which undoes the whole point of pagination. Were you thinking you'd make this a generator which returns the items item by item rather than page by page? You could do that like:
```
  for item in page.object_list:
    yield item
```
but IDK exactly what the goal was here.

arielsvn: good catch! yeah it was not doing anything.

PR #1712, opened 2019-10-02
Fix queryset iterator and setting of volume index

kurtwheeler:
## Issue Number

#1703 

We broke the queryset iterator so dispatching lots of compendia jobs fails.

For some reason the Foreman is also very broken. I saw some of it yesterday and tried to fix it with #1707 but that clearly didn't work. I tried to fix the issue in the downloader job utils but I don't think this'll fix the Foreman for jobs that already don't have volume index. However all it's telling me is that it couldn't queue the job so I've improved the logging there to get more visibility.

## Types of changes

## Functional tests

The unit tests cover this fairly well.

## Checklist

### DIFF
--- a/common/data_refinery_common/test_utils.py
+++ b/common/data_refinery_common/test_utils.py
@@ -120,3 +120,5 @@ class UtilsTestCase(TestCase):
         names = []
         for pipeline in utils.queryset_iterator(pipelines):
             names.append(pipeline.name)
+
+        self.assertEqual(len(names), 3000)
--- a/common/data_refinery_common/test_utils.py
+++ b/common/data_refinery_common/test_utils.py
@@ -118,5 +118,5 @@ class UtilsTestCase(TestCase):
         # Build a list of the names just to do something with the data
         # so we know the query actually resolved.
         names = []
-        for pipeline in queryset_iterator(pipelines):
+        for pipeline in utils.queryset_iterator(pipelines):
             names.append(pipeline.name)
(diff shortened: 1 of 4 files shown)


common/data_refinery_common/test_utils.py
```diff
@@ -102,3 +103,20 @@ def test_volume_index(self):
...
+        for i in range(3000):
+            Pipeline(name=str(i)).save()
+
+        pipelines = Pipeline.objects.all()
+
+        # Build a list of the names just to do something with the data
+        # so we know the query actually resolved.
+        names = []
+        for pipeline in utils.queryset_iterator(pipelines):
+            names.append(pipeline.name)
```
arielsvn: I like this test. Should we assert something with `names`? maybe check it's length?

common/data_refinery_common/utils.py
```diff
@@ -456,5 +456,5 @@ def queryset_page_iterator(queryset, page_size = 2000):
 def queryset_iterator(queryset, page_size = 2000):
     """ use the performant paginator to iterate over a queryset """
     for page in queryset_page_iterator(queryset, page_size):
-        for item in page.object_list:
+        for item in page:
```
arielsvn: nice catch!

common/data_refinery_common/test_utils.py
kurtwheeler: Sure!

PR #1717, opened 2019-10-02
Parallelize creating quantcompendia

arielsvn:
## Issue Number

N/A

This parallelizes the downloading of quant files when creating the quant-pendia.

## Types of changes

## Functional tests

staging

## Checklist


workers/data_refinery_workers/processors/smasher.py
```diff
@@ -478,19 +480,40 @@ def _quantile_normalize(job_context: Dict, ks_check=True, ks_stat=0.001) -> Dict
...
-        num_samples += 1
-        latest_computed_file.get_synced_file_path(path=output_file_path)
+    samples = [sample for (_, sample) in files_sample_tuple]
+    page_size = 100
+    # split the samples in groups and download each one individually
+    with Pool(processes=PROCESS_POOL_SIZE) as pool:
+        # for each sample we need it's latest quant.sf file we don't want to query the db
+        # for all of them, so we do it in groups of 100, and then download all of the computed_files
+        # in parallel
+        for sample_page in (samples[i:i+page_size] for i in range(0, len(samples), page_size)):
```
kurtwheeler:
```suggestion
        for sample_page in (samples[i*page_size:i+page_size] for i in range(0, len(samples), page_size)):
```

workers/data_refinery_workers/processors/smasher.py
```diff
@@ -478,19 +480,40 @@ def _quantile_normalize(job_context: Dict, ks_check=True, ks_stat=0.001) -> Dict
...
+            sample_and_computed_files = []
+            for sample in sample_page:
+                latest_computed_file = sample.get_most_recent_quant_sf_file()
+                if not latest_computed_file: continue
+                output_file_path = output_path + sample.accession_code + "_quant.sf"
+                sample_and_computed_files.append((output_file_path, latest_computed_file))
+
+            # download this set of files, this will take a few seconds that should also help the db recover
+            pool.map(downlad_computed_file, sample_and_computed_files)
+            num_samples += len(sample_and_computed_files)
```
kurtwheeler:
```suggestion
            num_samples += len(sample_and_computed_files)
            sample_and_computed_files = []
```

kurtwheeler:
Wait it just occurred to me: This seems like it could have a similar issue with the test database that @davidsmejia ran into. Does it?

arielsvn:
> Wait it just occurred to me: This seems like it could have a similar issue with the test database that @davidsmejia ran into. Does it?

Hmm hard to say, I don't see any errors related to that in the test logs. In this case, we aren't executing any db queries in other threads, only downloading the files from S3. What do you think @davidsmejia ?

workers/data_refinery_workers/processors/smasher.py
arielsvn: good catch! 👍

arielsvn:
@kurtwheeler @davidsmejia tests passed here, do you think this is good to be merged?

workers/data_refinery_workers/processors/smasher.py
```diff
@@ -50,6 +51,7 @@
 ### DEBUG ###
 logger.setLevel(logging.getLevelName('DEBUG'))
 
+PROCESS_POOL_SIZE = max(1, int(psutil.cpu_count()/2))
```
davidsmejia: maybe reduce it by 1 just so we don't over consume cpu resources ```max(1, int(psutil.cpu_count()/2 - 1)```

workers/data_refinery_workers/processors/smasher.py
```diff
@@ -117,20 +119,40 @@ def _prepare_files(job_context: Dict) -> Dict:
...
-        if not latest_computed_file: continue
-        accession_code = sample.accession_code
-        # copy file to the output path
-        output_file_path = output_path + accession_code + "_quant.sf"
-        num_samples += 1
-        latest_computed_file.get_synced_file_path(path=output_file_path)
+    samples = [sample for (_, sample) in files_sample_tuple]
+    page_size = 100
+    # split the samples in groups and download each one individually
+    with Pool(processes=PROCESS_POOL_SIZE) as pool:
```
davidsmejia: I think the more obvious way to do this is remove the context block and clean up explicitly 

```pool = Pool(processes=PROCESS_POOL_SIZE)```
... use pool ...
then at the end you can
```
pool.close()
pool.join()
```

**Comment by davidsmejia** [comment]
LGTM I don't see any issues with the DB connections and I'm pretty sure this was working on staging. I would just address the clean up of the pool so the workers don't outlive their usefulness

**Comment by arielsvn** [comment]
Thanks for reviewing @davidsmejia, I just addressed your comments. Let me know if you see anything else.

File: workers/data_refinery_workers/processors/smasher.py [hunk 5]
```diff
@@ -117,20 +119,43 @@ def _prepare_files(job_context: Dict) -> Dict:
     log_state("end prepare files", job_context["job"], start_prepare_files)
     return job_context
 
+def downlad_computed_file(item):
+    """ this function downloads the latest computed file sent in it's arguments,
+    its used to parallelize the download of all files. """
+    (output_file_path, latest_computed_file) = item
```
kurtwheeler: Could you improve the docstring and/or parameter name here? The docstring isn't grammatical and it's a tad vague and combined with a parameter named `item` it's rather vague. Perhaps you could typehint the Tuple parameter: https://docs.python.org/3/library/typing.html#typing.Tuple?

kurtwheeler: Btw, you're using `it's` and its` backwards:
  `it's` == `it is`
  `its` is possessive

workers/data_refinery_workers/processors/smasher.py
```diff
@@ -117,20 +119,45 @@ def _prepare_files(job_context: Dict) -> Dict:
...
-    for (_, sample) in files_sample_tuple:
-        latest_computed_file = sample.get_most_recent_quant_sf_file()
-        # we just want to output the quant.sf files
-        if not latest_computed_file: continue
-        accession_code = sample.accession_code
-        # copy file to the output path
-        output_file_path = output_path + accession_code + "_quant.sf"
-        num_samples += 1
-        latest_computed_file.get_synced_file_path(path=output_file_path)
+    samples = [sample for (_, sample) in files_sample_tuple]
```
kurtwheeler: Any reason not to use `job_context['samples']` at this point? I'd say that you could just modify `smasher._prepare_files()` to call the appropriate `get_most_recent_*` but function but I'm pretty sure that with https://github.com/AlexsLemonade/refinebio/issues/1751 that you're planning on not calling that at all, which means the key that you're referencing here won't exist at that point anyway.

arielsvn:
Tested on staging https://staging.refine.bio/dataset/11b0dc82-5b82-41f2-a2d3-cd4aab7df7a4

PR #1725, opened 2019-10-03
Prevent the foreman from killing compendia jobs

arielsvn:
## Issue Number

N/A

We noticed that the foreman was killing compendia jobs. This ensures that only happens with jobs that should have a volume assigned.

## Types of changes

## Functional tests

None

## Checklist

### DIFF
--- a/foreman/data_refinery_foreman/foreman/main.py
+++ b/foreman/data_refinery_foreman/foreman/main.py
@@ -1337,7 +1337,8 @@ def cleanup_the_queue():
             continue
 
         # ensure the job is one of the indexed_job_types
-        job_type = next(job_type.value for job_type in indexed_job_types if job["ParentID"].startswith(job_type.value), None)
+        possible_job_types = (job_type.value for job_type in indexed_job_types if job["ParentID"].startswith(job_type.value))
+        job_type = next(possible_job_types, None)
         if not job_type:
             continue
 
--- a/foreman/data_refinery_foreman/foreman/main.py
+++ b/foreman/data_refinery_foreman/foreman/main.py
@@ -1307,10 +1307,7 @@ def cleanup_the_queue():
     logger.info("Removing all jobs from Nomad queue whose volumes are not mounted.")
 
     # Smasher and QN Reference jobs aren't tied to a specific EBS volume.
-    indexed_job_types = [e.value for e in ProcessorPipeline if e.value not in ["SMASHER", "QN_REFERENCE"]]
-    # Special case for downloader jobs because they only have one
-    # nomad job type for all downloader tasks.
-    indexed_job_types.append("DOWNLOADER")
+    indexed_job_types = [pipeline for pipeline in ProcessorPipeline if pipeline not in SMASHER_JOB_TYPES]
 
     nomad_host = get_env_variable("NOMAD_HOST")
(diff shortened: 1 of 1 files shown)


foreman/data_refinery_foreman/foreman/main.py
```diff
@@ -1321,6 +1318,7 @@ def cleanup_the_queue():
         jobs = nomad_client.jobs.get_jobs()
     except:
         # If we cannot reach Nomad now then we can wait until a later loop.
+        logger.exception("Couldn't query Nomad about current jobs.")
```
kurtwheeler: Can you make this `warn` with `exc_info=1` so that we log the exception but it doesn't go to Sentry?

arielsvn: sure!

```suggestion
        logger.exception("Couldn't query Nomad about current jobs.", exc_info=1)
```

arielsvn:
Reviewed this with @kurtwheeler, we decided to hold merging it until the tests are fixed. I will add another PR with the fix for the foreman without refactoring it.

arielsvn:
All tests seem to pass, this should also be ready for review @kurtwheeler

foreman/data_refinery_foreman/foreman/main.py
```diff
@@ -1334,46 +1335,51 @@ def cleanup_the_queue():
...
+
+                if job_type == DOWNLOADER:
+                    job_record = DownloaderJob.objects.filter(nomad_job_id=job["ID"]).first()
+                    # If it's a downloader job, then it doesn't
+                    # have to run on the volume it was assigned
+                    # to. We can let the foreman reassign it.
+                    job_record.volume_index = None
+                    job_record.save()
+
+                logger.info('Foreman Killed nomad job because it did not have a volume assigned',
```
kurtwheeler: Why does this say "did not have a volume assigned" instead of "volume that was not active" like it used to? Seems like it was more accurate before.

foreman/data_refinery_foreman/foreman/main.py
```diff
@@ -1334,46 +1335,51 @@ def cleanup_the_queue():
...
+            try:
+                nomad_client.job.deregister_job(job["ID"], purge=True)
+
+                if job_type == DOWNLOADER:
+                    job_record = DownloaderJob.objects.filter(nomad_job_id=job["ID"]).first()
+                    # If it's a downloader job, then it doesn't
+                    # have to run on the volume it was assigned
+                    # to. We can let the foreman reassign it.
+                    job_record.volume_index = None
+                    job_record.save()
```
kurtwheeler: Hmm, I'm a tad torn on this because it doesn't hurt anything but it's also a bit unnecessary because the foreman doesn't bother trying to requeue downloader jobs onto the same volumes they were on before.

foreman/data_refinery_foreman/foreman/main.py
arielsvn: I think I misunderstood what was happening, let me change it.

```suggestion
                logger.info('Foreman Killed nomad job because it had a volume that was not active',
```

foreman/data_refinery_foreman/foreman/main.py
arielsvn: In that case let's remove it and make this simpler

PR #1726, opened 2019-10-03
Quick fix for the foreman killing compendia jobs

arielsvn:
## Issue Number

N/A

This is quick fix that prevents the foreman from killing compendia jobs. #1725 adds other improvements that we'll merge later.

## Types of changes

### DIFF
--- a/foreman/data_refinery_foreman/foreman/main.py
+++ b/foreman/data_refinery_foreman/foreman/main.py
@@ -1307,7 +1307,7 @@ def cleanup_the_queue():
     logger.info("Removing all jobs from Nomad queue whose volumes are not mounted.")
 
     # Smasher and QN Reference jobs aren't tied to a specific EBS volume.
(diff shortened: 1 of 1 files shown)


foreman/data_refinery_foreman/foreman/main.py
```diff
@@ -1354,21 +1354,19 @@ def cleanup_the_queue():
                 # will be incremented when it is requeued).
                 try:
                     nomad_client.job.deregister_job(job["ID"], purge=True)
-                    processor_jobs = ProcessorJob.objects.filter(nomad_job_id=job["ID"])
+                    logger.info('Foreman Killed nomad job because it did not have a volume assigned',
```
kurtwheeler: Can you make this DEBUG? This could get noisy.

kurtwheeler: Also the log statement isn't quite right. The job did have a volume assigned but the volume it was assigned to wasn't active.

PR #1727, opened 2019-10-03
Fix QN jobs

kurtwheeler:
## Issue Number

#1015 

I'm not exactly sure when but at some point we started using `prepare_original_files` instead of `prepare_dataset` for QN_REFERENCE jobs. This caused the job to look for original files so it didn't find any. This should fix it so we can actually prepare files.

I'm not exactly sure how we had recent successful QN_REFERENCE jobs...

## Types of changes

## Functional tests

None yet.

## Checklist


arielsvn:
> I'm not exactly sure how we had recent successful QN_REFERENCE jobs...

Should we assume the successful `QN_REFERENCE` jobs generated invalid files and remove them?

kurtwheeler:
> > I'm not exactly sure how we had recent successful QN_REFERENCE jobs...
> 
> Should we assume the successful `QN_REFERENCE` jobs generated invalid files and remove them?

I don't think that's necessary. My guess is that they weren't affected by this bug cause if they were they wouldn't have worked.

PR #1734, opened 2019-10-03
Don't create compendia for organisms that don't have a QN target.

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio/pull/449

This was always supposed to be a thing we checked. Part of creating compendia is quantile normalizing them, so if we don't have a QN target we should:
  A. Not try to create a compendia.
  B. Fail to create one if we did try.

## Methods

@jaclyn-taroni will be happy to hear that we're no longer going to create compendia for organisms that don't have a QN target.

## Types of changes

## Functional tests

WIP

## Checklist


jaclyn-taroni:
Woo!

foreman/data_refinery_foreman/foreman/management/commands/create_compendia.py
```diff
@@ -90,9 +90,12 @@ def handle(self, *args, **options):
         logger.debug(all_organisms)
 
         for organism in all_organisms:
-            logger.debug(organism)
-            job = create_job_for_organism(organism, quant_sf_only, svd_algorithm)
-            logger.info("Sending CREATE_COMPENDIA for Organism", job_id=str(job.pk), organism=str(organism))
-            send_job(ProcessorPipeline.CREATE_COMPENDIA, job)
+            if get_most_recent_qn_target_for_organism(organism):
```
arielsvn: We have `organism.qn_target`

common/data_refinery_common/utils.py
```diff
@@ -453,8 +454,25 @@ def queryset_page_iterator(queryset, page_size = 2000):
...
 
+
 def queryset_iterator(queryset, page_size = 2000):
     """ use the performant paginator to iterate over a queryset """
     for page in queryset_page_iterator(queryset, page_size):
         for item in page:
             yield item
+
+
+def get_most_recent_qn_target_for_organism(organism):
```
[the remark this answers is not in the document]
arielsvn: Just noticed, can we use `organism.qn_target` instead of this?

kurtwheeler:
Ok I think my latest commit addressed your comment. Here's hoping the tests pass :crossed_fingers:

common/data_refinery_common/models/organism.py
```diff
@@ -187,5 +191,20 @@ def get_objects_with_qn_targets(cls):
         """ Return a list of Organisms who already have valid QN targets associated with them. """
         return Organism.objects.all().filter(qn_target__isnull=False)
 
+    def get_most_recent_qn_target(self):
```
arielsvn: Do we need this method? I think we save the most recent QN Target in `organism.qn_target`

kurtwheeler: Good point.... that solves the issue I was having with import rings.

Good catch, wish we had realized that earlier lol

kurtwheeler:
I've made the requested changes on this PR and re-requested a review on it, why is its status still "Changes Requested"?

PR #1741, opened 2019-10-04
Tune smasher logging

kurtwheeler:
## Issue Number

#1258 

Trying to figure out where our human smasher job is dying. It might be in process frames, cause I think that was the last log messages that came from the compendia rather than the quantpendia. However looking the logs is currently a bit tricky cause the annotation function logs about 4-6 times a millisecond, it's really hard to tell what log is coming from what job, and the process frames wasn't logging cpu/ram so if that is where it died we can't see the RAM usage growing to the point of OOM-killing.

## Types of changes

- Tuning

## Functional tests

The unit tests are outputting logs like:
```
2019-10-04 15:36:00,324 local [volume: 0] data_refinery_workers.processors.smasher DEBUG [job_id: 1] [total_cpu: 0.0] [process_ram: 2.8238258361816406]: starting scaler
```
and
```
2019-10-04 15:39:01,399 local [volume: 0] data_refinery_workers.processors.smasher DEBUG [process_ram: 0.3532676696777344] [total_cpu: 36.8] [job_id: 1]: processing frame 2
```

## Checklist

### DIFF
--- a/workers/data_refinery_workers/processors/create_compendia.py
+++ b/workers/data_refinery_workers/processors/create_compendia.py
@@ -39,22 +39,23 @@ logger = get_and_configure_logger(__name__)
 logger.setLevel(logging.getLevelName('DEBUG'))
 
 
-def log_state(message, start_time=False):
+
+def log_state(message, job, start_time=False):
     if logger.isEnabledFor(logging.DEBUG):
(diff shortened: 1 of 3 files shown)


arielsvn:
seems like the smasher_tests are failing

PR #1749, opened 2019-10-07
Refactor and improve some smasher/compendia code to reduce its memory footprint.

kurtwheeler:
## Issue Number

#1258 

I found an optimization!!!! I think it's a pretty good one, should save 2 full copies of the data from being made. While I was in there I did some refactoring just because it was harder not to.

## Methods

The changes I made could have some scientific implications, but my goal was that this shouldn't change the end result at all @jaclyn-taroni 

What we were doing:
  * Copy the dataframe twice, once for RNA-seq and once for MA
  * Delete RNA-seq columns from MA dataframe and vice versa
  * Manipulate these new matrices to 

What we're doing now:
  * Sort the dataframe by technology (actually before we've created it).
  * Use slices so that we get views of the dataframe rather than a new copy.

## Types of changes

- Refactor

## Functional tests

WIP

## Checklist

### DIFF
--- a/workers/data_refinery_workers/processors/create_compendia.py
+++ b/workers/data_refinery_workers/processors/create_compendia.py
@@ -1,20 +1,15 @@
 import logging
 import os
 import psutil
-import random
 import shutil
-import string
-import subprocess
 import time
-import warnings
 
 import numpy as np
 import pandas as pd
-pd.set_option('mode.chained_assignment', None)
-from fancyimpute import KNN, BiScaler, SoftImpute, IterativeSVD
+from fancyimpute import IterativeSVD
 import simplejson as json
 
-from datetime import datetime, timedelta
+from datetime import datetime
 from django.utils import timezone
 from typing import Dict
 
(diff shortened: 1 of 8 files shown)


workers/data_refinery_workers/processors/create_compendia.py
```diff
@@ -56,30 +56,23 @@ def log_state(message, job, start_time=False):
...
-    # We're going to use the smasher outside of the smasher.
-    # I'm not crazy about this yet. Maybe refactor later,
-    # but I need the data now.
-    job_context = smasher._prepare_files(job_context)
-    job_context = smasher._smash(job_context, how="outer")
 
-    if not 'final_frame' in job_context.keys():
-        logger.warn("Unable to prepare files for creating compendia.",
-            job_id=job_context['job'].id)
+    def log_failure(failure_reason: str) -> None:
```
cgreene: Just curious since you're declaring this inside `_prepare_input` - is this function not useful in any other context?

arielsvn:
Also fixes #1752 🔥

workers/data_refinery_workers/processors/smasher.py
```diff
@@ -491,379 +134,262 @@ def sync_quant_files(output_path, files_sample_tuple, job_context: Dict):
...
-                computed_file=computed_file,
-                dataset_id=job_context['dataset'].id,
-            )
-            return unsmashable(computed_file_path)
+    Returns a new dict containing the metadata, not the job_context.
+    """
+    # Merge all of the frames we've gathered into a single big frame, skipping duplicates.
+    # TODO: If the very first frame is the wrong platform, are we boned?
+    # TODO: I think I'd like to not have all_frames be directly on the job context and overwritten.
+    # Hmm, now I'm less sure because I think that means we wouldn't overwrite them. Perhaps if I do in the
```
davidsmejia: is this cut off?

kurtwheeler: Oh yeah... There's a few notes to myself I need to go remove. I do plan on leaving some TODOs in for subsequent refactor PRs.

Issue #1736, opened 2019-10-03
We should create a single compendia for organisms that share platforms


kurtwheeler:
### Context

We have a compendium for `BOS_INDICUS` despite the fact that all the samples for it were collected using the platform for `BOS_TAURUS`. 

### Problem or idea

We should collapse organisms to a single compendium when they share a platform.

### Solution or next step

So it's not immediately obvious how we should implement this. The reason I say that is because when we build the dataset in the foreman management command `create_compendia.py` we don't look at platforms. We just add the accession code for every sample associated with the dataset and let the compendium figure out what to do.

I guess that while we're doing that we'll have to:
  * Check what the biggest platform is for that organism.
  * Check to see if it is used for any other organism.
  * If it's not then it's business as usual.
  * But if it is we should:
    * Check to see if that organism has any other platforms.
    * If it does then we should scream about it so we know these mixed cases exist.
    * If it doesn't then we should add all samples from that organism to the dataset.

arielsvn:
Does this mean that one compendium can be associated with more than one organism?

Right now they are `ComputedFiles` which can only have a single organism.

arielsvn:
Talked with @kurtwheeler, and we came up with a plan to solve this (The answer to the previous questions is yes).

1. Change the relationship to many-to-many between compendia computed files and Organisms
    - Add many-to-many relationship
    - Populate that using the existing relationship
    - Remove the old relationship (can only happen after step #2 below)
2. Correctly assign the organisms to the computed file when we generate a compendium. https://github.com/AlexsLemonade/refinebio/blob/dev/workers/data_refinery_workers/processors/create_compendia.py#L369
3. Modify the command that triggers the compendia to handle creating compendiums with multiple organisms. https://github.com/AlexsLemonade/refinebio/blob/dev/foreman/data_refinery_foreman/foreman/management/commands/create_compendia.py

arielsvn:
> If it does then we should scream about it so we know these mixed cases exist.

@kurtwheeler do we need to scream? (log an error?) I checked on prod and this seems to happen for a lot of Organisms. For example `GORILLA`'s biggest platform is `IlluminaHiSeq2000`, which is also used by `HOMO_SAPIENS`. 

These are a few other organisms whose biggest platform is also `IlluminaHiSeq2000`

```
PETROMYZON_MARINUS
ANAS_PLATYRHYNCHOS
CIONA_SAVIGNYI
ASTYANAX_MEXICANUS
[...]

PR #1753, opened 2019-10-09
Modify create compendia command to have a single compendia for organisms with the same platform

arielsvn:
## Issue Number

#1736 

This will create groups of organisms sharing the same platform as described in #1376. I ran the methods in prod, here're some of the groups I got:

- `HOMO_SAPIENS`, `TRYPANOSOMA_CRUZI`, `PAPIO_ANUBIS`, `PAN_TROGLODYTES`, `CERVUS_ELAPHUS`, `GRANULIBACTER_BETHESDENSIS`, `MICROCEBUS_MURINUS`
- `BOS_TAURUS`, `BOS_INDICUS`, `BUBALUS_BUBALIS`, `PROCAPRA_PRZEWALSKII`, `PANTHOLOPS_HODGSONII`, `BOS_GRUNNIENS`, 
- `MUS_MUSCULUS`, `MUS_SPRETUS`, `MUS_MUSCULUS_DOMESTICUS`, `MUS_MUSCULUS_X_MUS_SPRETUS`, `MARMOTA_MONAX`, `MUS_MUSCULUS_MUSCULUS_X_M_M._CASTANEUS`, `MUS_MUSCULUS_MUSCULUS_X_M_M._DOMESTICUS`, `MUS_MUSCULUS_CASTANEUS`, `MUS_MUSCULUS_MUSCULUS`, `HELICOBACTER_PYLORI`

## Types of changes

## Functional tests

Checked the functions `get_organism_platforms` and `group_organisms_by_biggest_platform` in prod.

## Checklist


arielsvn:
This should be ready for review. Generates the groups in https://github.com/AlexsLemonade/refinebio/issues/1736#issuecomment-541183861

foreman/data_refinery_foreman/foreman/management/commands/create_compendia.py
```diff
@@ -65,33 +35,143 @@ def add_arguments(self, parser):
...
+    """ We start with the organisms that have QN targets associated with them.
+    Here we are filtering out the ones with invalid QN targets. This can be removed once
+    https://github.com/AlexsLemonade/refinebio/issues/1757 is addressed """
+    # We have some invalid qn targets, that's why we check that each organism needs to have
+    # at least 100 processed samples. This condition can be removed once
+    # https://github.com/AlexsLemonade/refinebio/issues/1757 is addressed
+    return Organism.objects\
+        .filter(qn_target__isnull=False)\
+        .annotate(num_samples=Count('sample', filter=Q(sample__is_processed=True, sample__technology='MICROARRAY'), distinct=True))\
+        .filter(num_samples__gt=100)
```
kurtwheeler: This isn't quite going to do it. It's not just 100 microarray samples, it's 100 of them on the same platform.

arielsvn: That check will be redundant once the invalid QN targets are removed in https://github.com/AlexsLemonade/refinebio/issues/1757. Note we are only using the organisms that have valid QN Targets, and for those, we know that have >100 samples on the same platform.

Do you think we should make this more explicit here? Or maybe wait until the invalid QN targets are removed before merging this?

foreman/data_refinery_foreman/foreman/management/commands/create_compendia.py
```diff
@@ -65,33 +35,143 @@ def add_arguments(self, parser):
...
+    return job
+
+
+def get_dataset(organisms: List[Organism]):
+    """ Builds a dataset with the samples associated with the given organisms """
+    dataset = {}
+
+    filter_query = Q()
+    for organism in organisms:
+        filter_query = filter_query | Q(organisms=organism)
```
kurtwheeler: Is this the right way to query a many-to-many field? If it is cool it's just that `organisms=organism` looks like we're comparing a list to a single object so it should never be true, but I could see that just being how this is done in Django.

arielsvn: Yes, it was strange to me as well. I have seen it in several places and they use it in the [docs](https://docs.djangoproject.com/en/2.2/topics/db/examples/many_to_many/) as well.

foreman/data_refinery_foreman/foreman/management/commands/create_compendia.py
kurtwheeler: Well I don't know about waiting before merging this. As it is now we'll already create compendia jobs for organisms that shouldn't have it so we aren't getting worse.

However I think we should either remove this check because it won't be necessary after #1757 or we should do it right. It doesn't look like we prioritized #1757 for this sprint though, so unless we want to do that it looks like it could be an issue for a bit?

arielsvn: What if we make a note on #1757 to remove the additional check once that is addressed? As it is now, this will generate the groups at https://github.com/AlexsLemonade/refinebio/issues/1736#issuecomment-541183861 which all seem like will be valid compendiums.

kurtwheeler: Oh wait, https://github.com/AlexsLemonade/refinebio/pull/1759 is already open. Let's just get that out and this'll be a non-issue. I'm reviewing it now.

arielsvn:
@kurtwheeler removed the check with the number of samples. This should be fine as long as we merge and remove the qn targets in https://github.com/AlexsLemonade/refinebio/pull/1759 before running this

arielsvn:
🚑 We are waiting until we associate the compendias with multiple organisms before merging this.

foreman/data_refinery_foreman/foreman/management/commands/create_compendia.py
```diff
@@ -12,42 +15,6 @@
...
-
-    pjda = ProcessorJobDatasetAssociation()
-    pjda.processor_job = job
-    pjda.dataset = dset
-    pjda.save()
-
-    return job
-
-
 class Command(BaseCommand):
```
arielsvn: test

foreman/data_refinery_foreman/foreman/management/commands/create_compendia.py
```diff
@@ -58,33 +29,130 @@ def add_arguments(self, parser):
             help=("Specify SVD algorithm applied during imputation ARPACK, RANDOMIZED or NONE to skip."))
 
     def handle(self, *args, **options):
-        """Create a compendium for one or more organisms.
+        """Create a compendium for one or more organisms."""
+        target_organisms = self._get_target_organisms(options)
+        grouped_organisms = group_organisms_by_biggest_platform(target_organisms)
+
+        quant_sf_only = options["quant_sf_only"] is True
```
kurtwheeler: This is now never an option for this command.

arielsvn:
@kurtwheeler just resolved conflicts. This should be ready for review if you want to take another look. Compendia tests passed locally.

foreman/data_refinery_foreman/foreman/management/commands/create_compendia.py
arielsvn: good catch! just removed it

PR #1754, opened 2019-10-09
Scale smasher back down because the rat job finished.

kurtwheeler:
## Issue Number

#1743 

We finished the rat job! We're gonna try more compendia once #1749 goes out, but I don't expect that to be today and this big instance costs a lot of money so let's save money overnight!!


kurtwheeler:
@arielsvn I also just pushed a volume size change.

PR #1759, opened 2019-10-10
Add command to remove invalid QN Targets

### LINKED ISSUE #1735: Create command to delete invalid compendias
### Context

https://github.com/AlexsLemonade/refinebio/pull/1734

Compendia need a QN target so they can be quantile normalized. We need 100 samples on the sample platform to create a QN target. Therefore any organisms without 100 samples on the same platform should be ineligible for a compendium.

### Problem or idea

We created a compendium for Bos Inidicus which you can see only has 57 samples: https://www.refine.bio/search?filter_order=organism&organism=bos_indicus

### Solution or next step

In https://github.com/AlexsLemonade/refinebio/pull/1734 I should have prevented us from trying to create a compendium for this situation.

However, we should still:
  * Look into how the compendia was successful despite not having a QN target to do that step.
  * Delete this and any other compendia that should have failed to be created in the first place.

### LINKED ISSUE #1757: Remove QN targets that should not exist
### Context

We seem to have QN Targets for both `BOS_TAURUS ` and `BOS_INDICUS`.

### Problem or idea

`BOS_INDICUS` only has 54 processed samples.

```
>>> Organism.objects.get(name='BOS_TAURUS').sample_set.filter(is_processed=True).count()
5354
>>> Organism.objects.get(name='BOS_INDICUS').sample_set.filter(is_processed=True).count()
34
```

We require each organism to have at least 100 processed microarray samples in order to be able to create QN Targets.

### Solution or next step

We should find and remove all the QN Targets that don't meet the minimum threshold.

arielsvn:
## Issue Number

close #1757 
close #1735 

Adds a command to remove all QN targets associated with organisms that have less than `100` processed microarray samples.

## Types of changes

## Functional tests

None, added option to `dry-run`

## Checklist


foreman/data_refinery_foreman/foreman/management/commands/remove_invalid_data.py
```diff
@@ -0,0 +1,50 @@
...
+    ComputationalResult,
+    ComputationalResultAnnotation,
+    Organism
+)
+from data_refinery_common.logging import get_and_configure_logger
+
+logger = get_and_configure_logger(__name__)
+
+# Minimum number of microarray samples required to create a QN Target
+QN_TARGET_SAMPLE_THRESHOLD = 100
```
cgreene: Would it be better to make this an argument to the command with 100 used as the default in case we one day change our mind on 100?

arielsvn: sure, good idea!

arielsvn:
@kurtwheeler I updated the check for the number of samples in the biggest platform in the new command and also in `create_qn_target`. Let me know what do you think

workers/data_refinery_workers/processors/management/commands/remove_invalid_qn_targets.py
```diff
@@ -0,0 +1,46 @@
...
+        if not organism_can_have_qn_target(organism, min_samples):
+            # remove the referenced QN Target because it shouldn't have a qn target
+            logger.debug('Removing QN Target because it does not have enough samples on its biggest microarray platform',
+                         computational_result=annotation.result,
+                         organism=organism)
+            qn_target_ids.append(annotation.result.id)
+
+    if not dry_run:
+        # delete all invalid qn targets
+        ComputationalResult.objects.filter(id__in=qn_target_ids).delete()
```
kurtwheeler: So I'm assuming that django will do a cascading delete here that will remove the ComputedFile object. However will that also delete the file from S3? I'm almost certain we don't have anything that will make it. Maybe we can use the pre_delete signal to have django handle this smoothly for us? https://docs.djangoproject.com/en/dev/ref/signals/#pre-delete

kurtwheeler: Either that or just clean up the file manually. Whatever works for you :)

arielsvn:
Removing the computed files from S3 manually seemed easier. What do you think @kurtwheeler ?

arielsvn:
@kurtwheeler added changes here to remove invalid computational results as well and address #1735.

common/data_refinery_common/models/models.py
```diff
@@ -560,6 +560,11 @@ def save(self, *args, **kwargs):
         self.last_modified = current_time
         return super(ComputationalResult, self).save(*args, **kwargs)
 
+    def remove_computed_files_from_s3(self):
+        """ Removes all associated computed files from S3. Use this before deleting a computational result. """
```
cgreene: Would it be better to call this in `pre_delete`? If I recall correctly pre_delete doesn't get called for bulk deletions but it does whenever an individual instance is deleted.

arielsvn: yes, it's better... here I used the simplest option which was just to manually remove the files. I'm gonna look into `pre_delete`

arielsvn: I was reading [some docs](https://simpleisbetterthancomplex.com/tutorial/2016/07/28/how-to-create-django-signals.html), still not sure how to implement the signals. I think we should put this on a different PR and add it for all models that sync to S3. I'm gonna create an issue for it.

PR #1772, opened 2019-10-14
Write metadata after we do the processing for smasher/compendia.

kurtwheeler:
## Issue Number

Log progress on writing metadata.

Right now the human compendia is writing metadata. This takes a long time. We may not be able to complete the rest of the compendia processing though. This will make it so we don't write the metadata files until we've done everything else.

I also found http://www.lucainvernizzi.net/blog/2015/08/03/8x-speed-up-for-python-s-csv-dictwriter/ which may help us a lot in terms of the time we spend writing metadata so I added it here as well.

## Types of changes

- Optimization

## Functional tests

Unit tests should cover this fairly well.


davidsmejia:
Eager to see the performance improvement!

Issue #1537, opened 2019-08-26
Make compendium generation work with production-scale data


cgreene:
### Context

Building a compendium remains slow. We might want to consider making it faster.

### Problem or idea

We're using the smasher code to put together a large matrix. This smasher code uses a list append ( https://github.com/AlexsLemonade/refinebio/blob/f3b40c0d9eaee1e7326063f6de100c8c7915e3aa/workers/data_refinery_workers/processors/smasher.py#L574 ).

This then gets concated:
https://github.com/AlexsLemonade/refinebio/blob/f3b40c0d9eaee1e7326063f6de100c8c7915e3aa/workers/data_refinery_workers/processors/smasher.py#L667

Unless Pandas is doing something like looking ahead at all of the indices (which is what we're going to propose), it may get result in a significant hit to performance either looking for, reindexing, and remallocing data.

### Solution or next step

To figure out if this is worth investing time into:

1. We should benchmark how rapidly we're able to download files from S3. This can provide a nice lower bound on time. It should also be linear with respect to the number of samples in the compendium.
2. We should use timing to benchmark where time is being spent for the compilation of the merged file, filtering, imputation, and writing. We should do this from small (100) to larger (200, 400, 800, 1600, 3200, 6400, 12800) sizes.
3. We should plot the time spent in each stage as a function of size. What we're looking for is anything that goes up at a higher rate than our sample size (i.e., more than doubles).

If it turns out we've got places to optimize, we might be able to. The code I noted above is potentially suspicious. If it turns out to be an issue, we could pre-specify the result matrix, preallocate the numpy array of the appropriate size, and then fill it in. The approach would be something like:

* Store the title of each sample and an arbitrary index.
* Iterate through every sample and union the genes in the sample with all genes found thus far.
* After this, assign an arbitrary index to each gene.
* Store an arbitrary index for each gene.
* Create a NumPy array with NAs filling the entire array with the appropriate dimensions for the number of samples and genes.
* Iterate through each file, filling in the array above with the appropriate values at each index.
* After this, create the pandas dataframe from the numpy array. Assign the indices.

cgreene:
Completed

PR #1773, opened 2019-10-14
paginate multiprocessing pool maps

davidsmejia:
## Issue Number

#1537 

There were a few problems with the previous iteration of parallelizing the process_frame method in smasher.py. One of which was the amount of data being passed between threads was too large and this caused the smasher to crasher. This implements a nested generator which allows for the chunking of frames to be processed while keeping overhead at a minimum.

## Methods

```process_frames_for_key``` in ```smasher.py```

## Types of changes

## Functional tests

```./workers/run_tests.py -t smasher```
and 
```./workers/run_tests.py -t compendia```
are passing
also tested compendia with various ```MULTIPROCESS_CHUNK_SIZE``` which good results

## Checklist

### DIFF
--- a/workers/data_refinery_workers/processors/smashing_utils.py
+++ b/workers/data_refinery_workers/processors/smashing_utils.py
@@ -273,7 +273,10 @@ def process_frames_for_key(key: str, input_files: List[ComputedFile], job_contex
                              job_context["job"])
 
     def get_frame_inputs():
-        """Helper method to create a generator."""
+        """Helper method to create a generator object that when iterated on returns
+        a tuple for smashing_utils.process_frame method to process a single frame.
+        It returns a tuple becuase process_frame is called with multiprocessing.Pool.map
+        which restricts how arguments are passed"""
         for index, (computed_file, sample) in enumerate(input_files):
             # Don't pass job_context to worker threads because
             # for some reason it causes them to open database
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: workers/data_refinery_workers/processors/smashing_utils.py [hunk 1]
```diff
@@ -288,17 +288,44 @@ def get_frame_inputs():
...
+        MULTIPROCESSING_CHUNK_SIZE"""
+        source = get_frame_inputs()
+        iterator = itertools.islice(source, MULTIPROCESSING_CHUNK_SIZE)
+        while True:
+            item = next(iterator, None)
+            if item is None:
+                return
+            source, iterator = itertools.tee(itertools.chain([item], source))
+            yield itertools.islice(iterator, MULTIPROCESSING_CHUNK_SIZE)
+            # consume the source iterator to keep track of current chunk
```
kurtwheeler: I think we're gonna need a much more comprehensive set of docstrings/comments on this function. As it is it is really hard to tell by reading the code:
  * What we're trying to do, like the general strategy of the chunking/generators/multiprocessing.
  * Why we're trying to do that.
  * What kind of iterable object will be returned and what it's contents are. Are there multiple levels of iterables here? If so what purpose are they serving?
 
I think you tried to describe this in the docstring:
> Helper method to create generator for chunking get_frame_inputs chunked into generators that return a slice with size MULTIPROCESSING_CHUNK_SIZE

But that's rather terse, `get_frame_inputs` is a function (I'm assuming that you mean the output of it), and I think explaining this better might go a long way to making this more understandable.

davidsmejia: 👍 I think this might need another iteration (no pun intended) but let me know if the last commit helps clear things up

PR #1781, opened 2019-10-15
Ensure quant.sf are in s3 when returning them

arielsvn:
## Issue Number

N/A

Came up while running a big dataset with quantfiles. We were trying to download files where their s3_key was empty.

### DIFF
--- a/common/data_refinery_common/models/models.py
+++ b/common/data_refinery_common/models/models.py
@@ -1023,6 +1023,9 @@ class ComputedFile(models.Model):
         target_directory = os.path.dirname(self.absolute_file_path)
         os.makedirs(target_directory, exist_ok=True)
 
+        if not self.s3_bucket or not self.s3_key:
+            raise Exception('Tried to download a computed file with no s3_bucket or s3_key')
+
         try:
             S3.download_file(
                         self.s3_bucket,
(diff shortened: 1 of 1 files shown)


common/data_refinery_common/models/models.py
```diff
@@ -1022,6 +1023,9 @@ def sync_from_s3(self, force=False, path=None):
         target_directory = os.path.dirname(self.absolute_file_path)
         os.makedirs(target_directory, exist_ok=True)
 
+        if not self.s3_bucket or not self.s3_key:
+            raise Exception('Tried to download a computed file with no s3_bucket or s3_key')
```
kurtwheeler: Hmm, how about a KeyError or something instead of a raw Exception? You aren't really supposed to raise those directly.

arielsvn: Good point, what if we raise [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError)? 

`KeyError` seems to be for dicts

arielsvn: Good point, what if we raise ValueError? KeyError seems to be for dicts

kurtwheeler: Sounds good to me!

PR #1787, opened 2019-10-15
Create quantpendia processor job

### LINKED ISSUE #1751: Create separate processor job for the quantpendia
### Context

We reuse a lot of the `smasher` logic in the `create_compendia.py` processor job.

### Problem or idea

Seems like a lot of this logic is deeply tied to the dataset being aggregated by species. And it relies on `job_context['samples'][organism_key]`

https://github.com/AlexsLemonade/refinebio/blob/dev/workers/data_refinery_workers/processors/create_compendia.py#L307

In the case of the quantpendia, the datasets will be aggregated by experiment so all that logic will fail.

### Solution or next step

We should have logic specific for generating the quantpendia, we could modify the code in the current processor or add a new job that only takes care of generating the quantpendia.

I lean towards the second one, what do you think @kurtwheeler ?

arielsvn:
## Issue Number

close #1751 

This creates a new job that should take care of generating the quantpendias

## Types of changes

## Functional tests

Added tests

## Checklist


workers/data_refinery_workers/processors/test_create_quantpendia.py
```diff
@@ -0,0 +1,106 @@
...
+        pjda = ProcessorJobDatasetAssociation()
+        pjda.processor_job = job
+        pjda.dataset = ds
+        pjda.save()
+
+        final_context = create_quantpendia(job.id)
+
+        self.assertTrue(os.path.exists(final_context['output_dir'] + '/GSE51088/GSM1237818_quant.sf'))
+        self.assertTrue(os.path.exists(final_context['output_dir'] + '/README.md'))
+        self.assertTrue(os.path.exists(final_context['output_dir'] + '/LICENSE.txt'))
```
arielsvn:
```suggestion
        self.assertTrue(os.path.exists(final_context['output_dir'] + '/LICENSE.TXT'))
```

workers/nomad-job-specs/create_quantpendia.nomad.tpl
```diff
@@ -0,0 +1,99 @@
...
+
+        LOG_LEVEL = "${{LOG_LEVEL}}"
+      }
+
+      # The resources the job will require.
+      resources {
+        # CPU is in AWS's CPU units.
+        cpu =   4000
+        # Memory is in MB of RAM. Instance has 1,952GB of RAM.
+        memory = 8192
```
kurtwheeler: This is going to be enough for all organisms? I mean... I can see it cause we don't have to store too much data in memory at any point. It just seems really low, but it might be appropriate.

arielsvn: Not sure, seemed enough since the job will only download quant files and the compression should not take more than 2gb of ram. Do you think we should increase it?

kurtwheeler: Not necessarily. Do you know how large the metadata object for human will be? I think that might be the peak RAM usage actually.

The other thing to consider is that number of quantpendia jobs running on a single machine will be `(1000 or 2000 depending on instance type) / 8)` and the machine will only have 9TB of disk space available to write the files to. Would 125 or 250 of these jobs running simultaneously cause the volume to fill up?

common/data_refinery_common/job_lookup.py
```diff
@@ -129,12 +136,14 @@ class ProcessorPipeline(Enum):
...
     NONE = "NONE"
     CREATE_COMPENDIA = "CREATE_COMPENDIA"
+    CREATE_QUANTPENDIA = "CREATE_QUANTPENDIA"
 
 
 SMASHER_JOB_TYPES = [
     ProcessorPipeline.SMASHER,
     ProcessorPipeline.QN_REFERENCE,
     ProcessorPipeline.CREATE_COMPENDIA,
+    ProcessorPipeline.CREATE_QUANTPENDIA,
```
kurtwheeler: :+1:

workers/nomad-job-specs/create_quantpendia.nomad.tpl
arielsvn: In terms of disk space, we had calculated that 5TB should be enough for all organism quantpendias. (ref https://github.com/AlexsLemonade/refinebio/pull/1642)

The metadata was about 3.3Gb for the human, which is probably the biggest one.

```
$ du -sh /var/ebs/smashed/714ec9dc-b5c4-4465-8a45-d46c173d3252/output/*
3.3G    /var/ebs/smashed/714ec9dc-b5c4-4465-8a45-d46c173d3252/output/aggregated_metadata.json
958G    /var/ebs/smashed/714ec9dc-b5c4-4465-8a45-d46c173d3252/output/HOMO_SAPIENS
20K    /var/ebs/smashed/714ec9dc-b5c4-4465-8a45-d46c173d3252/output/LICENSE.TXT
4.0K    /var/ebs/smashed/714ec9dc-b5c4-4465-8a45-d46c173d3252/output/README.md
```

I think 8GB should be enough, but we can change it. What size do you think would be more appropiate?

kurtwheeler: Nah all that sounds good, I just wanted to make sure you had thought it through.

workers/data_refinery_workers/processors/test_create_quantpendia.py
```diff
@@ -1,28 +1,17 @@
...
-    SampleResultAssociation
-)
-from data_refinery_workers.processors.create_quantpendia import create_quantpendia
+from data_refinery_common.models import (ComputationalResult, ComputedFile,
+                                         Dataset, Experiment,
+                                         ExperimentSampleAssociation, Organism,
+                                         ProcessorJob,
+                                         ProcessorJobDatasetAssociation,
+                                         Sample, SampleComputedFileAssociation,
+                                         SampleResultAssociation)
```
kurtwheeler: Whhyyyyyyyy?!?

workers/data_refinery_workers/processors/create_quantpendia.py
```diff
@@ -0,0 +1,133 @@
+import os
+import shutil
+import time
+from django.utils import timezone
+from typing import Dict, List, Tuple
+
+from data_refinery_common.job_lookup import PipelineEnum
+from data_refinery_common.logging import get_and_configure_logger
+from data_refinery_common.models import (ComputationalResult,
+                                         ComputedFile, Organism, Pipeline,
```
kurtwheeler: Whyyyyyy?

PR #1811, opened 2019-10-20
compendia merge with dask

davidsmejia:
## Issue Number

#1537 

The idea here is to prevent either a MemoryError or OOM kill from docker as we try to merge substantially large dataframes into a compendium dataframe. This is happening before imputation as we prepare and merge the files. 

## Methods

`smashing_utils.py:_process_frame_for_key`

Keeps the pandas.concat for each chunk as this seems to fit into memory just fine. However the first concatenated frames are merged into a `dask.DataFrame` instead and we merge each chunk iteration into that. After we are done with all chunks we call `dask.DataFrame.compute` on the microarray_matrix dataframe and rnaseq_matrix dataframe to convert it back to pandas and place it in memory.

## Types of changes

## Functional tests

ran `./workers/run_tests.sh -t compendia` test cases and also set the MULTIPROCESSING_CHUNK_SIZE to 20 and ran `./workers/run_tests.sh -t compendia` test cases 

All of which passes.

## Checklist

### DIFF
--- a/workers/data_refinery_workers/processors/smashing_utils.py
+++ b/workers/data_refinery_workers/processors/smashing_utils.py
@@ -33,7 +33,7 @@ from data_refinery_workers.processors import utils
 # Use floor here because multiprocessing raises an exception if this isn't an int.
 MULTIPROCESSING_WORKER_COUNT = max(1, math.floor(multiprocessing.cpu_count()/2) - 1)
 MULTIPROCESSING_CHUNK_SIZE = 2000
-INDEX_DASK_START = 20000
+INDEX_DASK_START = 200000
 RESULTS_BUCKET = get_env_variable("S3_RESULTS_BUCKET_NAME", "refinebio-results-bucket")
 S3_BUCKET_NAME = get_env_variable("S3_BUCKET_NAME", "data-refinery")
 BODY_HTML = Path(
--- a/workers/data_refinery_workers/processors/smashing_utils.py
+++ b/workers/data_refinery_workers/processors/smashing_utils.py
@@ -20,6 +20,7 @@ from rpy2.robjects.packages import importr
 from typing import Dict, List
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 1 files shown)


workers/data_refinery_workers/processors/smashing_utils.py
```diff
@@ -33,6 +33,7 @@
 # Use floor here because multiprocessing raises an exception if this isn't an int.
 MULTIPROCESSING_WORKER_COUNT = max(1, math.floor(multiprocessing.cpu_count()/2) - 1)
 MULTIPROCESSING_CHUNK_SIZE = 2000
+INDEX_DASK_START = 20000
```
kurtwheeler: I thought you said you were gonna do this at 200k?

davidsmejia: hmm yeah good catch

workers/data_refinery_workers/processors/smashing_utils.py
```diff
@@ -387,6 +397,20 @@ def process_frames_for_key(key: str,
...
 
                 del rnaseq_chunk_frame
+                # start using dask to save ram at scale
+                if index > INDEX_DASK_START and type(job_context['rnaseq_matrix']) is pd.DataFrame:
+                    job_context['rnaseq_matrix'] = dd.from_pandas(job_context.pop('rnaseq_matrix'),
+                                                                      npartitions=MULTIPROCESSING_WORKER_COUNT)
+
+            log_state("end merging chunk of frames",
+                      job_context["job"],
+                      start_frame_chunk)
```
kurtwheeler: Can we get a counter on this and the `merging chunk of frames` message as well? Otherwise we're just gonna have 250 of these with no indication of progress.

davidsmejia: well it would be after the processing index line

PR #1820, opened 2019-10-22
Preallocate gene identifiers so compendia matrix only grows in one dimension

kurtwheeler:
## Issue Number

#1258 

Merging the matrix together has been giving us trouble RAM-wise. Our current theory is that it's because we're merging samples together iteratively, and as we do we discover more and more gene identifiers that we have to include for the entire matrix, causing the matrix to expand in two directions.

In this PR we instead go through all of the samples once to discover the gene identifiers, then once we have the full set we go through to actually populate the matrix. 

This PR also makes us reuse the same directory for every organism's compendia job and not clean it up when we finish. This way subsequent tests will go faster.

## Methods

To build the list of gene identifiers I do so with a set to remove any duplicates, then sort the list so that all the matrices have the same order of indices so that we don't have to reorder anything (especially later when we combine microarray and rnaseq matrices).

## Types of changes

- Optimization

## Functional tests

So the unit tests have better data than I do for testing, but with them I've made sure that:
  * It works with a chunk size of 200 (to make sure the multiple chunks still works well)
  * `list(job_context['rnaseq_matrix'].index) == list(job_context['microarray_matrix'].index)`

## Checklist

### DIFF
--- a/workers/data_refinery_workers/processors/create_compendia.py
+++ b/workers/data_refinery_workers/processors/create_compendia.py
@@ -64,6 +64,8 @@ def _prepare_input(job_context: Dict) -> Dict:
         # use the same directory every job.
         job_context["old_work_dir"] = job_context["work_dir"]
         job_context["work_dir"] = "/home/user/data_store/" + job_context["organism_name"] + "/"
+        if not os.path.exists(job_context["work_dir"]):
+            os.makedirs(job_context["work_dir"])
 
     log_state("prepare input done", job_context["job"], start_time)
     return job_context
--- a/workers/data_refinery_workers/processors/create_compendia.py
+++ b/workers/data_refinery_workers/processors/create_compendia.py
@@ -1,31 +1,24 @@
-import gc
 import logging
 import os
-import psutil
 import shutil
 import time
+from typing import Dict
 
+from django.utils import timezone
+from fancyimpute import IterativeSVD
 import numpy as np
 import pandas as pd
-from fancyimpute import IterativeSVD
-import simplejson as json
-
-from datetime import datetime
-from django.utils import timezone
-from typing import Dict
+import psutil
 
 from data_refinery_common.job_lookup import PipelineEnum
 from data_refinery_common.logging import get_and_configure_logger
-from data_refinery_common.models import (
-    ComputationalResult,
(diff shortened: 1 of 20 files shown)


kurtwheeler:
@jaclyn-taroni @cgreene this has a methods section if you're interested.

cgreene:
I am going to summarize how I read the methods section:

You are first iterating through all of the items going into the compendia. You are creating a set that has all of the gene identifiers. You then convert that into a list and sort it.

At the second stage, when you read files to create the dataframe, you specify the gene indices using the sorted list from before. This way, your dataframe never grows in the gene dimension.

Is that correct? If so, that sounds appropriate to me.

kurtwheeler:
@cgreene that is correct.

The exact way that the gene indices are specified (once we have the full set) is by reading in a sample's file to a dataframe, using the `reindex` method to set it's index to the full gene id set, `concat`ing these in chunks of 2000, then appending to the growing matrix. Because the index is set to the sorted full list of gene indentifiers on a sample by sample basis, all matrices everywhere should have the same index in the same order so operations should not have to increase or sort indices anywhere.

All of that is low level implementation detail, but I wanted to try and make that clear.

workers/data_refinery_workers/processors/create_compendia.py
```diff
@@ -66,6 +59,14 @@ def _prepare_input(job_context: Dict) -> Dict:
     if job_context['job'].success is not False:
         job_context["organism_name"] = job_context['group_by_keys'][0]
 
+        # TEMPORARY for iterating on compendia more quickly. Rather
+        # than downloading the data from S3 each run we're just gonna
+        # use the same directory every job.
+        job_context["old_work_dir"] = job_context["work_dir"]
+        job_context["work_dir"] = "/home/user/data_store/" + job_context["organism_name"] + "/"
```
arielsvn: Should we add `/smashed/` here?

```suggestion
        job_context["work_dir"] = SMASHING_DIR + job_context["organism_name"] + "/"
```

kurtwheeler: Sure, why not!

davidsmejia:
LGTM! great job!

Issue #1823, opened 2019-10-23
Quantpendias seem to be stuck


arielsvn:
### Context

Yesterday we ran quantpendia jobs for `HOMO_SAPIENS`, `MUS_MUSCULUS` and `DANIO_RERIO`

```
CREATE_QUANTPENDIA/dispatch-1571776183-d3109607   batch                80        running  2019-10-22T20:29:43Z
CREATE_QUANTPENDIA/dispatch-1571776420-889dce08   batch                80        running  2019-10-22T20:33:40Z
CREATE_QUANTPENDIA/dispatch-1571776659-8a07aa71   batch                80        running  2019-10-22T20:37:39Z
```

Only the one for `DANIO_RERIO` completed successfully [CREATE_QUANTPENDIA/dispatch-1571776183-d3109607](https://s3.amazonaws.com/data-refinery-s3-circleci-prod/v4uuojgiw7w8grng8tx7caiu_e762c998-3ecb-4a0e-a9cb-d55dc2fdf3b4_compendia.zip). The other two are still running after 20 hours.

### Problem or idea

The human and mouse jobs don't seem to be doing anything. I have been checking the number of downloaded quant files for both and they seem to be stuck.

```
$ find d4f84b68-9e55-495d-a913-9ece90d69939/output -maxdepth 2 -name "*_quant.sf" | wc -l
35789
$ find 8e3de724-7923-4b9a-bb77-4d0fdc64411b/output -maxdepth 2 -name "*_quant.sf" | wc -l
26976
```

It's likely in the part where we download the quant files in parallel https://github.com/AlexsLemonade/refinebio/blob/dev/workers/data_refinery_workers/processors/smashing_utils.py#L955

The sizes of the respective downloaded quant files:

```
$ du -hs d4f84b68-9e55-495d-a913-9ece90d69939
200G    d4f84b68-9e55-495d-a913-9ece90d69939
$ du -hs 8e3de724-7923-4b9a-bb77-4d0fdc64411b
213G    8e3de724-7923-4b9a-bb77-4d0fdc64411b
```

Also, looking and the instance's network usage. Looks that it hasn't downloaded anything in the last 10 hours.

![image](https://user-images.githubusercontent.com/1882507/67423964-3c339a80-f5a3-11e9-9c6f-95a11bc70b4f.png)

### Solution or next step

We need to investigate what might be causing this, some theories that we have had:

[...]

PR #1824, opened 2019-10-23
Increase ram for quantpendia and smasher constraint

arielsvn:
## Issue Number

#1823

Add more ram for the next time we try the quantpendias. Also, ensure it runs on the smasher since the job will be running for a couple of hours and we don't want it to be interrupted.

## Types of changes

- Parameter change

## Functional tests

None

## Checklist

### DIFF
--- a/workers/nomad-job-specs/create_quantpendia.nomad.tpl
+++ b/workers/nomad-job-specs/create_quantpendia.nomad.tpl
@@ -69,7 +69,7 @@ job "CREATE_QUANTPENDIA" {
       resources {
         # CPU is in AWS's CPU units.
         cpu =   4000
-        memory = 12288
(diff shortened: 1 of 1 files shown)


workers/nomad-job-specs/create_quantpendia.nomad.tpl
```diff
@@ -69,15 +69,16 @@ job "CREATE_QUANTPENDIA" {
       resources {
         # CPU is in AWS's CPU units.
         cpu =   4000
-        # Memory is in MB of RAM. Instance has 1,952GB of RAM.
-        memory = 8192
+        memory = 12288
```
arielsvn:
```suggestion
        memory = 16384
```

PR #1825, opened 2019-10-23
Remove multiprocessing, preallocate entire matrix.

kurtwheeler:
## Issue Number

#1258 

So it turns out that maxtasksperchild doesn't work like we thought it did. At least within the context of that parameter-name, `task` means one chunk of work, so really that parameter is specifying how many chunks each Process will do instead of tasks. Turns out that's about `2k/num_cores`....

Anyway this is causing multiprocessing to blow up our RAM usage. I'm looking into ways to use it without having it quintuple our RAM usage, but since we're no longer redownloading samples doing this sequentially should finish in a reasonable amount of time.

I also used the first pass over the data to count how many microarray and rnaseq samples would be found, so that I could actually preallocate the full matrix. This seems to make the `process_frames_for_key` function take a rather flat amount of RAM for the tests: right around .57 GB consistently. This makes me hopeful that we'll see flat RAM usage in prod as well.

## Methods

We're no longer using `concat` or `merge` anywhere in our code and instead we're preallocating the matrix and just directly setting the values. This should work well because we sort the indices on each matrix before doing this.

## Types of changes

## Functional tests

The unit tests look pretty good! I'm excited to see how it does in prod.

### DIFF
--- a/workers/data_refinery_workers/processors/create_compendia.py
+++ b/workers/data_refinery_workers/processors/create_compendia.py
@@ -87,8 +87,7 @@ def _prepare_frames(job_context: Dict) -> Dict:
         for key, input_files in job_context.pop('input_files').items():
             job_context = smashing_utils.process_frames_for_key(key,
                                                                 input_files,
-                                                                job_context,
-                                                                merge_strategy='outer')
+                                                                job_context)
             # if len(job_context['all_frames']) < 1:
             # TODO: Enable this check?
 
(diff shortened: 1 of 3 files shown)


kurtwheeler:
@davidsmejia I had to make some significant changes to keep the smasher doing an inner merge instead of an outer merge. Can you take another look?

davidsmejia:
OK looks good!

Issue #1733, opened 2019-10-03
Smasher instance only has 2TB available to it.


kurtwheeler:
### Context

I was wondering how the smasher was doing space-wise:
```
$ df -h
Filesystem      Size  Used Avail Use% Mounted on
udev            250G     0  250G   0% /dev
tmpfs            50G   25M   50G   1% /run
/dev/nvme0n1p1  2.0T  201G  1.8T  11% /
tmpfs           250G     0  250G   0% /dev/shm
tmpfs           5.0M     0  5.0M   0% /run/lock
tmpfs           250G     0  250G   0% /sys/fs/cgroup
tmpfs            50G     0   50G   0% /run/user/1000
```

### Problem or idea

Turns out that it's partition is only 2TB instead of the 6 we wanted and have a volume for:
```
$ lsblk
NAME        MAJ:MIN RM  SIZE RO TYPE MOUNTPOINT
nvme0n1     259:0    0  5.9T  0 disk 
└─nvme0n1p1 259:1    0    2T  0 part /
```

We should use the full 6TB that we have for that instance.

I tried to follow AWS' guide to doing this: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/recognize-expanded-volume-linux.html

However when I ran the command to do so it gave an error:
```
$ sudo growpart /dev/nvme0n1 1
WARNING: MBR/dos partitioned disk is larger than 2TB. Additional space will go unused.
NOCHANGE: partition 1 could only be grown by 1 [fudge=2048]
```

According to https://unix.stackexchange.com/questions/429507/cannot-extend-partition-beyond-2tb-on-aws-ubuntu

> Your drive is formatted as MBR.

> For drives larger than 2TB, they need to be partitioned as GPT as MBR is limited to 2TB regardless of the OS.

### Solution or next step

The internet has guides to do this: https://www.cyberciti.biz/tips/fdisk-unable-to-create-partition-greater-2tb.html

We should make `infrastructure/nomad-configuration/client-smasher-instance-user-data.tpl.sh` do this when the smasher starts up.

davidsmejia:
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/raid-config.html

```sudo mdadm --create --verbose /dev/md0 --level=0 --name=MY_RAID --raid-devices=number_of_volumes device_name1 device_name2```

kurtwheeler:
This was closed by #2552!

PR #1830, opened 2019-10-24
Increase the ram for the quantpendia 2x

arielsvn:
## Issue Number

#1823

The human quantpendia got stuck again, using all the ram that had available. This duplicates it to 32gb.

## Types of changes

## Functional tests

None

## Checklist

### DIFF
--- a/workers/data_refinery_workers/processors/create_quantpendia.py
+++ b/workers/data_refinery_workers/processors/create_quantpendia.py
@@ -28,15 +28,9 @@ def create_quantpendia(job_id: int) -> None:
                                      [utils.start_job,
                                       download_files,
                                       create_result_objects,
+                                      remove_job_dir,
                                       utils.end_job])
     return job_context
-
-
-def get_process_stats():
-    BYTES_IN_GB = 1024 * 1024 * 1024
-    process = psutil.Process(os.getpid())
-    ram_in_GB = process.memory_info().rss / BYTES_IN_GB
-    return { 'total_cpu': psutil.cpu_percent(), 'process_ram': ram_in_GB }
 
 
 def download_files(job_context: Dict) -> Dict:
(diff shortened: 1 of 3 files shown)


arielsvn:
Added a constraint to prevent the quantpendia job from running in the smasher, since we still need to address #1733

workers/data_refinery_workers/processors/create_quantpendia.py
```diff
@@ -19,25 +20,19 @@
...
-def get_process_stats():
-    BYTES_IN_GB = 1024 * 1024 * 1024
-    process = psutil.Process(os.getpid())
-    ram_in_GB = process.memory_info().rss / BYTES_IN_GB
-    return { 'total_cpu': psutil.cpu_percent(), 'process_ram': ram_in_GB }
-
-
 def download_files(job_context: Dict) -> Dict:
     job_context['time_start'] = timezone.now()
     _make_dirs(job_context)
```
kurtwheeler:
```suggestion
    job_context = _make_dirs(job_context)
```
You set some properties in there. Sure this is modifying `job_context` so it's not truly necessary, but doing this will make it explicit that `job_context` is being updated by `_make_dirs`.

workers/data_refinery_workers/processors/create_quantpendia.py
```diff
@@ -124,12 +120,25 @@ def create_result_objects(job_context: Dict) -> Dict:
...
 def _make_dirs(job_context: Dict):
     dataset_id = str(job_context["dataset"].pk)
-    job_context["work_dir"] = "/home/user/data_store/smashed/" + dataset_id + "/"
-    os.makedirs(job_context["work_dir"])
-    job_context["output_dir"] = job_context["work_dir"] + "output/"
-    os.makedirs(job_context["output_dir"])
+    job_context["job_dir"] = "/home/user/data_store/smashed/" + dataset_id + "/"
+    os.makedirs(job_context["job_dir"], exist_ok=True)
+    job_context["output_dir"] = job_context["job_dir"] + "output/"
+    os.makedirs(job_context["output_dir"], exist_ok=True)
```
kurtwheeler:
```suggestion
    os.makedirs(job_context["output_dir"], exist_ok=True)
    
    return job_context
```
This goes with the other comment above.

kurtwheeler:
That's an even better way to do it! :+1:

workers/data_refinery_workers/processors/create_quantpendia.py
arielsvn: good point! moved `make_dirs` into one of the functions in `run_pipeline`

Issue #1691, opened 2019-09-30
Missing Experiment GSE131617


cgreene:
### Context

A user requested this dataset.

### Problem or idea

This one is interesting - it was uploaded on May 26, 2019. I thought that perhaps we hadn't surveyed microarray datasets since then, except I think that @davidsmejia did just re-survey. We should see if our code to re-survey missed this one. It's pretty big (~425 human samples). If it got missed, maybe some others did as well.

### Solution or next step

Check to see if the re-survey code found this. The user who reported it does want updates, so when this gets closed we should let them know (see the experiment request form associated with this accession).

davidsmejia:
Looks like it was not in the last batch of re-run microarray experiments accession codes. It appears as though the sample's platform is `GPL5188` which is not on our list of supported platforms

cgreene:
BrainArray does support that platform, and there's ~10k samples on it. We should be able to process it, but it may need the extra RAM that the HTA2.0 jobs are using (what @arielsvn was just adding).

cgreene:
Interestingly this isn't yet available. It looks like we have ~18k samples from the platform processed, so perhaps these are still in the backlog:
https://www.refine.bio/search?p=3&ordering=-num_downloadable_samples&filter_order=organism,platform,platform,platform,platform&organism=homo_sapiens&platform=GPL5175&platform=HuEx-1_0-st&platform=GPL23541&platform=GPL15236

cgreene:
May be due to #1986

cgreene:
I have emailed the requester letting them know that we had made all but two samples available. I cc'd @dvenprasad in case they elect to share more about how they are using the data.

```
I'm writing to follow-up on your request for the experiment GSE131617. We've now added support for this platform and have been able to process the experiment successfully. As a quick heads up, there are two remaining samples for the platform that failed to process with the same error: "GSM3791176_Expression_BN_V-VI_13-FC_090612.CEL has a disproportionate number of zero values, so it cannot be processed."

It looks like those two samples may have some QC issues or other problems t hat cause SCAN to fail to process them.

Thank you for reaching out to us! We'd love to hear about how you're using data from refine.bio.

Sincerely,
Casey
```

davidsmejia:
I would say this is ready. Looks like there are 2 samples we won't be able to get

https://www.refine.bio/experiments/GSE131617/genes-associated-with-the-progression-of-neurofibrillary-tangles-in-alzheimers-disease

PR #1831, opened 2019-10-24
support microarray platform huex + add requested experiment GSE131617

davidsmejia:
## Issue Number

#1691 

This PR is to add experiment `GSE131617` to be surveyed. However, there were two reasons why it was not already processed.
1 - The platforms `GPL5175` and `GPL5188` were not included in `/config/supported_microarray_platforms.csv` so it would have been filtered out. (this PR remedies that by adding all HuEx accession_codes to that file)
2 - The GEOmetadb R package which actually consumes the above list for fetching new experiments is not updated continuously and as of today it is not being returned still. (This is not addressed in this PR and I will post an issue in [accession_retrieval](https://github.com/AlexsLemonade/accession_retrieval) for future benefit.)

And lastly,
Adds the requested experiment manually to the list so it should be queued after running `feed_the_beast` command.

## Methods

None.

## Types of changes

## Functional tests

None

## Checklist


davidsmejia:
This should create 635 experiments.

kurtwheeler:
@davidsmejia you were going to change this PR so that we're only appending new accessions to the list rather than completely regenerating it right?

kurtwheeler:
@davidsmejia I did set union on the list of accessions before this PR and after this PR and then did a natsort on them with this code:

```
from natsort import natsorted

old = set()
new = set()

with open('all_microarray_accessions.txt') as old_file:
    for accession in old_file:
        old.add(accession.strip())

with open('config/all_microarray_accessions.txt') as new_file:
    for accession in new_file:
        new.add(accession.strip())

all_accessions = old.union(new)

sorted_accessions = natsorted(all_accessions)
with open('new_accessions.txt', 'w') as all_file:
    for accession in sorted_accessions:
        all_file.write(accession + '\n')
```
Wanna take a look and see if this looks good to merge?

PR #1843, opened 2019-10-29
Add `ProcessorJobError` to be used when processor jobs fail

arielsvn:
## Issue Number

N/A

Got this idea to refactor how we handle errors on the processor jobs, instead of using the `job_context['failure_reason']` this uses an exception that is handled in a single place.

What do you think @kurtwheeler?

## Types of changes

## Functional tests

None

## Checklist

### DIFF
--- a/workers/data_refinery_workers/processors/utils.py
+++ b/workers/data_refinery_workers/processors/utils.py
@@ -64,10 +64,7 @@ def prepare_original_files(job_context):
     original_files = job.original_files.all()
 
     if original_files.count() == 0:
-        logger.error("No files found.", processor_job=job.id)
-        job_context["success"] = False
-        job.failure_reason = "No files were found for the job."
-        return job_context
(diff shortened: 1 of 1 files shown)


kurtwheeler:
I think I like this I just want to think on it a bit more!!

kurtwheeler:
BTW I think you should pull `dev` into this branch.

arielsvn:
Just merged `dev` I'm going to wait until the tests pass to close this.

Yeah, I think this will allow us to simplify the logic to handle errors in the processor jobs and raise exceptions instead of working with `job_context["success"]` and `job_context["failure_reason"]`

> I think we can even extend the exception class to include additional context to log in the error message!

How? This sounds like a good idea! We can do it in a different PR.

kurtwheeler:
> How? This sounds like a good idea! We can do it in a different PR.

I agree and I made an issue for it: https://github.com/AlexsLemonade/refinebio/issues/1874

PR #1852, opened 2019-10-31
Use threadpool when downloading quant.sf files

arielsvn:
## Issue Number

#1823 

The quantpendia jobs were OOM-killed because they ran out of memory when creating new processes. This uses Threads instead to download that data.

## Types of changes

## Functional tests

None

## Checklist

### DIFF
--- a/workers/data_refinery_workers/processors/smashing_utils.py
+++ b/workers/data_refinery_workers/processors/smashing_utils.py
@@ -4,6 +4,7 @@ import csv
 import logging
 import math
 import os
+import multiprocessing
 import shutil
 import time
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


workers/data_refinery_workers/processors/smashing_utils.py
```diff
@@ -913,25 +909,21 @@ def sync_quant_files(output_path, samples: List[Sample]):
...
-            latest_computed_file = sample.get_most_recent_quant_sf_file()
-            if not latest_computed_file:
-                continue
-            output_file_path = output_path + sample.accession_code + "_quant.sf"
-            sample_and_computed_files.append((latest_computed_file, output_file_path))
-
-        # download this set of files, this will take a few seconds that should also help the db recover
-        pool.map(downlad_computed_file, sample_and_computed_files)
-        num_samples += len(sample_and_computed_files)
+    with ThreadPoolExecutor(max_workers=16) as executor:
```
kurtwheeler: Can you move this number back to the top, define it as a constant, and then leave a comment explaining why you picked 16?

arielsvn: good idea, 16 seemed reasonable but not sure if it's the best number for this. Do you have any suggestions on how to calculate a better number for this?

kurtwheeler: Hmm, the way it was being calculated based on the number of cores of the instance seemed like a good way to do it to me. Not sure why we'd wanna switch away from that. Whether we're using processes or threads we should still base the number off the number of threads.

arielsvn: just updated it and used: `max(1, math.floor(multiprocessing.cpu_count()/2) - 1)`

Issue #1855, opened 2019-11-01
Some samples with non-normalized IDs have snuck through


cgreene:
### Context

We have been building a compendia for Humans, and it has required more RAM than has made sense. @kurtwheeler found that we were trying to build an initial matrix for filtering with more than 500k identifiers. We should have ~25-50k identifiers.

### Problem or idea

Many of the identifiers are either unconverted gene identifiers or otherwise IDs that we don't support in our human compendia `mmu-miR-665-star_st`, `hp_hsa-mir-30d_x_st` as well as quite a few SNP and other identifiers.

### Solution or next step

We should figure out where these are coming from because they increase our RAM requirements ~10 fold. The mouse ones in particular are confusing. These should get pruned before the compendia are created at the stage that prunes genes that appear in very few samples, but it's massively inefficient to have them around. We'll want to reprocess these datasets with repaired code, because it appears that in certain cases our identifier mapping code is not successful.

cgreene:
We should still figure out which samples these are because they are likely to break any individual attempts to combine datasets with the smasher (inner join is going to return nothing if they are combined with any other datasets).

PR #1856, opened 2019-11-01
Filter out samples that don't have raw data from the compendia.

kurtwheeler:
## Issue Number

#1855 

Our human compendium has almost 600k gene ids. There should only be ~50k of them. This seems to be because some pre-processed samples are making it through our pipelines without being gene converted. Until we can track down every last one of these and fix them, we'll instead just want to make sure that they don't blow up the set of gene identifiers. This accomplishes that by only including in the compendium those samples which we processed from raw data.

## Types of changes

## Functional tests

N/A, the unit tests cover this a bit but it's really rather minor.

## Checklist


kurtwheeler:
This is not going to be merged unless the current human job fails. This change will mean that pre-processed data won't be included in our compendia. This is better than not having compendia, but ideally we'd include all possible samples.

cgreene:
Will try a different approach for now. Can re-open if desired.

PR #1858, opened 2019-11-01
Merges on disk so as to avoid doubling up on RAM.

kurtwheeler:
## Issue Number

#1258 

We might not need this! However if we do this should avoid making a copy of our matrix because we do the merge on disk instead of in memory!

## Methods

I couldn't get pandas to let the index column be a string, so I just kept that column in memory and added it back after writing/reading the matrix to/from disk.

## Types of changes

## Functional tests

The unit tests are the best test I have for this!

## Checklist


kurtwheeler:
This PR shouldn't be necessary. If the current human job fails, #1856 should reduce the RAM footprint of the matrix to the point where this is unnecessary. However, should we ever need it we've got it now!

cgreene:
Will try a different route for now. Can re-open if we go this route.

PR #1867, opened 2019-11-06
Add OriginalFiles endpoint to API

arielsvn:
## Issue Number

N/A

This adds several small improvements to the API:

- A new original file endpoint.
- Returns original/computed file ids on the detailed sample endpoint
- Returns original file ids on the jobs endpoints -- instead of the full original file
- Adds a new parameter to the jobs endpoint to only return the jobs that are in the nomad queue.

## Types of changes

## Functional tests

Tested locally with https://github.com/AlexsLemonade/refinebio-frontend/pull/784

## Checklist

image


davidsmejia:
This looks good but why did you remove the s3 urls from here?

https://github.com/AlexsLemonade/refinebio/pull/1867/files#diff-415cd6002ae0888c9bee613a31905f87L245-L247

arielsvn:
> This looks good but why did you remove the s3 urls from here?

I thought those were supposed to be internal. Users should be able to download computed files with `download_url`, assuming they provide a token. Do you think we should keep them public?

PR #1870, opened 2019-11-06
Remove cached organism_names from experiments and use property instead

### LINKED ISSUE #1758: Improve accuracy of sample numbers by organism in the filters
### Context

https://github.com/AlexsLemonade/refinebio/issues/1756

### Problem or idea

It looks like we have a lot more chimp samples processed than we do. This is because the chimp samples belong to experiments that also contain human samples and those experiments are just reporting how many samples they have processed, despite the fact that (at least in some cases) all of those samples are actually human.

### Solution or next step

As mentioned here: https://github.com/AlexsLemonade/refinebio/issues/1756#issuecomment-540693551, @jashapiro had the idea that we should only index an experiment for an organism once that organism has at least one processed sample.

This won't make the numbers perfectly accurate, but it at least will prevent some experiments from contributing to the counts of an organism that they have zero processed samples for.

arielsvn:
## Issue Number

close #1758 

This removes the calculated `organism_names` from the `Experiment` model, and instead uses a property that returns the organism names associated with downloadable samples. This value is recalculated for each experiment when on every reindex. However, it shouldn't be too bad since we only do it for modified experiments now.

## Types of changes

## Functional tests

Tested api locally.

## Checklist


kurtwheeler:
Oh also it looks like the tests are unhappy specifically because of the changes in this PR:
```
ERROR: test_early_tximport (data_refinery_foreman.foreman.management.commands.test_run_tximport.RunTximportTestCase)
Tests that tximport jobs are created when the experiment is past the thresholds.
----------------------------------------------------------------------
  File "/usr/local/lib/python3.5/dist-packages/django/db/backends/utils.py", line 85, in _execute
    return self.cursor.execute(sql, params)
psycopg2.ProgrammingError: column "organism_names" does not exist
LINE 1: ..._modified" FROM "experiments" WHERE ((cardinality(organism_n...
                                                             ^


The above exception was the direct cause of the following exception:

  File "/usr/lib/python3.5/unittest/mock.py", line 1157, in patched
    return func(*args, **keywargs)
  File "/home/user/data_refinery_foreman/foreman/management/commands/test_run_tximport.py", line 242, in test_early_tximport
    run_tximport_at_progress_point(complete_accessions, incomplete_accessions)
  File "/home/user/data_refinery_foreman/foreman/management/commands/test_run_tximport.py", line 174, in run_tximport_at_progress_point
    run_tximport.run_tximport()
  File "/home/user/data_refinery_foreman/foreman/management/commands/run_tximport.py", line 52, in run_tximport
    page = paginator.page()
  File "/usr/local/lib/python3.5/dist-packages/data_refinery_common/performant_pagination/pagination.py", line 166, in page
    object_list = list(qs[:self.per_page + 1])
  File "/usr/local/lib/python3.5/dist-packages/django/db/models/query.py", line 268, in __iter__
    self._fetch_all()
  File "/usr/local/lib/python3.5/dist-packages/django/db/models/query.py", line 1186, in _fetch_all
    self._result_cache = list(self._iterable_class(self))
  File "/usr/local/lib/python3.5/dist-packages/django/db/models/query.py", line 54, in __iter__
    results = compiler.execute_sql(chunked_fetch=self.chunked_fetch, chunk_size=self.chunk_size)
  File "/usr/local/lib/python3.5/dist-packages/django/db/models/sql/compiler.py", line 1065, in execute_sql
    cursor.execute(sql, params)
  File "/usr/local/lib/python3.5/dist-packages/raven/contrib/django/client.py", line 127, in execute
    return real_execute(self, sql, params)
  File "/usr/local/lib/python3.5/dist-packages/django/db/backends/utils.py", line 68, in execute
    return self._execute_with_wrappers(sql, params, many=False, executor=self._execute)
  File "/usr/local/lib/python3.5/dist-packages/django/db/backends/utils.py", line 77, in _execute_with_wrappers
    return executor(sql, params, many, context)
  File "/usr/local/lib/python3.5/dist-packages/django/db/backends/utils.py", line 85, in _execute
    return self.cursor.execute(sql, params)
  File "/usr/local/lib/python3.5/dist-packages/django/db/utils.py", line 89, in __exit__
    raise dj_exc_value.with_traceback(traceback) from exc_value
  File "/usr/local/lib/python3.5/dist-packages/django/db/backends/utils.py", line 85, in _execute
    return self.cursor.execute(sql, params)
django.db.utils.ProgrammingError: column "organism_names" does not exist
LINE 1: ..._modified" FROM "experiments" WHERE ((cardinality(organism_n...
```

arielsvn:
Addressed your comments @kurtwheeler. Test pass locally.

kurtwheeler:
It doesn't look like you addressed:

> Also it looks like the field you removed, organism_names, is still used by the DatasetDetailsExperimentSerializer, so it looks like as this is it'll break stuff.

https://github.com/AlexsLemonade/refinebio/blob/dev/api/data_refinery_api/serializers.py#L620

arielsvn:
Yes, it's addressed, it'll use the `organisms` property in `Experiment` instead of the old `organism_names`. Check the property in the fields:

https://github.com/AlexsLemonade/refinebio/blob/dev/api/data_refinery_api/serializers.py#L625-L631

PR #1876, opened 2019-11-08
Removing duplicated code in the methods to create processor jobs

arielsvn:
## Issue Number

N/A

Looks like this code is duplicated. This removes some of it, let's see if the tests complain.

## Types of changes

## Functional tests

None

## Checklist


common/data_refinery_common/job_management.py
```diff
@@ -256,13 +195,21 @@ def create_processor_job_for_original_files(original_files: List[OriginalFile],
...
+
             assoc = ProcessorJobOriginalFileAssociation()
             assoc.original_file = original_file
             assoc.processor_job = processor_job
             assoc.save()
 
         logger.debug("Queuing processor job.",
-                     processor_job=processor_job.id)
+                     processor_job=processor_job.id,
+                     downloader_job=downloader_job.id)
```
kurtwheeler: Note that the `downloader_job` parameter defaults to None, this could cause an error.

common/data_refinery_common/job_management.py
```diff
@@ -147,75 +147,15 @@ def create_downloader_job(undownloaded_files: List[OriginalFile],
...
-
-        # Fix for: https://github.com/AlexsLemonade/refinebio/issues/968
-        # Basically, we incorrectly detected technology/manufacturers
-        # for many Affymetrix samples and this is a good place to fix
-        # some of them.
-        if original_file.is_affy_data():
-            # Only Affymetrix Microarrays produce .CEL files
-            sample_object.technology = 'MICROARRAY'
-            sample_object.manufacturer = 'AFFYMETRIX'
-            sample_object.save()
```
kurtwheeler: This code block isn't present in the other function that you're calling. I think we still want to be doing this right?

arielsvn: Looks like we still have `2719` samples affected by this. If the original bug is fixed, can we update them directly and remove this code from here?

```
>>> Sample.objects.exclude(technology='MICROARRAY', manufacturer='AFFYMETRIX').filter(Q(original_files__source_filename__iendswith='.cel') | Q(original_files__source_filename__iendswith='.cel.gz')).distinct().count()
2719
```

kurtwheeler: Yeah that sounds good to me!

arielsvn: ok I ran the following code and corrected them:

```
>>> from data_refinery_common.models import Sample
>>> from django.db.models.expressions import F, Q
>>> affected_samples = Sample.objects.exclude(technology='MICROARRAY', manufacturer='AFFYMETRIX').filter(Q(original_files__source_filename__iendswith='.cel') | Q(original_files__source_filename__iendswith='.cel.gz')).distinct()
>>> affected_samples.count()
2719
>>> affected_samples.update(technology='MICROARRAY', manufacturer='AFFYMETRIX')
2719
```

Issue #1872, opened 2019-11-07
Don't remove processor jobs from db


arielsvn:
### Context

Processor Jobs remove themselves from the DB when they detect that original files are missing.

https://github.com/AlexsLemonade/refinebio/blob/5843afd/workers/data_refinery_workers/processors/utils.py#L104-L111

### Problem or idea

This makes it really hard to track down cycles, like #1818 and #1828. Because every time we lose the information in the DB.

### Solution or next step

Is there any other property we can use to mark the processor jobs? maybe `abort` or `no_retry` so that the foreman does not pick it up. @kurtwheeler

PR #1881, opened 2019-11-08
Possible solution to avoid removing jobs from db

arielsvn:
## Issue Number

close https://github.com/AlexsLemonade/refinebio/issues/1872

This sets `no_retry=True` and `success=False` for jobs that have missing original files. Not sure if this has any other implications elsewhere.

@kurtwheeler do you think this could work?

## Types of changes

## Functional tests

None

## Checklist


arielsvn:
Looks like this might cause problems with our ability to determine if an `OriginalFile` needs processing. Because we check that it doesn't have any jobs with `no_retry=True`

https://github.com/AlexsLemonade/refinebio/blob/arielsvn/1872-dont-remove-processor-jobs/common/data_refinery_common/models/models.py#L784-L787

arielsvn:
@kurtwheeler tests passed! This should be ready for review.

As we discussed, this adds a new field named `abort` to the processor jobs. I added multiple steps to the migration to avoid a timeout.

common/data_refinery_common/models/jobs.py
```diff
@@ -171,6 +171,7 @@ class Meta:
     original_files = models.ManyToManyField('OriginalFile', through='ProcessorJobOriginalFileAssociation')
     datasets = models.ManyToManyField('DataSet', through='ProcessorJobDataSetAssociation')
     no_retry = models.BooleanField(default=False)
+    abort = models.BooleanField(default=False)
```
kurtwheeler: Can you add a comment describing `abort` here?

foreman/data_refinery_foreman/surveyor/test_end_to_end.py
```diff
@@ -530,17 +530,14 @@ def test_transcriptome_redownloading(self):
             successful_processor_jobs = []
             for processor_job in processor_jobs:
                 # One of the calls to wait_for_job will fail if the
```
kurtwheeler: We're no longer expecting one of the calls to fail, I think this comment should be updated.

foreman/data_refinery_foreman/surveyor/test_end_to_end.py
```diff
@@ -611,7 +608,7 @@ def test_sra_redownloading(self):
                             break
 
             # There's a chance that the processor job with a missing
-            # file deletes itself before the last downloader job
+            # file is aborted before the last downloader job
             # completes, therefore just check that there's at least 3
             # processor jobs.
```
kurtwheeler: We no longer have to check for at least three, because there will always be 4 since the one with a missing file will not delete itself.

kurtwheeler: I mean that we can just check for exactly 4 jobs here.

workers/data_refinery_workers/processors/utils.py
```diff
@@ -401,11 +396,13 @@ def run_pipeline(start_value: Dict, pipeline: List[Callable]):
 
 class ProcessorJobError(Exception):
     """ General processor job error class. """
-    def __init__(self, failure_reason, *, success=None, no_retry=None, **context):
+    def __init__(self, failure_reason, *, success=None, no_retry=None, retried=None, abort=None, **context):
```
kurtwheeler: Wouldn't it make more sense for `abort` to default to False than None?

workers/data_refinery_workers/processors/utils.py
```diff
@@ -366,16 +364,13 @@ def run_pipeline(start_value: Dict, pipeline: List[Callable]):
...
-            logger.info('Processor Job deleted itself.', reason=e.reason, processor_job=job_id)
-            break
         except ProcessorJobError as e:
             e.update_job(job)
             logger.exception(e.failure_reason, processor_job=job.id, **e.context)
             if e.success is False:
                 # end_job will use this and set the value
                 last_result['success'] = False
-            return end_job(last_result)
+            return end_job(last_result, abort=bool(e.abort))
```
kurtwheeler: I think this cast can be removed once `e.abort` is always True or False.

common/data_refinery_common/migrations/0043_processorjob_abort.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def batch_update(queryset, batch=1000, **changes):
+    """ Update per batch """
+    is_done = False
+    current = 0
+
+    while not is_done:
+        current = current + 1
+        start = (current - 1) * batch
```
kurtwheeler: This would be a bit simpler if you incremented `current` at the end of the loop instead of the start because then you wouldn't have to decrement it for this operation.

common/data_refinery_common/migrations/0043_processorjob_abort.py
```diff
@@ -0,0 +1,60 @@
...
+def set_default_abort_value(apps, schema_editor):
+    ProcessorJob = apps.get_model('data_refinery_common', 'ProcessorJob')
+    queryset = ProcessorJob.objects.all()
+    batch_update(queryset, abort=False)
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('data_refinery_common', '0042_remove_experiment_organism_names'),
```
kurtwheeler: This will need to be updated because we've had migrations go out. Sorry about taking so long on this :(

workers/data_refinery_workers/processors/utils.py
arielsvn: I used `None` to differentiate when the user wants to explicitly set a value for the job in the database. For example: `ProcessorJobError('message', abort=False)` will set `job.abort=False`, on the other hand `ProcessorJobError('message')` would leave the `job.abort` to the default value.

cgreene:
This looks like it's ready to go with an updated migration. Do I have that right?

kurtwheeler:
@cgreene That seems correct to me :+1: :shipit:

arielsvn:
Updated the migration, I'm going to let the test run and then merge this.

Issue #1882, opened 2019-11-11
Compendium Identifiers -> Start with "approved" set


cgreene:
### Context

We currently have 10 times as many human (and maybe other organism) gene identifiers as we could plausibly include in the compendium. This is because our no-op processors seem to be occasionally leaking through data that is not identifier remapped.

### Problem or idea

We should restrict our compendium-building procedure to a set of genes that we could plausibly include. This will help to avoid blowing up our memory requirements as we're putting the compendium together.

One idea -> filter each incoming sample to only genes that are in our RNA-seq set. It seems unlikely that we would be able to have enough measurements for things that are not found in RNA-seq.

Another idea (probably more flexible/preferable) -> Since we're now doing two passes over the data, in the first pass don't create a set of all identifiers. Instead, create a dictionary. The keys would be the identifiers. The values would be number of samples in which that key was observed. We'd also store the number of samples processed.

Then, after the first pass we'd only retain those identifiers that are in >= 50% of the samples. This should leave somewhat more identifiers than we remove here:
https://github.com/AlexsLemonade/refinebio/blob/dev/workers/data_refinery_workers/processors/create_compendia.py#L230
because it both filters to a lower threshold and also because we drop certain genes from the RNA-seq step if they are very lowly expressed:
https://github.com/AlexsLemonade/refinebio/blob/dev/workers/data_refinery_workers/processors/create_compendia.py#L170

If we're going to stick with two-pass, I think the second approach is better. It's more flexible for cases where the transcriptome indices don't yet exist. For example, those currently referenced in https://github.com/AlexsLemonade/refinebio/issues/1722.

### Solution or next step

@jaclyn-taroni @kurtwheeler curious as to your thoughts.

kurtwheeler:
Ok so the only question I have about this solution is what to do about samples that have the wrong gene ids. Specifically in terms of deciding when to drop the samples entirely.

Here's the method for dropping these that I've come up with so far:
  - Instead of the dictionary having its values be the number of samples with the gene id, instead build a list of the samples that have the gene id.
  - Then once that dict is built we can count the number of samples for each gene id, and filter out the gene id if that count isn't 50% of total samples.
  - Then we can just use every sample that is still in the dict for the full matrix (with a set union across all the gene ids' samples.)


[...]

PR #1889, opened 2019-11-12
Only use gene identifiers that are present in >50% of the samples.

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/1882

This implements the second solution suggested in #1882:

> Another idea (probably more flexible/preferable) -> Since we're now doing two passes over the data, in the first pass don't create a set of all identifiers. Instead, create a dictionary. The keys would be the identifiers. The values would be number of samples in which that key was observed. We'd also store the number of samples processed.

> Then, after the first pass we'd only retain those identifiers that are in >= 50% of the samples. This should leave somewhat more identifiers than we remove here:
https://github.com/AlexsLemonade/refinebio/blob/dev/workers/data_refinery_workers/processors/create_compendia.py#L230
because it both filters to a lower threshold and also because we drop certain genes from the RNA-seq step if they are very lowly expressed:
https://github.com/AlexsLemonade/refinebio/blob/dev/workers/data_refinery_workers/processors/create_compendia.py#L170

We will have to make sure we blow away our cache so we make a new one rather than resuing the existing cache.

## Types of changes

## Functional tests

Unit tests.

### DIFF
--- a/workers/data_refinery_workers/processors/create_compendia.py
+++ b/workers/data_refinery_workers/processors/create_compendia.py
@@ -150,6 +150,10 @@ def _perform_imputation(job_context: Dict) -> Dict:
     # We potentially can have a microarray-only compendia but not a RNASeq-only compendia
     log2_rnaseq_matrix = None
     if job_context['rnaseq_matrix'] is not None:
+        # Drop any genes that are entirely NULL in the RNA-Seq matrix
+        job_context['rnaseq_matrix'] = job_context['rnaseq_matrix'].dropna(axis='columns',
+                                                                           how='all')
+
         # Calculate the sum of the lengthScaledTPM values for each row
         # (gene) of the rnaseq_matrix (rnaseq_row_sums)
         rnaseq_row_sums = np.sum(job_context['rnaseq_matrix'], axis=1)
@@ -207,8 +211,11 @@ def _perform_imputation(job_context: Dict) -> Dict:
     # Perform a full outer join of microarray_matrix and
     # log2_rnaseq_matrix; combined_matrix
     if log2_rnaseq_matrix is not None:
+        # Use how='right' because we've filtered some gene identifiers
+        # out of the the rnaseq_matrix and we want to keep those
+        # filtered out of the full matrix.
         combined_matrix = job_context.pop('microarray_matrix').merge(log2_rnaseq_matrix,
-                                                                     how='outer',
+                                                                     how='right',
                                                                      left_index=True,
                                                                      right_index=True)
     else:
--- a/workers/data_refinery_workers/processors/create_compendia.py
+++ b/workers/data_refinery_workers/processors/create_compendia.py
@@ -211,11 +211,8 @@ def _perform_imputation(job_context: Dict) -> Dict:
     # Perform a full outer join of microarray_matrix and
     # log2_rnaseq_matrix; combined_matrix
     if log2_rnaseq_matrix is not None:
-        # Use how='right' because we've filtered some gene identifiers
-        # out of the the rnaseq_matrix and we want to keep those
-        # filtered out of the full matrix.
         combined_matrix = job_context.pop('microarray_matrix').merge(log2_rnaseq_matrix,
-                                                                     how='right',
+                                                                     how='outer',
                                                                      left_index=True,
                                                                      right_index=True)
     else:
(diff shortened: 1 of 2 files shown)


workers/data_refinery_workers/processors/create_compendia.py
```diff
@@ -150,6 +150,10 @@ def _perform_imputation(job_context: Dict) -> Dict:
     # We potentially can have a microarray-only compendia but not a RNASeq-only compendia
     log2_rnaseq_matrix = None
     if job_context['rnaseq_matrix'] is not None:
+        # Drop any genes that are entirely NULL in the RNA-Seq matrix
+        job_context['rnaseq_matrix'] = job_context['rnaseq_matrix'].dropna(axis='columns',
```
cgreene: 👍 this should get us back to the prior behavior (before preallocation)

workers/data_refinery_workers/processors/create_compendia.py
```diff
@@ -207,8 +211,11 @@ def _perform_imputation(job_context: Dict) -> Dict:
     # Perform a full outer join of microarray_matrix and
     # log2_rnaseq_matrix; combined_matrix
     if log2_rnaseq_matrix is not None:
+        # Use how='right' because we've filtered some gene identifiers
```
cgreene: I'm not sure we want to keep those out of the full matrix. I suspect they will be dropped in the later steps but I don't know for sure that we could guarantee that. Is there a major downside to keeping this from `outer`? I think it should not reallocate much, since the full set is given by the left matrix.

workers/data_refinery_workers/processors/smashing_utils.py
```diff
@@ -360,7 +360,11 @@ def process_frames_for_key(key: str,
             # Count how many frames are in each tech so we can preallocate
             # the matrices in both directions.
             if not frame['unsmashable']:
-                all_gene_identifiers = all_gene_identifiers.union(frame['dataframe'].index)
+                for gene_id in frame['dataframe'].index:
+                    if gene_id in gene_identifier_counts:
```
cgreene: Don't you mean if it's not in gene_identifier_counts?

workers/data_refinery_workers/processors/smashing_utils.py
```diff
@@ -369,8 +373,16 @@ def process_frames_for_key(key: str,
                 elif frame['technology'] == 'rnaseq':
                     rnaseq_columns.extend(frame['dataframe'].columns)
 
-        all_gene_identifiers = list(all_gene_identifiers)
+        total_samples = len(microarray_columns) + len(rnaseq_columns)
+        all_gene_identifiers = []
+        for gene_id, sample_count in gene_identifier_counts.items():
+            # We only want to use gene identifiers which are present
+            # in >50% of the samples.
+            if sample_count > total_samples / 2:
```
cgreene: Do you want to convert this to the same methodology as the other filtering steps (`* 0.5`) instead of divide by 2? Should be the same, but I think in the compendia we always multiply by the decimal fraction.

workers/data_refinery_workers/processors/smashing_utils.py
```diff
@@ -369,8 +373,16 @@ def process_frames_for_key(key: str,
                 elif frame['technology'] == 'rnaseq':
                     rnaseq_columns.extend(frame['dataframe'].columns)
 
-        all_gene_identifiers = list(all_gene_identifiers)
+        total_samples = len(microarray_columns) + len(rnaseq_columns)
+        all_gene_identifiers = []
+        for gene_id, sample_count in gene_identifier_counts.items():
+            # We only want to use gene identifiers which are present
+            # in >50% of the samples.
```
cgreene: It would probably be good to note that we're doing this because a large number of gene identifiers present in only a modest number of experiments have leaked through. We wouldn't necessarily _want_ to do this if we'd mapped all the data to ENSEMBL identifiers successfully.

workers/data_refinery_workers/processors/create_compendia.py
kurtwheeler: I would think that would be fine. I just made an assumption that a gene that had no representation in the RNA-Seq data would be one we we'd wanna keep out of the final matrix. I'll change this back!

cgreene: I'd guess it'll get filtered later and this is almost certainly a safe change, but if we don't need it - my preference is to avoid it until we can examine whether or not it makes a difference. 👍

workers/data_refinery_workers/processors/smashing_utils.py
kurtwheeler: Haha I just had a test complain about this.

cgreene: Glad I've still got some eyes 👀

kurtwheeler:
@cgreene I've made the requested changes, and marked each one as resolved, but github won't let me merge until you come back or I dismiss your review :(

PR #1891, opened 2019-11-12
Disable the QN test until we can figure out why it OOM-kills the human compendia job.

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio/pull/1861

We don't know why #1861 failed. Rather than iterating on that problem in prod, this PR disables the QN test so we can just get it done. I'll run the QN tests on the compendia manually, and in the process of doing so I should also be able to figure out what's breaking it and come up with a solution that doesn't take so much RAM.


davidsmejia:
I support this assuming we can remove `is_public` until we QC this ourselves.
👍

PR #1897, opened 2019-11-13
Additional caching for quantpendias

arielsvn:
## Issue Number

#1823 

Cache the major steps for the quantpendia: downloading files, creating metadata, making the archive.

Wrote this with a decorator, do you think it's confusing like this @kurtwheeler?

## Types of changes

## Functional tests

Tested locally

## Checklist

### DIFF
--- a/workers/data_refinery_workers/processors/create_quantpendia.py
+++ b/workers/data_refinery_workers/processors/create_quantpendia.py
@@ -36,7 +36,7 @@ def create_quantpendia(job_id: int) -> None:
                                       utils.end_job])
     return job_context
 
-
+@utils.cache_keys('time_start', 'num_samples', 'time_end', 'formatted_command', work_dir_key='job_dir')
 def download_files(job_context: Dict) -> Dict:
     job_context['time_start'] = timezone.now()
 
--- a/workers/data_refinery_workers/processors/create_quantpendia.py
+++ b/workers/data_refinery_workers/processors/create_quantpendia.py
@@ -67,7 +67,7 @@ def download_files(job_context: Dict) -> Dict:
     return job_context
 
 
-@utils.cache_keys(work_dir_key='job_dir')
+@utils.cache_keys('metadata', work_dir_key='job_dir')
 def add_metadata(job_context: Dict) -> Dict:
     logger.debug("Writing metadata for quantpendia.",
             job_id=job_context['job_id'],
--- a/workers/data_refinery_workers/processors/create_quantpendia.py
+++ b/workers/data_refinery_workers/processors/create_quantpendia.py
@@ -31,6 +31,8 @@ def create_quantpendia(job_id: int) -> None:
                                      [utils.start_job,
                                       make_dirs,
                                       download_files,
+                                      add_metadata,
+                                      make_archive,
                                       create_result_objects,
                                       remove_job_dir,
                                       utils.end_job])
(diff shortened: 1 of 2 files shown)


workers/data_refinery_workers/processors/utils.py
```diff
@@ -668,3 +670,48 @@ def handle_processor_exception(job_context, processor_key, ex):
...
+        # and the cached keys
+        cache_name = hashlib.md5(key_names.encode()).hexdigest()
+
+        def pipeline(job_context):
+            cache_path = os.path.join(job_context[work_dir_key], cache_name)
+            if os.path.exists(cache_path):
+                # cached values exist, load keys from cacke
+                try:
+                    values = pickle.load(open(cache_path, "rb" ))
+                    return {**job_context, **values}
```
kurtwheeler: Any reason to not just [update](https://docs.python.org/3/library/stdtypes.html#dict.update) `job_context` and return it?

workers/data_refinery_workers/processors/utils.py
```diff
@@ -668,3 +670,48 @@ def handle_processor_exception(job_context, processor_key, ex):
...
     return job_context
+
+
+def cache_keys(*keys, work_dir_key='work_dir'):
+    """ Decorator to be applied to a pipeline function.
+    Returns a new function that calls the original one and caches the given
+    keys into the `work_dir`. On the next call it will load those keys (if they
+    exist) and add them to the job_context instead of executing the function. """
+    def inner(func):
+        key_names = ''.join(list(keys) + [func.__name__])
```
kurtwheeler: Hmm, just for readability if we ever want to inspect these caches or something.... 
```suggestion
        key_names = '__'.join(list(keys) + [func.__name__])
```

workers/data_refinery_workers/processors/utils.py
arielsvn: Not really, just an old habit of not modifying the parameters. Do you think this has any disadvantages?

workers/data_refinery_workers/processors/utils.py
arielsvn: hmm I wonder if the filename of the cache should be this, instead of the `md5` hash, so that they are readable. This might be better :)

workers/data_refinery_workers/processors/create_quantpendia.py
```diff
@@ -65,30 +67,42 @@ def download_files(job_context: Dict) -> Dict:
...
-    _add_metadata(job_context)
+            job_id=job_context['job_id'],
+            **get_process_stats())
+    smashing_utils.write_non_data_files(job_context)
+    shutil.copy("/home/user/README_QUANT.md", job_context["output_dir"] + "/README.md")
+    return job_context
+
+
+@utils.cache_keys('archive_path', work_dir_key='job_dir')
+def make_archive(job_context: Dict):
```
kurtwheeler: Any reason `make_archive` and `add_metadata` no longer are prefixed with a `_` denoting them as private functions of this namespace?

workers/data_refinery_workers/processors/utils.py
kurtwheeler: > just an old habit of not modifying the parameters

I do generally love this. `job_context` gets mutated directly all over the place though. We can leave it like this, works for me.

workers/data_refinery_workers/processors/create_quantpendia.py
arielsvn: because I made them steps in the pipeline, instead of calling them from `create_result_objects`.

kurtwheeler: So? They still aren't being called from outside this namespace.

workers/data_refinery_workers/processors/utils.py
kurtwheeler: Oh yeah I missed that you were doing a md5 hash, yeah I like this!

workers/data_refinery_workers/processors/create_quantpendia.py
arielsvn: hmm, only the main function `create_quantpendia` is being called from outside the module. Should I prefix all functions except that one with `_`?

kurtwheeler: Yup!!

PR #1909, opened 2019-11-14
[WIP] Upload/Download Dev Data Scripts

davidsmejia:
## Issue Number

WIP/NA

This is to allow us to stash processed data and a populated database. Currently trying to create the database for the initial upload. I want to test this script before merging but I wanted to share the direction I was headed in.

New community and in-house developers that want to work on a specific task or work on the front end but not against the production database can run `./scripts/download-dev-data.sh` and work against a populated local DB which is 😎 

If we want to work on a specific task like a processor job that isn't working in production and run it locally we can run the downloader job locally and then run `./scripts/upload-dev-data.sh` and push the updated state to s3 which will then be available to anyone who wants to run that processor job locally. 

Note: `upload-dev-data.sh` requires that you are logged into `aws-cli` to handle the upload the s3

Also will need to add a filter to single out only processed data and not include raw data in the tar file.

### DIFF
--- a/scripts/download-dev-data.sh
+++ b/scripts/download-dev-data.sh
@@ -44,11 +44,6 @@ while getopts ":h" opt; do
     esac
 done
 
-if [ -z "$SYSTEM_VERSION" ]; then
-    echo 'Error: must specify the version repo with -v'
-    exit 1
-fi
-
 VOLUME_FOLDER=./volume
 POSTGRES_VOLUME_FOLDER=./volumes_postgres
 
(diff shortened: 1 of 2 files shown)


arielsvn:
@davidsmejia I love this idea! Is there something missing from this PR to get it merged?

kurtwheeler:
I'm going to close this until we're ready to resume it.

(I think there was quite a bit more work to be done, but we should think about prioritizing it soon.)

PR #1910, opened 2019-11-15
Add endpoints to return the top 10 failure reasons

### LINKED ISSUE #1556: Add the biggest `failure_reason` to the dashboard?
### Context

I have been debugging some sample processing errors.

### Problem or idea

I think it would be useful to have a list of the top failure reasons and the number of samples that they affect. I'm not sure if the dashboard is a good place for this.

```
  ccc   |                                 job_failure_reason                                  
--------+-------------------------------------------------------------------------------------
 646262 | 
  47680 | Salmon timed out because it failed to complete within 3 hours.
  40654 | Encountered error in R code while running illumina.R pipeline during processing of 
  29566 | Unhandled exception caught while running processor function _prepare_files in pipel
  26986 | Shell call to salmon failed because: ### salmon (mapping-based) v0.13.1            +
        | ### [ progr
  10347 | Missing file for processor job but unable to recreate downloader jobs!
   8912 | Encountered error in R code while running AFFY_TO_PCL pipeline during processing of
   3682 | The tx index for this organism is broken right now.
   3010 | list index out of range
   2629 | ProcessorJob has already completed with a fail - why are we here again? Bad Nomad!
   2556 | Unhandled exception caught while running processor function start_job in pipeline: 
   2440 | Missing transcriptome index. (TRANSCRIPTOME_LONG)
   1919 | Unhandled exception caught while running processor function _convert_genes in pipel
   1886 | Missing transcriptome index. (TRANSCRIPTOME_SHORT)
   1699 | Failed to download or extract transcriptome index for organism ARABIDOPSIS_THALIANA
   1424 | Told to convert AFFY genes without an internal_accession - how did this happen?
   1200 | Unhandled exception caught while running processor function _detect_platform in pip
   1138 | Unhandled exception caught while running processor function _run_scan_upc in pipeli
    868 | Encountered error in R code while running gene_convert_illumina.R pipeline during p
    791 | Status code -9 from gene_convert.R: b"\nAttaching package: 'dplyr'\n\nThe following
(20 rows)
```

That's the top 20 failure reasons with a total of `835,649` potential samples. It's also very strange that there are so many samples with no `failure_reason` on their last processor job.

<details><summary>query</summary>

```sql
select count(*) as ccc, (
    select left(processor_jobs.failure_reason, 83)
    from original_file_sample_associations, processorjob_originalfile_associations, processor_jobs
    where samples.id=original_file_sample_associations.sample_id
        and original_file_sample_associations.original_file_id=processorjob_originalfile_associations.original_file_id
        and processorjob_originalfile_associations.processor_job_id=processor_jobs.id
    order by processor_jobs.start_time desc nulls last
    limit 1
  ) as job_failure_reason
from samples
where samples.is_processed='f'
group by job_failure_reason
order by ccc desc
limit 20;
```

</details>

I have been working on both:
```
   8912 | Encountered error in R code while running AFFY_TO_PCL pipeline during processing of
   2629 | ProcessorJob has already completed with a fail - why are we here again? Bad Nomad!
```

Maybe we can re-queue some of the samples with these errors? With https://github.com/AlexsLemonade/refinebio/pull/1541

### Solution or next step

Should we add this to the dashboard?

arielsvn:
## Issue Number

close #1556 

Adds two new endpoints: `/stats/failed-downloader` and `/stats/failed-processor`, which return the top 10 failure reasons for both downloaders and processors jobs respectively. They accept a parameter `range` to only consider the jobs that have failed in the last day/week/year.

Suggestions for better URLs welcome.

## Types of changes

## Functional tests

Tested the queries for the last year in prod to ensure they don't time out.

## Checklist


api/data_refinery_api/views.py
```diff
@@ -916,6 +916,64 @@ def get_queryset(self):
...
+            .annotate(reason=Left('failure_reason', 80))\
+            .values('reason')\
+            .annotate(
+                job_count=Count('reason'),
+                sample_count=Count('original_files__samples', distinct=True, filter=Q(original_files__samples__is_processed=False))
+            )\
+            .order_by('-job_count')
+
+        # return the top 10 failure reasons
+        return Response(list(jobs[:10]))
```
davidsmejia: Maybe we should consider paginating this and have defualt page set to `10`?

api/data_refinery_api/urls.py
```diff
@@ -102,6 +104,8 @@ class AccessUser:
 
         # Dashboard Driver
         url(r'^stats/$', Stats.as_view(), name='stats'),
+        url(r'^stats/failed-downloader$', FailedDownloaderJobStats.as_view(), name='stats_failed_downloader'),
+        url(r'^stats/failed-processor$', FailedProcessorJobStats.as_view(), name='stats_failed_processor'),
```
davidsmejia: What about adding this to `stats/failures/[type]`

arielsvn: I like this! If we were to go this route, do we need to have something on `/failures`?

arielsvn:
Addressed the comments @davidsmejia, let me know what do you think.

PR #1911, opened 2019-11-15
Quantile normalize the compendia matrix in chunks to avoid OOM.

kurtwheeler:
## Issue Number

#1258 

We got OOM-killed in the `quantile_normalize` function. Not entirely sure why, but this should make things a lot more performant both by performing QN in chunks and also by making less copies of the data as we do so.

Specifically, after we QN each chunk we reassign the normalized values right back onto the original matrix so we don't have to make additional copies.

## Methods

It was originally @jaclyn-taroni's suggestion to do this in chunks so I think she'll be a fan of the changes.

## Types of changes

- Optimization

## Functional tests

The compendia unit tests pass!

### DIFF
--- a/workers/data_refinery_workers/processors/smashing_utils.py
+++ b/workers/data_refinery_workers/processors/smashing_utils.py
@@ -458,7 +458,7 @@ def process_frames_for_key(key: str,
 
 # Modified from: http://yaoyao.codes/pandas/2018/01/23/pandas-split-a-dataframe-into-chunks
 def _index_marks(num_columns, chunk_size):
-    return range(1 * chunk_size, math.ceil(num_columns / chunk_size) * chunk_size, chunk_size)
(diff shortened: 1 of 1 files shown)


workers/data_refinery_workers/processors/smashing_utils.py
```diff
@@ -455,6 +456,16 @@ def process_frames_for_key(key: str,
     return job_context
 
 
+# Modified from: http://yaoyao.codes/pandas/2018/01/23/pandas-split-a-dataframe-into-chunks
+def _index_marks(num_columns, chunk_size):
+    return range(1 * chunk_size, math.ceil(num_columns / chunk_size) * chunk_size, chunk_size)
```
arielsvn: `1 * chunk_size`?

kurtwheeler: woah good call, da fuck?!?

Issue #1849, opened 2019-10-31
Strategy for updating the data models to support multiple organism compendia


arielsvn:
### Context

https://github.com/AlexsLemonade/refinebio/issues/1736#issuecomment-539642057

### Problem or idea

We need to represent compendia with multiple organisms in the database. https://github.com/AlexsLemonade/refinebio/pull/1753 Takes care of modifying the command that triggers the compendia to include multiple organisms.

### Solution or next step

We'll need several changes to the database models:

1. Create a new model `CompendiaResult` with the following fields
  - `result` is a  `ComputationalResult`
  - `primary_organism` to `Organism`
  - `organisms` is a list with all the organisms associated with the compendia and will always include the primary organism
  - `quant_sf_only`
  - `svd_algorithm`
  - `compendia_version`

2. Add a migration to populate the new `CompendiaResult` from the existing data in the `ComputedFile`

3. Use the new models on the `create_compendia` processor job https://github.com/AlexsLemonade/refinebio/blob/0847a876c2894167d1e37fd997e6e62ad40c3f85/workers/data_refinery_workers/processors/create_compendia.py#L139-L182

4. Drop the columns related to the compendia from `ComputedFile`

5. Modify the `/v1/compendia` endpoint to return the compendia. With a serializer that includes a download URL if there's an API token.

/cc @cgreene

cgreene:
👍 Exciting!

davidsmejia:
I'm currently working on API endpoints for this and I think we might have overlooked removing the is_compendia attribute from `ComputedFile`. I'm not sure how we can single out the computed file from the other two that are generated. @arielsvn @kurtwheeler 

<img width="642" alt="Screen Shot 2019-11-07 at 2 44 24 PM" src="https://user-images.githubusercontent.com/1075609/68421975-4b067980-016d-11ea-80a2-c9d4e9e73e5c.png">

davidsmejia:
I think we should add to `CompendiaResult` a `compenida_computed_file` association or something along those lines. Thoughts?

davidsmejia:
Alternatively we could just get the `.zip` file 🤷‍♂

kurtwheeler:
Hmm yeah, that does seem to be a problem. I think we forgot that we had multiple ComputedFiles per compendia ComputationalResult while we were planning.

[...]

PR #1915, opened 2019-11-18
Multiple Organisms Compendia w/ CompendiumResult

davidsmejia:
## Issue Number

#1849 

Dependent on #1753

This PR allows us to handle passing multiple organisms to the `create_compendia` command. 

It implements most of what was discussed in #1849 but fixes the name `compendia_result` to the grammatically correct `compendium_result`.

 - Creates a `CompendiumResult` model
 - Adds migrations to move computed files that have `is_compendia=True`
 - Adds Serializers for `CompendiaResults` with and without URLS
 - Updates `/compendia` using this model and adds `/compendia/id` for a single compendium
 - Adds `/compendia?compendium_version=latest` to get the latest version for each kind of compendium
 - Stops saving multiple computed files and only adds the 
 - updates tests to check for organism consistency and correct number of computed files

NOTE:
*It also stops deletion of processed files after processing*
which was helpful for local testing

## Methods

Prepare files in `create_compendia` and `create_result_objects` science is untouched.. this is more of an api change.. we collect all of the samples until the primary_organism key instead of actually aggregating it by species and we used to.

## Types of changes

- New feature
- Breaking change

## Functional tests

`./workers/run_tests.sh -t compendia`

Running the api locally works. This should and will be tested on staging before deploying.

## Checklist

- [ x] Lint and unit tests pass locally with my changes
- [ x] I have added tests that prove my fix is effective or that my feature works
- [ x] I have added necessary documentation (if appropriate)
- [ x] Any dependent changes have been merged and published in downstream modules


foreman/data_refinery_foreman/foreman/main.py
```diff
@@ -1381,14 +1381,15 @@ def monitor_jobs():
...
 
-        if timezone.now() - last_janitorial_time > JANITOR_DISPATCH_TIME:
-            send_janitor_jobs()
-            cleanup_the_queue()
-            last_janitorial_time = timezone.now()
-
-        if timezone.now() - last_dbclean_time > DBCLEAN_TIME:
-            clean_database()
-            last_dbclean_time = timezone.now()
+        if not settings.RUNNING_IN_CLOUD:
```
kurtwheeler: We want to run janitor and dbclean jobs in the cloud and not locally. 

```suggestion
        if settings.RUNNING_IN_CLOUD:
```

api/data_refinery_api/serializers.py
```diff
@@ -790,49 +791,50 @@ class Meta:
...
 
-class CompendiaSerializer(serializers.ModelSerializer):
-    organism_name = serializers.CharField(source='compendia_organism.name', read_only=True)
+    def to_representation(self, value):
+        return value.name
 
     class Meta:
-        model = ComputedFile
+        model = Organism
+        fields = '__all__'
```
arielsvn: do we need anything besides `name`?

common/data_refinery_common/models/models.py
```diff
@@ -598,6 +598,61 @@ def save(self, *args, **kwargs):
...
+        choices=SVD_ALGORITHM_CHOICES,
+        default="NONE",
+        help_text='The SVD algorithm that was used to impute the compendium result.'
+    )
+
+    # Common Properties
+    is_public = models.BooleanField(default=True)
+
+    #helper
+    def get_computed_file():
```
arielsvn: `self` argument missing

foreman/data_refinery_foreman/foreman/main.py
```diff
@@ -1381,14 +1381,15 @@ def monitor_jobs():
...
 
-        if timezone.now() - last_janitorial_time > JANITOR_DISPATCH_TIME:
-            send_janitor_jobs()
-            cleanup_the_queue()
-            last_janitorial_time = timezone.now()
-
-        if timezone.now() - last_dbclean_time > DBCLEAN_TIME:
-            clean_database()
-            last_dbclean_time = timezone.now()
+        if settings.RUNNING_IN_CLOUD:
```
arielsvn: Why do we need this here?

workers/data_refinery_workers/processors/create_compendia.py
```diff
@@ -49,6 +53,12 @@ def log_state(message, job_id, start_time=False):
 def _prepare_input(job_context: Dict) -> Dict:
     start_time = log_state("prepare input", job_context["job"].id)
 
+    job_context["primary_organism"] = max(job_context["samples"],
+                                          key=lambda organism:len(job_context["samples"][organism]))
+    job_context["all_organisms"] = job_context["samples"].keys()
+    all_samples = itertools.chain(*job_context["samples"].values())
```
arielsvn: should we convert this to a list? I think it returns an iterator

api/data_refinery_api/serializers.py
davidsmejia: removed and used `serilizers.StringRelatedField`

common/data_refinery_common/models/models.py
davidsmejia: good catch!

foreman/data_refinery_foreman/foreman/main.py
davidsmejia: This is to prevent local jobs from being cleaned up which I had to do while testing. It could be its own PR but I kept it since it was loosely related.

workers/data_refinery_workers/processors/create_compendia.py
davidsmejia: I have tested this and it works. So it's a matter of preference

workers/data_refinery_workers/processors/test_compendia.py
```diff
@@ -272,5 +272,12 @@ def test_create_compendia_danio(self):
         for file in final_context['computed_files']:
             self.assertTrue(os.path.exists(file.absolute_file_path))
 
+        # test compendia_result
+        self.assertEqual(final_context['compendia_result'].svd_algorithm, 'ARPACK')
+        self.assertEqual(final_context['compendia_result'].primary_organism.name,
```
arielsvn: Can we assert against the actual organism name here? is it `DANIO_RERIO`?

api/data_refinery_api/views.py
```diff
@@ -1235,26 +1236,55 @@ def get_object(self):
...
 
+    def get_serializer_class(self):
         try:
+            token_id = self.request.META.get('HTTP_API_KEY', None)
             token = APIToken.objects.get(id=token_id, is_activated=True)
-            serializer = CompendiaWithUrlSerializer(computed_files, many=True)
+            serializer_class = CompendiumResultWithUrlSerializer
         except Exception: # General APIToken.DoesNotExist or django.core.exceptions.ValidationError
-            serializer = CompendiaSerializer(computed_files, many=True)
+            serializer_class = CompendiumResultSerializer
```
arielsvn: I think we need to return `serializer_class`

workers/data_refinery_workers/processors/create_compendia.py
davidsmejia: evaluating it now because we check it at annotations step

api/data_refinery_api/views.py
davidsmejia: 👍

Issue #1927, opened 2019-11-20
Comendium Results Svd Algorithm had a typo 


davidsmejia:
### Context

svd_algorithm was not correctly set on compendium_result records

### Problem or idea

We need to write a migration that copies the `computed_file.svd_agorithm` to `compendium_result.svd_algorithm` before we remove those columns. 

### Solution or next step

Make a PR that batch updates the ~250 compendium results.

cgreene:
Fixed by #1928

PR #1928, opened 2019-11-20
add migration to correct svd attributes

davidsmejia:
## Issue Number

#1927 

This adds a migration to iterate over the `computed_files` and reassign the `svd_algorithm`

### DIFF
--- a/common/data_refinery_common/migrations/0045_svd_algorithm.py
+++ b/common/data_refinery_common/migrations/0045_svd_algorithm.py
@@ -1,0 +1,28 @@
+from django.db import migrations, models
+
+def svd_algorithm_fix(apps, schema_editor):
+    """ Fetch existing compendia and create compendium results"""
+    CompendiumResult = apps.get_model('data_refinery_common', 'CompendiumResult')
+    ComputedFile = apps.get_model('data_refinery_common', 'ComputedFile')
+
+    compendium_computed_files = ComputedFile.objects.filter(is_compendia=True,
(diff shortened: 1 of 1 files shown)


common/data_refinery_common/migrations/0045_svd_algorithm.py
```diff
@@ -0,0 +1,28 @@
+from django.db import migrations, models
+
+def svd_algorithm_fix(apps, schema_editor):
+    """ Fetch existing compendia and create compendium results"""
+    CompendiumResult = apps.get_model('data_refinery_common', 'CompendiumResult')
+    ComputedFile = apps.get_model('data_refinery_common', 'ComputedFile')
+
+    compendium_computed_files = ComputedFile.objects.filter(is_compendia=True,
+                                                            result__compendium_result__isnull=False)\
```
kurtwheeler: I think this query will be fine, but now that I've had a bit of time to think about it I don't think you need this null check necessarily. The reason I think this is that you're filtering on `is_compendia=True`, so you're already only looking at computed files that should have one and only one chain of ComputedFile->ComputationalResult->CompendiaResult, so just doing a prefetch_related should pull all of that in a single query.

However I don't think these extra filter will really hurt anything, so you can just leave it as it is if you want.

davidsmejia: Ok, I think we can keep it because the query looks like it's performing the Inner Join with it there but won't if it's not there.

Issue #1838, opened 2019-10-28
Record which samples end up making it into a compendia


kurtwheeler:
### Context

As we build compendia we drop some samples that aren't suitable for various reasons. However the only record in our database that tracks what samples are in it is the dataset object which has every sample including those that get dropped.

### Problem or idea

We should track this... _somehow_

### Solution or next step

We should already have this information in the compendia's data because during the first pass over the samples we keep track of the gene ids, the microarray samples, and the rnaseq samples. It's just a matter of actually getting this into the database.

cgreene:
If we want this to be for users of the compendium but not the front end we could include one or more files that list the samples dropped and the stage at which they were dropped.

kurtwheeler:
> If we want this to be for users of the compendium but not the front end we could include one or more files that list the samples dropped and the stage at which they were dropped.

Ooooh I like that! We should do that! That'll require keeping track of what samples we drop at each step that can drop samples.

But also, we should look at the matrix we create rather than the dataset in places like 
https://github.com/AlexsLemonade/refinebio/blob/dev/workers/data_refinery_workers/processors/smashing_utils.py#L884-L899 and https://github.com/AlexsLemonade/refinebio/blob/dev/workers/data_refinery_workers/processors/smashing_utils.py#L661-L666

arielsvn:
I went over the code and found 4 places where we drop samples, @kurtwheeler let me know if I'm missing any:

- on `_prepare_input` we drop any samples without a smashable file or if we have an error downloading the file from S3

- on `_prepare_frames` we do two passes over all `computed_files, samples` and drop all the samples where we can't load the data from the computed file.

- on `_perform_imputation` we drop all the samples "with <50% present values"

Do we want to include the samples that are dropped on all of these steps? The ones that are dropped in the imputation step seem more important.

Also, since we have multiple reasons why samples can be dropped, how should we write this information into the metadata. Instead of using separate files I'm thinking we could also add an additional property to the metadata on the sample details.

arielsvn:
I discussed this with @kurtwheeler and we came up with several options:

1. Only include the samples that made into the compendia in `aggregated_metadata.json`, and then have a different file (eg `missing_samples_metadata.json`) with the samples that didn't make it and the reason why. Something like:
    ```
    {
    "GSM1232": {
        "reason": "We failed while downloading the computed file",
    },
    "sample_accession_code": {
        "reason": "Sample had <50% present values"
    },
    [...]
    }
[...]

PR #1941, opened 2019-11-22
Arielsvn/1838 record samples in compendia

arielsvn:
## Issue Number

https://github.com/alexslemonade/refinebio/issues/1838

Started reading into the smasher code while trying to understand it, and made a few refactors. Mostly split the qn normalization functions into smaller ones.

## Types of changes

## Functional tests

Tests pass locally.

## Checklist

### DIFF
--- a/workers/data_refinery_workers/processors/smasher.py
+++ b/workers/data_refinery_workers/processors/smasher.py
@@ -538,20 +538,19 @@ def _update_result_objects(job_context: Dict) -> Dict:
 
     return job_context
 
+
 def smash(job_id: int, upload=True) -> None:
     """ Main Smasher interface """
     pipeline = Pipeline(name=PipelineEnum.SMASHER.value)
-    job_context = utils.run_pipeline({
-                                        "job_id": job_id,
-                                        "upload": upload,
-                                        "pipeline": pipeline
-                                     },
+    job_context = utils.run_pipeline({"job_id": job_id,
+                                      "upload": upload,
+                                      "pipeline": pipeline},
                                      [utils.start_job,
-                                        smashing_utils.prepare_files,
-                                        _smash_all,
-                                        _upload,
-                                        _update_result_objects,
-                                        utils.end_job])
+                                      smashing_utils.prepare_files,
+                                      _smash_all,
+                                      _upload,
+                                      _update_result_objects,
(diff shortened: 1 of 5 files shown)


workers/data_refinery_workers/processors/smasher.py
```diff
@@ -541,16 +540,18 @@ def _update_result_objects(job_context: Dict) -> Dict:
...
+                                        "job_id": job_id,
+                                        "upload": upload,
+                                        "pipeline": pipeline
+                                     },
+                                     [utils.start_job,
+                                        smashing_utils.prepare_files,
+                                        _smash_all,
+                                        _upload,
+                                        _update_result_objects,
+                                        utils.end_job])
```
kurtwheeler: I don't think this is an improvement to this formatting, but that's just like, my opinion.

kurtwheeler: I do think it was messed up before though, it should have been lined up better.

arielsvn: yes, not sure how to format this part. I'll note that I removed `_notify` from the pipelines methods and added it afterward so that it always gets executed. And we can throw failure exceptions while processing.

arielsvn: fixed it!

workers/data_refinery_workers/processors/smashing_utils.py
```diff
@@ -470,166 +451,184 @@ def _split_dataframe_columns(dataframe, chunk_size):
...
-                                " either bad smash or single saple smash."),
-                               dataset_id=job_context['dataset'].id)
-
-        # Remove un-quantiled normalized matrix from job_context
-        # because we no longer need it.
-        job_context.pop('merged_no_qn')
-
-        # And add the quantile normalized matrix to job_context.
-        job_context['merged_qn'] = new_merged
+        logger.warning("Not enough columns to perform KS test - either bad smash or single saple smash.",
```
kurtwheeler: "saple"

arielsvn:
```suggestion
        logger.warning("Not enough columns to perform KS test - either bad smash or single sample smash.",
```

PR #1945, opened 2019-11-25
Don't use samples with NULL values in QN targets.

kurtwheeler:
## Issue Number

Our human QN target is mostly NULL values. This makes a lot of things bad.

This will prevent QN targets from being made like this in the future, because any samples that have any NULL values will not be used for the QN target.

## Methods

These methods were discussed in advance with @jaclyn-taroni and @cgreene 

## Types of changes

## Functional tests

We probably should get some better unit tests for this, but for now let's get this out so we can fix this big bug.

## Checklist


kurtwheeler:
I will say that I downloaded a file I knew to have ~21k null genes, loaded it into pandas with `_load_and_sanitize_file()`, and then ran the NA-counting code on it to make sure worked. It gave the exactly number of nulls.

Issue #599, opened 2018-09-09
Use the zebrafish crunch as an opportunity to evaluate quantile normalization


jaclyn-taroni:
### Context

In #519, we introduced quantile normalization. That PR was developed using test data and we have not used it on an organism-wide collection of samples.

### Problem or idea

We should use the zebrafish crunch as a opportunity to evaluate QN.

### Solution or next step

- [x] How big is the Affymetrix Zebrafish Genome Chip compendium, i.e. how much memory does it require to calculate the reference? How much memory would it take for the organism with the largest number of samples (human)?
- [x] Are the reference values as expected?
- [ ] What do platform (e.g., specific _chip_) effects look like post-QN? If we're able to process sequencing data, let's examine cross-technology.

### New Issue Checklist

- [x] The title is short and descriptive.
- [x] You have explained the context that led you to write this issue.
- [x] You have reported a problem or idea.
- [x] You have proposed a solution or next step.

jaclyn-taroni:
@Miserlou sent me two files:
* `ALL.tsv` - the gene expression matrix output of smashing all the Affymetrix zebrafish samples we've processed 
* `target.tsv` - the target distribution calculated using the refine.bio QN processor 

I verified that the target vector was as expected by doing the following:
```R
# distribution calculated from refine.bio
target_df <- readr::read_tsv("target.tsv", col_names = FALSE)
rb_target_vector <- as.vector(as.matrix(target_df))

# "smashed" gene expression data used to determine target distribution
exprs_df <- readr::read_tsv("ALL.tsv")
# convert to matrix for use with preprocessCore, no gene ids
exprs_mat <- as.matrix(exprs_df[, 2:ncol(exprs_df)])
# calculate target reference using this matrix
qn_target <- preprocessCore::normalize.quantiles.determine.target(exprs_mat)

# same values?
all.equal(qn_target, rb_target_vector)
```
And got:
```
[1] TRUE
```

Some thoughts on QA at smash time: we could use a [Kolmogorov-Smirnov test](https://en.wikipedia.org/wiki/Kolmogorov%E2%80%93Smirnov_test) ([`ks.test`](https://stat.ethz.ch/R-manual/R-devel/library/stats/html/ks.test.html) in R )to make sure samples are drawn from the same distribution _after_ quantile normalization.

[...]

PR #1955, opened 2019-11-26
Remove unused and broken _verify_result function from qn_reference.py

kurtwheeler:
## Issue Number

#1954 

While working on #1954 I cleaned up the qn_reference.py file, and noticed that there was an undefined variable name in `_verify_result`. Looking into it more, I realized there was no data object in that job that could be used as `smashed_frame`. This just looks super invalid. The comment it got that code from (https://github.com/AlexsLemonade/refinebio/issues/599#issuecomment-422132009), used a microarray zebrafish compendia as the input. I have no idea what this was supposed to be comparing against.

## Methods

@jaclyn-taroni this is a code cleanup PR, so it's not actually changing what we're doing. However, should we be doing any kind of verification of QN targets? @davidsmejia had the idea of comparing against the previous QN target that was generated for that organism, but it's not clear how stringent we should be or what to do in the case where the QN target is supposed to be changing a lot (like for example, going from ~9k genes to ~20k genes).

## Types of changes

- Cleanup

## Functional tests

This code isn't even called, unit tests should be sufficient.

## Checklist


jaclyn-taroni:
Oooo this is a good question. I would say that I would be surprised, at least for popular organisms, if the largest platform changed. However, if/when we update the Brainarray version, I expect that we might see a change in the _number of genes_ by a few percent which makes things challenging.

jaclyn-taroni:
Can we create a ticket for the verification stuff? I don't think we'll arrive at a solution before the holiday and I don't want it to get lost.

cgreene:
I updated merged (one line got commented in a block that had fixed indentation). Let's see if test pass now.

cgreene:
Tests passed. I think this is good to go out :+1:

Issue #1962, opened 2019-11-28
Compendium Missing Generated File


davidsmejia:
### Context

`RATTUS_NORVEGICUS/rattus_norvegicus.tsv` is missing.

<img width="832" alt="Screen Shot 2019-11-27 at 10 29 50 PM" src="https://user-images.githubusercontent.com/1075609/69775087-ae306e00-1165-11ea-9a64-1035c56ea079.png">

### Problem or idea

This should be included. I looked it up, I removed this when removing the s3 copy to remove the extra computed files.
This means that any compendia generated in the past 9 days are not available as well.

### Solution or next step

Deploy fix for compendia output file.
I will add a list of compendia here that need to be regenerated.
Regenerate those compendia.

cgreene:
Closed by #1963

PR #1963, opened 2019-11-28
fix compendia output file

davidsmejia:
## Issue Number

#1962 

This adds back the 2 lines that send the csv to the output directory before zipping the `output_dir`

## Methods

```create_compendia.py```

## Types of changes

## Functional tests

None.

## Checklist

## Screenshots


davidsmejia:
Ok it actually looks like we have only written RAT since this change is `RATTUS_NORVEGICUS`

So it will be rat, mouse, human that will need to be re-run. I am also finding out all of the archived files that are missing the tsv so we can set `is_public to=False`

cgreene:
btw test failing is:
> requests.exceptions.RetryError: HTTPSConnectionPool(host='rest.ensembl.org', port=443): Max retries exceeded with url: /info/eg_version?content-type=application/json (Caused by ResponseError('too many 500 error responses',))

Looks upstream at ensembl

PR #1969, opened 2019-12-03
Keep track of which samples are filtered in the compendia

### LINKED ISSUE #1838: Record which samples end up making it into a compendia

arielsvn:
## Issue Number

close #1838 

This adds `job_context['filtered_samples']` to keep track of the samples that are dropped at each step during the smashing pipeline. Then creates the file `filtered_samples_metadata.json` with the samples that were filtered and the reasons why - as described by https://github.com/alexslemonade/refinebio/issues/1838#issuecomment-559143625

## Types of changes

## Functional tests

Added tests for a compendium where we skip one sample.

## Checklist

### DIFF
--- a/workers/data_refinery_workers/processors/smashing_utils.py
+++ b/workers/data_refinery_workers/processors/smashing_utils.py
@@ -312,7 +312,7 @@ def process_frames_for_key(key: str,
                                job_id=job_context["job"].id)
                 job_context['filtered_samples'][sample.accession_code] = {
                     'reason': 'We were unable to smash the file associated with this sample during the first pass.',
-                    'source_url': computed_file.source_url
+                    'filename': computed_file.filename
                 }
                 continue
 
@@ -387,7 +387,7 @@ def process_frames_for_key(key: str,
             job_context['unsmashable_files'].append(computed_file.filename)
             job_context['filtered_samples'][sample.accession_code] = {
                 'reason': 'We were unable to smash the file associated with this sample during the first pass.',
-                'source_url': computed_file.source_url
+                'filename': computed_file.filename
             }
             continue
 
@@ -631,7 +631,7 @@ def compile_metadata(job_context: Dict) -> Dict:
 
     samples = {}
     for sample in job_context["dataset"].get_samples():
-        if job_context['filtered_samples'] and sample.accession_code in job_context['filtered_samples']:
+        if sample.accession_code in filtered_samples:
             # skip the samples that were filtered
             continue
         samples[sample.accession_code] = sample.to_metadata_dict()
@@ -675,7 +675,7 @@ def write_non_data_files(job_context: Dict) -> Dict:
         with open(aggregated_metadata_path, 'w', encoding='utf-8') as metadata_file:
             json.dump(job_context['metadata'], metadata_file, indent=4, sort_keys=True)
 
-        if 'filtered_samples' in job_context and job_context['filtered_samples']:
+        if job_context['filtered_samples']:
             # generate filtered samples file only if some samples were skipped
             filtered_samples_path = os.path.join(job_context["output_dir"],
                                                  'filtered_samples_metadata.json')
--- a/workers/data_refinery_workers/processors/smashing_utils.py
+++ b/workers/data_refinery_workers/processors/smashing_utils.py
@@ -691,6 +691,16 @@ def write_non_data_files(job_context: Dict) -> Dict:
                                                  'filtered_samples_metadata.json')
             with open(filtered_samples_path, 'w',encoding='utf-8') as metadata_file:
                 json.dump(job_context['filtered_samples'], metadata_file, indent=4, sort_keys=True)
+
+            columns = get_tsv_columns(job_context['filtered_samples'])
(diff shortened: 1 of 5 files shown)


workers/data_refinery_workers/processors/smashing_utils.py
```diff
@@ -643,18 +667,22 @@ def write_non_data_files(job_context: Dict) -> Dict:
...
+                                                'aggregated_metadata.json')
+        with open(aggregated_metadata_path, 'w', encoding='utf-8') as metadata_file:
             json.dump(job_context['metadata'], metadata_file, indent=4, sort_keys=True)
+
+        if job_context['filtered_samples']:
+            # generate filtered samples file only if some samples were skipped
+            filtered_samples_path = os.path.join(job_context["output_dir"],
+                                                 'filtered_samples_metadata.json')
+            with open(filtered_samples_path, 'w',encoding='utf-8') as metadata_file:
+                json.dump(job_context['filtered_samples'], metadata_file, indent=4, sort_keys=True)
```
kurtwheeler: I'm not sure this metadata file is sufficient. For one thing, we tend to include a JSON version and a TSV version of every metadata file we include in the compendia, in case a user is only comfortable with one of the formats.

Secondly, in https://github.com/AlexsLemonade/refinebio/issues/1838#issuecomment-559143625 you mention

> We could also add all sample metadata there, and the experiment relations.

I would think we probably want to do this because if a user is actually using this file then they may be interested in analyzing why some samples got filtered. Would you agree with this @jaclyn-taroni? Should we include every metadata field we have about every sample that got filtered including it's experiment_accession_code?

workers/data_refinery_workers/processors/smashing_utils.py
```diff
@@ -304,6 +310,10 @@ def process_frames_for_key(key: str,
                                computed_file=computed_file.id,
                                dataset_id=job_context['dataset'].id,
                                job_id=job_context["job"].id)
+                job_context['filtered_samples'][sample.accession_code] = {
+                    'reason': 'We were unable to smash the file associated with this sample during the first pass.',
```
kurtwheeler: This reason is going to go into the metadata file that we give our users. They won't know what "first pass" and "second pass" mean and we should avoid the use of the word "smash". A better reason to me would be: `The file associated with this sample did not pass the QC checks we apply before aggregating.` or something.

workers/data_refinery_workers/processors/smashing_utils.py
```diff
@@ -74,6 +76,10 @@ def prepare_files(job_context: Dict) -> Dict:
                 smashable_files = smashable_files + [(smashable_file, sample)]
                 seen_files.add(smashable_file)
                 found_files = True
+            else:
+                job_context['filtered_samples'][sample.accession_code] = {
+                    'reason': 'We could not find a valid smashable file for this sample.'
```
kurtwheeler: We shouldn't use `smashable` in copy destined for user consumption. Maybe `This sample did not have a processed file associated with it in our database.` would be better?

workers/data_refinery_workers/processors/smashing_utils.py
```diff
@@ -375,9 +385,13 @@ def process_frames_for_key(key: str,
 
         if frame_data is None:
             job_context['unsmashable_files'].append(computed_file.filename)
+            job_context['filtered_samples'][sample.accession_code] = {
+                'reason': 'We were unable to smash the file associated with this sample during the second pass.',
```
kurtwheeler: Once again, they won't know what second pass means. Perhaps `The file associated with this sample did not contain a vector that fit the expected dimensions of the matrix.`? I'm less sure about this one. I honestly wouldn't think we'd see many of these.

workers/data_refinery_workers/processors/smashing_utils.py
arielsvn: > we tend to include a JSON version and a TSV version of every metadata file we include in the compendia, in case a user is only comfortable with one of the formats.

We only have a tsv version for the samples inside the compendia. The main `aggregated_metadata.json` file doesn't have a tsv version.

image

I'd say the new `filtered_samples_metadata.json` file should be similar to `aggregated_metadata.json`. Do you think this makes sense?

> Should we include every metadata field we have about every sample that got filtered including it's experiment_accession_code?

Ok, I'll also include this information for each sample.

arielsvn:
Thanks for reviewing @kurtwheeler. I think I addressed all your comments and included a new file named `filtered_samples_metadata.tsv`. Let me know if you see something else that can be improved

kurtwheeler:
These changes look good but looks like the tests aren't passing :(

arielsvn:
@kurtwheeler tests passing now :)

Issue #1795, opened 2019-10-16
Redirect to latest version of the API when people use unversioned endpoint


arielsvn:
### Context

Our API is versioned with endpoints like 

https://api.refine.bio/v1/transcriptome_indices/

### Problem or idea

The docs have a lot of endpoints that reference unversioned endpoints like

https://api.refine.bio/transcriptome_indices/?organism=DANIO_RERIO&length=LONG

Which now returns 404, we should automatically redirect to the latest endpoint version if someone uses those.

### Solution or next step

We need to investigate if there's a recommended approach to implement this.

cgreene:
Update the docs, but if someone goes to an unversioned endpoint potentially return something more specific than a generic 404.

cgreene:
Closed by #2001

PR #2001, opened 2019-12-13
Unversioned API response

davidsmejia:
## Issue Number

#1795 

This PR ensures that the response type for 404 and 500 errors will be JSON. 
If a 404 error will be returned it will check to see if the request began with `r"^/v[1-9]"` and if so it will respond with a message mentioning that the api is versioned.
Also, the 404 errors will return a link to the docs.

## Methods

`data_refinery_api/urls.py`

## Types of changes

## Functional tests

Tested with debug = False

### DIFF
--- a/api/data_refinery_api/middleware.py
+++ b/api/data_refinery_api/middleware.py
@@ -13,22 +13,21 @@ class SentryCatchBadRequestMiddleware(MiddlewareMixin):
         if not client.is_enabled():
             return response
 
+        # format error message
+        message = (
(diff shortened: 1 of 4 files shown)


api/data_refinery_api/middleware.py
```diff
@@ -22,13 +23,13 @@ def process_response(self, request, response):
         content = ""
         try:
             content = "with message: {}".format(str(response.content))
-        except:
+        except Exception:
```
arielsvn: Remove the `try...except`?

Issue #1284, opened 2019-04-26
Remove staging smasher instance


kurtwheeler:
### Context

We have a persistent smasher instance for the staging stack.

### Problem or idea

No one requests datasets from staging, so we don't ever run smash jobs.

### Solution or next step

Disable the smasher instance for staging deploys, make it so that we can bring it back easily if we ever do need it to test smashing/compendia in staging.

PR #2008, opened 2019-12-17
Add full_stack env var to be able to disable infra for non-prod stacks.

kurtwheeler:
## Issue Number

#1284

We don't wanna run a smasher instance in staging all the time so we don't have to pay for it as much. This disables that, but it also introduces a pattern that can be used to disable pieces of infrastructure that we don't need in staging all the time.

This is the `full_stack` env var, and it's set to True for prod and dev stacks and False for staging. If we ever want to re-enable the infra on staging all we have to do is set it to True, run our tests, and then reset it back to false to spin down non-essential infra.

Also clean up some deprecated env vars.

## Types of changes

## Functional tests

I've tested this with my dev stack. Spun up with full_stack=True, saw the instance come up, set it to False,  saw it go away, set it back to True, saw it come back. This last one is still WIP but it will be true by the time I request a reviewer.

## Checklist


kurtwheeler:
Ok, I've finished my tests of this! It works super well. That field dictates whether or not a smasher instance exists. It should be easy to make that work for other resources too!

infrastructure/instances.tf
```diff
@@ -247,6 +247,8 @@ data "template_file" "nomad_client_script_smasher_smusher" {
 }
 
 resource "aws_instance" "smasher_instance" {
+  # This is an expense we don't need in the staging stack.
+  count = "${var.stage == "prod" || var.full_stack == "True" ? 1 : 0}"
```
arielsvn: Why do we need to check for `var.stage == "prod"` here? `var.full_stack` should be `True` in prod right?

kurtwheeler: I agree that it should be. However I don't think that we'll ever want prod to not have the full stack so this adds some protection against us ever accidentally changing that value in the `prod.environment` file.

PR #2009, opened 2019-12-18
Bump django versions

kurtwheeler:
## Issue Number

This replaces some of the auto-PRs that got generated.

Django released some security updates. This lets us take advantage of them.

We can't go up to Django 3.0.1 until https://github.com/sabricot/django-elasticsearch-dsl/pull/233 is released.

## Types of changes

- Version Increase

## Functional tests

I've run the API and hit it to make sure all the versions work  together.

## Checklist

### DIFF
--- a/workers/data_refinery_workers/downloaders/requirements.txt
+++ b/workers/data_refinery_workers/downloaders/requirements.txt
@@ -4,12 +4,13 @@
 #
 #    pip-compile --output-file requirements.txt requirements.in
 #
+asgiref==3.2.3            # via django
 certifi==2018.8.24        # via requests
 chardet==3.0.4            # via requests
 coverage==4.5.1           # via coveralls
 coveralls==1.5.1
 django-elasticsearch-dsl==0.5.1
-django==2.2.7
+django==3.0.1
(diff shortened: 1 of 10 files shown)


workers/data_refinery_workers/downloaders/requirements.txt
```diff
@@ -4,12 +4,13 @@
...
 #    pip-compile --output-file requirements.txt requirements.in
 #
+asgiref==3.2.3            # via django
 certifi==2018.8.24        # via requests
 chardet==3.0.4            # via requests
 coverage==4.5.1           # via coveralls
 coveralls==1.5.1
 django-elasticsearch-dsl==0.5.1
-django==2.2.7
+django==3.0.1
```
arielsvn: Should we use the same version of Django in all projects?

```suggestion
django>=2.2.9
```

kurtwheeler: Good catch. I changed the version number in requirements.in but I didn't change `>=` to `==` so recompiling didn't downgrade it. I've fixed it here and in the processors as well!

PR #2014, opened 2019-12-24
Ensure datasets are marked with errors when processing fails

### LINKED ISSUE #1923: Smasher jobs don't mark datasets as failures when they fail
### Context

There's a couple places in the smasher code where we delay failing the job so that the job can make it to the `_notify()` function so it can let the user know that their dataset failed to process.

### Problem or idea

The job eventually gets marked as a failure like we want. However the dataset object does not, nor does the failure_reason get set on the dataset object. This means that the frontend just forever shows the dataset as processing.

### Solution or next step

Fix this bug by fixing the smasher's flow. Doing it like that is such a hack. We should have two notify functions, one for successes, one for failures. If the job fails we should call the `_notify_failure()` function and then properly exit the processor, setting the `failure_reason` and `success` fields of the dataset object. If we succeed then the pipeline should be allowed to continue to the `notify_success()` function.

arielsvn:
## Issue Number

close #1923 

Datasets failure reason is now set on `end_job`, for jobs associated with a dataset.

Also reduced the scope of several `try...except` since all exceptions are handled on `run_pipeline`.

## Types of changes

## Functional tests

None

## Checklist

### DIFF
--- a/workers/data_refinery_workers/processors/smasher.py
+++ b/workers/data_refinery_workers/processors/smasher.py
@@ -226,40 +226,17 @@ def _smash_key(job_context: Dict, key: str, input_files: List[ComputedFile]) -> 
 
     # Quantile Normalization
     if job_context['dataset'].quantile_normalize:
-        try:
-            job_context['merged_no_qn'] = merged
-            job_context['organism'] = job_context['dataset'].get_samples().first().organism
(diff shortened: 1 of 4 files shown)


arielsvn:
@kurtwheeler resolved conflicts here, this should be ready for review

Issue #1986, opened 2019-12-09
The annotation package, pd.huex.1.0.st.v1, could not be loaded


kurtwheeler:
### Context

I'm seeing this error message in our logs:
```
2019-12-09 01:39:11,785 i-056f1616d982e5f85 [volume: 0] data_refinery_workers.processors.utils ERROR [processor_job: 29527291] [failure_reason: Encountered error in R code while running AFFY_TO_PCL pipeline during processing of /home/user/data_store/GSE56583/GSM1364698_h_10282009_5-21-10_S2.CEL: Error in { :
  task 1 failed - "The annotation package, pd.huex.1.0.st.v1, could not be loaded."
]: Processor function _run_scan_upc failed. Terminating pipeline.
```

Most likely related to #1831 

### Problem or idea

We have the HUEX v2, but not v1 it looks like: https://github.com/AlexsLemonade/refinebio/blob/dev/workers/affymetrix_dependencies.R#L126

Bioconductor has v2 but not v1 it seems:
  * https://bioconductor.org/packages/release/data/annotation/html/pd.huex.1.0.st.v2.html works but
  * https://bioconductor.org/packages/release/data/annotation/html/pd.huex.1.0.st.v1.html doesn't
  * A search for pd.huex.1.0.st https://bioconductor.org/help/search/index.html?q=pd.huex.1.0.st/ only returns results for v2.

### Solution or next step

I'm not sure.

@jaclyn-taroni any ideas? Do you know another place we could try to get the CDF for v1 for this platform?

cgreene:
This makes it sound like there will not be a CDF for v1 of this platform, but it may be safe to override and use v2:
http://aroma-project.org/chipTypes/HuEx-1_0-st-v2/

cgreene:
I talked with @jaclyn-taroni and she says we should do this override 👍

cgreene:
Figure out if this is why #1691 didn't fully process. We would ideally like to find an experiment run with v1 and v2 together. We'd like to process that on staging by renaming the CDFs. Then we'd like to look at gene expression from both sets and make sure the values aren't widely divergent.

cgreene:
@jaclyn-taroni says there's an instance where we sanitize what comes off the CEL header and this is a similar issue

kurtwheeler:
[...]

PR #2015, opened 2019-12-25
Add Array Express Annotation Package Name Overrides Functionality

davidsmejia:
## Issue Number

#1986 

Adds support for adding overrides to specific brainarray platforms.
Adds Huex v1 override to use v2
Adds test for Huex v1 to use v2 annotation package name
Fixed linter errors/warning on files touched

Basically, since `pd.huex.1.0.st.v1` does not exist this allows us to specify passing `pd.huex.1.0.st.v2` instead to scan_upc to use as the `annotationPackageName`

You can add additional overrides to `./config/affymetrix_annotation_package_name_overrides.csv` and they will be checked and applied when the brainarray package is determined.

## Methods

`./workers/data_refinery_workers/processors/array_express.py:_determine_brainarray_package`
`./workers/data_refinery_workers/processors/array_express.py:_run_scan_upc`

## Types of changes

## Functional tests

`./workers/run_tests.sh -t affymetrix `

## Checklist


davidsmejia:
This should probably be documented somewhere that we override the annotation package.
Also I'm not really sure what the implications are of doing this or if I did it right so a science review/test would be good on staging first. Let me know what you would need to test this before we go to prod. @cgreene

cgreene:
I like this solution! Our ideal test case would be a dataset that contains both v1 samples (failed to process) and v2 samples (successfully processed). I wonder if the experiment that was requested ( https://www.refine.bio/experiments/GSE131617/genes-associated-with-the-progression-of-neurofibrillary-tangles-in-alzheimers-disease ) does just that. It looks like about half of the samples failed to process. If we're lucky those are on v1.

Then we'd plot each gene (ensembl ID's) average expression for v1 samples against its average expression for v2 samples and add a diagonal 1-1 line. If the answers from Affymetrix are correct, we'd see a more or less diagonal line here. If the answers from Affy in the various threads are wrong (spots change position on the array) we'll see essentially a random positioning of points.

The first step is ideally to find an experiment with some V1 and some V2 though. 🤞 that it's the one that was requested.

common/data_refinery_common/utils.py
```diff
@@ -12,6 +12,7 @@
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 from retrying import retry
+from multiprocessing import current_process
```
cgreene: unused import?

davidsmejia: I saw an error that it wasn't being imported or defined here, so I imported it.. https://github.com/AlexsLemonade/refinebio/pull/2015/files/9f39a868e98b2cb52000b5ad4f9fe8d67ddc67a3#diff-851165ed4bf22b784eb9cf6b338d69ecR66

cgreene: gotcha - i missed current_process being used there 👍

davidsmejia:
I'd like to merge this in, see if we can process v1 samples, and perform the following test

```graph 1 Compare v1/v2

all datasets are affy huex

dataset 1
51 experiments
approx 1300 total samples
aka. all experiments using v1

datset 2
51 random v2 experiments that contain approx same samples as v1 

dataset 3 and 4
51 random v2 experiments with the same sample count as dataset 2 all v2

graph 1
x: dataset 1 avg gene expression 
y: dataset 2 avg gene expression

graph 2
x: dataset 3 avg gene expression
y: dataset 4 avg gene expression

see how close both graphs are to x=y 

????

profit
```

common/data_refinery_common/utils.py
```diff
@@ -235,14 +244,41 @@ def get_readable_affymetrix_names(mapping_csv: str="config/readable_affymetrix_n
...
+    Annotation Package Name | Inernal Accession
+    """
+    global ANNOTATION_PACKAGE_OVERRIDES
+
+    if ANNOTATION_PACKAGE_OVERRIDES is not None:
+        return ANNOTATION_PACKAGE_OVERRIDES
+
+    ANNOTATION_PACKAGE_OVERRIDES = {}
+    with open(overrides_csv, encoding='utf-8') as overrides_file:
+        reader = csv.reader(overrides_file, )
```
cgreene: were you intending to put something after `overrides_file`?

common/data_refinery_common/utils.py
```diff
@@ -235,14 +244,41 @@ def get_readable_affymetrix_names(mapping_csv: str="config/readable_affymetrix_n
...
+        for line in reader:
```
cgreene: Is skipping preferred under our style guide to using `next(reader)` before starting the loop? I'd probably do the `next...` approach.

common/data_refinery_common/utils.py
```diff
@@ -296,14 +337,16 @@ def calculate_sha1(absolute_file_path):
...
 def get_sra_download_url(run_accession, protocol="fasp"):
     """Try getting the sra-download URL from CGI endpoint"""
-    #Ex: curl --data "acc=SRR6718414&accept-proto=fasp&version=2.0" https://www.ncbi.nlm.nih.gov/Traces/names/names.cgi
+    # Ex: curl --data "acc=SRR6718414&accept-proto=fasp&version=2.0" \
+    #   https://www.ncbi.nlm.nih.gov/Traces/names/names.cgi
     cgi_url = "https://www.ncbi.nlm.nih.gov/Traces/names/names.cgi"
     data = "acc=" + run_accession + "&accept-proto=" + protocol + "&version=2.0"
     try:
         resp = requests.post(cgi_url, data=data)
-    except Exception as e:
```
cgreene: I'm guessing you changed this because `e` was unused.

common/data_refinery_common/utils.py
```diff
@@ -358,7 +402,7 @@ def load_blacklist(blacklist_csv: str = "config/RNASeqRunBlackList.csv"):
         for line in reader:
```
cgreene: Same comment as above about the next() question but I guess this answers it & this is a pattern we use in our code.

workers/data_refinery_workers/processors/test_array_express.py
```diff
@@ -82,3 +103,18 @@ def test_affy_to_pcl_no_brainarray(self):
         self.assertEqual(ComputedFile.objects.all()[0].filename, 'GSM45588.PCL')
 
         os.remove(ComputedFile.objects.all()[0].absolute_file_path)
+
+    @tag("affymetrix", "huex")
```
cgreene: Do you want a test in here somewhere that makes sure things fail without the override?

common/data_refinery_common/utils.py
davidsmejia: just convention to have trailing commas

common/data_refinery_common/utils.py
davidsmejia: yup

common/data_refinery_common/utils.py
davidsmejia: yeah I had only planned on fixing the pep8 issues but we I can open a PR, honestly since its being repeated we can with a single function

workers/data_refinery_workers/processors/test_array_express.py
davidsmejia: hmm I'm not sure how i would be able to test that since it reads from the file... would need to have another example that we don't want to override and should fail?

common/data_refinery_common/utils.py
davidsmejia: Yeah I think that would make more sense too

workers/data_refinery_workers/processors/test_array_express.py
cgreene: I was imagining that you'd have the same file without the override & it would fail + the version with the override we would test for success. Maybe it's overkill for this case.

Issue #1944, opened 2019-11-25
Too many emails when dataset fails to smash


dvenprasad:
### Context
I tried to download a dataset from refine.bio which failed. (Nov 19, 20) 

### Problem or idea
I received 5 emails that day and 4 of them in a span of 20 minutes, all with the same failure reason. 
For another one, I tried to download I got 3 emails in a span of 2 minutes. 

I don't know if this is the normal behavior or there was something else going on that day which might have triggered the additional emails.

### Solution or next step

Users should receive one failure email.

arielsvn:
I need some help here... when a dataset fails to process we have to send an email to the user that requested it, the problem is that datasets jobs are executed multiple times and from the smasher, we don't have a way to know which one will be the last time.

I see two possible solutions:

1. Add a way to determine when a smasher job won't be retried again, ie it has `num_retries=2`, and send the emails before finishing the job. 
The logic to determine if a job should be retried is in the foreman, so we would have to either duplicate it or add a dependency between these modules.

2. Execute the smasher jobs **a single time**, this way we can always send the email if processing fails.

I lean towards the second one since we allow users to retry datasets easily when they fail from the frontend. 

Does this make sense? Am I missing something here? @davidsmejia @kurtwheeler

arielsvn:
- Have the foreman queue a new job type that should handle failures for other jobs, in the case of the smasher here is where we would send the failure notification. But it could be useful for other jobs as well.

- Don't send failures notifications at all

Talked with @davidsmejia and we came up with additional options.

arielsvn:
Also, if a dataset fails to smash, what's the likelihood that re-trying it would have a different effect?

arielsvn:
I looked into all smasher failures to see if there might be some that make sense to retry. These are all the **different reasons**

```
'Caught either a SIGTERM or SIGINT signal.'
'Failure reason: Failed to upload /home/user/data_store/smashed/9b0127b5-e3c1-42ea-ae5a-58a945131a9c.zip to data-refinery-s3-results-circleci-prod/9b0127b5-e3c1-42ea-ae5a-58a945131a9c.zip: An error occurred (BadDigest) when calling the PutObject operation (reached max retries: 4): The Content-MD5 you specified did not match what we received.'

'Could not find QN target for Organism: MUS_MUSCULUS'
"Couldn't get any files to smash for Smash job!!"
"Couldn't get any files to smash for Smash job - empty all_sample_files"
[...]

PR #2021, opened 2019-12-27
Improve slack notifications when datasets are submitted

arielsvn:
## Issue Number

#1944

Include additional information in slack notifications when new datasets are submitted or they fail processing.

image

image

## Types of changes

## Functional tests

Tested sending notifications to #robots

## Checklist

### DIFF
--- a/api/data_refinery_api/views.py
+++ b/api/data_refinery_api/views.py
@@ -512,6 +512,7 @@ class DatasetView(generics.RetrieveUpdateAPIView):
                             city = "COULD_NOT_DETERMINE"
 
                         user_agent = self.request.META.get('HTTP_USER_AGENT', None)
+                        total_samples = len(set([accession_code for experiment in new_data.values() for accession_code in experiment]))
                         response = requests.post(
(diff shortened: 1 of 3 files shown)


api/data_refinery_api/views.py
```diff
@@ -511,21 +511,42 @@ def perform_update(self, serializer):
...
+                            json={
+                                'channel': 'ccdl-general', # Move to robots when we get sick of these
+                                'username': 'EngagementBot',
+                                'icon_emoji': ':halal:',
+                                'attachments': [
+                                    {
+                                        'color': 'good',
+                                        'title': 'New dataset download',
+                                        'fallback': 'New dataset download',
+                                        'title_link': 'http://www.refine.bio/dataset/' + str(old_object.id),
```
davidsmejia: Can we do string concat like this instead?

```"http://www.refine.bio/dataset/{0}".format(old_object.id)```

Comment (arielsvn) [reply api/data_refinery_api/views.py]:
```suggestion
                                        'title_link': 'http://www.refine.bio/dataset/'.format(old_object.id),
```

Issue #1692, opened 2019-09-30
Make LGTM Happy


cgreene:
### Context

We recently enabled LGTM for refine.bio.

### Problem or idea

We have 323 issues identified by LGTM ( https://lgtm.com/projects/g/AlexsLemonade/refinebio/ ). @kurtwheeler notes that code quality measures are more effective if they start from a good place. Many (all?) of these are pretty straightforward to fix (unused imports, etc).

### Solution or next step

We should make a focused effort to resolve all of these that we can so that the pressure is higher on code quality (going from 0 issue to 1 issue is a bigger mental step than 323 to 324 issues).

arielsvn:
Is there any way to fix them automatically? In js with Eslint, we can fix the unused import alerts automatically. If we have to do it manually, I would lean towards making the fixes gradually.

PR #2024, opened 2019-12-27
Make static method static

cgreene:
## Issue Number

#1692 

LGTM noticed that this wasn't a static method but that it was being used as such. It does look like it was intended to be a static method:
http://ses4j.github.io/2015/11/23/optimizing-slow-django-rest-framework-performance/

This may be dead code though. Unless there's some magic where this gets called (the blog post suggests one must intentionally call it if I'm reading it correctly) the code does not appear to get used in our project.

## Methods

N/A

## Types of changes

## Functional tests

## Checklist

## Screenshots

### DIFF
--- a/api/data_refinery_api/serializers.py
+++ b/api/data_refinery_api/serializers.py
@@ -425,6 +425,7 @@ class ExperimentSerializer(serializers.ModelSerializer):
                     'technologies'
                 )
 
+    @staticmethod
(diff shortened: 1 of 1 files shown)


cgreene:
Ahh - we don't run tests on PRs from forks. No wonder everyone works on branches in the main repo :)

PR #2022, opened 2019-12-27
[Experimental] Reformat all code

arielsvn:
## Issue Number

N/A

## Purpose/Implementation Notes

@davidsmejia and I found a [code formatter for Python](https://black.readthedocs.io/en/latest/index.html), not sure if it works well or if we like its style. It says it's opinionated and doesn't have a lot of customization options. However it would be nice to have an automatic tool that keeps the formatting of the codebase consistent. In their own words:

> Black gives you speed, determinism, and freedom from pycodestyle nagging about formatting. You will save time and mental energy for more important matters.

If tests pass we should discuss if this is a style we like @kurtwheeler @davidsmejia, or maybe there're other options.

Here's the link to the Black repository https://github.com/psf/black

## Types of changes

- Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Functional tests

None

## Checklist

- [ ] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

PR #2033, opened 2020-01-02
Reformat all code: with a line limit of 100.

kurtwheeler:
## Issue Number

#2022

This is a remake of #2022, run with a line limit of 100 instead of black's default 88. This is because we've been using a line limit of 100 for a while so a lot of code was written to fit neatly within 100 characters, so black had broken a lot of these up in unfortunate ways.

Also this shaves ~1600 lines of code out of our project compared to the 88 character line limit.

## Types of changes

- Reformat

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,12 +126,12 @@ The following services will need to be installed:
 - [git-crypt](https://www.agwa.name/projects/git-crypt/)
 - [iproute2mac](https://github.com/brona/iproute2mac)
 - [jq](https://stedolan.github.io/jq/)
-- [black](https://black.readthedocs.io/en/stable/) can be installed on Mac clients with `pip3 install black`
+- [black](https://black.readthedocs.io/en/stable/)
 
 Instructions for installing [Docker](https://www.docker.com/docker-mac) and [Homebrew](https://brew.sh/) can be found by
 on their respective homepages.
 
-Once Homebrew is installed, the other required applications can be installed by running: `brew install iproute2mac git-crypt nomad terraform jq`.
+Once Homebrew is installed, the other required applications can be installed by running: `brew install iproute2mac git-crypt nomad terraform jq black`.
 
 Many of the computational processes running are very memory intensive. You will need
 to [raise the amount of virtual memory available to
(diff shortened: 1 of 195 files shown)


kurtwheeler:
@arielsvn I added some manual changes to the README, install_all, and I added a script for us to copy over to the pre-commit hook. Wanna take a look? (I'm not sure the way I'm installing black on linux is general enough, should it sometimes be pip instead of just pip3?)

README.md
```diff
@@ -122,11 +126,12 @@ The following services will need to be installed:
 - [git-crypt](https://www.agwa.name/projects/git-crypt/)
 - [iproute2mac](https://github.com/brona/iproute2mac)
 - [jq](https://stedolan.github.io/jq/)
+- [black](https://black.readthedocs.io/en/stable/)
 
 Instructions for installing [Docker](https://www.docker.com/docker-mac) and [Homebrew](https://brew.sh/) can be found by
 on their respective homepages.
 
-Once Homebrew is installed, the other required applications can be installed by running: `brew install iproute2mac git-crypt nomad terraform jq`.
+Once Homebrew is installed, the other required applications can be installed by running: `brew install iproute2mac git-crypt nomad terraform jq black`.
```
arielsvn: I installed `black` using `pip3` on Mac. Also tested it, and it doesn't work with `brew`.

scripts/install_all.sh
```diff
@@ -152,6 +152,18 @@ if ! command -v nomad > /dev/null; then
     fi
 fi
 
+if ! command -v black > /dev/null; then
+    echo "Installing black..."
+    if [ $BREW ]; then
+        $INSTALL_CMD black > $OUTPUT
```
arielsvn: should we always use `pip3` for `black`?

kurtwheeler: Ok thanks for testing that!! I've made that change.

kurtwheeler:
Ok, I just tested using `install_all.sh` to configure the pre-commit hook and then made a commit and it worked!!!

README.md
arielsvn: We should remove black from here too.

```suggestion
Once Homebrew is installed, the other required applications can be installed by running: `brew install iproute2mac git-crypt nomad terraform jq`.
```

arielsvn:
Ups my bad, seems that black can be installed with `brew`

PR #2034, opened 2020-01-02
[Experimental] Testing isort to organize all imports automatically

arielsvn:
## Issue Number

N/A

Since we liked `black` (#2033) what about going one step further and also run `isort` to organize all our imports automatically and into sections automatically.

This tool is recommended on the [Django contributing guides](https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/#imports)

@davidsmejia @kurtwheeler what do you think of this?

## Types of changes

## Functional tests

None

## Checklist


cgreene:
FYI there is a ticket for black/isort interoperability that looks like it has a config file:
https://github.com/timothycrosley/isort/issues/694

It looks like there are also ways to remove unused imports automatically as well:
https://github.com/psf/black/issues/333

arielsvn:
> there is a ticket for black/isort interoperability that looks like it has a config file

Yes, I used the configuration black recommended on [their docs](https://black.readthedocs.io/en/latest/the_black_code_style.html#how-black-wraps-lines)

> It looks like there are also ways to remove unused imports automatically as well

That's really cool, I didn't know that existed. I wouldn't go as far as to add this automatically, but we could run it and remove all imports in one PR

arielsvn:
@kurtwheeler added `isort` to the pre-commit hook.

Also tried to add it to circleci but it didn't work, it doesn't pick up the configuration correctly for some reason.

arielsvn:
I spent some time playing with the configuration of `isort` and there're some issues that I haven't been able to solve. 

I started using [seed-isort-config](https://github.com/asottile/seed-isort-config/) which generates a config option that informs `isort` of which dependencies are 3rd parties - those get organized into their own group.

All good until that part, the problem is that `seed-isort-config` depends on the python version that it's installed in order to determine the 3rd party packages. In circleci we have `Python 3.6.4` and locally I have `Python 3.7.4`. This means that the generated list of 3rd party packages is different in my local environment (when I commit code) vs circle ci, so the `syntax_tests` fail

I see two options to move this forward:

1. Don't enable import automatic sorting on pre-commit. We could still have the command available and do big PR to sort all imports once.

2. Don't run `seed-isort-config` on `pre-commit`. This would allow us to keep the imports sorted based on a fixed list of known 3rd party packages.

@davidsmejia @kurtwheeler what do you think?

arielsvn:
Ok, syntax tests passed. In the config, I enumerated all our `FIRSTPARTY` packages and made everything else `THIRDPARTY`. This removed the need to use `seed-isort-config` which was causing all the issues with the python versions.

@kurtwheeler @davidsmejia let me know what do you think.

scripts/install_all.sh
```diff
@@ -165,6 +165,16 @@ using an additional ~185MB of disk space."
...
 
+if ! command -v isort > /dev/null; then
+    echo "Installing isort..."
+    if [ $APT ] || confirm "Would you like to automatically install isort?"; then
+        pip3 install isort seed-isort-config
+    else
+        echo "You need to manually install isort (code formatter) before continuing..." >&2
+        exit 1
+    fi
+fi
```
kurtwheeler: I thought `pre-commit` was going to handle installation of its packages for us?

arielsvn: it does, forgot to remove this.

arielsvn: done

Issue #1995, opened 2019-12-12
Expand data model to support organism strains


kurtwheeler:
### Context

In https://github.com/AlexsLemonade/refinebio/issues/1722 we're starting to tackle the problem of strains. However our data model has no concept of strains nor does it currently have a way to support it.

### Problem or idea

Regardless of whether we decide to go with one strain per organism or multiple, we should figure out a data model to support multiple strains so that we don't lock ourselves into one-strain-per-organism.

### Solution or next step

We should figure out how to represent this in the database.

I have two initial ideas:
  * One would be to create a separate Strains model and have a one-to-many relationship between organims and strains. The tricky part about this is that it apppears NCBI gives each strain a different taxonomy id, and that currently is a property of our organisms table.
  * The other would be to treat each strain like a separate organism. This seems to be more in line with how NCBI handles it. We potentially could use this as an opportunity to make the organism model more robust in the following ways:
    * Rather than using a `get_genus()` function to get an organism's genus, we could just store that as a separate property.
    * While we're at it we could just store each of `kingdom, phylum, class, order, family, genus, species, strain` as their own property. This could be useful later for other things?
    * We could even build out some additional behavior to organisms such that you could ask an organism to give experiments for the full species, or genus, rather than just the one relationship. This would let us then choose what grouping we want to work with in different places.

Here are two strains of E. coli that have different taxonomy ids for reference:
  * https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Info&id=1281005&lvl=3&lin=f&keep=1&srchmode=1&unlock
  * https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Info&id=1432549&lvl=3&lin=f&keep=1&srchmode=1&unlock

Issue #1722, opened 2019-10-02
Microbial transcriptome indices


georgiadoing:
### Context

For many microbes, it is not clear which transcriptome indices are best for building compendia. 

### Problem or idea

The first hurdle is that genome assemblies are not in the main Ensemble! database. The second is that many microbes have many genome assemblies without a standard naming scheme to point out the 'best' reference. Part of this comes from the multitude of strains for each microbial species and in turn the multitude of assemblies for each strain.

### Solution or next step

1. A first step would be if transcriptome indices can be called for EnsembleBacteria in a manner similar to how the README instructs for Surveyor Jobs like so:

./foreman/run_surveyor.sh survey_all --accession "Pseudomonas aeruginosa, EnsemblBacteria"

2. Next we would need to decide and specify which **strain** to use. This is a bit harder to solve and my only short-term solution is to manually specify  for each microbe of interest based on expert knowledge. For example, if our preferred reference strain of Pseudomonas aeruginosa PAO1 could be specified, perhaps like so:

./foreman/run_surveyor.sh survey_all --accession "Pseudomonas aeruginosa PAO1, EnsemblBacteria"

3. Thirdly would be to make sure we are getting the best **assembly** for that strain. For example, for Pseudomonas aeruginosa PAO1 there are 7 genomes in EnsemblBacteria and we would like to specifically get ASM676v1, the most complete reference genome that would be best for building a compendium. 

In summary, if there is a way, for a given microbial organism, to specify the strain and assembly in the accession call, and if we curated a list of preferred strains and assemblies we could begin to tackle this.

Below is a link to a working gsheet of organisms, strains and assemblies that I will curate, in case it is useful.

https://docs.google.com/spreadsheets/d/1Lbi68UP2dQtfp-KoxtXpE7jhCOxgP_FweGznbbOiMkw/edit?usp=sharing

dvenprasad:
Got a request for `bacillus subtilis` via hotjar

kurtwheeler:
Hi @georgiadoing! I've finally gotten a chance to dig into this a little. I've come up with a few questions for you.

When you say "Next we would need to decide and specify which strain to use." it makes me think that you're suggesting that we use a single strain per organism, but then for _Candida albicans_ the spreadsheet has two rows with different strains. Are you thinking that we should handle multiple strains per organism or just do one per?

[...]

PR #2035, opened 2020-01-03
Support "blessed" strains of specific organisms

kurtwheeler:
## Issue Number

#1722 

In #1722 @georgiadoing a mapping from organism to strain/assembly for a limited number of organisms. This gave us enough information to be able to build transcriptome indices for these organisms, whereas before we did not know which strain's genome should be used for each organism.

This PR extends our transcriptome index surveyor so that we can now discover the locations of these files. They were not exactly easy to find. I had to email the ensembl dev mailing list and this was the response:

```
Unforunately, you are into rather inconvenient methods to do this. We have plans for a service that indexes the FTP content and will make it searchable, but you'll have to wait for that.

In the meantime, you can:

Get ftp://ftp.ensemblgenomes.org/pub/bacteria/current/species_EnsemblBacteria.txt
Extract coloumn 13 for the rows you are interested in
The column contains values like bacteria_177_collection_core_45_98_1
Regex out the tail end after _collection and then you have the collection name: bacteria_177_collection
Put that into your URL, then add "species" name as per the tab-separated file above

It's not pretty but it's better than crawling the FTP site! My answer to your other question might be a good source of unique species names depending on how you're working with our services.

Hopefully you are already aware of the "current" link in the path should you not wish to work on a specific release.
```

This PR does what he says, so it's not pretty, but it's functional and I've successfully built transcriptome indices with it! As more experts contribute to https://docs.google.com/spreadsheets/d/1Lbi68UP2dQtfp-KoxtXpE7jhCOxgP_FweGznbbOiMkw/edit?usp=sharing we can update the config file this PR adds to be able to support addtional organisms.

## Methods

The general method is explained above. However other scientific considerations that @jaclyn-taroni, @jashapiro, and @cgreene may want to be aware of are:
  * This does not address https://github.com/AlexsLemonade/refinebio/issues/1995, so our data model is not capable of keeping track of strains yet. I think this should be fine for now. When we later want to make them aware, we can use the mapping config file to figure out which strains have been assumed for all the organisms.
  * As we survey new experiments/samples that we've previously been unable to process, we won't keep track of their strains. I think this is less than ideal, but once we figure out a more comprehensive approach to #1995 I think this will be a small enough number of organisms that we should be able to migrate their samples without too much trouble.

## Types of changes

## Functional tests

- [x] Added unit tests which survey an organism from both the Fungi and Bacteria divisions of ensembl.
- [x] Ran the downloader/processor jobs for an organism from both of those divisions to make sure that there's not an issue once we get to those steps.
- [x] Existing unit tests are passing
- [x] Rerun a full resurveying of all divisions to ensure that there's not any edge cases that we'll break on. I won't let the downloader/processor jobs run for these. I just want to check that the surveying is still good.

## Checklist


jaclyn-taroni:
The science bits of this seem reasonable to me 👍  I'll note that when we're thinking about supporting strains (#1995) we may essentially end up with two strains for a microbial _sample_: 1) the strain from which the sample originates and 2) the strain that we used for quantification (e.g., building the txome index). And further down the line -- I can imagine a situation where we'd want to process each sample twice (1 - actual strain, 2 - 'blessed' strain for compendium purposes) for some organisms. This is somewhat different from what I understand is captured on #1995.

PR #2050, opened 2020-01-07
use _names suffix when only returning organisms' name

davidsmejia:
## Issue Number

https://github.com/AlexsLemonade/refinebio-frontend/pull/825

We tend to interchangeably use `organisms` and `organism_names`

This PR intends on making the use of `organism_name` or `organism_names` when the attribute does not contain a serialized model but instead contains only a string of the name. ie `HOMO_SAPIENS`

## Types of changes

## Functional tests

Loading experiments locally

## Checklist

### DIFF
--- a/api/data_refinery_api/serializers.py
+++ b/api/data_refinery_api/serializers.py
@@ -766,8 +766,7 @@ class APITokenSerializer(serializers.ModelSerializer):
 
 class CompendiumResultSerializer(serializers.ModelSerializer):
     primary_organism_name = serializers.StringRelatedField(
-        read_only=True,
-        source="primary_organism"
+        read_only=True, source="primary_organism"
     )
     organism_names = serializers.StringRelatedField(many=True, source="organisms", read_only=True)
     computed_file = ComputedFileSerializer(source="get_computed_file", read_only=True)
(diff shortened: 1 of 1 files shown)


api/data_refinery_api/serializers.py
```diff
@@ -447,7 +447,7 @@ class DetailedExperimentSerializer(serializers.ModelSerializer):
     annotations = ExperimentAnnotationSerializer(many=True, source="experimentannotation_set")
     samples = DetailedExperimentSampleSerializer(many=True)
     sample_metadata = serializers.ReadOnlyField(source="sample_metadata_fields")
-    organisms = serializers.StringRelatedField(many=True, read_only=True)
+    organism_names = serializers.StringRelatedField(many=True, source="organisms", read_only=True)
```
kurtwheeler: I would expect this endpoint to return the full Organism model instead of just the name.

davidsmejia: This would mean we would need to change the search document serializer to also return the object as well.

davidsmejia: https://github.com/AlexsLemonade/refinebio/issues/2055

davidsmejia:
Oh right on, will fix before merging!

Issue #89, opened 2018-01-17
Detect Agilent one color vs. two color experiments


jaclyn-taroni:
One color and two color experiments will be the same "platform" in GEO or ArrayExpress -- there should be some metadata field that indicates whether a second channel is present

jaclyn-taroni:
I think looking at the _protocol_ metadata is likely to be a good way to proceed. I suggest we might randomly select tens of experiments from the GEO platform GPL4133 & take a look at the metadata.

Miserlou:
By looking at the protocols we can see the presence of both 'Cy3' and 'Cy5' strings:
https://www.ebi.ac.uk/arrayexpress/experiments/E-MTAB-4636/protocols/

Although it is not present in the API metadata:
https://www.ebi.ac.uk/arrayexpress/json/v3/experiments/E-MTAB-4636/

Miserlou:
A better example (previous one wasn't Agilent): https://www.ebi.ac.uk/arrayexpress/experiments/E-GEOD-77820

Still holds for Cy3/5 criteria.

Miserlou:
Protocol has an API endpoint so we don't need to scrape: https://www.ebi.ac.uk/arrayexpress/json/v3/experiments/E-MTAB-4636/protocols

Miserlou:
Verdict is that we should snarf basically the entire `/protocols` API response because it might be scientifically useful. So, related: #96

kurtwheeler:
To clarify, if we detect that a sample is Agilent 1 Color then we should just grab the user-submitted processed data, if it is 2 Color then we should grab the raw data and process it with SCAN.UPC.

kurtwheeler:
Ok so it looks like the next thing to be done for this is to go through that list of accessions, pull the protocol for each experiment from the above listed protocols endpoint, and determine if any are Agilent 1-color. I believe the heuristic for determining 1-color vs 2-color is that if "Cy5" appears anywhere in the protocol information then it is a 2-color experiment.

@Miserlou or @jaclyn-taroni can you confirm if all of the above is accurate?

jaclyn-taroni:
Yep, there should also be some fields that reference "Channel 2" (often ending in `_ch2` in GEO) that are non-empty in 2-color experiments.

dongbohu:
Here is the statistics of the 756 experiments:

285 IDs have both Cy3 and Cy5:
--------------------------------------------
[...]

Issue #87, opened 2018-01-17
Processor for Agilent 2-color data


jaclyn-taroni:
For Agilent two color experiments that have raw data available, we can use [`SCAN.UPC::SCAN_TwoColor`](https://bioconductor.org/packages/release/bioc/manuals/SCAN.UPC/man/SCAN.UPC.pdf#Rfn.SCAN.Rul.TwoColor.1) for processing. This gives us the output of the two channels separately.

 Would be great to run a handful of experiments through a processor with `SCAN_TwoColor` in development and gain some insight into the following:
* Are paired channels highly correlated (within a sample)? [Reported](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2375008/) to be of particular concern in probes with high GC content
* Does there appear to be strong dye bias (i.e., do the two channels cluster apart from one another)? (This "dye bias" could be driven by the use of a common reference like [Universal Human Reference](https://www.genomics.agilent.com/en/References-Controls/Universal-Reference-RNAs/?cid=AG-PT-172&tabId=AG-PR-1217).) Could any dye bias be mitigated by quantile normalizing (using Affymetrix reference) the channels separately? 

Looking a bit into the future -- could be useful to _automatically_ detect control/reference channel.

cc @srp33

kurtwheeler:
Assuming we can determine a method for differentiating between one and two channel samples, I will run `SCAN.UPC::SCAN_TwoColor` on 20 experiments overnight.

jaclyn-taroni:
Two color experiment example: https://www.ebi.ac.uk/arrayexpress/experiments/E-GEOD-35186/

jaclyn-taroni:
Now that we can process GSE35186 end to end thanks to @Miserlou (https://github.com/AlexsLemonade/refinebio/commit/179a2c6b1bd98768e22d197b4247dd920aa24f66) and @dongbohu has identified hundreds of two-color experiments, can we run 20 randomly selected experiments from the list of IDs with `Cy5` and `Cy3` on https://github.com/AlexsLemonade/refinebio/issues/89#issuecomment-381185928?

Miserlou:
Okay, here it is: https://s3.amazonaws.com/crunch-outputs/random_twocolor.tar.gz

jaclyn-taroni:
As of #212, SCAN_TwoColor is implemented but we have not fully evaluated it, so I am inclined to keep this open and not labeled for Keytar Kurt.

PR #2052, opened 2020-01-07
Remove all unused variables

arielsvn:
## Issue Number

N/A

Initially suggested in https://github.com/AlexsLemonade/refinebio/pull/2034#issuecomment-570360849

This uses `autoflake` to remove all unused variables in the project.

I executed the following commands 

```
$ autoflake -r --in-place --remove-unused-variables api/
$ autoflake -r --in-place --remove-unused-variables common/
$ autoflake -r --in-place --remove-unused-variables foreman 
$ autoflake -r --in-place --remove-unused-variables infrastructure 
$ autoflake -r --in-place --remove-unused-variables workers 
```

## Types of changes

## Functional tests

None

## Checklist

### DIFF
--- a/api/data_refinery_api/tests.py
+++ b/api/data_refinery_api/tests.py
@@ -628,7 +628,6 @@ class APITestCases(APITestCase):
         )
         self.assertEqual(response.json()["count"], 1)
 
-        Experiment.processed_public_objects
         self.assertEqual(len(experiment.processed_samples), 1)
 
         experiment.delete()
--- a/api/data_refinery_api/tests.py
+++ b/api/data_refinery_api/tests.py
@@ -1,5 +1,4 @@
 import json
-import random
 import time
 from unittest.mock import Mock, patch
 
@@ -629,7 +628,7 @@ class APITestCases(APITestCase):
         )
         self.assertEqual(response.json()["count"], 1)
 
-        qs = Experiment.processed_public_objects
+        Experiment.processed_public_objects
         self.assertEqual(len(experiment.processed_samples), 1)
 
         experiment.delete()
@@ -827,7 +826,6 @@ class APITestCases(APITestCase):
         result.commands.append("create_qn_target.py")
         result.is_ccdl = True
         result.is_public = True
-        processor_key = "QN_REFERENCE"
         result.processor = None
         result.save()
 
--- a/common/data_refinery_common/migrations/0045_svd_algorithm.py
+++ b/common/data_refinery_common/migrations/0045_svd_algorithm.py
@@ -3,7 +3,7 @@ from django.db import migrations, models
 
 def svd_algorithm_fix(apps, schema_editor):
     """ Fetch existing compendia and create compendium results"""
-    CompendiumResult = apps.get_model("data_refinery_common", "CompendiumResult")
+    apps.get_model("data_refinery_common", "CompendiumResult")
     ComputedFile = apps.get_model("data_refinery_common", "ComputedFile")
 
     compendium_computed_files = ComputedFile.objects.filter(
--- a/common/data_refinery_common/migrations/0045_svd_algorithm.py
+++ b/common/data_refinery_common/migrations/0045_svd_algorithm.py
@@ -3,7 +3,6 @@ from django.db import migrations, models
 
 def svd_algorithm_fix(apps, schema_editor):
     """ Fetch existing compendia and create compendium results"""
-    apps.get_model("data_refinery_common", "CompendiumResult")
     ComputedFile = apps.get_model("data_refinery_common", "ComputedFile")
 
     compendium_computed_files = ComputedFile.objects.filter(
--- a/foreman/data_refinery_foreman/foreman/management/commands/run_tximport.py
+++ b/foreman/data_refinery_foreman/foreman/management/commands/run_tximport.py
@@ -7,9 +7,6 @@ split experiments into multiple tximport jobs.
 """
 
 import random
-import sys
-import time
-from typing import Dict, List
 
 from django.core.management.base import BaseCommand
 from django.db.models import Count
@@ -77,7 +74,7 @@ def run_tximport():
                 # experiment will work. Tximport is somewhat special
                 # in that it doesn't actuallhy use original files so
                 # this is just used to point to the experiment.
-                samples = experiment.samples.all()
+                experiment.samples.all()
                 assoc.original_file = experiment.samples.all()[0].original_files.all()[0]
                 assoc.processor_job = processor_job
                 assoc.save()
--- a/foreman/data_refinery_foreman/foreman/management/commands/run_tximport.py
+++ b/foreman/data_refinery_foreman/foreman/management/commands/run_tximport.py
@@ -74,7 +74,6 @@ def run_tximport():
                 # experiment will work. Tximport is somewhat special
                 # in that it doesn't actuallhy use original files so
                 # this is just used to point to the experiment.
-                experiment.samples.all()
                 assoc.original_file = experiment.samples.all()[0].original_files.all()[0]
                 assoc.processor_job = processor_job
                 assoc.save()
--- a/foreman/data_refinery_foreman/foreman/management/commands/test_retry_samples.py
+++ b/foreman/data_refinery_foreman/foreman/management/commands/test_retry_samples.py
@@ -33,8 +33,6 @@ def setup_experiment() -> Dict:
 
     # Create the experiment
     experiment_accession = "SRP095529"
-    data_dir = "/home/user/data_store/"
-    data_dir + experiment_accession
     experiment = Experiment.objects.create(
         accession_code=experiment_accession, technology="RNA-SEQ"
     )
--- a/foreman/data_refinery_foreman/foreman/management/commands/test_retry_samples.py
+++ b/foreman/data_refinery_foreman/foreman/management/commands/test_retry_samples.py
@@ -1,8 +1,4 @@
-import datetime
-import json
-import os
-from typing import Dict, List
-from unittest.mock import MagicMock, Mock, call, patch
+from typing import Dict
 
 from django.test import TestCase
 from django.utils import timezone
@@ -38,7 +34,7 @@ def setup_experiment() -> Dict:
     # Create the experiment
     experiment_accession = "SRP095529"
     data_dir = "/home/user/data_store/"
-    experiment_dir = data_dir + experiment_accession
+    data_dir + experiment_accession
     experiment = Experiment.objects.create(
         accession_code=experiment_accession, technology="RNA-SEQ"
     )
--- a/foreman/data_refinery_foreman/foreman/test_main.py
+++ b/foreman/data_refinery_foreman/foreman/test_main.py
@@ -1,6 +1,5 @@
 import datetime
 import math
-import time
 from test.support import EnvironmentVarGuard  # Python >=3
 from unittest.mock import MagicMock, patch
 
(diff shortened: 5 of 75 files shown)


foreman/data_refinery_foreman/surveyor/geo.py
```diff
@@ -116,11 +115,11 @@ def set_platform_properties(
             # Example of it being a list: GSE113945
             channel1_temp = sample_metadata.get("label_protocol_ch1", "")
             if type(channel1_temp) != list:
```
cgreene: This looks like a bug. I imagine we were hoping to do something with channel1_protocol (otherwise, why are we doing any of this?).

api/data_refinery_api/tests.py
```diff
@@ -629,7 +628,7 @@ def test_processed_samples_only(self):
         )
         self.assertEqual(response.json()["count"], 1)
 
-        qs = Experiment.processed_public_objects
+        Experiment.processed_public_objects
```
cgreene: Bug? Looks like we would have intended to use this.

common/data_refinery_common/migrations/0045_svd_algorithm.py
```diff
@@ -3,7 +3,7 @@
 
 def svd_algorithm_fix(apps, schema_editor):
     """ Fetch existing compendia and create compendium results"""
-    CompendiumResult = apps.get_model("data_refinery_common", "CompendiumResult")
+    apps.get_model("data_refinery_common", "CompendiumResult")
```
cgreene: Either a bug or dead code?

foreman/data_refinery_foreman/foreman/management/commands/run_tximport.py
```diff
@@ -77,7 +74,7 @@ def run_tximport():
                 # experiment will work. Tximport is somewhat special
                 # in that it doesn't actuallhy use original files so
                 # this is just used to point to the experiment.
-                samples = experiment.samples.all()
+                experiment.samples.all()
```
cgreene: This was a dead query. Did we want it to do something?

foreman/data_refinery_foreman/foreman/management/commands/test_retry_samples.py
```diff
@@ -38,7 +34,7 @@ def setup_experiment() -> Dict:
     # Create the experiment
     experiment_accession = "SRP095529"
     data_dir = "/home/user/data_store/"
-    experiment_dir = data_dir + experiment_accession
+    data_dir + experiment_accession
```
cgreene: This isn't getting used. Also looks like a bug.

foreman/data_refinery_foreman/foreman/test_main.py
```diff
@@ -292,7 +291,7 @@ def mock_init_nomad(host, port=0, timeout=0):
         main.retry_lost_downloader_jobs()
         self.assertEqual(len(mock_send_job.mock_calls), 0)
 
-        jobs = DownloaderJob.objects.order_by("id")
+        DownloaderJob.objects.order_by("id")
```
cgreene: Dead code here (and later in this file) all looks safe to remove.

foreman/data_refinery_foreman/surveyor/test_harmony.py
```diff
@@ -215,7 +210,7 @@ def test_sdrf_big(self):
             if not metadata:
                 continue
             # No assertions, just making sure we don't barf.
-            harmonized = harmonize(metadata)
+            harmonize(metadata)
```
cgreene: This actually looks right

foreman/data_refinery_foreman/surveyor/test_harmony.py
```diff
@@ -270,8 +265,8 @@ def test_sra_lots(self):
         for accession in lots:
             try:
                 metadata = SraSurveyor.gather_all_metadata(accession)
-                harmonized = harmonize([metadata])
-            except UnsupportedDataTypeError as udte:
+                harmonize([metadata])
```
cgreene: On the other hand - not sure what this pattern is trying to do. Run harmonize but then we're ok with UnsupportedDataTypeError? Looks like not a great test pattern. Seems better to test one where we expect success and another where we assert that this error is raised.

foreman/data_refinery_foreman/surveyor/transcriptome_index.py
```diff
@@ -313,7 +313,7 @@ def _generate_files(self, species: Dict) -> None:
         fasta_download_url = url_builder.build_transcriptome_url()
         gtf_download_url = url_builder.build_gtf_url()
 
-        platform_accession_code = species.pop("division")
+        species.pop("division")
```
cgreene: It seems like we would have wanted this? Maybe bug?

workers/data_refinery_workers/processors/illumina.py
```diff
@@ -319,7 +317,7 @@ def _run_illumina(job_context: Dict) -> Dict:
     Expects a job_context which has been pre-populated with inputs, outputs
     and the column identifiers which the R script needs for processing.
     """
-    input_file_path = job_context["input_file_path"]
+    job_context["input_file_path"]
```
cgreene: It seems like we have a bug here or dead code.

workers/data_refinery_workers/processors/management/commands/run_tximport.py
```diff
@@ -67,6 +67,6 @@ def handle(self, *args, **options):
         pjofa.original_file = original_file
         pjofa.save()
 
-        final_context = tximport.tximport(job.id)
+        tximport.tximport(job.id)
```
cgreene: It seems like we'd probably want that final_context...

workers/data_refinery_workers/processors/test_salmon.py
```diff
@@ -236,7 +234,7 @@ def test_no_salmon_on_geo(self):
         sample_object.source_database = "GEO"
         sample_object.save()
 
-        job_context = salmon.salmon(job.pk)
+        salmon.salmon(job.pk)
```
cgreene: Job context goes unused through here. Probably ok.

workers/data_refinery_workers/processors/test_smasher.py
```diff
@@ -560,7 +559,7 @@ def test_no_smash_all_diff_species(self):
         experiment.accession_code = "GSE51084"
         experiment.save()
 
-        mus_mus = Organism.get_object_for_name("MUS_MUSCULUS")
+        Organism.get_object_for_name("MUS_MUSCULUS")
```
cgreene: I assume mus_mus was intended to be used here. Otherwise I can't see what this is helping to test.

foreman/data_refinery_foreman/surveyor/geo.py
arielsvn: looks like we could remove the whole `if...` since `channel1_temp` is also not used below

api/data_refinery_api/tests.py
arielsvn: I think we can safetly remove the whole line, since `qs` wasn't used here. And we don't need to test `Experiment.processed_public_objects`

common/data_refinery_common/migrations/0045_svd_algorithm.py
arielsvn: dead code, will remove the whole line

foreman/data_refinery_foreman/foreman/management/commands/run_tximport.py
arielsvn: looks like we wanted to use it below but added the whole value instead. Will remove the whole line

foreman/data_refinery_foreman/surveyor/test_harmony.py
arielsvn: I agree, I think we should add this on a different PR and leave this one only for unused variables.

foreman/data_refinery_foreman/surveyor/transcriptome_index.py
arielsvn: hmm not sure why we do this, looks like `species` is no longer used in this method after this.

workers/data_refinery_workers/processors/management/commands/run_tximport.py
arielsvn: for what? doesn't look like is needed

workers/data_refinery_workers/processors/illumina.py
arielsvn: dead code, will remove the whole line

workers/data_refinery_workers/processors/test_smasher.py
arielsvn: `Organism.get_object_for_name` also creates the organism in the db. Still it looks safe to delete the whole line since the organism `MUS_MUSCULUS` is not used after this.

arielsvn:
@cgreene thanks for the review, I addressed all your comments, and in some cases removed the whole line when the variable values didn't do anything.

@davidsmejia @kurtwheeler This should be ready for review now.

foreman/data_refinery_foreman/surveyor/geo.py
kurtwheeler: If I had to guess I'd guess it had to with https://github.com/AlexsLemonade/refinebio/issues/87 or https://github.com/AlexsLemonade/refinebio/issues/89

foreman/data_refinery_foreman/surveyor/transcriptome_index.py
kurtwheeler: Hmm yeah this is odd, not sure why we'd call this platform_accession_code. It's also not used after the method either. Probably safe to remove the line entirely.

workers/data_refinery_workers/processors/management/commands/run_tximport.py
kurtwheeler: I think most of instances of `final_context` were relics from Rich functional testing his code. Occasionally they'd be used in unit tests as well, but not always.

foreman/data_refinery_foreman/surveyor/transcriptome_index.py
arielsvn: and also the call below to `self._clean_metadata(species)`? Since `species` is no longer used here.

kurtwheeler: Yes. My guess is we used more things from the species dict for the original file metadata before but now that's handled by the URL builder classes.

arielsvn: ok removed `self._clean_metadata`. Will let the tests pass before merging.

davidsmejia:
This looks really great. One question/concern though, would it be better to only remove unused imports and not remove unused variables in this PR. Looks like the tests will fail if an error is raised which is good but could be improved to make sure things aren't silently broken by checking that the functions/methods that return say a `job_context` is actually a `job_context` or if it returns a `processor_id` that the return value looks like a `processor_id`?

arielsvn:
> One question/concern though, would it be better to only remove unused imports and not remove unused variables in this PR.

I think there's value in removing unused variables as well, there're many that were removed that weren't doing anything.

> Looks like the tests will fail if an error is raised which is good but could be improved to make sure things aren't silently broken by checking that the functions/methods that return say a job_context is actually a job_context or if it returns a processor_id that the return value looks like a processor_id?

This would be also increasing the scope of this PR, but as you said is a good opportunity to also improve our tests. I was thinking just checking that the return values were not null. Is there anything else that you think we should assert @davidsmejia?

arielsvn:
Added additional assertions on https://github.com/AlexsLemonade/refinebio/pull/2052/commits/172abfc16f8842e3861976e0b69f16312569e3f7. @davidsmejia let me know if you think we should add any other.

foreman/data_refinery_foreman/surveyor/test_harmony.py
```diff
@@ -210,7 +210,8 @@ def test_sdrf_big(self):
             if not metadata:
                 continue
             # No assertions, just making sure we don't barf.
```
davidsmejia: This comment can go!

Issue #1878, opened 2019-11-08
Mock 3rd party API responses in the tests


arielsvn:
### Context

Our tests depend on several 3rd party APIs, sometimes they fail when those services are down.

### Problem or idea

Running the tests multiple times should give the same result. Sometimes I see failed tests and I'm not sure if it's because there's something wrong with our code, or because some service is down.

```
2019-11-08 18:27:57,805 local [volume: 0] data_refinery_foreman.surveyor.external_source INFO [downloaded_urls: ['second_url', 'third_url', 'fourth_url']] [survey_job: 26] [downloader_job: 10308]: Queuing downloader job.
....2019-11-08 18:30:48,214 local [volume: 0] data_refinery_foreman.surveyor.harmony ERROR [response_code: 404]: Unable to fetch URL: https://www.ebi.ac.uk/arrayexpress/files/E-GEOD-7753/E-GEOD-7753.sdrf.txt
...Too long with no output (exceeded 10m0s)
```

```
======================================================================
FAIL: test_geo_celgz_redownloading (data_refinery_foreman.surveyor.test_end_to_end.GeoCelgzRedownloadingTestCase)
Survey, download, then process an experiment we know is Affymetrix.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/user/data_refinery_foreman/surveyor/test_end_to_end.py", line 425, in test_geo_celgz_redownloading
    self.assertTrue(recreated_job.success)
AssertionError: None is not true
```

```
2019-11-08 19:36:45,327 local [volume: 0] data_refinery_foreman.surveyor.external_source ERROR [survey_job: 1]: Exception caught while discovering samples. Terminating survey job.
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/requests/adapters.py", line 449, in send
    timeout=timeout
  File "/usr/local/lib/python3.5/dist-packages/urllib3/connectionpool.py", line 731, in urlopen
    body_pos=body_pos, **response_kw)
  File "/usr/local/lib/python3.5/dist-packages/urllib3/connectionpool.py", line 731, in urlopen
    body_pos=body_pos, **response_kw)
  File "/usr/local/lib/python3.5/dist-packages/urllib3/connectionpool.py", line 731, in urlopen
    body_pos=body_pos, **response_kw)
  File "/usr/local/lib/python3.5/dist-packages/urllib3/connectionpool.py", line 711, in urlopen
[...]

PR #2053, opened 2020-01-08
Kurtwheeler/api tests no internet

kurtwheeler:
## Issue Number

#1878 

This PR makes it so the API tests don't hit external APIs so that our tests' success isn't dependent on them being up. This is the first of a few PRs to get us to the point where we don't hit external APIs from our test suite.

The API sub-project was easy because there were just a few places where the API made external calls: when initializing organism models for the tests. I just changed the way we initialized these to not required hitting NCBI's API and it worked!

## Types of changes

## Functional tests

I turned off my internet and ran the API tests.

## Checklist


arielsvn:
I like this way of creating organisms! Is it possible to abstract it in some way? Maybe have a test helper file somewhere with a `create_organism` function?

```
ailuropoda = Organism(name="AILUROPODA_MELANOLEUCA", taxonomy_id=9646, is_scientific_name=True)
ailuropoda.save()
```

Looks like we are repeating this logic in multiple places.

davidsmejia:
🔥 🔥 🔥

scripts/common.sh
```diff
@@ -8,7 +8,12 @@ export CCDL_WORKER_IMAGES="smasher compendia illumina affymetrix salmon transcri
...
     if [ "$(uname)" = "Linux" ]; then
-        ip route get 8.8.8.8 | grep -oE 'src ([0-9]{1,3}\.){3}[0-9]{1,3}' | awk '{print $2; exit}'
+        # If we're not connected to the internet return an error code.
+        ip_out=$(ip route get 8.8.8.8)
+        if [ $? -ne 0 ]; then
+            return 1
+        fi
+        echo $ip_out | grep -oE 'src ([0-9]{1,3}\.){3}[0-9]{1,3}' | awk '{print $2; exit}'
     elif [ "$(uname)" = 'Darwin' ]; then # MacOS
         ifconfig | grep "inet " | grep -v 127.0.0.1 | cut -d\  -f2 | tail -1
```
davidsmejia:
```suggestion
         ip_out=$(ifconfig | grep "inet " | grep -v 127.0.0.1 | cut -d\  -f2 | tail -1)
         if [ -z "$ip_out" ]; then
                 return 1
         fi
         echo $ip_out 

```

davidsmejia: Might want to have this also work on 🍎

kurtwheeler:
@arielsvn I'm not sure about that one. This is creation of test data, which is something we tend to repeat a lot... It's also not clear what the function would do for taxonomy ids. If you feel strongly about this one we can discuss it more but I feel like I'm generally okay with keeping test code a bit DAMP.

arielsvn:
Yeah let's talk it, I'm not talking about removing repetition but making it more descriptive. I would prefer if we could have `test_utils.create_organism()`, instead of the whole to logic create the organism. From the tests it looks like we only care about the organism, the actual name and taxonomy id are important?

PR #2058, opened 2020-01-09
More checks on pre-commit

arielsvn:
## Issue Number

N/A

Adds a few other pre-commit hooks for several useful things:

- block commits on dev and master (I have made them by mistake)
- Remove trailing whitespaces
- Fix end of line
- Fail if files larger than 500kb are added
- Check merge conflicts
- Check file docstrings
- check yaml syntax

These are all of the pre-commit checks that I wanted to add.

## Types of changes

## Functional tests

None

## Checklist


arielsvn:
Some of these are failing, marked as WIP.

.pre-commit-config.yaml
```diff
@@ -1,12 +1,30 @@
...
+      - id: check-docstring-first
+      - id: check-yaml
 
   - repo: https://github.com/pre-commit/mirrors-isort
     rev: v4.3.21
     hooks:
       - id: isort
+
+  - repo: https://github.com/psf/black
+    rev: stable
```
kurtwheeler: Can we actually pin this to a specific version so we don't have to worry about someone getting malicious code into this repo?

.pre-commit-config.yaml
```diff
@@ -1,12 +1,30 @@
...
-      - id: black
-        args: [--line-length=100]
+      - id: no-commit-to-branch
+        args: [--branch, dev, --branch, master]
+      - id: end-of-file-fixer
+      - id: mixed-line-ending
+        args: [--fix=lf]
+      - id: trailing-whitespace
+        args: [--markdown-linebreak-ext=md]
+      - id: check-added-large-files
```
kurtwheeler: I think we should nix this!

common/data_refinery_common/models/models.py
```diff
@@ -1524,13 +1523,6 @@ def terms_and_conditions(self):
         return settings.TERMS_AND_CONDITIONS
 
 
-"""
-# Associations
-
-These represent the relationships between items in the other tables.
-"""
-
-
```
kurtwheeler: I don't like these changes. These docstring denoted different parts of the file.

We should probably handle this issue by breaking this file up into different namespaces, but just moving all these docstrings to the top doesn't really make much sense.

arielsvn: yeah, I wasn't sure what to do with that rule. I'll revert these and comment it until we simplify `models.py`

arielsvn:
Addressed your comments @kurtwheeler, let me know if you see something else.

PR #2059, opened 2020-01-09
fix slack webhook on slackpost.sh script

arielsvn:
## Issue Number

N/A

We have a notification on slack when a deploy finishes successfully that it wasn't working since we updated the slack webhooks.

This fixes the url.

## Types of changes

## Functional tests

Tested script and saw notification on #robots

## Checklist


cgreene:
> We recently discovered a publicly accessible incoming webhook associated with your Slack workspace, alexslemonade.slack.com. This can happen when a webhook URL is posted on a public site, such as a code-sharing forum. None of your data is at risk — webhooks can only send messages into Slack and cannot access any data.
> 
> Though your data is not at risk of being exposed through a webhook, we've invalidated the publicly exposed webhook URL and generated a new one on your behalf. We realize that this may cause some disruption for you or your team; however, protecting your team's security and privacy is important to us. Here is the webhook we found (with its previous URL):

cgreene:
We should probably make these types of URLs and contact email addresses configurable. If someone deployed their own instance of refine.bio they probably would want the information to go to them and not us.

arielsvn:
We moved the URL to an environment variable in circle ci.

cgreene:
Is there documentation somewhere of what environment variables one needs to set on Circle CI to run the deploy process/etc? Just curious - not to make for this PR. If not, is this something we should have docs for? If so, that might be a good ticket.

arielsvn:
We have some documentation on the circleci environment variables on https://github.com/AlexsLemonade/refinebio/blob/dev/infrastructure/deploy_box_instance_data.sh. Updated them now :)

arielsvn:
Good catch @kurtwheeler, added a check now

PR #2060, opened 2020-01-09
Fixes the order in which we run parts of the smasher pipeline 

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio/pull/2014

Deepa noticed her datasets failed this morning.

We were deleting the dataset before calculating its size. This switches it so we calculate the size of the dataset before deleting it.

This has affected datasets since the 4th of January. I will rerun those jobs once this change is deployed.

## Types of changes

## Functional tests

This stuff won't even happen locally, which is probably why it never came up before. We should probably improve these tests more but right now I wanna get this bug fixed.

### DIFF
--- a/api/data_refinery_api/views.py
+++ b/api/data_refinery_api/views.py
@@ -744,7 +744,9 @@ class SampleList(generics.ListAPIView):
         """
         queryset = (
             Sample.public_objects.prefetch_related("organism")
-            .prefetch_related(Prefetch("results", queryset=ComputationalResult.objects.order_by('time_start')))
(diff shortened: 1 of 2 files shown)


arielsvn:
On `_update_result_objects` we mark the dataset as processed, what if we fail to `upload` afterward?

kurtwheeler:
Good call. We should probably instead move the delete to after the update function.

kurtwheeler:
Tests aren't passing :(
I'll fix them after lunch/my meeting with Josh.

PR #2065, opened 2020-01-10
Add migration to remove invalid tximport associations

### LINKED ISSUE #2054: Processing info: Sample is not available but only tximport is run on it
Ex:
https://www.refine.bio/experiments/SRP074593/gene-expression-profiling-of-drosophila-s2r-cells-following-rnai-mediated-knockdown-of-transcription-factors

Screen Shot 2020-01-03 at 2 58 39 PM

We should just show the submitter supplied info, since the samples aren't really processed through our pipeline

arielsvn:
## Issue Number

close #2054 

Removes associations between tximport computational results and samples that are not processed.

## Types of changes

## Functional tests

Ran migrations locally

## Checklist

### DIFF
--- a/common/data_refinery_common/migrations/0052_remove_invalid_tximport_associations.py
+++ b/common/data_refinery_common/migrations/0052_remove_invalid_tximport_associations.py
@@ -1,5 +1,4 @@
-from django.db import migrations, models
-from django.db.models import Count, F
+from django.db import migrations
 
 
 def remove_tximport_invalid_associations(apps, schema_editor):
--- a/common/data_refinery_common/migrations/0052_remove_invalid_tximport_associations.py
+++ b/common/data_refinery_common/migrations/0052_remove_invalid_tximport_associations.py
@@ -1,0 +1,24 @@
+from django.db import migrations, models
+from django.db.models import Count, F
+
+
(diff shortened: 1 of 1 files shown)


common/data_refinery_common/migrations/0052_remove_invalid_tximport_associations.py
```diff
@@ -0,0 +1,24 @@
+from django.db import migrations, models
```
davidsmejia:
```suggestion
from django.db import migrations
```

common/data_refinery_common/migrations/0052_remove_invalid_tximport_associations.py
```diff
@@ -0,0 +1,24 @@
...
+from django.db.models import Count, F
```
davidsmejia:
```suggestion

```

arielsvn: removed, thanks!

arielsvn:
This will remove `11671` associations in total:

```
>>> SampleResultAssociation.objects.filter(sample__is_processed=False, result__processor__name__iexact="tximport").count()
11671
```

Issue #2031, opened 2019-12-31
Re-queueing samples from GEO are not being extracted


davidsmejia:
### Context

Can't process this sample [GSM708631](https://www.refine.bio/samples/GSM708631) with the error `Failed to extract any downloaded files.`

As far as I can tell this is only occuring with microarray/affy processor jobs.
#1986 
https://github.com/AlexsLemonade/refinebio/issues/1538#issuecomment-530124920

Trying to re-requeue sample via [requeue_samples](https://github.com/AlexsLemonade/refinebio/blob/996bcafaa1077ad9697918f06c9578279fb825ba/foreman/data_refinery_foreman/foreman/management/commands/retry_samples.py#L24) which is erroring out [here](https://github.com/AlexsLemonade/refinebio/blob/dev/workers/data_refinery_workers/downloaders/geo.py#L393).

### Problem or idea

The geo downloader correctly identifies that the original file for this sample exists but I believe that it incorrectly returns `sample.needs_processing() -> false` because
 `sample.has_blocking_jobs() -> true`

So original file is downloaded, geo downloaded sees that it exists, tries to see if it should be processed, says no, skips it in the results set, then says there was nothing to extract....

### Solution or next step

Either we are not checking existing jobs correctly [here](https://github.com/AlexsLemonade/refinebio/blob/dev/common/data_refinery_common/models/models.py#L840) or we are not correctly setting one of the checked attributes when jobs fail. (In the issue that lead me here it most likely would be somewhere around [here](https://github.com/AlexsLemonade/refinebio/blob/dev/workers/data_refinery_workers/processors/array_express.py#L178).

We should dig a little deeper and find out which one and either fix the check or run a migration to fix historical errors.

kurtwheeler:
As part of verifying this fix, please test by rerunning the samples affected by https://github.com/AlexsLemonade/refinebio/issues/1986

PR #2073, opened 2020-01-15
Only prevent retry of jobs if no_retry job was current version.

davidsmejia:
## Issue Number

#2031 

It was impossible to retry samples where a potential fix was created. This PR adresses that.

## Methods

`OriginalFile::has_blocking_jobs`

When checking if an original file has blocking jobs we first check if there are processor jobs where `no_retry=True` and  exclude where `abort=True`

However if we think that the error in processing that sample/original file has been resolved this prevents us from being able to queue up that file again. So this PR adds checking against the current version (same as `processor_job_instance.worker_version`) so the job will not be retried unless their has been a new deploy.

## Types of changes

## Functional tests

None.

## Checklist

### DIFF
--- a/common/data_refinery_common/models/models.py
+++ b/common/data_refinery_common/models/models.py
@@ -33,6 +33,9 @@ LOCAL_ROOT_DIR = get_env_variable("LOCAL_ROOT_DIR", "/home/user/data_store")
 # we need something with the pattern: 'salmon X.X.X'
 CURRENT_SALMON_VERSION = "salmon " + get_env_variable("SALMON_VERSION", "0.13.1")
 CHUNK_SIZE = 1024 * 256  # chunk_size is in bytes
+# Let this fail if SYSTEM_VERSION is unset.
+SYSTEM_VERSION = get_env_variable("SYSTEM_VERSION")
+
 
 """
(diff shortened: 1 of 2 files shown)


common/data_refinery_common/models/models.py
```diff
@@ -916,7 +918,9 @@ def has_blocking_jobs(self, own_processor_id=None) -> bool:
         # If the file has a processor job that should not have been
         # retried, then it still shouldn't be retried.
         # Exclude the ones that were aborted.
-        no_retry_processor_jobs = self.processor_jobs.filter(no_retry=True).exclude(abort=True)
+        no_retry_processor_jobs = self.processor_jobs.filter(
+            no_retry=True, worker_version=SYSTEM_VERSION
+        ).exclude(abort=True)
```
arielsvn: I'm wondering if we should also remove `exclude(abort=True)` which was an attempt to achieve the same?

davidsmejia: If the version stays the same ie the job is retried without a re-deploy then excluding `abort=True` would return 0 which to would mean the job is retried. This code only says to respect `no_retry` for the current version.

PR #2074, opened 2020-01-15
Enable cloudtrail logging for S3 compendia bucket and others

### LINKED ISSUE #2043: Set up AWS CloudTrail to get more detailed info on compendia downloads
### Context

We set up google analytics for our compendia page. However this doesn't tell us much more than the fact that someone started downloading the compendia.

### Problem or idea

It'd be nice to know if someone finished downloading it, where they were from, and anything else we can.

### Solution or next step

Use AWS CloudTrail to track as much info about compendia downloads as possible.

arielsvn:
## Issue Number

close #2043 

This should add the configuration to start logging read events for the buckets `data_refinery_compendia_bucket`, `data_refinery_transcriptome_index_bucket`, `data_refinery_qn_target_bucket`.

All events are saved in S3 to the bucket `data_refinery_bucket` and also they are pushed to cloudfront. To a stream named [`589864003899_CloudTrail_us-east-1`](https://console.aws.amazon.com/cloudwatch/home?region=us-east-1#logEventViewer:group=data-refinery-log-group-arielsvn-dev;stream=589864003899_CloudTrail_us-east-1) I wasn't able to find a way to rename this stream. We could alternatively create a different group for cloudtrail events?

@kurtwheeler let me know what do you think, spent some time with the IAM roles, not sure if that's the right way to have them.

## Types of changes

## Functional tests

Created a dev environment and made sure that the trail configuration was correct and that we were logging to cloudfront in the right log group.

## Checklist

## Screenshots

image

[created with [blast-radius](https://github.com/28mm/blast-radius)]

### DIFF
--- a/infrastructure/deploy.sh
+++ b/infrastructure/deploy.sh
@@ -281,7 +281,7 @@ docker run \
 
 # Make sure to clear out any old nomad job specifications since we
 # will register everything in this directory.
-rm -r nomad-job-specs
+rm -rf nomad-job-specs
 
 # Template the environment variables for production into the Nomad Job
 # specs and API confs.
--- a/infrastructure/logging.tf
+++ b/infrastructure/logging.tf
@@ -6,8 +6,11 @@ resource "aws_cloudtrail" "data_refinery_s3_cloudtrail" {
   include_global_service_events = false
   depends_on                    = ["aws_s3_bucket.data_refinery_bucket"]
 
+  cloud_watch_logs_group_arn    = "${aws_cloudwatch_log_group.data_refinery_log_group.arn}"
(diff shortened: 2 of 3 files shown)


arielsvn:
> Why are we logging to both cloudwatch and S3? I would think just cloudwatch would be sufficient.

I wondered the same... It's required for cloudtrail, maybe we can find a way around it?

> What kinds of events are going to be logged? Looks like the only filter is include_management_events = false, does that mean we'll get other non-management related events? We really only care about GetObject events as far as I can tell.

Cloudtrail doesn't seem to offer a lot of control on what get's logged, [the docs](https://www.terraform.io/docs/providers/aws/r/cloudtrail.html) only list that option.

kurtwheeler:
We're going to close this in favor of doing this with CloudWatch metrics instead of CloudTrail.

Issue #2076, opened 2020-01-16
SRA downloader does not handle ENA's FTP server being down correctly


kurtwheeler:
### Context

`.SRA` files with unmated reads break our salmon processor. We've therefore implemented a workaround where we check ENA's FTP server to see how many files they have. This works because if there are unmated reads then ENA will have three files, two for the paired reads and one for the unmated reads.

### Problem or idea

The whole check for the unmated reads is done in a [try-catch block](https://github.com/AlexsLemonade/refinebio/blob/b303b9789656cde7465dca20f2537fa1e04dc6bc/workers/data_refinery_workers/downloaders/sra.py#L238) so if the server is down we'll catch it and just return False, meaning that we don't think the sample has unmated reads (even if it does). This means that for samples that have unmated reads we will later fail to run salmon on the `.sra` files.

This also breaks our end-to-end tests for SRA because when the FTP server is down the expected behavior doesn't happen and the tests fail.

### Solution or next step

There are two parts of the solution for this problem:
  1. Before checking for unmated reads, first check if the FTP server is even up. Fail SRA downloader jobs if it's not.
  2. Change the end-to-end SRA tests so that they expect the downloader jobs to fail if the FTP server is down. This will prevent our tests from failing and actually will be a good check that we handle that scenario gracefully.

PR #2078, opened 2020-01-16
Minimize the dependence on external services of end-to-end tests

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/1878

We're trying to make our tests less likely to fail when external services that we rely upon are down. This PR improves the end-to-end tests by using `pyvcr` for API responses and by switching the `source_url` to S3 for file downloads. Unfortunately it's not so easy to mock out the FTP and aspera server calls for the end-to-end tests. #2076 deals with the FTP servers. I'm currently hoping that the aspera servers are just reliable enough to not cause an issue. We'll see if that's the case though.

Note that this does not work across the full foreman test suite yet. This is just for end-to-end tests. I have other branches in progress for using `vcr` in more places and for using local files for downloads.

## Types of changes

- Improvement

## Functional tests

I'm only changing test code.

## Checklist


kurtwheeler:
Still working on getting circelci tests to pass. May need a couple more iterations, but I think things work locally.

foreman/data_refinery_foreman/surveyor/test_end_to_end.py
```diff
@@ -38,10 +39,99 @@
...
+
+def build_surveyor_init_mock(source_type):
+    if source_type == "ARRAY_EXPRESS":
+        original_surveyor = ORIGINAL_ARRAY_EXPRESS_SURVEYOR
+    if source_type == "SRA":
+        original_surveyor = ORIGINAL_SRA_SURVEYOR
+    if source_type == "TRANSCRIPTOME_INDEX":
+        original_surveyor = ORIGINAL_TRANSCRIPTOME_SURVEYOR
+    if source_type == "GEO":
+        original_surveyor = ORIGINAL_GEO_SURVEYOR
```
arielsvn: Should we throw an error if an unknown `source_type` is provided?

foreman/data_refinery_foreman/surveyor/test_end_to_end.py
```diff
@@ -38,10 +39,99 @@
...
+
+    def mock_init_surveyor(survey_job):
+        ret_value = original_surveyor(survey_job)
+
+        def mock_queue_downloader_job_for_original_files(
+            original_files, experiment_accession_code: str = None, is_transcriptome: bool = False,
+        ):
+            for original_file in original_files:
+                original_file.source_url = EXTERNAL_FILE_URL_MAPPING[original_file.source_url]
```
arielsvn: `EXTERNAL_FILE_URL_MAPPING[original_file.source_url]` will probably fail if we a new url is found. Can we throw a more meaningful error? And maybe point the developer to some instructions on how to fix it?

foreman/data_refinery_foreman/surveyor/test_end_to_end.py
```diff
@@ -69,10 +159,17 @@ def wait_for_job(job, job_class: type, start_time: datetime, loop_time: int = No
 # job in the database because it'd be stuck in a transaction.
 class NoOpEndToEndTestCase(TransactionTestCase):
     @tag("slow")
-    def test_no_op(self):
+    @vcr.use_cassette(
+        "/home/user/data_store/cassettes/surveyor.test_end_to_end.no_op.yaml",
```
arielsvn: what about using a constant or something else to store the cassettes path? `/home/user/data_store/cassettes/`

foreman/data_refinery_foreman/surveyor/test_end_to_end.py
```diff
@@ -69,10 +159,17 @@ def wait_for_job(job, job_class: type, start_time: datetime, loop_time: int = No
...
+        ignore_hosts=["nomad"],
+    )
+    @patch("data_refinery_foreman.surveyor.surveyor.ArrayExpressSurveyor")
```
arielsvn: What about having an additional decorator that does the `@patch` with the side effects below? It would be nice if we could do `@patch_surveyor("ARRAY_EXPRESS")` at the top of the tests. We can also explore this down the road.

foreman/data_refinery_foreman/surveyor/test_end_to_end.py
kurtwheeler: An error will be thrown, because `original_surveyor` will be undefined. Do you think that won't be specific enough? This is test code so presumably that would only happen if a developer was modifying the tests and wrote code that expected the source_type to be in here and it wasn't.

foreman/data_refinery_foreman/surveyor/test_end_to_end.py
kurtwheeler: Good call :+1:

foreman/data_refinery_foreman/surveyor/test_end_to_end.py
kurtwheeler: Oh yeah good call! When I originally did it like this I was planning on it but forgot.

foreman/data_refinery_foreman/surveyor/test_end_to_end.py
kurtwheeler: Mmm yeah, that's not a bad idea but I think it could get pretty complicated because it'd be a wrapper around a wrapper. I'd rather not shoehorn it into this PR.

PR #2079, opened 2020-01-16
Add command to transfer compendias to the right bucket

arielsvn:
## Issue Number

#2072 

This adds a new command that will transfer all compendias/quantpendias to the bucket `data-refinery-s3-compendia-circleci-prod`

This should be merged after #2077, as it relies in one of the environment variables declared there.

## Types of changes

## Functional tests

None, thinking of testing it on staging where we have a couple of compendias

## Checklist


foreman/data_refinery_foreman/foreman/management/commands/fix_compendia.py
```diff
@@ -0,0 +1,58 @@
...
+    computed_file.save()
+
+    # 3. Remove file from original bucket
+    S3.delete_object(Bucket=S3_BUCKET_NAME, Key=s3_key)
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **options):
+        compendia_files = ComputedFile.objects.filter(
+            result__compendium_result__isnull=False, s3_bucket=S3_BUCKET_NAME
```
kurtwheeler: Instead of checking for compendia that are in the wrong bucket, can we check for compendia that are not in the right bucket?
```suggestion
            result__compendium_result__isnull=False
        ).exclude(s3_bucket__=S3_COMPENDIA_BUCKET_NAME)
```

foreman/data_refinery_foreman/foreman/management/commands/fix_compendia.py
```diff
@@ -0,0 +1,58 @@
...
+        s3_key,
+        ExtraArgs={"ACL": "public-read", "StorageClass": "STANDARD_IA"},
+    )
+
+    # 2. update computed file
+    computed_file.s3_bucket = S3_COMPENDIA_BUCKET_NAME
+    computed_file.save()
+
+    # 3. Remove file from original bucket
+    S3.delete_object(Bucket=S3_BUCKET_NAME, Key=s3_key)
```
kurtwheeler: When you set `copy_source` you don't assume that the bucket is `S3_BUCKET_NAME`, but here you do. Why not store the original bucket the file is in as `old_bucket` and then you can use that variable name here as well?

foreman/data_refinery_foreman/foreman/management/commands/fix_compendia.py
arielsvn: I like this, good idea!

arielsvn:
Addressed your comments @kurtwheeler, in case you want to take another look.

PR #2088, opened 2020-01-21
Use terraform variables for database and elsticsearch

arielsvn:
## Issue Number

N/A

Looks like we can use these variables instead of the actual values in terraform


arielsvn:
We talked about this, it's better to use the actual properties than the variables. Closing...

PR #2093, opened 2020-01-22
[Master] Deploy

arielsvn:
We need to run the command `fix_compendia` in prod to move compendiums to the right bucket. #2079


davidsmejia:
merged in #2073 also 👊

Issue #2087, opened 2020-01-21
Run fungi and bacteria for supported strains


kurtwheeler:
### Context

#1722 #2035  enabled us to process RNASeq samples from specific species/strains of bacteria and fungi. We haven't run any of them in production yet though.

### Problem or idea

We should run the organisms that we have strains selected for by expert opinions, which is every row of https://github.com/AlexsLemonade/refinebio/blob/dev/config/organism_strain_mapping.csv

### Solution or next step

Kick off surveyor jobs for the Fungi and Bacteria strains. Let them get processed. Look at the successes/failures and determine if there are additional issues that need to be filed or if we hit about the success rate we would expect given that there's some bad data out there in the world.

arielsvn:
I'm about to run the surveyor for those organisms. Looks like we have some samples processed for some of them:

```
organism PSEUDOMONAS_AERUGINOSA has 1127 samples (992 processed)
organism CANDIDA_ALBICANS has 354 samples (0 processed)
organism ESCHERICHIA_COLI has 1143 samples (1009 processed)
```

<summary>

```py
for name in ["PSEUDOMONAS_AERUGINOSA", "CANDIDA_ALBICANS", "ESCHERICHIA_COLI"]:
    samples = Organism.objects.get(name=name).sample_set.all()
    print(
        "organism {0} has {1} samples ({2} processed)".format(
            name, samples.count(), samples.filter(is_processed=True).count()
        )
    )
```

</summary>

arielsvn:
I ran the following command

```
nomad job dispatch -meta FILE=https://data-refinery-test-assets.s3.amazonaws.com/survey-ensembl.txt SURVEYOR_DISPATCHER
```

And `survey-ensembl.txt` content is:

```
PSEUDOMONAS_AERUGINOSA, EnsemblBacteria
CANDIDA_ALBICANS, EnsemblFungi
ESCHERICHIA_COLI, EnsemblBacteria
```

arielsvn:
That failed :/ Looks like there's something strange happening with the organism names, saw these errors on the logs:

```
Unable to retrieve NCBI taxonomy ID number for organism with name: ESCHERICHIA_COLI_GCA_001012635

ERROR: Unable to retrieve NCBI taxonomy ID number for organism with name: CANDIDA_ALBICANS_SC5314_GCA_000784635
```

<details><summary>all logs</summary>

```
[...]

PR #2096, opened 2020-01-24
Fix creating transcriptome indices for bacteria

kurtwheeler:
## Issue Number

#1722 
https://github.com/AlexsLemonade/refinebio/issues/2087

There was an issue with organism names coming out of the processor job. I think it's because I tested this end-to-end with LEISHMANIA_MAJOR but only the surveying for ESCHERICHIA_COLI and ESCHERICHIA_COLI comes back from the ensembl API with some addtional junk in its organism name.

This fixes that issue by using the taxonomy_id that ensembl returns instead of the organism name, and then also by creating the organism object at surveyor time instead of waiting until the processor job to do it.

## Types of changes

## Functional tests

I created organism indices locally for E. coli:
```
 id |     index_type      | source_version | assembly_name | salmon_version |                            absolute_directory_path                             | is_public |          created_at           |         last_modified         | organism_id | result_id 
----+---------------------+----------------+---------------+----------------+--------------------------------------------------------------------------------+-----------+-------------------------------+-------------------------------+-------------+-----------
  3 | TRANSCRIPTOME_LONG  | 46             | 0             | salmon 0.13.1  | /home/user/data_store/TRANSCRIPTOME_INDEX/ESCHERICHIA_COLI_GCA_001012635/long  | t         | 2020-01-24 15:52:54.764306+00 | 2020-01-24 15:52:54.764306+00 |           2 |         6
  4 | TRANSCRIPTOME_SHORT | 46             | 0             | salmon 0.13.1  | /home/user/data_store/TRANSCRIPTOME_INDEX/ESCHERICHIA_COLI_GCA_001012635/short | t         | 2020-01-24 15:53:12.491601+00 | 2020-01-24 15:53:12.491601+00 |           2 |         7

```

I also added unit tests to make sure the organism gets created by the surveyor.

## Checklist

### DIFF
--- a/common/data_refinery_common/models/organism.py
+++ b/common/data_refinery_common/models/organism.py
@@ -147,6 +147,8 @@ class Organism(models.Model):
             organism = Organism(name=name, taxonomy_id=taxonomy_id, is_scientific_name=True)
             organism.save()
 
+        return organism
+
     @classmethod
     def get_name_for_id(cls, taxonomy_id: int) -> str:
         organism = cls.get_object_for_id(taxonomy_id)
--- a/common/data_refinery_common/models/organism.py
+++ b/common/data_refinery_common/models/organism.py
@@ -137,7 +137,7 @@ class Organism(models.Model):
         return self.name.split("_")[0]
 
     @classmethod
-    def get_name_for_id(cls, taxonomy_id: int) -> str:
+    def get_object_for_id(cls, taxonomy_id: int):
         try:
             organism = cls.objects.filter(taxonomy_id=taxonomy_id).order_by("-is_scientific_name")[
                 0
@@ -146,6 +146,10 @@ class Organism(models.Model):
             name = get_scientific_name(taxonomy_id).upper().replace(" ", "_")
             organism = Organism(name=name, taxonomy_id=taxonomy_id, is_scientific_name=True)
             organism.save()
+
+    @classmethod
+    def get_name_for_id(cls, taxonomy_id: int) -> str:
+        organism = cls.get_object_for_id(taxonomy_id)
 
         return organism.name
 
--- a/common/data_refinery_common/models/organism.py
+++ b/common/data_refinery_common/models/organism.py
@@ -137,7 +137,7 @@ class Organism(models.Model):
         return self.name.split("_")[0]
 
     @classmethod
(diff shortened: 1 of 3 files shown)


foreman/data_refinery_foreman/surveyor/transcriptome_index.py
```diff
@@ -290,7 +290,7 @@ def _generate_files(self, species: Dict) -> None:
         gtf_download_url = url_builder.build_gtf_url()
 
         # Gettiing the object will ensure it is created in the DB.
```
cgreene:
```suggestion
        # Getting the object will ensure it is created in the DB.
```

common/data_refinery_common/models/organism.py
```diff
@@ -137,7 +137,7 @@ def get_genus(self):
         return self.name.split("_")[0]
 
     @classmethod
-    def get_name_for_id(cls, taxonomy_id: int) -> str:
+    def get_object_for_id(cls, taxonomy_id: int):
```
arielsvn: I would rename it to `get_or_create_...` so that it's more explicit that this method also creates an object.

foreman/data_refinery_foreman/surveyor/test_transcriptome_index.py
```diff
@@ -51,6 +51,9 @@ def test_correct_index_location(self):
         surveyor = TranscriptomeIndexSurveyor(survey_job)
         files = surveyor.discover_species()[0]
 
+        # Make sure the organism object got created.
+        Organism.objects.filter(name="DANIO_RERIO")
```
arielsvn: is there an assertion missing here?

kurtwheeler: Nope! I wanna make sure we don't get an exception, and there's no assertion for that because it's kinda asserted by default. I can leave a comment to make that more explicit though.

arielsvn: hmm but `Organism.objects.filter` should never throw an exception, or I'm missing something?

kurtwheeler: Good call, I meant to use `get()` instead of filter, fixed now.

PR #2098, opened 2020-01-24
Add code coverage reports

### LINKED ISSUE #2066: Get coverage reports working
### Context

Our current coverage reports are split across sub-projects and test suites. This makes them not very useful, because we only see how much of our code is covered by one suite at a time.

@davidsmejia found https://codecov.io/, which might be a good tool to help us solve this.

### Problem or idea

However we do this I think the first step will be to use CircleCI's [workspace feature](https://circleci.com/docs/2.0/workflows/#using-workspaces-to-share-data-among-jobs) to store the output of each test suite's code coverage reports. Some test suites run multiple tests, we may want to combine all of these before persisting them to the workspace.

Then in a downstream job we should compile all the different coverage reports and send some kind of notification.

### Solution or next step

How we "compile all the different coverage reports and send some kind of notification" may need to be explored a bit. [codecov](https://codecov.io/) is free for open source, so I think we should first try using it and see if it makes things easier. If not, the python coverage tool (which is what we use to generate the sub-reports) has a command line sub-command to do it: https://coverage.readthedocs.io/en/latest/cmd.html#combining-data-files

arielsvn:
## Issue Number

close #2066 

Add code coverage reports


arielsvn:
This should be ready for review @kurtwheeler, I made it so that the test don't fail if we fail to upload the coverage reports.

.circleci/upload_test_coverage.sh
```diff
@@ -0,0 +1,37 @@
...
+then
+    echo "Coverage file wasn't found, were the tests run before?"
+    exit 0 # exit this script but don't fail the tests for this.
+fi
+
+output_file="${test_volume}/${project}_coverage.xml"
+
+# In the test coverage report, all file paths are absolute to each project
+# folder. We need to be relative to refinebio/ that's why we append the project
+# folder name to each file path in coverage.xml
```
kurtwheeler:
```suggestion
# In the test coverage report, all file paths are relative to each project
# folder. We need to be relative to the repo's root directory. That's why we 
# append the project folder name to each file path in coverage.xml
```

kurtwheeler: Can't do this as a suggestion because it's multiple lines, but I think this might be a bit more clear:
```
# In the test coverage report, all file paths are relative to each project
# folder. We need to be relative to the repo's root directory. That's why we 
# append the project folder name to each file path in coverage.xml
```

.circleci/upload_test_coverage.sh
```diff
@@ -0,0 +1,37 @@
...
+sed "s/filename=\"/filename=\"$project\//g" $coverage_file > $output_file
+
+curl -s https://codecov.io/bash | bash -s -- -f "$output_file" -Z -F $project
```
kurtwheeler: Evaluating code that is pulled dynamically from the internet is a pretty big security flaw. Might I downloading that bash script, adding it to the repo, and then executing it from there. This way we know what we're executing and if codecov ever got hacked the hackers couldn't put a malicious script at that URL for us to run.

.circleci/upload_test_coverage.sh
arielsvn: that's way better, thanks!

.circleci/upload_test_coverage.sh
arielsvn: This can also be installed as a [python package](https://github.com/codecov/codecov-python) do you think it's better if I do `pip install codecov` before this line?

kurtwheeler: Works for me :+1:

PR #2110, opened 2020-01-29
Fix transcriptome indices for microbes.

kurtwheeler:
## Issue Number

#2096 

We were trying to ask NCBI for taxonomy ids for names like Candida_albicans_sc5314_gca_000784635. This didn't work but the lookup code returned a taxonomy id of 0 for some dumb reason. (Almost certainly my fault from a long time ago.)

This meant that when I tested one locally it would go into the DB with a taxonomy_id of 0, which worked once but the second one would violate a DB constraint and fail the job.

## Types of changes

## Functional tests

I've created a transcriptome index for E. coli and am currently running processing SRP242240 locally to test it.

## Checklist

### DIFF
--- a/common/data_refinery_common/models/organism.py
+++ b/common/data_refinery_common/models/organism.py
@@ -65,6 +65,7 @@ def get_taxonomy_id(organism_name: str) -> int:
         raise
 
     if len(id_list) == 0:
+        # If we can't find the taxonomy id, it's likely a bad organism name.
         error_message = (
             "Unable to retrieve NCBI taxonomy ID number for organism "
             + "with name: {}".format(organism_name)
--- a/common/data_refinery_common/models/organism.py
+++ b/common/data_refinery_common/models/organism.py
@@ -28,6 +28,10 @@ class InvalidNCBITaxonomyId(Exception):
     pass
 
 
+class UnknownOrganismId(Exception):
+    pass
+
+
 def get_scientific_name(taxonomy_id: int) -> str:
     parameters = {"db": TAXONOMY_DATABASE, "id": str(taxonomy_id), "api_key": NCBI_API_KEY}
     response = requests.get(EFETCH_URL, parameters)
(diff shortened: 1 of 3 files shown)


common/data_refinery_common/models/organism.py
```diff
@@ -61,11 +65,13 @@ def get_taxonomy_id(organism_name: str) -> int:
...
-            "Unable to retrieve NCBI taxonomy ID number for organism " + "with name: %s",
-            organism_name,
+        # If we can't find the taxonomy id, it's likely a bad organism name.
+        error_message = (
+            "Unable to retrieve NCBI taxonomy ID number for organism "
+            + "with name: {}".format(organism_name)
         )
-        return 0
+        logger.error(error_message)
+        raise UnknownOrganismId(error_message)
```
arielsvn: Custom exceptions? I love it!

workers/data_refinery_workers/processors/transcriptome_index.py
```diff
@@ -66,8 +66,13 @@ def _compute_paths(job_context: Dict) -> str:
...
     os.makedirs(job_context["rsem_index_dir"], exist_ok=True)
 
-    # I think this is a bit sketchy.
-    job_context["organism_name"] = job_context["base_file_path"].split("/")[-1]
+    # We don't actually associated the organism with the job, so we
+    # have to determine the organism name from the file
+    # path. Sometimes these files have strain information appended to
+    # them, so we want to make sure we only use the first two parts of
+    # a name like Candida_albicans_sc5314_gca_000784635
+    organism_directory = job_context["base_file_path"].split("/")[-1]
```
arielsvn: I added a helper a while ago to get the filename from a path https://github.com/AlexsLemonade/refinebio/blob/dev/common/data_refinery_common/utils.py#L548-L550. Is it helpful here?

workers/data_refinery_workers/processors/transcriptome_index.py
```diff
@@ -66,8 +66,13 @@ def _compute_paths(job_context: Dict) -> str:
...
+    job_context["organism_name"] = "_".join(organism_directory.split("_")[:2]).upper()
```
arielsvn: Also, is it worthwhile to extract this logic into a separate function? `get_organism_name_from_path`?

workers/data_refinery_workers/processors/transcriptome_index.py
kurtwheeler: It's not actually the filename that this is going for.... It's the directory name. I think this is a bit confusing but I think your function idea will help.

Issue #2075, opened 2020-01-16
Aggregate cloudtrail statistics into something more meaningful


arielsvn:
### Context

https://github.com/AlexsLemonade/refinebio/issues/2043#issuecomment-573909971

### Problem or idea

With #2074 we'll start logging all activity in the compendia bucket to S3 and CloudFront, this is just a bunch of JSON objects. 

```
{
    "eventVersion": "1.05",
    "userIdentity": {...},
    "eventTime": "2020-01-13T21:28:37Z",
    "eventSource": "s3.amazonaws.com",
    "eventName": "GetObject",
    "awsRegion": "us-east-1",
    "sourceIPAddress": "73.30.64.75",
    "userAgent": "[Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/79.0.3945.88 Safari/537.36]",
    "requestParameters": {
        "X-Amz-Date": "20200113T212837Z",
        "bucketName": "data-refinery-s3-circleci-prod",
        "X-Amz-Algorithm": "AWS4-HMAC-SHA256",
        "X-Amz-SignedHeaders": "host",
        "Host": "data-refinery-s3-circleci-prod.s3.amazonaws.com",
        "X-Amz-Expires": "604800",
        "key": "l6kam620qlxgf3885njojymb_0c4a0af0-8364-43b6-962c-991d133bfe3c_compendia.zip"
    },
    "responseElements": null,
    "additionalEventData": {
        "SignatureVersion": "SigV4",
        "CipherSuite": "ECDHE-RSA-AES128-GCM-SHA256",
        "bytesTransferredIn": 0,
        "AuthenticationMethod": "QueryString",
        "x-amz-id-2": "UzEbOORoNr9jz0WFiMLN7BhzXckIdZh05vVqhfJ2m9g9TUiBy3yUW+0Zk8J1nB6TVywjiSdjP6w=",
        "bytesTransferredOut": 2397392
    },
    "requestID": "13F6C73D19C5F90A",
    "eventID": "ad9ddf43-aaf3-40db-be81-250107a23d29",
    "readOnly": true,
    "resources": [
        {
            "type": "AWS::S3::Object",
            "ARN": "arn:aws:s3:::data-refinery-s3-circleci-prod/l6kam620qlxgf3885njojymb_0c4a0af0-8364-43b6-962c-991d133bfe3c_compendia.zip"
        },
        {
[...]

PR #2132, opened 2020-02-11
Implement object level statistics for compendia

davidsmejia:
## Issue Number

#2075 

This PR adds the terraform code to add a aws cloudwatch rule, event target, and log group to capture compendia downloads.

- enables metrics for the compendia bucket so we can get aggregate stats for the compendia
- enables cloudtrail and sets up permissions to write logs to the data_refinery bucket
- sets up a custom rule that reads from those events and sends api `getObject` events with object details and sends that to the `"/aws/events/data-refinery-compendia-log-group-${var.user}-${var.stage}"` log group

## Methods

Infrastructure

## Types of changes

## Functional tests

None, need to test on staging.

## Checklist

## Screenshots


infrastructure/disk.tf
```diff
@@ -114,11 +114,54 @@ resource "aws_s3_bucket" "data_refinery_compendia_bucket" {
...
 }
 
+
+resource "aws_cloudwatch_event_target" "compendia_object_metrics_target" {
+  rule = "${aws_cloudwatch_event_rule.compendia_object_metrics.name}"
+  target_id = "compendia-object-logs-target-${var.user}-${var.stage}"
+  arn = "${aws_cloundwatch_log_group.compendia_object_metrics_log_group.arn}"
+}
+
+data "aws_cloudwatch_log_group" "compendia_object_metrics_log_group" {
```
kurtwheeler: Why is this `data` instead of `resource`?

kurtwheeler:
Have you tested this on a dev stack yet?

infrastructure/disk.tf
davidsmejia: This should have been resource!

davidsmejia:
> Have you tested this on a dev stack yet?

tested and working

infrastructure/disk.tf
```diff
@@ -114,9 +114,51 @@ resource "aws_s3_bucket" "data_refinery_compendia_bucket" {
   }
 }
 
-resource "aws_s3_bucket_metric" "compendia_metrics" {
+resource "aws_s3_bucket_metric" "compendia_bucket_metrics" {
   bucket = "${aws_s3_bucket.data_refinery_compendia_bucket.bucket}"
-  name   = "data-refinery-compendia-metric-${var.user}-${var.stage}"
+  name   = "EntireBucket"
```
kurtwheeler: EntireBucket?

kurtwheeler: Not sure what that means and shouldn't this be namespaced with the user/stack so that there won't be conflicts between dev/staging/prod?

infrastructure/disk.tf
```diff
@@ -114,9 +114,51 @@ resource "aws_s3_bucket" "data_refinery_compendia_bucket" {
...
+  "detail": {
+    "eventSource": [
+      "s3.amazonaws.com"
+    ],
+    "eventName": [
+      "GetObject"
+    ],
+    "requestParameters": {
+      "bucketName": [
+        "data-refinery-s3-compendia-${var.user}-${var.stage}"
```
kurtwheeler: This is the bucket that compendia are stored in? I think this should be `"${aws_s3_bucket.data_refinery_compendia_bucket.bucket}"` (assuming that's the correct bucket).

infrastructure/disk.tf
davidsmejia: Yeah... funny that, the bucket parameter is where tell it which bucket to create a metrics rule for (which is dev/stage/prod) depending on env, EntireBucket is some codeword they use that allows us to enable object level logs.

infrastructure/disk.tf
davidsmejia: I think that would get the ID/arn no?

infrastructure/disk.tf
```diff
@@ -114,9 +114,51 @@ resource "aws_s3_bucket" "data_refinery_compendia_bucket" {
...
+}
+PATTERN
+}
+
+# arn needs to be stripped of trailing `:*`
+# aws appends this when creating the event_target
+resource "aws_cloudwatch_event_target" "compendia_object_metrics_target" {
+  rule = "${aws_cloudwatch_event_rule.compendia_object_metrics.name}"
+  target_id = "compendia-object-logs-target-${var.user}-${var.stage}"
+  arn = "${substr(aws_cloudwatch_log_group.compendia_object_metrics_log_group.arn,0,length(aws_cloudwatch_log_group.compendia_object_metrics_log_group.arn) - 2)}"
```
kurtwheeler: Geez that's obnoxious. Solid way to deal with it though :+1:

infrastructure/disk.tf
```diff
@@ -114,9 +114,51 @@ resource "aws_s3_bucket" "data_refinery_compendia_bucket" {
...
+}
+
+# Must start with `/aws/events` in order to connect to a cloudwatch_event_target.
+resource "aws_cloudwatch_log_group" "compendia_object_metrics_log_group" {
+  name = "/aws/events/data-refinery-compendia-log-group-${var.user}-${var.stage}"
```
kurtwheeler: I think `logging.tf` might be a slightly more appropriate home for this resource.

infrastructure/disk.tf
davidsmejia: yeah, not sure if the comment above could be improved

infrastructure/logging.tf
```diff
@@ -1,3 +1,29 @@
+# CloudTrail
+
+resource "aws_cloudtrail" "data_refinery_s3_cloudtrail" {
+  name                          = "data-refinery-s3-cloudtrail-${var.user}-${var.stage}"
+  depends_on                    = ["aws_s3_bucket.data_refinery_bucket"]
+  s3_bucket_name                = "${aws_s3_bucket.data_refinery_bucket.id}"
```
kurtwheeler: This bucket is where we store our processed data. Could you create a new bucket for storing these logs?

infrastructure/permissions.tf
```diff
@@ -216,3 +216,32 @@ resource "aws_iam_role_policy_attachment" "cloudwatch" {
   role = "${aws_iam_role.data_refinery_instance.name}"
   policy_arn = "${aws_iam_policy.cloudwatch_policy.arn}"
 }
+
+resource "aws_s3_bucket_policy" "cloudtrail_access_policy" {
+  bucket = "${aws_s3_bucket.data_refinery_bucket.id}"
```
kurtwheeler: This also will need to change to reference the same logging bucket as `aws_cloudtrail.data_refinery_s3_cloudtrail`

davidsmejia: 👍

infrastructure/disk.tf
kurtwheeler: https://www.terraform.io/docs/providers/aws/r/s3_bucket.html#bucket says:
```
bucket - (Optional, Forces new resource) The name of the bucket. If omitted, Terraform will assign a random, unique name.
```

kurtwheeler: I mean that you should use the variable rather than the value.

infrastructure/disk.tf
kurtwheeler: Ah, in that case can you leave a comment explaining that?

infrastructure/disk.tf
kurtwheeler: It was helpful enough for me to figure out what was happening here :man_shrugging:

infrastructure/disk.tf
```diff
@@ -114,9 +114,51 @@ resource "aws_s3_bucket" "data_refinery_compendia_bucket" {
   }
 }
 
-resource "aws_s3_bucket_metric" "compendia_metrics" {
+resource "aws_s3_bucket_metric" "compendia_bucket_metrics" {
```
davidsmejia:
```suggestion
# Passing the name attribute as EntireBucket enables request metrics for the bucket. 
resource "aws_s3_bucket_metric" "compendia_bucket_metrics" {
```

davidsmejia: is this clearer @kurtwheeler

kurtwheeler: Maybe also a link to documentation explaining this if possible?

infrastructure/disk.tf
davidsmejia: updated the comment

infrastructure/logging.tf
```diff
@@ -54,3 +80,10 @@ resource "aws_cloudwatch_log_stream" "log_stream_api_nginx_error" {
   name           = "log-stream-api-nginx-error-${var.user}-${var.stage}"
   log_group_name = "${aws_cloudwatch_log_group.data_refinery_log_group.name}"
 }
+
+# Cloudtrail Logs
+
+# Must start with `/aws/events` in order to connect to a cloudwatch_event_target.
+resource "aws_cloudwatch_log_group" "compendia_object_metrics_log_group" {
+  name = "/aws/events/data-refinery-compendia-log-group-${var.user}-${var.stage}"
+}
```
davidsmejia: moved to logging from disk.tf

infrastructure/permissions.tf
```diff
@@ -216,3 +216,32 @@ resource "aws_iam_role_policy_attachment" "cloudwatch" {
   role = "${aws_iam_role.data_refinery_instance.name}"
   policy_arn = "${aws_iam_policy.cloudwatch_policy.arn}"
 }
+
+resource "aws_s3_bucket_policy" "cloudtrail_access_policy" {
+  bucket = "${aws_s3_bucket.data_refinery_cloudtrail_logs_bucket.id}"
```
davidsmejia: updated bucket name

davidsmejia:
Ok was able to deploy this again @kurtwheeler

'
```diff
@@ -0,0 +1,89 @@
+# CloudTrail
+
+resource "aws_cloudtrail" "data_refinery_s3_cloudtrail" {
+  name                          = "data-refinery-s3-cloudtrail-${var.user}-${var.stage}"
```
kurtwheeler: Oooh! I know the terraform docs write it with the spacing like this but the convention for our project is to not put more than one space before the `=`. The reasoning is that if you added a longer line to this block then you have to change every other line in the block.

'
```diff
@@ -0,0 +1,89 @@
...
+  depends_on                    = ["aws_s3_bucket.data_refinery_cloudtrail_logs_bucket"]
```
kurtwheeler: Isn't this implied when you reference the id of this resource in the line below?

You don't need to retest it over just this, so you can leave it, but it's just a thought.

'
```diff
@@ -0,0 +1,89 @@
+# CloudTrail
```
kurtwheeler: Wait a second. What is this file? It's just called `'`?

davidsmejia: i removed this file it was a scratch from logging.tf

infrastructure/disk.tf
```diff
@@ -114,9 +114,59 @@ resource "aws_s3_bucket" "data_refinery_compendia_bucket" {
...
 }
 
-resource "aws_s3_bucket_metric" "compendia_metrics" {
+resource "aws_s3_bucket" "data_refinery_cloudtrail_logs_bucket" {
+  bucket = "data-refinery-s3-cloudtrail-logs-${var.user}-${var.stage}"
+  acl = "private"
+  force_destroy = "${var.static_bucket_prefix == "dev" ? true : false}"
+
+  tags {
+    Name        = "data-refinery-s3-cloudtrail-logs-${var.user}-${var.stage}"
```
kurtwheeler: One more whitespace jawn here.

davidsmejia: nuked

Issue #2114, opened 2020-01-30
Beadchip samples look like noise in human compendium


ben-heil:
### Context

I was doing some exploratory data analysis for the samples I'm working with in whistl, and noticed bands of what appeared to be random noise through the data. On further inspection, each of these bands appears to come from a study using a beadchip as the sequencing platform.

The x-axis entries of these heatmaps are samples, while the y-axis entries are genes. The three noise bands below correspond to GSE58411, GSE40553, and GSE83456.

![image](https://user-images.githubusercontent.com/15317110/73485580-5a165580-4371-11ea-8bcb-f023d3b547f1.png)

*Zoom, enhance*

![image](https://user-images.githubusercontent.com/15317110/73485590-613d6380-4371-11ea-878b-d80a1ff022ba.png)

### Problem or idea

I know that some issues with beadchips have been mentioned in the past, but I'm unsure whether they are supposed to be this noisy.

### Solution or next step

I'll pull these samples out of my analysis, but wanted to make that the noise wasn't due to a pipeline issue before doing so

kurtwheeler:
For this sprint we will just focus on investigating what is going on, and then we will fix/address whatever issues we identify in the following sprint.

davidsmejia:
@ben-heil Looks like the detection pvalue for those experiments are very high. I don't think anyone would recommend using them in your analysis.

davidsmejia:
We will be regenerating this and other compendium soon, without submitter-processed data until we can ensure the integrity of the data.

ben-heil:
Sounds good, thanks!

PR #2148, opened 2020-02-20
Compendia without submitter processed data

davidsmejia:
## Issue Number

#2114 

Remove Submitter Processed Data from compendium datasets. 

## Methods

In the management command we loop through all of the samples to make sure they are processed. This adds an extra clause to that to exclude any samples where the result's process name is `Submittter-processed`. Which is as far as I can tell the best way to check for this sort of thing.

## Types of changes

## Functional tests

Was unable to run nomad locally.. something about unable to bind to port... so running tests here and will request review after I am able to run them locally.

## Checklist

### DIFF
--- a/foreman/data_refinery_foreman/foreman/management/commands/create_compendia.py
+++ b/foreman/data_refinery_foreman/foreman/management/commands/create_compendia.py
@@ -161,9 +161,11 @@ def get_dataset(organisms: List[Organism]):
     experiments = Experiment.objects.filter(filter_query).prefetch_related("samples")
 
     for experiment in queryset_iterator(experiments):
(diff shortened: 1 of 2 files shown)


davidsmejia:
local test for create_compendia command passed on my machine

kurtwheeler:
Sorry, I should have mentioned this sooner, but why not just use `sample.has_raw`?

davidsmejia:
> Sorry, I should have mentioned this sooner, but why not just use `sample.has_raw`?

Good call! @kurtwheeler I can just filter by `has_raw=False`?

kurtwheeler:
Yep! Well, `exclude` by `has_raw=False`, `filter` by `has_raw=True` ;)

PR #2158, opened 2020-02-25
Davidsmejia/pgbouncer public

davidsmejia:
## Issue Number

N/A

I need to test on a dev stack but I believe this is why routing everything though the public ip for pg bouncer was failing. We needed to allow for the response from pg bouncer to come in. This PR should fix that.


kurtwheeler:
@davidsmejia can you also approve this since I've pushed since you opened it?

Issue #2175, opened 2020-02-28
Reduce size of the EBS volumes for the Nomad cluster


kurtwheeler:
### Context

https://github.com/AlexsLemonade/refinebio/issues/2042#issuecomment-578818717

### Problem or idea

The EBS volumes we currently use are 9TB so that we don't have to worry about them filling up whilst processing data. However we never run for days at a time anymore so they won't have a chance to fill up.

### Solution or next step

Reduce the size down to at least 2TB. We can consider going even lower if that doesn't cause any issues for a few different crunches.

cgreene:
Closed with #2179

Issue #2171, opened 2020-02-28
Reduce the max size of our Nomad queue and of our Nomad servers


kurtwheeler:
### Context

https://github.com/AlexsLemonade/refinebio/issues/2042

Idea 3

### Problem or idea

We don't run at such large scales any more because we've now processed the bulk of the data. Therefore we can reduce the size of our Nomad servers and decrease the limits of how many jobs we can put in the queue at once.

### Solution or next step

Reduce max number of jobs that can be queued at a single time by 50-75% and then drop the size of the Nomad servers by a proportional amount. The size of servers that we're currently using have a cap of \~20k jobs, but we don't need to run at such extreme scales any more so we could probably be fine with a limit of \~5k jobs.

Issue #2170, opened 2020-02-28
Use only a single Nomad server


kurtwheeler:
### Context

https://github.com/AlexsLemonade/refinebio/issues/2042

Ideas 1 and 2 cause I think we should just do them both.

### Problem or idea

We could use a single Nomad server. This would have less redundancy when we're running, but we already don't trust Nomad to not mess up so it being perfect isn't very important for our system.

### Solution or next step

Reduce the number of Nomad servers to just one, taking out the other two from each stack.

cgreene:
Closed with #2179

PR #2179, opened 2020-03-02
Decrease Nomad resources to save money

kurtwheeler:
Decrease nomad server/queue size, remove redundancy for Nomad servers, scale down EBS volumes.

## Issue Number

#2171 
#2175 
#2170 

This should save us a lot of money because we're paying 1/3rd as many Nomad servers that are each 1/4th as large. Also we won't have such large EBS volumes to pay for.

## Types of changes

- Infrastructure tuning

## Functional tests

I ran a dev stack cluster at 80 capacity for about an hour. The nomad job queue got right up to the limit multiple times, at which point the foreman stopped queuing jobs and it dropped lower. Throughout this test, the nomad lead server rode the 80% utilization line, which seems great. It didn't struggle at all so we should be good to run at this scale long term in the future.

## Checklist

## Screenshots

new_nomad_performance


foreman/data_refinery_foreman/foreman/main.py
```diff
@@ -49,7 +49,7 @@
 MAX_NUM_RETRIES = 2
 
 # This can be overritten by the env var "MAX_TOTAL_JOBS"
-DEFAULT_MAX_JOBS = 20000
+DEFAULT_MAX_JOBS = 5000
```
arielsvn: Are there specific reasons why we are setting `5000` here? should we add a comment for it?

kurtwheeler: Good point!

kurtwheeler:
We have a billing report going back so we can compare across time!

Issue #2140, opened 2020-02-17
EngagementBot: Summarize into weekly digest 


dvenprasad:
### Context

Engagement bot is noisy! 🎉 

### Problem or idea

Too noisy. Summarize the download information into a weekly digest instead.
Summarize:
- number of downloads
- Total number of unique downloaders
- Number of new downloaders and their locations. 

### Solution or next step
Include other stats I may have not thought of.

kurtwheeler:
Basic plan: create a cron job that summarizes the dataset and compendia downloads for each week and pushes it to slack. This can either run on the API or the Foreman.

arielsvn:
Do we really need the weekly digest on slack? We have some of this information on the [dashboard](https://www.refine.bio/dashboard) that we could update with location information

![image](https://user-images.githubusercontent.com/1882507/76693543-497a2000-663d-11ea-9ec8-1d5a6f2a021d.png)

cgreene:
There are stakeholders who may not check the dashboard but who are in slack, and this is a way to communicate more broadly with them about the impact. That includes some data science team members as well as leadership at ALSF.

arielsvn:
https://github.com/AlexsLemonade/refinebio/issues/2140 Adds a command to post the summary in the last N days.

In order for this to be closed, we'll also need to set up a cron job, similar to:

https://github.com/AlexsLemonade/refinebio/blob/dev/infrastructure/api-configuration/api-server-instance-user-data.tpl.sh#L137

I'll add it in another PR after testing #2140

PR #2193, opened 2020-03-16
Post a summary of the latest downloads in slack

arielsvn:
## Issue Number

#2140 

Creates a command in slack to post to slack a summary of the downloads in the last N days.

This still needs to be scheduled to run weekly with a cron job. I want to test this on staging and production before adding the cron job.

## Types of changes

## Functional tests

Tested locally posting to `#robots`

## Checklist

## Screenshots

image


api/data_refinery_api/management/commands/post_downloads_summary.py
```diff
@@ -0,0 +1,129 @@
...
+def should_display_email(email: str) -> bool:
+    """ Returns true if the given email is not associated with the CCDL suers """
+    return (
+        email is not None
+        and email.find("cansav09") != 0
+        and email.find("arielsvn") != 0
+        and email.find("jaclyn.n.taroni") != 0
+        and email.find("kurt.wheeler") != 0
+        and email.find("greenescientist") != 0
+        and email.find("@alexslemonade.org") == -1
```
cgreene: Curious why this one is different than the others

api/data_refinery_api/management/commands/post_downloads_summary.py
```diff
@@ -0,0 +1,129 @@
...
+        and email.find("miserlou") != 0
+        and email.find("michael.zietz@gmail.com") != 0
```
cgreene: Michael has been gone for ~8 months now, so if he is still downloading those should probably count

api/data_refinery_api/management/commands/post_downloads_summary.py
arielsvn: Yes, this was a bit confusing to me too. The ones with `!= 0` are testing that the email doesn't start with the provided string. And this one is checking that it doesn't contain `@alexslemonade.org`. I'll change it to use `startswith` and `in`

api/data_refinery_api/management/commands/post_downloads_summary.py
```diff
@@ -0,0 +1,130 @@
...
+        parser.add_argument(
+            "--channel",
+            type=str,
+            default="ccdl-general",
+            help=("Optional parameter to choose the channel where the message will be posted."),
+        )
+
+    def handle(self, *args, **options):
+        """
+        """
```
kurtwheeler: This docstring is empty. I feel like the help covers it though so this is fine and we can just remove these empty strings.

common/data_refinery_common/models/dataset_annotation.py
```diff
@@ -0,0 +1,36 @@
...
+    public_objects = PublicObjectsManager()
+
+    # Relations
+    dataset = models.ForeignKey("Dataset", blank=False, null=False, on_delete=models.CASCADE)
+
+    # Properties
+    data = JSONField(default=dict)
+
+    # Common Properties
+    is_public = models.BooleanField(default=True)
```
kurtwheeler: Hmm, would it be an issue if people could query our API and see what data was downloaded by which IPs @jaclyn-taroni ? I think I remember overhearing discussions about people potentially being scared to use us in fear that what data they are downloading could be used to help scoop them? Is that actually a concern?

It's not actually an issue because this new model doesn't have a view so it's not accessible via the API. However, if it was a concern we could set public to False by default and leave a comment about how a view should not be created for this model in the future.

arielsvn: But that `is_public` variable is only used internally right? Does it have anything to do with the objects being in the API? My interpretation of it was that if `is_public=False` is as if the object is removed from the db.

We also never expose the `email_address` field of the datasets, which is part of the model.

kurtwheeler: My understanding is that is that `is_public` just controls whether or not the `DatasetAnnotation.public_objects` manager returns them or not. If you use `DatasetAnnotation.objects` it should return them whether they are public or not.

api/data_refinery_api/management/commands/post_downloads_summary.py
```diff
@@ -0,0 +1,130 @@
...
+        or email is ("daniel.himmelstein@gmail.com")
+        or email is ("dv.prasad991@gmail.com")
+    )
+
+
+def get_ip_location(remote_ip):
+    try:
+        city = requests.get("https://ipapi.co/" + remote_ip + "/json/", timeout=10).json()["city"]
+    except Exception:
+        city = "COULD_NOT_DETERMINE"
```
kurtwheeler: If we can't get the city it might be because the ipapi is down. In this case can we then include the IP address in the slackbot message so that we can try looking it up later?

common/data_refinery_common/models/dataset_annotation.py
jaclyn-taroni: > would it be an issue if people could query our API and see what data was downloaded by which IPs

Gut reaction here: I do not like the idea of that being a possibility.

davidsmejia: The base manager is the public objects manager so if it was false it shouldn't show up with `DatasetAnnotation.objects` I believe.

arielsvn:
I think I addressed your comments, thanks for reviewing. WIll merge this once the tests pass.

kurtwheeler:
:+1:

Issue #2006, opened 2019-12-16
Recover samples w/ mismatched salmon versions / transcriptome indices


cgreene:
### Context

@arielsvn was looking into a user request in issue https://github.com/AlexsLemonade/refinebio/issues/1996#issuecomment-566190000 .

### Problem or idea

In at least some cases, we had at least one `quant.sf` file already processed for a sample. When we upgraded salmon's transcriptome indices we ended up with `quant.sf` files that cannot be merged.

### Solution or next step

We should query for any cases where have samples from the same experiment processed with different transcriptome indices. We should reprocess all of these with the latest version and attempt to run tximport over samples processed with the latest version.

arielsvn:
We should have a command for this `rerun_salmon_old_samples`, but I couldn't find any record from when we ran it. Looks like we have `530` experiments with mismatched salmon versions, with ~250k unprocessed samples still.

<details>

```
>>> eligible_experiments = (
...     Experiment.objects.filter(technology="RNA-SEQ", num_processed_samples=0)
...     .annotate(
...         num_salmon_versions=Count(
...             "samples__results__organism_index__salmon_version",
...             distinct=True,
...             filter=Q(
...                 samples__results__processor__name=ProcessorEnum.SALMON_QUANT.value["name"]
...             ),
...         )
...     )
...     .filter(num_salmon_versions__gt=1)
... )
>>> eligible_experiments.count()
530
>>> eligible_experiments.values('samples__accession_code').distinct().count()
25368
```

</details>

I'm going to wait until #2210 can be deployed to test this command and try to see why it didn't work for those experiments.

arielsvn:
Retried sample [DRR013807](https://www.refine.bio/samples/DRR013807) with the downloader job [id=7097213](https://api.refine.bio/v1/jobs/downloader/?format=json&id=7097213). We'll need to investigate why it didn't create any processor jobs afterward, the downloader job completed successfully.

arielsvn:
Did some debugging of this, but still haven't found the problem. I manually set `is_downloaded=False` for the `OriginalFile` associated with that sample and now the downloader jobs are failing with the error:

> Was told to redownload file(s) that are already downloaded!

Which is set on https://github.com/AlexsLemonade/refinebio/blob/dev/workers/data_refinery_workers/downloaders/utils.py#L76-L90

`original_file.needs_downloading()` is `True` so I don't yet understand why we're hitting that condition.

arielsvn:
[...]

PR #2210, opened 2020-03-26
Bump max_clients to 10 to start reprocessing samples

arielsvn:
## Issue Number

#2006 

This sets `max_clients=10`, we'll need to re-process some samples for #2006.

Adding the PR for when we are ready to deploy.


kurtwheeler:
I think 10 is too big for two reasons. One is that I don't know that we need to churn through this data THAT fast. What's the rush? The other more important one is that we scaled back our max capacity in https://github.com/AlexsLemonade/refinebio/pull/2179. I think we dropped it to about 1/4 its original size, so I'd say 2 instances should be big enough. If we know we're going to have big fat jobs coming in 5 would be the max. I base this off the fact that our old max used to be 10 for normal operation and 20 for lots of big fat jobs.

arielsvn:
ok I set it to `2` @kurtwheeler that makes sense

PR #2218, opened 2020-04-01
Allow downloader jobs to run if file is partially downloaded

arielsvn:
## Issue Number

#2006

We found another corner case for the method `original_file.needs_downloading()`. If a downloader job gets OOM-Killed and it's not able to download the file fully, then on further retries we assume that the file is successfully downloaded.

This fixes it by adding an additional check to ensure that the SHA1 of the downloaded file matches what we have in the database. Files that we have never downloaded will get re-downloaded again since we don't have a way to check if it's complete.

@kurtwheeler We also need to increase the memory for the SRA jobs, it seems 1Gb isn't enough for these. Or do we re-try them with higher memory?

## Types of changes

## Functional tests

None

## Checklist

### DIFF
--- a/workers/data_refinery_workers/downloaders/sra.py
+++ b/workers/data_refinery_workers/downloaders/sra.py
@@ -306,14 +306,6 @@ def download_sra(job_id: int) -> None:
     downloaded_files = []
     success = None
     for original_file in original_files:
-        if original_file.is_downloaded:
-            logger.info(
-                "File already downloaded!", original_file=original_file.id, downloader_job=job_id
-            )
-            success = True
-            downloaded_files.append(original_file)
-            continue
-
         exp_path = LOCAL_ROOT_DIR + "/" + job.accession_code
         samp_path = exp_path + "/" + sample.accession_code
         os.makedirs(exp_path, exist_ok=True)
--- a/workers/data_refinery_workers/downloaders/sra.py
+++ b/workers/data_refinery_workers/downloaders/sra.py
@@ -311,6 +311,7 @@ def download_sra(job_id: int) -> None:
                 "File already downloaded!", original_file=original_file.id, downloader_job=job_id
             )
             success = True
+            downloaded_files.append(original_file)
             continue
 
         exp_path = LOCAL_ROOT_DIR + "/" + job.accession_code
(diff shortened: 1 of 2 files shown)


cgreene:
Good detective work!

workers/data_refinery_workers/downloaders/sra.py
```diff
@@ -296,28 +296,22 @@ def download_sra(job_id: int) -> None:
...
-        # _replace_dotsra_with_fastq_files returns a list of
-        # OriginalFiles so turn the queryset of
-        # DownloaderJobOriginalFileAssociations into a list of
-        # OriginalFiles to be consistent.
-        original_files = [assoc.original_file for assoc in file_assocs]
 
     downloaded_files = []
     success = None
     for original_file in original_files:
         if original_file.is_downloaded:
```
kurtwheeler, line 309: Where is this being set?

arielsvn:
> If the sha does match and we return False from needs_downloading then the job is still just going to think it needs to do nothing. It's not going to queue a processor job because it won't have realized that it needs to.

Yes, in this case, I assumed that if the file was fully downloaded then a previous downloader job was successful in downloading the file and very likely in triggering a processor job. A solution for this would be to remove [that check](https://github.com/AlexsLemonade/refinebio/blob/9b54b7d446baf112680e1cdb2037af2f1889e2fe/workers/data_refinery_workers/downloaders/utils.py#L76-L90) completely and let the job run even when we know the files are already downloaded.

> I think you tried to address this by putting: `downloaded_files.append(original_file)` in the condition that the file already has `is_downloaded==True`

That was mostly for completeness, should we also check `needs_downloading()` there besides `is_downloaded`? It seemed to me, that if we are running a downloader job and a file is already downloaded, then it should be included in the list of files that we use to trigger the processor job. Does this seem reasonable?

kurtwheeler:
> It seemed to me, that if we are running a downloader job and a file is already downloaded, then it should be included in the list of files that we use to trigger the processor job.

My concern is that if we include it in the list because the database says it's downloaded, but it's not actually on the disk, then the processor job would be doomed to fail.

Do we really need that check for `is_downloaded` there at all? We're already checking if the file is downloaded by actually looking at the sha of the file on disk, so checking the property doesn't seem like it adds much.

arielsvn:
Tests are starting to pass, I think I found the problem. The modification I did for the `needs_downloading` method was returning true for all original files in the test data that didn't have a SHA1, which caused the tests to fail.

arielsvn:
> Do we really need that check for is_downloaded there at all? We're already checking if the file is downloaded by actually looking at the sha of the file on disk, so checking the property doesn't seem like it adds much.

Yes, you're right, we also check for that on the `start_job` method. I removed the check. This should be ready for another review @kurtwheeler

PR #2223, opened 2020-04-08
Different HTTP download method

arielsvn:
## Issue Number

N/A

I have been trying to process several samples, all with similar original files and some of the SRA jobs are being OOM-Killed.

image

This uses a different method for the HTTP downloads, @kurtwheeler do you think this could be more stable? I got it from [here](https://stackoverflow.com/a/39217788/763705)

## Types of changes

## Functional tests

None

## Checklist

### DIFF
--- a/workers/data_refinery_workers/downloaders/sra.py
+++ b/workers/data_refinery_workers/downloaders/sra.py
@@ -9,6 +9,8 @@ from typing import List
 
 from django.utils import timezone
 
+import requests
+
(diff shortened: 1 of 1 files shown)


kurtwheeler:
Ohhhhhh, this explains why some Array Express downloader jobs got OOM-killed and we needed tiers on downloader jobs. We didn't know how downloader jobs could OOM-kill, but we ran into that back when we were in crunch time so we fixed it with the RAM-tiering solution.

PR #2227, opened 2020-04-09
Increase timeout before mounting volume

arielsvn:
## Issue Number

N/A

After the last deploy, the volume wasn't mounted on the spot instance. I tested the commands and they worked afterward.

So I assume the `sleep` before selecting the disk wasn't big enough? @kurtwheeler what do you think?


arielsvn:
> You think 10 seconds was making the difference?

I'm not sure how to test it, hopefully it will be enough?

PR #2230, opened 2020-04-10
Add helper to download url and retry 3 times on error

arielsvn:
## Issue Number

#2006

I'm running out of theories for this, [reading online](https://stackoverflow.com/a/17473795/763705) I saw that there might be multiple reasons why these `OSError Cannot allocate memory` are raised and that for some of them there isn't much to do from the python script.

This will be capturing the exception, waiting a few secs and retrying the download 3 times before failing the job.

## Types of changes

## Functional tests

None

## Checklist

### DIFF
--- a/common/data_refinery_common/utils.py
+++ b/common/data_refinery_common/utils.py
@@ -545,7 +545,7 @@ def download_file(
     retry=1,
     *,
     backoff_factor=8,
-    max_retries=5,
+    max_retries=10,
     max_sleep_timeout=120  # 2 mins
 ):
     """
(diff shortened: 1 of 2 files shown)


arielsvn:
Nice, I didn't know that existed :) Is exponential backoff used mostly to retry network requests? What would be the benefit here? We get the error when writing into the disk.

kurtwheeler:
Well given that we don't actually know why we're getting this error it seems like it might be prudent. I also think that it could potentially help because if you put a random element to it then it could maybe stagger the jobs some so not so many run simultaneously since you've said that they seem to fail more when run simultaneously.

arielsvn:
@kurtwheeler I modified the algorithm to integrate exponential backoff. What do you think?

PR #2235, opened 2020-04-15
Slack downloads cron remove */ before 0 12 * * MON

davidsmejia:
Slack Posts weren't firing on mondays. This should fix them not being scheduled at all.


kurtwheeler:
Nice find indeed! That's subtle.

Issue #2237, opened 2020-04-16
Instance volume switching to Read-only while processing


arielsvn:
### Context

We are trying to re-process `~30k` SRA samples for https://github.com/AlexsLemonade/refinebio/issues/2006#issuecomment-613538306, some need to download large files (`>10gb`).

Downloader jobs were failing with out-of-memory errors so we introduced https://github.com/AlexsLemonade/refinebio/pull/2223 to stream downloads. But the number of Out of memory errors increased, so we tried https://github.com/AlexsLemonade/refinebio/pull/2230 to re-try downloads within the same job if any error was raised and sleep in between re-tries.

Also, added some changes to the script that mounts the volume https://github.com/AlexsLemonade/refinebio/pull/2222 because it wasn't correctly identified. As well as https://github.com/AlexsLemonade/refinebio/pull/2227 because the volume wasn't being mounted when the instance started. That didn't solved it, and we have an issue for it https://github.com/AlexsLemonade/refinebio/issues/2234

### Problem or idea

Now we are having another issue, the instance is switching to readonly and we are getting the following error on the downloader jobs:

```
OSError: [Errno 30] Read-only file system: '/home/user/data_store/ERP006862'
```

Also on the instance:

```
ubuntu@ip-10-0-84-77:/var/ebs$ touch xx
touch: cannot touch 'xx': Read-only file system
ubuntu@ip-10-0-84-77:/var/ebs$ grep "[[:space:]]ro[[:space:],]" /proc/mounts 
tmpfs /sys/fs/cgroup tmpfs ro,nosuid,nodev,noexec,mode=755 0 0
/dev/xvdf /var/ebs ext4 ro,relatime,data=ordered 0 0
ubuntu@ip-10-0-84-77:/var/ebs$ df -h
Filesystem      Size  Used Avail Use% Mounted on
udev            481G     0  481G   0% /dev
tmpfs            97G  1.1G   95G   2% /run
/dev/xvda1       97G  2.4G   95G   3% /
tmpfs           481G     0  481G   0% /dev/shm
tmpfs           5.0M     0  5.0M   0% /run/lock
tmpfs           481G     0  481G   0% /sys/fs/cgroup
/dev/xvdf       2.0T  459G  1.4T  25% /var/ebs
tmpfs            97G     0   97G   0% /run/user/1000
```

### Solution or next step

by @cgreene 

[...]

PR #2238, opened 2020-04-16
Use instance c5d.9xlarge for spot fleet with 900gb disk

arielsvn:
## Issue Number

#2237

EBS volumes seem to be giving problems when processing because they are going `Read-Only`. This changes the instance type to `c5d.9xlarge` which should come with it's own storage.

This PR should be reverted once the tests pass.

## Types of changes

## Functional tests

None

### DIFF
--- a/infrastructure/instances.tf
+++ b/infrastructure/instances.tf
@@ -217,13 +217,10 @@ resource "aws_spot_fleet_request" "cheap_ram" {
             "aws_instance.pg_bouncer"
   ]
 
-  ##
-  # x1e.8xlarge
-  ##
   launch_specification {
 
     # Client Specific
-    instance_type             = "x1e.8xlarge"
+    instance_type             = "c5d.9xlarge"
     weighted_capacity         = 10 # via https://aws.amazon.com/ec2/instance-types/
     spot_price                = "${var.spot_price}"
     ami                       = "${data.aws_ami.ubuntu.id}"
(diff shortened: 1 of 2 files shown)


arielsvn:
Created a dev stack to test this, but there's no capacity now


> c5d.9xlarge, ami-03ffa9b61e8d2cfda, Linux/UNIX: There is no Spot capacity available that matches your request.


Will wait for a bit to see if it gets launched to ensure that the drive is correctly mounted.

arielsvn:
I tested this on a dev stack to ensure that the volume was correctly mounted on the instance in `/var/ebs`. 

```
ubuntu@ip-10-0-94-163:~$ df -h
Filesystem      Size  Used Avail Use% Mounted on
udev             35G     0   35G   0% /dev
tmpfs           6.9G  8.8M  6.9G   1% /run
/dev/nvme0n1p1   97G  1.5G   96G   2% /
tmpfs            35G     0   35G   0% /dev/shm
tmpfs           5.0M     0  5.0M   0% /run/lock
tmpfs            35G     0   35G   0% /sys/fs/cgroup
/dev/loop0       94M   94M     0 100% /snap/core/8935
/dev/loop1       18M   18M     0 100% /snap/amazon-ssm-agent/1566
tmpfs           6.9G     0  6.9G   0% /run/user/1000
/dev/nvme1n1    825G   73M  783G   1% /var/ebs
```

I wanted to also test the jobs but the nomad jobs were not correctly registered in the dev stack. We can also test with the jobs we have in prod as well. @kurtwheeler what do you think?

PR #2239, opened 2020-04-17
[Deploy] Master

arielsvn:
Use different instance type https://github.com/AlexsLemonade/refinebio/pull/2238, I'll revert this once the test is finished.

It also includes engagement bot.

@kurtwheeler do you think we can hotfix deploy this?


kurtwheeler:
Yeah this is 100% infra, I'm cool with a hotfix.

PR #2248, opened 2020-04-21
Set max clients to 1

arielsvn:
## Issue Number

#2237 

This sets `max_clients=1`. Looks like when `max_clients=0` only smasher jobs are registered in nomad.

https://github.com/AlexsLemonade/refinebio/blob/dev/scripts/format_nomad_with_env.sh#L278-L297


arielsvn:
Already merged with https://github.com/AlexsLemonade/refinebio/commit/f9249db36006f2b314c66c68d6f6bcde493705ef but github didn't close the PR

Issue #2246, opened 2020-04-21
Downloaded RNA-seq compendium for Pseudomonas aeruginosa missing quant.sf files


georgiadoing:
### Context
I went to download the RNA-seq (not normalized) compendium for Pseudomonas aeruginosa through the web server

https://www.refine.bio/compendia?c=rna-seq-sample

### Problem or idea
I got a 586.8 KB zip file that contained sub directories for individual experiments but in those directories was only metadata .tsv files, no quant.sf files or raw data files. 

The metadata files also appeared to be from microarray experiments (at least he few I checked). I'm not sure if this is a separate issue/ticket or related, so thought I'd mention it.

### Solution or next step
I have not tried downloading the compendium through the API or any other means, just this one way. I did have success downloading the normalized compendium of Pseudomonas aeruginosa microarray data from the web server though.

kurtwheeler:
Hi Georgia!

Thanks for reporting this!!! That doesn't look right at all to me. I've checked out our database and it looks like the file we created and uploaded was that same size. It looks like something went wrong, but I'm not entirely sure what yet.

I'm going to start by reprocessing this compendium and seeing if it Just Works. If not @arielsvn will dig into why not further!

kurtwheeler:
Hi Georgia!

Thanks for reporting this!!! That doesn't look right at all to me. I've checked out our database and it looks like the file we created and uploaded was that same size. It looks like something went wrong, but I'm not entirely sure what yet.

I'm going to start by reprocessing this compendium and seeing if it Just Works. If not @arielsvn will dig into why not further!

arielsvn:
@georgiadoing I generated a new version of the Rna-Seq compendium for Pseudomonas aeruginosa, try downloading it again. It's now ~86mb. I made sure that all experiment folders contain quant files inside.

I'm closing this, feel free to re-open if you find any other issues with this compendium and thanks for filing the issue!

georgiadoing:
Awesome - just downloaded new compendium -thanks!

PR #2251, opened 2020-04-21
Reduce the RAM requirements for the quantpendia

arielsvn:
## Issue Number

#2246 

We will need to re-process some rna-seq compendias. However, right now they can't run because the smasher uses an `m5.4xlarge` instance with 64GB of RAM and that's the exact amount of RAM required for the job.

This PR reduces the amount of ram required by the RNA-seq compendia so that it can run on the smasher instance.

## Types of changes

## Functional tests

None

## Checklist

### DIFF
--- a/infrastructure/variables.tf
+++ b/infrastructure/variables.tf
@@ -116,8 +116,9 @@ variable "nomad_server_instance_type" {
 }
 
 variable "smasher_instance_type" {
-  # 64GiB Memory, smasher and compendia jobs need 30.
-  default = "m5.4xlarge"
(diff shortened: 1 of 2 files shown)


arielsvn:
At 32gb the job failed for human and mouse the last time we tried. Do you think we should try again?

kurtwheeler:
If the job needs to be that big then I think we might wanna scale up the smasher instance while we run it. Especially since those jobs take days. We don't want downloads to be blocked in the meantime.

arielsvn:
ok I increased the smasher instance to 128gb with `m5.8xlarge`, what do you think @kurtwheeler

PR #2257, opened 2020-04-22
WIP Remove public access from buckets

### LINKED ISSUE #731: Lock down S3 buckets

arielsvn:
**Work in progress:**

- [ ] Additional permission to allow the processor jobs to upload files when the buckets are blocked
- [ ] Confirm that the terraform configuration works
-  [ ] Update to a newer version of terraform that supports [aws_s3_bucket_public_access_block](https://www.terraform.io/docs/providers/aws/r/s3_bucket_public_access_block.html). Looks like it was introduced with [v1.54.0](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1540-december-21-2018) of the terraform aws provider. 

## Issue Number

close #731 

Files were still publicly available because the buckets had a public read only policy.

image

This should make them private. In addition to adding documentation to the computed files endpoint with details to retrieve the download url.

## Types of changes

## Functional tests

Tested in Prod, to ensure everything worked after making the buckets private. This will cause issues when uploading computed files from the jobs.

Still need to test the terraform configuration, for which we'll have to update to a newer version.

## Checklist

### DIFF
--- a/api/data_refinery_api/serializers.py
+++ b/api/data_refinery_api/serializers.py
@@ -250,7 +250,6 @@ class ComputedFileListSerializer(serializers.ModelSerializer):
             "sha1",
             "s3_bucket",
             "s3_key",
-            "s3_url",
             "download_url",
             "created_at",
             "last_modified",
--- a/infrastructure/disk.tf
+++ b/infrastructure/disk.tf
@@ -34,6 +34,13 @@ resource "aws_s3_bucket" "data_refinery_bucket" {
   }
 }
 
+resource "aws_s3_bucket_public_access_block" "data_refinery_bucket" {
+  bucket = "${aws_s3_bucket.data_refinery_bucket.id}"
+
+  block_public_acls   = true
+  block_public_policy = true
+}
+
 resource "aws_s3_bucket" "data_refinery_results_bucket" {
   bucket = "data-refinery-s3-results-${var.user}-${var.stage}"
   acl    = "private"
@@ -59,6 +66,13 @@ resource "aws_s3_bucket" "data_refinery_results_bucket" {
       days = 1
     }
   }
+}
+
+resource "aws_s3_bucket_public_access_block" "data_refinery_results_bucket" {
+  bucket = "${aws_s3_bucket.data_refinery_results_bucket.id}"
+
+  block_public_acls   = true
+  block_public_policy = true
 }
 
 resource "aws_s3_bucket" "data-refinery-static" {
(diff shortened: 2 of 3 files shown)


api/data_refinery_api/serializers.py
```diff
@@ -250,7 +250,6 @@ class Meta:
             "sha1",
             "s3_bucket",
             "s3_key",
-            "s3_url",
```
kurtwheeler, line 253: This can stay! Some computed files can be downloaded with just an S3_url such as transcriptome indices and quantile normalization targets. They are in their own buckets with different permissions.

infrastructure/disk.tf
```diff
@@ -92,6 +106,13 @@ resource "aws_s3_bucket" "data_refinery_transcriptome_index_bucket" {
   }
 }
 
+resource "aws_s3_bucket_public_access_block" "data_refinery_transcriptome_index_bucket" {
+  bucket = "${aws_s3_bucket.data_refinery_transcriptome_index_bucket.id}"
+
+  block_public_acls   = true
+  block_public_policy = true
+}
```
kurtwheeler, line 114: This bucket does not need to be blocked. (See https://github.com/AlexsLemonade/refinebio/issues/731#issuecomment-469443058)

infrastructure/disk.tf
```diff
@@ -103,6 +124,13 @@ resource "aws_s3_bucket" "data_refinery_qn_target_bucket" {
   }
 }
 
+resource "aws_s3_bucket_public_access_block" "data_refinery_qn_target_bucket" {
+  bucket = "${aws_s3_bucket.data_refinery_qn_target_bucket.id}"
+
+  block_public_acls   = true
+  block_public_policy = true
+}
```
kurtwheeler, line 132: This also does not need to be locked down.

arielsvn:
Tested the terraform configuration on a dev stack, and made sure that the bucket public policy was disabled. However, with those settings the processor jobs fail to upload files to the buckets. 

image

PR #2258, opened 2020-04-23
Increase downloads buffer size to 4mb

arielsvn:
## Issue Number

#2237

Increase downloads buffer size to 4mb to see if that prevents the disk from going read-only


cgreene:
can you make the title and comment of this 4MiB though - just so we have the right number

cgreene:
whoops :)

PR #2266, opened 2020-04-27
Add check for empty experiments in datasets.

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio-frontend/pull/890

The frontend no longer posts datasets with empty experiments, but these will cause errors for our smasher jobs so we shouldn't allow them in at the API level either.

## Types of changes

## Functional tests

Relying on unit tests here.

## Checklist


davidsmejia:
I thought we had a shortcut to add all experiments with `{'xxx': 'ALL'}`

davidsmejia:
I guess this won't break that thought

PR #2271, opened 2020-05-13
Remove old contributors from git-crypt

kurtwheeler:
## Issue Number

https://app.zenhub.com/workspaces/refinebio-5c92300a9d1bd30b8d3101f2/issues/alexslemonade/adminitems/254

Only @davidsmejia, @dongbohu, and I have git-crypt access now.

## Functional tests

David and I both tested this works.


davidsmejia:
One thing I want to check is if the permissions for the `.pem` file get reset after we update git-crypt users.

davidsmejia:
Looks like if you `chmod 600 file.pem` then `git-crypt lock` and then `git-crypt unlock` it sets file permissions to `file.pem` will be set to `644`
This is not an issue for this PR though but we should consider adding something that mentions this in the docs.

kurtwheeler:
I added some docs for git crypt to the README. They also explain the `.pem` file.

PR #2281, opened 2020-05-28
Fix git crypt deploy

kurtwheeler:
## Issue Number

Hmm, I don't think there is one. The deploy failed yesterday and this is why.

We re-inited git-crypt when we removed Ariel from it after he left. This meant that the old git-crypt key we had in the repo for circleci to use was no longer valid. This replaces it and adds a script to make doing so easier in the future.

## Types of changes

## Functional tests

I'm able to unlock git-crypt, run the script, relock git-crypt, run the commands output by the script to unlock the git_crypt key, which I can then use to unlock git-crypt again.

Anything more than that is just gonna have to be tested with a deploy.

## Checklist


kurtwheeler:
Oh I should note that I already changed OPENSSL_KEY's value in CircleCI.

PR #2287, opened 2020-05-29
Store where each dataset was downloaded from

### LINKED ISSUE #2217: Store locations from where our users download data
### Context

https://github.com/AlexsLemonade/refinebio/pull/2216#discussion_r401233850

### Problem or idea

Our engagement bot only stores IPs from the users, but we are also interested in their locations which can change over the years.

### Solution or next step

We should modify the `post_engagement_bot_summary` command to also store the location from where people are downloading datasets.

wvauclain:
## Issue Number

Closes #2217 

Store the location of each IP address as calculated the first time we process that dataset annotation by the management script. I put this in the management script instead of in the API because we have to call an external API to get the IP location, which could significantly slow down our response times on the API.

We will have to fix the migration after #2282 lands. I had to change the `base_manager` of DatasetAnnotation from `public_objects` to `objects` because otherwise Django's ORM would not try to modify private datasets. Instead it would try to create a new entry with the same PK and fail.

## Types of changes

## Functional tests

I ran refine.bio locally and downloaded datasets, then ran the management command and made sure that the changes looked right in the database.

## Checklist

## Screenshots


api/data_refinery_api/management/commands/post_downloads_summary.py
```diff
@@ -39,8 +39,19 @@ def handle(self, *args, **options):
...
+        # Save the locations permanently, since IP addresses can cycle over time
+        location_cache = dict()
+        for annotation in annotation_queryset:
+            if not "location" in annotation.data:
+                ip = annotation.data["ip"]
+                print(ip)
+                if not ip in location_cache:
+                    location_cache[ip] = get_ip_location(ip)
+                annotation.data["location"] = location_cache[ip]
+                annotation.save()
```
kurtwheeler, line 51: Why wait until we post a summary to do this? It seems like we could add these annotations when the dataset is created, then just report on them from this command.

wvauclain: `get_ip_location` calls an external API to get the location of each ip address, and generally I think people try to avoid calling external APIs before their own API returns because it causes extra request latency.

kurtwheeler: Ah yes. That's a very good point!

api/data_refinery_api/management/commands/post_downloads_summary.py
```diff
@@ -39,8 +39,19 @@ def handle(self, *args, **options):
             if annotation.data["start"] and should_display_email(annotation.dataset.email_address)
         ]
 
+        # Save the locations permanently, since IP addresses can cycle over time
+        location_cache = dict()
+        for annotation in annotation_queryset:
+            if not "location" in annotation.data:
+                ip = annotation.data["ip"]
+                print(ip)
```
kurtwheeler, line 47: Can we drop this print?

wvauclain: Oh yeah my bad

PR #2312, opened 2020-06-08
Fix deploys, allowing many changes to be released

kurtwheeler:
It's been a while since we've had a deploy work. A lot of changes have built up. I think the change I had blocked by this was about invalid datasets.


wvauclain:
I reran the salmon and API tests, we might want to make sure they pass before we deploy to prod.

kurtwheeler:
I agree. If they don't pass now we can try them again in the morning.

I was going to wait until morning to deploy anyway, just in case.

Issue #2124, opened 2020-02-03
Add pagination for the organisms endpoint.


arielsvn:
### Context

We now have `2249` Organisms in our database.

### Problem or idea

The [organisms endpoint](https://api.refine.bio/v1/#tag/organisms) is not paginated.

### Solution or next step

Add pagination for the organisms endpoint.

wvauclain:
Fixed by https://github.com/AlexsLemonade/refinebio/pull/2319

PR #2319, opened 2020-06-11
Paginated organisms endpoint #2124

BEW111:
## Issue Number

#2124 

Paginated the organisms endpoint using existing default pagination class.

## Types of changes

## Functional tests

Added a test which creates 26 test organisms and checks the name of the first one on the second page, as well as the number on that page (29 organisms are created total so it checks for 4 on the second page).

## Checklist


api/data_refinery_api/tests.py
```diff
@@ -281,6 +285,18 @@ def test_sample_pagination(self):
...
+        response = self.client.get(reverse("organisms", kwargs={"version": API_VERSION}))
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(len(response.json()["results"]), 25)
+
+        # First organism on second page should be TEST_ORGANISM_25
+        response = self.client.get(
+            reverse("organisms", kwargs={"version": API_VERSION}), {"offset": 25}
+        )
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.json()["results"][0]["name"], "TEST_ORGANISM_25")
```
kurtwheeler, line 298: Can you test that the second page has only 1 result?

kurtwheeler: Do the other tests make more organisms or something? If you make 26 organisms and then make the offset 25, shouldn't there only be 1 on the second page?

Issue #2187, opened 2020-03-12
Fix the search to use alternate accessions


kurtwheeler:
### Context

We store an alternate_accession field in the database for experiments because a lot of experiments are in both GEO and Array Express. AlexsLemonade/refinebio#2186 is also going to populate this field with GEO accessions for SRA experiments.

### Problem or idea

I believe the original reason we added that field was so that it didn't matter which kind of accession people used to search. However searching on alternate accessions doesn't seem to work. GSE44094's alternate_accession is E-GEOD-44094 but it doesn't turn up if I search for it: https://www.refine.bio/search?q=E-GEOD-44094.

### Solution or next step

Tweak our elastic search to search on alternate accessions.

PR #2328, opened 2020-06-16
Fix accession code filtering, allow searching by alternate_accession_code

wvauclain:
## Issue Number

#2322
Back-end fixes for #2187 

Allow filtering by accession_code. I also changed accession_code and alternate_accession_code to keyword fields so that filtering by them in the search works better so that we can do #2187. I think the dashes in `E-GEOD-*` was messing with Elasticsearch when they were text fields so `/search/?search=alternate_accession_code:E-GEOD-44094` was not filtering properly.

## Types of changes

## Functional tests

I added a test for `alternate_accession_code`, and I ran the API locally to make sure.

## Checklist

## Screenshots

### DIFF
--- a/api/data_refinery_api/views.py
+++ b/api/data_refinery_api/views.py
@@ -223,6 +223,7 @@ class ExperimentDocumentView(DocumentViewSet):
         "technology": "technology",
         "has_publication": "has_publication",
         "accession_code": "accession_code",
+        "alternate_accession_code": "alternate_accession_code",
         "platform": "platform_accession_codes",
         "organism": "organism_names",
         "num_processed_samples": {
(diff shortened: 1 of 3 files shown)


api/data_refinery_api/views.py
```diff
@@ -222,6 +222,7 @@ class ExperimentDocumentView(DocumentViewSet):
         "id": {"field": "_id", "lookups": [LOOKUP_FILTER_RANGE, LOOKUP_QUERY_IN],},
         "technology": "technology",
         "has_publication": "has_publication",
+        "accession_code": "accession_code",
```
kurtwheeler, line 225: Can we also add `alternate_accession_code` here?

PR #2331, opened 2020-06-16
Do not allow non-downloadable samples to be added to a dataset

### LINKED ISSUE #2329: Validate that samples added to the dataset are downloadable
### Context

We allow datasets like [this](https://www.refine.bio/dataset/98e449b3-b485-4786-8529-79dc50e4cd5b) to be created.

### Problem or idea

This dataset has *at least one experiment* and it also has *at least one sample per experiment*, but it still isn't valid because the samples are not downloadable.

### Solution or next step

The API should also verify that all the samples added to the dataset are downloadable.

wvauclain:
## Issue Number

Fixes #2329

Add more validation to the `validate_dataset` method to disallow more ways to create an invalid dataset.

## Methods

## Types of changes

## Functional tests

I added unit tests

## Checklist

## Screenshots

### DIFF
--- a/api/data_refinery_api/serializers.py
+++ b/api/data_refinery_api/serializers.py
@@ -592,35 +592,53 @@ class ProcessorJobSerializer(serializers.ModelSerializer):
 
 
 def validate_dataset(data):
-    """ Basic dataset validation. Currently only checks formatting, not values. """
-    if data.get("data") is not None:
-        if type(data["data"]) != dict:
-            raise serializers.ValidationError("`data` must be a dict of lists.")
-
-        if data.get("start") and len(data["data"]) == 0:
-            raise serializers.ValidationError("`data` must contain at least one experiment..")
-
-        for key, value in data["data"].items():
-            if type(value) != list:
-                raise serializers.ValidationError(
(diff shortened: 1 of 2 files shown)


api/data_refinery_api/serializers.py
```diff
@@ -592,35 +592,53 @@ class Meta:
...
+        # Otherwise, we will check that all the samples they requested are downloadable
+        else:
+            accessions.extend(value)
 
-    else:
-        raise serializers.ValidationError("`data` must be a dict of lists.")
+    if len(accessions) > 0:
+        processed_samples = Sample.processed_objects.filter(accession_code__in=accessions)
+        if len(accessions) != processed_samples.count():
+            raise serializers.ValidationError("Non-downloadable samples in " + str(accessions))
```
kurtwheeler, line 641: I think we can be more helpful than this. If we changed the query to be `Sample.public_objects.filter(accession_code__in=accessions, is_processed=False)` then we could check if the result is empty. If it's not then we know exactly which accessions weren't processed and we can instead say "The following samples are not downloadable: "

kurtwheeler:
Sweet, that looks good!

Issue #2204, opened 2020-03-24
Potential improvements to the API to make easier to use programatically


kurtwheeler:
### Context

I recently wrote a script that uses our API to help debug https://github.com/AlexsLemonade/refinebio/issues/2191. Here's the script:

```python
"""This script uses refine.bio's API to create a CSV containing a row
for each RNASeq sample which has been processed by salmon and has a
quant.sf file. For each sample the row will contain the sample
accesssion, the experiment accession, and the size of the quant.sf
file."""

import csv

import dateutil.parser
import requests

csv_headers = ["sample_accession_code", "experiment_accession_code", "size_in_bytes"]
api_root = "https://api.refine.bio/v1/"
search_endpoint = api_root + "search/"
experiments_endpoint = api_root + "experiments/"
samples_endpoint = api_root + "samples/"
computed_files_endpoint = api_root + "computed_files/"
results_endpoint = api_root + "computational_results/"


def get_quant_sf_size(sample_accession_code):
    """Looks up the size of the most recently created quant.sf file for the sample.

    If there isn't one None is returned.
    """
    sample_details = requests.get(samples_endpoint + sample_accession_code + "/").json()

    quant_sf_files = []

    for comp_result in sample_details["results"]:
        comp_result_json = requests.get(results_endpoint, params={"id": comp_result["id"]}).json()
        # We filter by id so there's only one result in the query.
        for computed_file_json in comp_result_json["results"][0]["files"]:
            if computed_file_json["filename"] == "quant.sf":
                quant_sf_files.append(
                    {
                        "size_in_bytes": computed_file_json["size_in_bytes"],
                        "created_at": computed_file_json["created_at"],
                    }
                )

    if len(quant_sf_files) == 0:
        return None
    elif len(quant_sf_files) == 1:
        return quant_sf_files[0]["size_in_bytes"]
    else:
        # Return the size for the file that was created most recently.
        latest = dateutil.parser.parse(quant_sf_files[0]["created_at"])
        latest_size = quant_sf_files[0]["size_in_bytes"]

        for quant_sf_file in quant_sf_files:
            parsed_date = dateutil.parser.parse(quant_sf_file["created_at"])
            if parsed_date > latest:
                latest = parsed_date
                latest_size = quant_sf_file["size_in_bytes"]

        return latest_size


# Hit the search endpoint, get all RNASeq experiments.
response_json = requests.get(search_endpoint, params={"limit": 100}).json()

[...]

PR #2351, opened 2020-06-25
#2204 API improvements

BEW111:
## Issue Number

[#2204](https://github.com/AlexsLemonade/refinebio/issues/2204)

* Added read views for some models (which use the same serializers that their list views do):
  * Computational results (by ID), computed files (by ID), jobs (by ID), organisms (by name), original files (by ID), and processors (by ID)
* Added filters for some models
  * Compendia can be filtered by computational result ID (uses `result__id`)
  * Computational results can be filtered by processor ID (uses `processor__id`)
  * Computed files can be filtered by computational result ID (uses `result__id`)
  * Transcriptome indices can be filtered by organism name (uses `organism__name`) or by computational result ID (uses `result_id`)
    * I used Django's double underscore for most of the filters but the `result_id` filter for transcriptome indices ended up working with that name so I just used that, so that's something to note

## Types of changes

## Functional tests

Checked local API endpoints and ran API tests.

## Checklist

## Screenshots


api/data_refinery_api/serializers.py
```diff
@@ -284,6 +388,16 @@ class Meta:
         )
 
 
+class OriginalFileListSerializer(OriginalFileSerializer):
+    pass
```
kurtwheeler, line 392: Can we remove this one since it doesn't do much?

kurtwheeler:
I will say that all the code looks good though, I just want to make sure it works as expected when called.

Issue #2252, opened 2020-04-21
Downloading a dataset with the option `quant_sf_only` only includes processed samples


arielsvn:
### Context

I was trying to debug some missing `quant.sf` files from the quantpendia https://github.com/AlexsLemonade/refinebio/issues/2246

### Problem or idea

Noticed that if the option `quant_sf_only=True` then the dataset will only include `quant.sf` files for processed samples.

I [generated one](https://www.refine.bio/dataset/380eecce-8abf-4fbc-9755-a149845b221d) with two samples, both with associated `quant.sf` files but the processing failed.

### Solution or next step

This seems to be related to the way we set `job_context["input_files"]` in `prepare_files`

https://github.com/AlexsLemonade/refinebio/blob/dev/workers/data_refinery_workers/processors/smashing_utils.py#L69-L88

Where we only include samples that have a smashable result file and throw an error if no one is found.

This should be updated to include all samples with quant files if the option is set. Alternatively we could add a separate nomad job spec to process these, similarly to how we implemented the two compendias.

PR #2370, opened 2020-07-01
Stop failing quant_sf_only jobs with no smashable samples

wvauclain:
## Issue Number

Fixes https://github.com/AlexsLemonade/refinebio/issues/2252

Check for `quant.sf` files when processing datasets with `quant_sf_only=True`. Previously we were only checking for smashable files, which caused processing datasets with `quant_sf_only=True` to fail if the samples were all not smashable, because the smasher would think there are no processable samples.

## Types of changes

## Functional tests

I added a new test for a `quant_sf_only` smasher job, made sure that it failed in the method described in the issue, then fixed the problem.

## Checklist

## Screenshots


api/data_refinery_api/serializers.py
```diff
@@ -682,7 +682,26 @@ def validate_dataset(data):
...
+            results__computedfile__s3_bucket__isnull=False,
+        )
+        if samples_without_quant_sf.count() > 0:
+            raise serializers.ValidationError(
+                "Sample(s) '"
+                + ", ".join([s.accession_code for s in samples_without_quant_sf])
+                + "' in dataset are missing quant.sf files"
+            )
+
+    else:
```
kurtwheeler, line 704: Hmm, I don't think this will work if the experiment specifies `ALL`. In our handling of the `ALL` keyword above we specifically check `processed_public_objects`, and return an error if we don't find any, so we wouldn't make it down here.

Issue #2371, opened 2020-07-01
Make the dashboard more efficient


wvauclain:
### Context

I have been using the dashboard a lot recently while debugging the failed processor jobs.

### Problem or idea

The dashboard still takes like a minute to refresh the cache and times out.

### Solution or next step

Are all of our queries actually necessary? Are we hitting the index properly? Would implementing https://github.com/AlexsLemonade/refinebio/issues/1842 help?

wvauclain:
Ideally we would make the dashboard performant enough that we can drop the cache time and get a fairly real-time look at the running jobs.

wvauclain:
I ran `cProfile` on `Stats.calculate_stats`, and these were the results inside the view file:
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
       10    0.000    0.000    0.000    0.000 views.py:1152(get_start_date)
        1    0.001    0.001   20.186   20.186 views.py:1336(calculate_stats)
        1    0.000    0.000    0.038    0.038 views.py:1395(_get_dataset_stats)
        1    0.000    0.000    1.384    1.384 views.py:1423(_samples_processed_last_hour)
        1    0.000    0.000    4.718    4.718 views.py:1429(_get_input_data_size)
        1    0.000    0.000    1.983    1.983 views.py:1436(_get_output_data_size)
        3    0.000    0.000    5.999    2.000 views.py:1445(_get_job_stats)
        4    0.000    0.000    2.759    0.690 views.py:1497(_get_object_stats)
        7    0.000    0.000    0.007    0.001 views.py:1508(_get_intervals)
```

It might make sense to add query parameters to the `stats` endpoint filtering the output. `_get_input_data_size` and `_get_output_data_size` seem to be for the executive dashboard and not the dashboard, so there's ~7 seconds of wins right there if we can split out the information.

Looking at other lines with a high cumulative time, there is:
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
       16    0.000    0.000   20.054    1.253 compiler.py:1112(execute_sql)
```
so it looks like the vast majority of our time is spent in the database.

wvauclain:
Update: I just merged  #2375, so once that is done we can do some tests and check how much faster the dashboard is, then make changes on the frontend or keep tweaking things on the backend.

wvauclain:
I tried the new dashboard endpoint and this was the result:
```
$ time curl "https://api.refine.bio/v1/stats/?range=day"
[...]

PR #2375, opened 2020-07-02
Tweaks to improve dashboard performance

wvauclain:
## Issue Number

#2371

I split out the stats code more so that if you pass `?dashboard=true` to `/stats/`, it excludes all of the stats that are not on the dashboard. From the profile I took, the rough estimate is that this should result in the dashboard being ~2x faster to load once we update the frontend, but I'll keep testing once this gets deployed.

## Functional tests

I ran the frontend locally, patched it to send `?dashboard=true`, and made sure that none of my changes broke the dashboard.

## Checklist

## Screenshots

### DIFF
--- a/api/data_refinery_api/views/stats.py
+++ b/api/data_refinery_api/views/stats.py
@@ -32,7 +32,7 @@ from data_refinery_common.models import (
 )
 from data_refinery_common.utils import get_active_volumes, get_nomad_jobs_breakdown
 
-JOB_CREATED_AT_CUTOFF = datetime(2019, 6, 5, tzinfo=timezone.utc)
+JOB_CREATED_AT_CUTOFF = datetime(2019, 9, 19, tzinfo=timezone.utc)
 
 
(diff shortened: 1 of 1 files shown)


api/data_refinery_api/views/stats.py
```diff
@@ -32,7 +32,7 @@
 )
 from data_refinery_common.utils import get_active_volumes, get_nomad_jobs_breakdown
 
-JOB_CREATED_AT_CUTOFF = datetime(2019, 6, 5, tzinfo=timezone.utc)
+JOB_CREATED_AT_CUTOFF = datetime(2019, 9, 19, tzinfo=timezone.utc)
```
kurtwheeler, line 35: Any particular reason for this date?

wvauclain: This matches the date on the backend, I think that's why the dashboard shows so many pending jobs that don't exist

PR #2381, opened 2020-07-06
Better error messages for the dataset validator

### LINKED ISSUE #2378: Make the response for non-downloadable samples more programmer-friendly
### Context

Right now the message is a string with an error message and accession codes inside of it.

### Problem or idea

We don't want users to have to parse the error message to get the accessions they need to remove from their dataset.

### Solution or next step

Can we make the serializer error return a dict with the error message and contents instead of a string?

wvauclain:
## Issue Number

Fixes #2378

Made changes to the dataset validation errors so that users of the API do not have to parse the strings of our error messages to get the non-downloadable accessions out.

## Types of changes

## Functional tests

I fixed the unit tests to check for the new error style.

## Checklist

## Screenshots

### DIFF
--- a/api/data_refinery_api/tests.py
+++ b/api/data_refinery_api/tests.py
@@ -631,9 +631,10 @@ class APITestCases(APITestCase):
             content_type="application/json",
         )
         self.assertEqual(response.status_code, 400)
-        self.assertEqual(
-            response.json()["non_field_errors"][0], "Non-downloadable sample(s) '456' in dataset"
+        self.assertIn(
+            "Non-downloadable sample(s) in dataset", response.json()["message"][0],
         )
+        self.assertEqual(response.json()["non_downloadable_samples"], ["456"])
 
         # Good, 789 is processed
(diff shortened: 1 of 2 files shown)


api/data_refinery_api/tests.py
```diff
@@ -631,9 +631,10 @@ def test_dataset_adding_non_downloadable_samples_fails(self):
             content_type="application/json",
         )
         self.assertEqual(response.status_code, 400)
-        self.assertEqual(
-            response.json()["non_field_errors"][0], "Non-downloadable sample(s) '456' in dataset"
+        self.assertIn(
+            "Non-downloadable sample(s) in dataset", response.json()["message"][0],
```
kurtwheeler, line 635: Why is the `[0]` necessary? Is it returning a list of strings instead of just a single string?

wvauclain: Yeah, for some reason Django Rest Framework is transforming the string into a list of strings of length 1

PR #2394, opened 2020-07-08
srastats sometimes lies about the number of reads

wvauclain:
## Issue Number

Salmon continued to not work on prod.

I looked into why, and it turns out that we were detecting that the sample had two reads when it really only had one. I dug into why and found out that srastats was returning that `sra_num_reads` should be 2, but we interpret that as having more than one read. Instead, let's trust the library strategy first because that should hopefully be more accurate.

## Methods

## Types of changes

## Functional tests

## Checklist

## Screenshots


wvauclain:
Salmon was hanging before and now it finishes in minutes.

cgreene:
🤞 🕯️

cgreene:
Do we want to make some sort of test around this? I'm imagining a situation where we use this for a while and then identify another set of samples for which the opposite is true, leading us to eventually flip this back to the other state. Either a test or comment, perhaps, so that we don't later say the other strategy was more accurate :)

PR #2402, opened 2020-07-09
Upgrade to 18.04

wvauclain:
## Issue Number

#2237 

From https://github.com/AlexsLemonade/refinebio/issues/2237#issuecomment-656157201, it looks like there is a kernel bug that is causing the ext4 driver to hit an OOM condition. Looking at issues on the docker GitHub page it looks like people are having issues with 3.x and 4.x kernels, so let's try upgrading to 18.04 which has linux 5.3.

## Types of changes

## Functional tests

## Checklist

## Screenshots

### DIFF
--- a/infrastructure/instances.tf
+++ b/infrastructure/instances.tf
@@ -8,7 +8,7 @@ data "aws_ami" "ubuntu" {
     name   = "name"
     # This is a HVM, EBS backed SSD Ubuntu LTS AMI with Docker version 17.12.0 on it in the US,
     # the stock Ubuntu cloud image in the EU.
-    values = ["${var.region == "us-east-1" ? "ubuntu-16.04-docker-5-19.03.2-*" : "ubuntu/images/hvm-ssd/ubuntu-xenial-16.04-amd64-server-*" }"]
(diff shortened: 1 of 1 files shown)


infrastructure/instances.tf
```diff
@@ -8,7 +8,7 @@ data "aws_ami" "ubuntu" {
     name   = "name"
     # This is a HVM, EBS backed SSD Ubuntu LTS AMI with Docker version 17.12.0 on it in the US,
     # the stock Ubuntu cloud image in the EU.
-    values = ["${var.region == "us-east-1" ? "ubuntu-16.04-docker-5-19.03.2-*" : "ubuntu/images/hvm-ssd/ubuntu-xenial-16.04-amd64-server-*" }"]
+    values = ["${var.region == "us-east-1" ? "ubuntu-18.04-docker" : "ubuntu/images/hvm-ssd/ubuntu-xenial-16.04-amd64-server-*" }"]
```
kurtwheeler, line 11: Why is this a conditional? Shouldn't we update `ubuntu-xenial-16.04-amd64-server` as well?

Issue #2191, opened 2020-03-15
Metadata differences and truncated quant.sf output


erflynn:
I'm working with the Mus Musculus and Homo Sapiens RNA-seq data and I wanted to check on a couple things:

1. For some studies, the samples in the aggregated_metadata.json file do not match the samples in the <experiment_id>_metadata.tsv file. Which should I be looking at? What is the difference between the two?
One example is human study SRP006970, which has four samples listed in the json experiment data (however none in the json sample-level data), but three in the in the directory metadata file.

2. A subset of the quant.sf files are much smaller than the rest and lack headers - they appear to be truncated somehow? What should I do with these?
 Examples of this: SRP166447/SRR8140707_quant.sf, SRP087577/SRR4195593_quant.sf, SRP092166/SRR4455112_quant.sf
 (human)

Thanks!

cgreene:
This is up for this sprint. Ideally we would identify these and re-run them. If we are unable to do this for some reason, we could prune them for now and rerun them later.

kurtwheeler:
Our rerun hasn't addressed the issues. I think I see what got skipped. We need to reprocess these samples from the start. To make sure I got them all I ran the following script on the human RNA-Seq compendium to check the headers of every file:

```python
import json
import os

with open("aggregated_metadata.json") as f:
    metadata = json.load(f)

for experiment_accession_code, experiment in metadata["experiments"].items():
    for filename in [path for path in os.listdir(experiment_accession_code) if "quant.sf" in path]:
        full_path = os.path.join(experiment_accession_code, filename)
        with open(full_path, "r") as f:
            if not f.readline().startswith("Name"):
                print(full_path)
```

The results of running that are in this file: [header_check.txt](https://github.com/AlexsLemonade/refinebio/files/5343793/header_check.txt)

erflynn:
Is this separate from the missing files?

kurtwheeler:
See my comment on that issue: https://github.com/AlexsLemonade/refinebio/issues/2211#issuecomment-705712410

PR #2414, opened 2020-07-13
Filter truncated quant.sf files

wvauclain:
## Issue Number

#2191 

Since the quant.sf files that were causing issues are the ones that are missing their headers, let's filter out files that are missing the correct header.

## Types of changes

## Functional tests

I added a new quant test file that is one of the truncated files from our human quantpendia and modified the quant.sf only smash test to make sure it was filtered.

## Checklist

## Screenshots

### DIFF
--- a/workers/data_refinery_workers/processors/smashing_utils.py
+++ b/workers/data_refinery_workers/processors/smashing_utils.py
@@ -1043,6 +1043,9 @@ def download_quant_file(download_tuple: Tuple[Sample, ComputedFile, str]) -> str
         with open(output_file_path, "r") as f:
             header = f.readline()
             # Filter out files that have been truncated and are missing their header
+            # so that we can avoid https://github.com/AlexsLemonade/refinebio/issues/2191
+            #
+            # From the salmon docs, the header is always the same:
             # https://salmon.readthedocs.io/en/latest/file_formats.html#quantification-file
             if header.rstrip("\n") != "Name\tLength\tEffectiveLength\tTPM\tNumReads":
                 logger.warn(
--- a/workers/data_refinery_workers/processors/smashing_utils.py
+++ b/workers/data_refinery_workers/processors/smashing_utils.py
@@ -1026,16 +1026,39 @@ def write_tsv_json(job_context):
         return [tsv_path]
 
 
-def download_computed_file(download_tuple: Tuple[ComputedFile, str]):
-    """ this function downloads the latest computed file. Receives a tuple with
-    the computed file and the path where it needs to be downloaded
-    This is used to parallelize downloading quantsf files. """
-    (latest_computed_file, output_file_path) = download_tuple
+def download_quant_file(download_tuple: Tuple[Sample, ComputedFile, str]) -> str:
+    """ this function downloads the latest computed file and then returns the
+    failure reason as a string, if there was one. Receives a tuple with the
+    computed file and the path where it needs to be downloaded This is used to
+    parallelize downloading quantsf files. """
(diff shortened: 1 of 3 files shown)


workers/data_refinery_workers/processors/smashing_utils.py
```diff
@@ -1026,16 +1026,39 @@ def write_tsv_json(job_context):
...
+    parallelize downloading quantsf files. """
+    (sample, latest_computed_file, output_file_path) = download_tuple
     try:
         latest_computed_file.get_synced_file_path(path=output_file_path)
-    except:
+    except Exception as e:
         # Let's not fail if there's an error syncing one of the quant.sf files
-        logger.exception("Failed to sync computed file", computed_file_id=latest_computed_file.pk)
+        logger.exception(
+            "Failed to sync computed file", computed_file_id=latest_computed_file.pk, exception=e
```
kurtwheeler, line 1040: `exception=e` isn't necessary here as `logger.exception` will use the current stack traceback info to log the stacktrace and the exception.

workers/data_refinery_workers/processors/smashing_utils.py
```diff
@@ -1026,16 +1026,39 @@ def write_tsv_json(job_context):
...
+        )
+        return "This sample's quant.sf file failed to download"
+
+    try:
+        with open(output_file_path, "r") as f:
+            header = f.readline()
+            # Filter out files that have been truncated and are missing their header
+            # https://salmon.readthedocs.io/en/latest/file_formats.html#quantification-file
+            if header.rstrip("\n") != "Name\tLength\tEffectiveLength\tTPM\tNumReads":
```
kurtwheeler, line 1049: Can you also include a link to the issue for this?

workers/data_refinery_workers/processors/smashing_utils.py
wvauclain: Oh sweet

workers/data_refinery_workers/processors/smashing_utils.py
```diff
@@ -1064,10 +1087,17 @@ def sync_quant_files(output_path, samples: List[Sample], filtered_samples: Dict)
...
             # download this set of files, this will take a few seconds that should also help the db recover
-            executor.map(download_computed_file, sample_and_computed_files)
-            num_samples += len(sample_and_computed_files)
+            for error_reason in executor.map(download_quant_file, sample_and_computed_files):
+                if error_reason is not None:
+                    filtered_samples[sample.accession_code] = {
+                        **sample.to_metadata_dict(),
+                        "reason": error_reason,
+                    }
+                    continue
```
kurtwheeler, line 1099: I don't think this code is nested where you think it is. We loop over the samples in the `sample_page` and add them to `sample_and_computed_files`. Once we've gotten them all we map over the list and run `download_computed_files` on it. In this for loop, you're looping over every error reason from those downloads, and attributing them to `sample`, which is still defined as the last sample in `sample_page`.

I think you should instead be logging from `download_quant_file`. The rest of this doesn't really need to change I don't think.

wvauclain: Good catch, I was originally doing more of the processing in this loop, but I missed that I removed where I updated sample.

kurtwheeler: Oh wait, I forgot this is actually writing to the metadata file. In that case we do need to loop over it like this, however we need to have `download_quant_file` return the sample and the error message, so you can do `for (sample, error_reason) in executor.map(download_quant_file, sample_and_computed_files):`

PR #2425, opened 2020-07-15
Refactor the token API endpoint

### LINKED ISSUE #2423: Make the token endpoint more programmer-friendly
### Context

I was writing up an example of how to create a dataset using the API

### Problem or idea

Just like it should only take one request to create a dataset if you have everything ready, I don't see a good reason why we need two requests to create an API token.

### Solution or next step

Refactor the token routes to be like the refactored dataset viewset.

wvauclain:
## Issue Number

Fixes #2423

Make the same changes I made to the dataset endpoint so we don't have two separate views for creating and updating API keys.

## Types of changes

## Functional tests

I added a test to verify that both the old and the new way of getting API keys works.

## Checklist

## Screenshots

### DIFF
--- a/api/data_refinery_api/tests.py
+++ b/api/data_refinery_api/tests.py
@@ -459,16 +459,11 @@ class APITestCases(APITestCase):
 
         # Create a token first
         response = self.client.post(
-            reverse("token", kwargs={"version": API_VERSION}), content_type="application/json"
-        )
-        token = response.json()
-        token["is_activated"] = True
-        token_id = token["id"]
-        response = self.client.put(
-            reverse("token_id", kwargs={"id": token_id, "version": API_VERSION}),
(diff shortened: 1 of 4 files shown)


cgreene:
We need the person to take an affirmative step to agree to the T&C. Is there a way to make this into an affirmative step agreeing to the T&C? Perhaps the user must pass a separate boolean marking agreement for this to be allowed?

kurtwheeler:
> Is there a way to make this into an affirmative step agreeing to the T&C? Perhaps the user must pass a separate boolean marking agreement for this to be allowed?

The user does in fact need to pass `is_activated=True` in this implementation, so it sounds like we're good! (We're still allowing the previous flow, which is to create an inactive token and then activate it with a separate request.)

Issue #2426, opened 2020-07-16
Narrow all of our exception handling


wvauclain:
### Context

We've had a lot of cases like https://github.com/AlexsLemonade/refinebio/pull/2416 where missing imports or undefined variables have made it to prod because we are catching those exceptions.

### Problem or idea

We are catching things inside `try catch` blocks that really should error out so our tests can catch them.

### Solution or next step

At a minimum, we should stop using `except:`, but probably also `except Exception:` depending on which unintended exceptions that catches.

Make https://lgtm.com/projects/g/AlexsLemonade/refinebio/alerts/?mode=list&id=py%2Fcatch-base-exception have 0 results.

PR #2428, opened 2020-07-16
Clean up exception handling

wvauclain:
## Issue Number

Fixes https://github.com/AlexsLemonade/refinebio/issues/2426

Reduces our LGTM error count #1692

Stop using `except:` and `except Exception:` unless there is a good reason to. Instead, let's just catch the exceptions that we expect will get thrown, and otherwise there's probably a bug anyway.

I started with the API project, and I'll mark this as ready for review once I finish the other projects.

## Types of changes

## Functional tests

The tests still pass.

## Checklist

## Screenshots


wvauclain:
I finished the other projects. Unfortunately we have a lot of functions where I'm not sure what exceptions we are trying to catch, so we still have a lot of `except Exception:` blocks. This isn't ideal, since these still hide errors with missing imports, but they do not catch SIGTERM and other serious errors.

Issue #2424, opened 2020-07-15
Our disk is filling up


wvauclain:
### Context

Ever since we switched to running on SSDs, disk space has been kind of tight, but we could also run into this before.

### Problem or idea

If too many downloader jobs run before the backlog is cleared by processor jobs, the disk of our instance can fill up.

We tried to fix this using a nomad constraint to block downloader jobs from running when disk space dropped below a certain threshold, but unfortunately nomad only calculates disk usage at startup so the constraint was never violated

### Solution or next step

@kurtwheeler suggested limiting how many downloader jobs get placed into the nomad queue, either rate-limiting them or limiting them if there are too many processor jobs in the queue.

wvauclain:
We just deployed https://github.com/AlexsLemonade/refinebio/pull/2430 and are working through some samples. I will keep monitoring the disk usage on our instance.

kurtwheeler:
We were able to run for a few days without this disk filling up! This too seems to be fixed!

PR #2430, opened 2020-07-16
Use the volume work depth to keep track of processor jobs

wvauclain:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/2424

The disk is filling up on our instances because we keep queuing downloader jobs to run even though we have a large backlog of processor jobs. Since we already have logic that stops us from queuing too many downloader jobs at once, let's just also include open processor jobs in our calculations of a volume's work depth.

We may have to fiddle with the `DESIRED_WORK_DEPTH` again, I dropped it down to 200 because I was worried 500 might be too much. We have ~800GB of useable space (a bit more than that but just to be safe), so how much space does a worst-case job take up in terms of downloaded raw data and processor output? 

## Types of changes

## Functional tests

I am running it on a dev stack right now. Things are processing successfully, and the disk is staying empty.

```
ubuntu@ip-10-0-41-123:~$ df -h
Filesystem      Size  Used Avail Use% Mounted on
/dev/xvda1      873G   44G  829G   6% /
```

It would probably be filling up a bit more if I was running a larger instance that could run more jobs in parallel, but I'm on a `4xlarge` since before we filled up the disk easily on it.

## Checklist

## Screenshots

### DIFF
--- a/foreman/data_refinery_foreman/foreman/main.py
+++ b/foreman/data_refinery_foreman/foreman/main.py
@@ -62,18 +62,13 @@ PAGE_SIZE = 2000
 VOLUME_WORK_DEPTH = dict()
 TIME_OF_LAST_WORK_DEPTH_CHECK = timezone.now() - datetime.timedelta(minutes=10)
 
-# The number of processor jobs in the entire database queue
-# TODO: make this by volume
-PROCESSOR_JOBS_IN_QUEUE = 0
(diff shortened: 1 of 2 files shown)


wvauclain:
There is one test in the foreman that we have to fix because it was written assuming that `DESIRED_WORK_DEPTH` is equal to 500.

cgreene:
Worst case raw data could be very large. 100m reads - maybe 30GB or so? Most will be quite a bit smaller though.

PR #2435, opened 2020-07-17
Lots of fixes for LGTM

wvauclain:
## Issue Number

#1692

This PR and the exceptions PR don't totally fix the issue because there are a few that are a bit trickier to fix, but we will be almost there.

Mostly just deleting imports and removing/rewording assignments for variables that are never used.

## Types of changes

## Functional tests

Hopefully the tests still all pass.

## Checklist

## Screenshots

### DIFF
--- a/api/data_refinery_api/views/compendium_result.py
+++ b/api/data_refinery_api/views/compendium_result.py
@@ -125,7 +125,7 @@ class CompendiumResultListView(generics.ListAPIView):
         try:
             token_id = self.request.META.get("HTTP_API_KEY", None)
             # Verify that a token with that id exists
-            _ = APIToken.objects.get(id=token_id, is_activated=True)
+            APIToken.objects.get(id=token_id, is_activated=True)
             return CompendiumResultWithUrlSerializer
         except (APIToken.DoesNotExist, ValidationError):
             return CompendiumResultSerializer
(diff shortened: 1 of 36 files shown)


api/data_refinery_api/views/compendium_result.py
```diff
@@ -125,7 +123,8 @@ def get_queryset(self):
     def get_serializer_class(self):
         try:
             token_id = self.request.META.get("HTTP_API_KEY", None)
-            token = APIToken.objects.get(id=token_id, is_activated=True)
+            # Verify that a token with that id exists
+            _ = APIToken.objects.get(id=token_id, is_activated=True)
```
kurtwheeler, line 127: Honestly we probably don't even need the assignment at all, but this is fine.

common/data_refinery_common/job_management.py
```diff
@@ -117,8 +110,6 @@ def create_downloader_job(
         accession_code = original_downloader_job.accession_code
         original_files = original_downloader_job.original_files.all()
 
-        sample_object = original_files[0].samples.first()
```
kurtwheeler, line 120: Hey this isn't just cleanup! This is saving a database call!

api/data_refinery_api/views/compendium_result.py
cgreene: If you don't need the assignment, just ditch it :)

cgreene, line 127:
```suggestion
            APIToken.objects.get(id=token_id, is_activated=True)
```

cgreene: I see this pattern in a few places. I much prefer just doing the function call instead of having an unused assignment.

Issue #2439, opened 2020-07-21
Upgrade all docker images to use Ubuntu 18.04 as a base


kurtwheeler:
### Context

In #2415 we changed our instance Ubuntu version to 18.04.

### Problem or idea

It seems like it'd be good to match the instance version with the docker image, and it would probably speed up our build times since `apt` will have less to update.

### Solution or next step

Switch all our docker images to use `FROM ubuntu:18.04` and test them thoroughly.

cgreene:
If we're going to make a jump, do we want to go to 20.04 instead so that we don't have to do this again for some time?

kurtwheeler:
Hmm, that seems to make sense. I can't think of any reason not to that we won't run into while testing it.

PR #2458, opened 2020-07-30
Update all of our docker containers

wvauclain:
## Issue Number

Fixes https://github.com/AlexsLemonade/refinebio/issues/2439

Upgrade all of our docker containers to Ubuntu 20.04. Unfortunately CRAN doesn't make 20.04 packages with R less than version 4.0, so we also have to upgrade some of the bioconductor packages.

## Methods

We need to make sure that upgrading R packages/version does not impact our processing. Unfortunately two tests are failing right now for our `no_op` pipeline because the output filesize is not what was expected. This would seem to suggest that something changed from version to version, but I am not sure if this change was benign or serious.

## Types of changes

## Functional tests

All of the docker images build and all (but 2) of the tests pass.

## Checklist

## Screenshots


jaclyn-taroni:
@wvauclain do you have any pointers for how I should direct my attention re: 

> Unfortunately two tests are failing right now for our `no_op` pipeline because the output filesize is not what was expected. This would seem to suggest that something changed from version to version, but I am not sure if this change was benign or serious.

It looks like `main_tests`, `salmon_and_api_tests` and `common_smasher_tests` are all failing and I'm not familiar enough with the test suite to know which one to look at.

jaclyn-taroni:
Looks like `main_tests` -> [`workers/data_refinery_workers/processors/test_no_op.py#L99`](https://github.com/AlexsLemonade/refinebio/blob/dev/workers/data_refinery_workers/processors/test_no_op.py#L99) and [workers/data_refinery_workers/processors/test_no_op.py#L148](https://github.com/AlexsLemonade/refinebio/blob/dev/workers/data_refinery_workers/processors/test_no_op.py#L148).

wvauclain:
Yeah @jaclyn-taroni those are the two tests that are failing. I'll have to look at why those other CircleCI runs are failing but all the other tests passed locally.

jaclyn-taroni:
My _gut reaction_ is that these failures may have to do with our platform detection step (`workers/data_refinery_workers/processors/detect_database.R`), which could be due these packages being updated:

```
illuminaHumanv1.db
illuminaHumanv2.db
illuminaHumanv3.db
illuminaHumanv4.db
illuminaMousev1.db
illuminaMousev1p1.db
illuminaMousev2.db
illuminaRatv1.db
```

I know that you did not _explicitly_ update these annotation packages (https://github.com/AlexsLemonade/sci-projects/issues/51#issuecomment-668007815). [Bioconductor versions are tied to R versions](https://bioconductor.org/about/release-announcements/), so my expectation is that the Illumina annotation packages may be getting updated, too. These packages are not in the release notes for Bioconductor 3.11 but I expect we're a number of minor versions behind.  It could also be an `AnnotationDbi` change, relevant line here: https://github.com/AlexsLemonade/refinebio/blob/f59517ad932e4a78ee16553a460a7f2f77130840/workers/data_refinery_workers/processors/detect_database.R#L59

jaclyn-taroni:
Thanks for getting this started and for providing the list of Bioconductor packages we changed explicitly! Sitting with this a bit more, I think I will need to accomplish some version of #2470 before we get this `sci review`-ed and merged.

api/data_refinery_api/management/commands/post_downloads_summary.py
```diff
@@ -128,8 +128,8 @@ def should_display_email(email: str) -> bool:
         or email.startswith("d.prasad")
         or email.endswith("@alexslemonade.org")
         or email.endswith("@ccdatalab.org")
-        or email is ("daniel.himmelstein@gmail.com")
-        or email is ("dv.prasad991@gmail.com")
+        or email == ("daniel.himmelstein@gmail.com")
```
cgreene, line 131: Daniel doesn't work with us any longer and hasn't for some time - not sure it still makes sense to filter these if he turns out to be using it.

workers/CRAN.gpg
```diff
@@ -1,40 +1,44 @@
 -----BEGIN PGP PUBLIC KEY BLOCK-----
```
cgreene, line 1: Just making sure this is intended.

cgreene (approved):
This all looks fine to me as long as we are happy with that GPG key.

workers/CRAN.gpg
kurtwheeler: Indeed, this got updated.

PR #2472, opened 2020-08-06
#2076 SRA Downloader Handling FTP server correctly

BEW111:
## Issue Number

#2076 

When ENA's FTP server is down, the downloader job now fails instead of just saying that the sample doesn't have unmated reads. I had to change `_has_unmated_reads` function a little by passing in the job so it can give the failure reason for it, and to also return whether or not the function was successful (meaning whether or not it actually connected to the server). I originally had a try-except in `download_sra` to check if it worked, which is shorter, but I thought that that wouldn't be as specific since any error in the function would make it seem like the server was down.

## Types of changes

## Functional tests

Ran downloader tests locally

## Checklist

## Screenshots


workers/data_refinery_workers/downloaders/sra.py
```diff
@@ -297,11 +311,21 @@ def download_sra(job_id: int) -> None:
...
     original_file = original_files.first()
     sample = original_file.samples.first()
-    if _has_unmated_reads(sample.accession_code):
-        original_files = _replace_dotsra_with_fastq_files(sample, job, original_file)
 
     downloaded_files = []
     success = None
+
+    # Try accessing FTP server, see if that was successful
+    unmated_reads, success = _has_unmated_reads(sample.accession_code, job)
```
kurtwheeler, line 319: Hmm, I prefer to use functions with multiple return values as little as possible. I think that maybe you could just remove the first try-catch block from the above function and wrap this function call with it instead. That way this function handles failure to login, and then if the function doesn't raise an exception then the return value can be used to indicate the presence/absence of the files.

workers/data_refinery_workers/downloaders/test_sra.py
```diff
@@ -80,12 +82,48 @@ def test_download_file_ncbi(self):
...
         result, downloaded_files = sra.download_sra(dlj.pk)
         utils.end_downloader_job(dlj, result)
-        self.assertTrue(result)
-        self.assertEqual(downloaded_files[0].sha1, "e7ad484fe6f134ba7d1b2664e58cc15ae5a958cc")
-        self.assertTrue(os.path.exists(downloaded_files[0].absolute_file_path))
+
+        # If the FTP server is down, then we expect that the downloader job should have failed
+        ftp_server = "ftp.sra.ebi.ac.uk"
+        ftp = FTP(ftp_server)
+        if ftp.login()[0:3] == "550":
```
kurtwheeler, line 89: Oh nice, checking for the failure in the tests! I like that, it'll make them less brittle!

workers/data_refinery_workers/downloaders/test_sra.py
```diff
@@ -80,12 +82,48 @@ def test_download_file_ncbi(self):
...
+        assoc.downloader_job = dlj
+        assoc.original_file = og
+        assoc.save()
+        result, downloaded_files = sra.download_sra(dlj.pk)
+        dlj.refresh_from_db()
+        utils.end_downloader_job(dlj, result)
+        self.assertFalse(result)
+        self.assertEqual(dlj.failure_reason, "Failed to connect to ENA server.")
+
+    # @tag("downloaders")
```
kurtwheeler, line 125: Any reason this tag is commented out?

BEW111: that was an accident, thanks for noticing that!

workers/data_refinery_workers/downloaders/sra.py
```diff
@@ -297,11 +298,23 @@ def download_sra(job_id: int) -> None:
...
-        original_files = _replace_dotsra_with_fastq_files(sample, job, original_file)
 
     downloaded_files = []
     success = None
+
+    # Try accessing FTP server, see if that was successful
+    try:
+        if _has_unmated_reads(sample.accession_code, job):
+            original_files = _replace_dotsra_with_fastq_files(sample, job, original_file)
+    except:
```
kurtwheeler, line 309: Can you make sure it's an issue with connecting to the server:
```suggestion
    except ftplib.all_errors:
```
(from https://stackoverflow.com/a/3169751/6095378)

kurtwheeler: (you'll also need to import ftplib instead of FTP.

workers/data_refinery_workers/downloaders/sra.py
```diff
@@ -306,7 +305,7 @@ def download_sra(job_id: int) -> None:
     try:
         if _has_unmated_reads(sample.accession_code, job):
             original_files = _replace_dotsra_with_fastq_files(sample, job, original_file)
-    except:
+    except ftplib.all_errors:
```
wvauclain, line 308: You also need to import `ftplib`.

BEW111: I think I imported it on the first line but the file view hides it

kurtwheeler: I guess it was already imported? I checked and it is in fact on line 1.

wvauclain: Oh weird

Issue #2459, opened 2020-07-31
Transcriptome Index endpoint is 500ing [but only when viewed in the browser]


wvauclain:
### Context

I was trying to make sure we have a transcriptome index for candida, so I hit https://api.refine.bio/v1/transcriptome_indices/?organism_name=CANDIDA_ALBICANS.

### Problem or idea

The above URL 500s, and so does hitting just https://api.refine.bio/v1/transcriptome_indices.

### Solution or next step

Figure out why this is happening. Can I reproduce it locally?

wvauclain:
It isn't happening locally, but it is happening on both staging and prod.

wvauclain:
We did a lot of debugging on this on Friday, and it looks like it is just an issue with the way DRF tries to render the HTML for the API endpoints. If you hit https://api.staging.refine.bio/v1/transcriptome_indices/?format=json it returns the data as requested. This means it's lower priority, because it doesn't impact people consuming our API normally.

wvauclain:
I can also recreate the problem of the API timing out by ssh-ing onto the API instance then running
```
curl -v -H "Accept: text/javascript, text/html, application/xml, */*" http://127.0.0.1:8081/v1/transcriptome_indices/
```
but the server still replies with 0 bytes.

wvauclain:
It's finally fixed! https://api.staging.refine.bio/v1/transcriptome_indices/

PR #2476, opened 2020-08-07
Fix filterset fields on transcriptome index view

wvauclain:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/2459

I was trying to see if there was anything even slightly suspicious about the `TranscriptomeIndexListView`, and I found this. `result_id` is a field defined on the serializer but not on the model itself, which could be why the endpoint is timing out. Maybe.

## Types of changes

## Functional tests

None, the bug only happens on staging and prod.

## Checklist

## Screenshots


wvauclain:
Whoops, never mind. More Django magic; apparently Django creates this field behind-the-scenes because `result` is a `ForeignKey`.

Issue #2467, opened 2020-08-03
Change metadata model to use key-value-confidence


kurtwheeler:
### Context

https://github.com/AlexsLemonade/refinebio/issues/1333
https://github.com/AlexsLemonade/refinebio/pull/2207

Our metadata model is not currently sufficient to represent all the metadata that we need to be able to store. In [this comment](https://github.com/AlexsLemonade/refinebio/issues/2127#issuecomment-591651893), I laid out a flexible representation of metadata fields.

### Problem or idea

I think the JSON structure I laid out above maps nicely to a database table. I'm imaging a `sample_attributes` table with the following fields:
* sample (foreign key to sample)
* mapped_ontology_type (string)
* value_type (string, either bool, numeric, or ontology)
* value (string)
* probability (float)
* unit (string of ontology, optional)

### Solution or next step

Create a model SampleAttribute and create a management command that will populate it with metadata provided as a file. Use the sex labels from https://github.com/AlexsLemonade/refinebio/pull/2207 to test it.

PR #2483, opened 2020-08-11
Implement the new metadata model

wvauclain:
## Issue Number

Closes https://github.com/AlexsLemonade/refinebio/issues/2467

I implemented the key-value-probability metadata model, using my old `OntologyTerm` branch as a starting point.

I mainly focused on sample metadata, since that is all of the external sources of metadata we have right now, but I wrote it in such a way that we could add experiment metadata easily if someone provided us with some.

I also wrote a management command to allow us to import external metadata using the `json` file format @kurtwheeler came up with.

## Types of changes

## Functional tests

I wrote a bunch of tests, and running the management command on some externally supplied metadata behaved as expected locally, skipping all of the samples. We should definitely test this on staging first.

## Checklist

## Screenshots


api/data_refinery_api/test/test_api_general.py
```diff
@@ -310,6 +325,17 @@ def test_experiment_alternate_accession(self):
         self.assertEqual(len(response.json()["results"]), 1)
         self.assertEqual(response.json()["results"][0]["alternate_accession_code"], "E-GEOD-000")
 
+    # Test if we can filter based on metadata supplied by an external contributor
```
kurtwheeler, line 328: Can this comment just be a docstring?

common/data_refinery_common/models/ontology_term.py
```diff
@@ -0,0 +1,116 @@
+import xml.etree.ElementTree as ET
+
+from django.db import models
+
+import requests
+
+from data_refinery_common.logging import get_and_configure_logger
+
+logger = get_and_configure_logger(__name__)
+OLS_URL_TEMPLATE = "http://www.ebi.ac.uk/ols/api/terms?id={}"
```
kurtwheeler, line 10: Can you mention what OLS stands for in a comment? Ontology lookup service?

common/data_refinery_common/models/ontology_term.py
```diff
@@ -0,0 +1,116 @@
...
+            about = child.attrib.get("{" + namespace["rdf"] + "}about")
+            ontology_term = about.split("/")[-1].replace("_", ":")
+            human_readable_name = child.find("rdfs:label", namespace).text
+
+            if OntologyTerm._get_ontology_prefix(ontology_term) == ontology_prefix:
+                term, _ = OntologyTerm.objects.get_or_create(ontology_term=ontology_term)
+                term.human_readable_name = human_readable_name
+                term.save()
+
+        # The her way is <rdf:Description> tags with an <rdfs:label> child
```
kurtwheeler, line 79:
```suggestion
        # The other way is <rdf:Description> tags with an <rdfs:label> child
```

common/data_refinery_common/models/ontology_term.py
```diff
@@ -0,0 +1,116 @@
...
+            ontology_term = about.split("/")[-1].replace("_", ":")
+            human_readable_name = child.find("rdfs:label", namespace).text
+
+            if OntologyTerm._get_ontology_prefix(ontology_term) == ontology_prefix:
+                term, _ = OntologyTerm.objects.get_or_create(ontology_term=ontology_term)
+                term.human_readable_name = human_readable_name
+                term.save()
+
+    @classmethod
+    def poke_term(cls, ontology_term: str):
```
kurtwheeler, line 91: This method is basically identicall to `get_or_create_from_api`. Can that just be called instead of this and just ignore its return value?

wvauclain: Good point

common/data_refinery_common/models/attributes.py
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        if self.value_type == ONTOLOGY_TERM:
+            rendered["value"] = rendered["value"].to_dict()
+
+        return rendered
+
+    def set_value(self, value):
+        """This method sets the attribute value and assigns the correct
+        value_type. NOTE: we assume that all provided strings are ontology terms."""
```
kurtwheeler, line 50: @cgreene is this a good assumption to make? Would we ever want to allow people to just have the value of this metadata be a non-ontology string?

common/data_refinery_common/models/attributes.py
```diff
@@ -0,0 +1,90 @@
...
+        else:
+            raise ValueError("Invalid value_type '{}'".format(self.value_type))
+
+
+class SampleAttribute(AbstractAttribute):
+    sample = models.ForeignKey("Sample", on_delete=models.CASCADE)
+
+
+class ExperimentAttribute(AbstractAttribute):
+    experiment = models.ForeignKey("Experiment", on_delete=models.CASCADE)
```
kurtwheeler, line 90: I think `attributes` would be a good `related_name` for both Sample and Experiment.

common/data_refinery_common/models/attributes.py
```diff
@@ -0,0 +1,90 @@
...
+    value = models.TextField()
+
+    class Meta:
+        abstract = True
+
+    def to_dict(self):
+        rendered = {
+            "name": self.name.to_dict(),
+            "unit": None if self.unit is None else self.unit.to_dict(),
+            "probability": "unknown" if self.probability is None else self.probability,
```
cgreene, line 38: Is this better than reporting none?

common/data_refinery_common/models/attributes.py
cgreene: What did your design document state? I don't understand "the value of this metadata" - can you give a concrete example so that I can try to answer?

common/data_refinery_common/models/attributes.py
```diff
@@ -0,0 +1,90 @@
...
+
+class AbstractAttribute(models.Model):
+    """This is an abstract class that defines all of the properties of a single
+    attribute on either a sample or an experiment. We then subclass this to
+    associate attributes with either an experiment or a sample."""
+
+    name = models.ForeignKey("OntologyTerm", on_delete=models.CASCADE, related_name="+")
+    unit = models.ForeignKey("OntologyTerm", on_delete=models.CASCADE, related_name="+", null=True)
+    probability = models.FloatField(null=True)
+    source = models.TextField()
```
cgreene, line 19: Does this mean that every annotation will duplicate the text in source? If so, we should keep an eye on how many entries there are and how long the string is.

common/data_refinery_common/models/experiment.py
```diff
@@ -124,36 +124,46 @@ def to_metadata_dict(self):
...
-            "cell_line",
-            "treatment",
-            "race",
-            "subject",
-            "compound",
-            "time",
-        ]
+        fields = set()
+
+        possible_fields = set(
```
cgreene, line 140: Do we want to keep this this way, or do we want to start using the ontology terms?

common/data_refinery_common/models/experiment.py
```diff
@@ -124,36 +124,46 @@ def to_metadata_dict(self):
...
-            for sample in samples:
+        for sample in samples:
+            for field in possible_fields - fields:
                 if getattr(sample, field) != None and getattr(sample, field) != "":
-                    fields.append(field)
-                    break
+                    fields.add(field)
+
+            fields |= set(
+                sample.sampleattribute_set.values_list("name__human_readable_name", flat=True)
```
kurtwheeler, line 163: These fields are what show up in the frontend while searching as the metadata fields that the experiment has to offer. @cgreene do we want to include these additional attributes in this list?

common/data_refinery_common/models/attributes.py
wvauclain: That is a good question, I am not sure. I felt like reporting `None` might make it seem like the probability was 100% rather than unreported.

common/data_refinery_common/models/experiment.py
wvauclain: I think this information is only ever used on the frontend where we want human-readable names.

cgreene: Can we get human readable names from the ontology terms though? I'm just imagining a world where we could list any of the available metadata elements.

common/data_refinery_common/models/attributes.py
wvauclain: About the "value of this metadata", an example of a piece of metadata would be `{'name': 'sex', 'value': 'male'}`. The question is if we always want them to report this value as an ontology term, or if there could be cases where they would report it as a string (maybe for different kinds of metadata).

In terms of the source, that is a good point. We can look into ways to intern the string so that the full text is only stored once.

common/data_refinery_common/models/attributes.py
wvauclain: About the "value of this metadata", an example of a piece of metadata would be `{'name': 'sex', 'value': 'male'}`. The question is if we always want them to report this value as an ontology term, or if there could be cases where they would report it as a string (maybe for different kinds of metadata).

foreman/data_refinery_foreman/foreman/management/commands/import_external_sample_attributes.py
```diff
@@ -0,0 +1,160 @@
...
+            )
+            continue
+
+        for name, value in attribute.items():
+            # We want importing sample attributes to be idempotent, so if an
+            # attribute with the same submitter and name already exists, let's
+            # update that instead.
+            #
+            # NOTE: Are there any situations where a sample could have multiple
+            # attributes with the same ontology term name?
```
kurtwheeler, line 60: I don't think so? @cgreene can you think of any?

common/data_refinery_common/models/attributes.py
kurtwheeler: Hmm, I'm pretty sure I imagined these not being strings. I just wanted to make sure we weren't going to box ourselves in here. I can't think of any examples that would really make sense. `{'name': 'name', 'value': 'Kurt Wheeler'}` is the idea, but I don't think we'd ever want to store names... `{'name': 'notes', 'value': 'I sneezed on this sample'}` might also make sense, but I'm not sure where we'd get that kinda data.

Unless you can think of anything that would make sense, I think this is probably fine as is.

common/data_refinery_common/models/experiment.py
wvauclain: Yeah, all those associations are stored in the database. Another option would be to return a list/set of `OntologyTerm`s, then let the API render that as the list of dicts `[{'term': <ontology term>, 'name': <human-readable name>}...]` and let the API consumer decide what to do with it.

wvauclain: I guess that would break querying though, hmm

common/data_refinery_common/models/keywords.py
```diff
@@ -0,0 +1,9 @@
+from django.db import models
+
+
+class SampleKeyword(models.Model):
+    """An ontology term associated with a sample in our database"""
+
+    name = models.ForeignKey("OntologyTerm", on_delete=models.CASCADE, related_name="+")
+    sample = models.ForeignKey("Sample", on_delete=models.CASCADE, related_name="keywords")
+    source = models.TextField()
```
kurtwheeler, line 9: Should this be a contributor?

common/data_refinery_common/models/test_models.py
```diff
@@ -63,3 +70,31 @@ def test_get_sample_metadata_fields_age(self):
...
+        experiment_sample_association.save()
+
+        length = OntologyTerm()
+        length.ontology_term = "EFO:0002939"
+        length.human_readable_name = "medulloblastoma"
+        length.save()
+
+        sk = SampleKeyword()
+        sk.name = length
+        sk.submitter = Contributor.objects.get_or_create(name="Refinebio Tests")
```
kurtwheeler, line 96: This doesn't seem to match the model's `source` field.

foreman/data_refinery_foreman/foreman/management/commands/import_ontology.py
```diff
@@ -0,0 +1,30 @@
...
+        )
+
+    def handle(self, *args, **options):
+        if options.get("ontology", None) is None:
+            logger.error("You must provide an ontology term with --ontology")
+            sys.exit(1)
+
+        OntologyTerm.import_entire_ontology(options["ontology"])
+
+        print(OntologyTerm.objects.all().count())
```
kurtwheeler, line 30: This output seems like it might be a bit unclear.

wvauclain: Good catch, I left that in after debugging something

common/data_refinery_common/models/contributors.py
```diff
@@ -0,0 +1,10 @@
+from django.db import models
+
+
+class Contributor(models.Model):
+    """This model represents an external contributer and stores their name so
+    we can use it as a foreign key on information (e.g. keywords and
+    attributes) supplied by that contributor
+    """
+
+    name = models.TextField(unique=True)
```
kurtwheeler, line 10: Ok so based on the discussion in slack, I'm thinking this model should be:
```
Contributions:
  * source_name
  * methods_url
  * created_at
```

Issue #2240, opened 2020-04-20
It's not possible to run a dev stack outside `us-east-1`


arielsvn:
### Context

I was testing an infrastructure change with a dev-stack deployed on `us-east-2`

### Problem or idea

Looks like some scripts are dependant on the region, and they assume that the app is deployed on `us-east-1`. For example: https://github.com/AlexsLemonade/refinebio/blob/dev/infrastructure/nomad-configuration/client-instance-user-data.tpl.sh#L31-L35

When looking into one of the instances I got the following error:

```
+ EBS_VOLUME_ID=null
+ aws ec2 attach-volume --volume-id null --instance-id i-00162d77206f82d6a --device /dev/sdf --region us-east-2

An error occurred (InvalidParameterValue) when calling the AttachVolume operation: Value (null) for parameter volume is invalid. Expected: 'vol-...'.
```

### Solution or next step

Find the places where `us-east-1` is used and replace it with the region variable.

wvauclain:
If we want to run outside `us-east-1`, we also might have to find a different AMI for the new region. We used to try to do this here:
https://github.com/AlexsLemonade/refinebio/blob/ed0988fedeb18adc74ff26cb4a0cbfbef0778ecd/infrastructure/instances.tf#L11
but since that other AMI doesn't have docker installed it wouldn't have worked anyway.

kurtwheeler:
Closed by https://github.com/AlexsLemonade/refinebio/pull/2498

PR #2498, opened 2020-08-18
Allow dev stacks to run outside us-east-1

BEW111:
## Issue Number

#2240

Replaced the places where `us-east-1` was assumed with the region variable in infrastructure, except in `backend.tf` and default values.

If someone tries to use `deploy.sh` in a region which doesn't have a `ccdl-ubuntu-18.04` ami, it makes a copy of the `us-east-1` one to the new region.

`destroy_terraform.sh` now also takes a region argument.

Remade the `ccdl-ubuntu-18.04` AMI because it was missing python3.5 (there are some older AMIs, one extra that I made and some from July, so I'm thinking that it would be good to keep one from July as a backup in case something goes wrong with the new one I made)

## Types of changes

## Functional tests

- [x] `deploy.sh` and `destroy_terraform.sh` work in `us-east-1` and `us-east-2`
- [x] Nomad runs jobs successfully
- [x] Smasher has region env var

## Checklist


infrastructure/deploy.sh
```diff
@@ -138,6 +138,21 @@ done
 # Copy ingress config to top level so it can be applied.
 cp deploy/ci_ingress.tf .
 
+# Check if a new ccdl-ubuntu ami will be needed for this region
+if [[ $(aws ec2 describe-images --region $TF_VAR_region --owners 589864003899 --filters 'Name=name,Values=ccdl-ubuntu-18.04-*' --query 'length(Images)') -eq 0 ]]; then
+    echo "No ccdl-ubuntu-18.04 AMI found for this region, creating a new one"
+
+    # Find most recent ccdl-ubuntu ami from us-east-1
+    template_ami_id=$(aws ec2 describe-images --region us-east-1 --owners 589864003899 --filters 'Name=name,Values=ccdl-ubuntu-18.04-*' --query 'sort_by(Images,&CreationDate)[-1].ImageId' --output text)
```
kurtwheeler, line 146: These are some long lines. Makes it harder to read. When bash commands get really long like this I like using `\` to put one option per line. I believe there's some examples of this in this file.

infrastructure/destroy_terraform.sh
```diff
@@ -50,11 +56,15 @@ if [[ -z $TF_VAR_user ]]; then
     exit 1
 fi
 
+if [[ -z $TF_VAR_region ]]; then
+    TF_VAR_region=us-east-1
```
kurtwheeler, line 60: I think you'll need `export` here as well.

infrastructure/list_hosts.sh
```diff
@@ -8,10 +8,10 @@ if [[ $# -eq 0 ]] ; then
...
+    AWS_REGION=us-east-1
 else
-    REGION=$2
+    AWS_REGION=$2
 fi
 
-# /Users/rjones/Library/Python/2.7/bin/aws ec2 describe-instances --region=$REGION --filters "Name=tag:User,Values=$1" | grep PublicDnsName | tr -d '"' | sed "s/PublicDnsName: //g" | tr -d "," | awk '{$1=$1};1' | uniq
-aws ec2 describe-instances --region="$REGION" --filters "Name=tag:User,Values=$1" | grep PublicDnsName | tr -d '"' | sed "s/PublicDnsName: //g" | tr -d "," | awk '{$1=$1};1' | uniq
+# /Users/rjones/Library/Python/2.7/bin/aws ec2 describe-instances --region=$AWS_REGION --filters "Name=tag:User,Values=$1" | grep PublicDnsName | tr -d '"' | sed "s/PublicDnsName: //g" | tr -d "," | awk '{$1=$1};1' | uniq
+aws ec2 describe-instances --region="$AWS_REGION" --filters "Name=tag:User,Values=$1" | grep PublicDnsName | tr -d '"' | sed "s/PublicDnsName: //g" | tr -d "," | awk '{$1=$1};1' | uniq
```
kurtwheeler, line 17: Can you also break this line up a bit and remove the comment? That's definitely no longer needed.

Issue #1813, opened 2019-10-21
Remove files from S3 when models are deleted with `pre_delete` signal


arielsvn:
### Context

Some models have data on S3.

https://github.com/AlexsLemonade/refinebio/pull/1759#discussion_r337095064

### Problem or idea

When they are deleted we need to also delete the files on S3. This applies for `OriginalFile` and `ComputedFile`

### Solution or next step

We should add a `pre_delete` signal to have these files deleted automatically when the db model are deleted.

arielsvn:
Example code for this https://github.com/gothinkster/django-realworld-example-app/blob/master/conduit/apps/articles/signals.py

davidsmejia:
Closed with #2524

PR #2524, opened 2020-08-25
Remove S3 files on pre_delete

BEW111:
## Issue Number

#1813 

S3 files from `computed_file` or `original_file` models are deleted on the `pre_delete` signal for either of those. Also added a `delete_s3_file` function for `original_file.py` which is the same as the one from `computed_file.py`.

## Types of changes

## Functional tests

- [x] Deleting a computational result will delete the S3 files of the associated computed files
- [ ] Deleting an original file deletes its S3 files

## Checklist


common/data_refinery_common/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+
+class CommonAppConfig(AppConfig):
+    name = "data_refinery_common"
+
+    # This will be called multiple times
+    def ready(self):
+        import data_refinery_common.signals
+
+
+default_app_config = "data_refinery_common.CommonAppConfig"
```
kurtwheeler, line 12: What's this file doing?

BEW111: This is modifying the app config so that it imports `signals.py`

kurtwheeler: Oh I see... In resources portal we've just been adding the signals to the model file. If you do that does this become unnecessary? I think I actually prefer having the signals in the model file because it's more obvious why a model would behave differently.

kurtwheeler (approved):
My last concern wasn't a big deal, let's just get this merged!

Issue #2525, opened 2020-08-26
`source_version` for Ensembl release could have a better name!


jaclyn-taroni:
### Context

https://github.com/AlexsLemonade/refinebio/issues/2497#issuecomment-680907888

### Problem or idea

We report the Ensembl release version used to generate transcriptome indices in a field called `source_version`. This name could be more descriptive.

### Solution or next step

Change this name to be more descriptive/informative. Here are some ideas:

- `release_version` if we want to keep things pretty general (e.g., not tied to Ensembl explicitly)
- `ensembl_release` if we want to be specific

PR #2527, opened 2020-08-28
Transcriptome Index Field Changes

BEW111:
## Issue Number

#2525 

Renamed `source_version` to `release_version` for transcriptome indices

Added a field `database_name` which will be `EnsemblMain` for the main Ensembl, and `EnsemblPlants`, `EnsemblMetazoa`, etc. for the other divisions. The field defaults to just `Ensembl` for now, so I used `EnsemblMain` to distinguish between new transcriptome indices in the main division and any of the old ones.

The method of getting the division is a little weird but from what I've seen the ftp url contains the division unless it is the main division.

## Types of changes

## Functional tests

Ran transcriptome jobs for EnsemblMetazoa and EnsemblProtists locally

## Checklist


workers/data_refinery_workers/processors/transcriptome_index.py
```diff
@@ -155,6 +155,30 @@ def _extract_assembly_information(job_context: Dict) -> Dict:
...
+                else:
+                    database_name = "Ensembl" + str.capitalize(division_name)
+            except:
+                # This may be a test without a source_url
+                if not og_file.source_url:
+                    logger.info(
+                        "Failed to retrieve/check for division name from url (looks like there was no source_url for the original file)"
+                    )
+                else:
+                    logger.error("Failed to retrieve/check for division name from url")
```
kurtwheeler, line 178: Hmm, I think it might better to fail the job, so we know that we need to fix something. It also sounds like the tests should set `source_url` on the original file so they don't have to be handled specially.

kurtwheeler (approved):
I just realized the only failing test on this was codecov!!! :raises_fist:

Issue #2535, opened 2020-09-03
Swagger is broken


cgreene:
### Context

I was trying to use swagger, which is linked from our API page.

### Problem or idea

Everything fails. @davidsmejia found that it was due to issues with the redirect.

### Solution or next step

@davidsmejia points to https://drf-yasg.readthedocs.io/en/stable/settings.html#default-api-url

PR #2536, opened 2020-09-03
swagger https on prod always

davidsmejia:
## Issue Number

#2535 

Adds a url for swagger to determine the supported schemes

I would also recommend basing this off of the env down the road but this will make the local, stage, and prod all create requests against prod.


kurtwheeler (approved):
Looks good for now!

Issue #2539, opened 2020-09-21
Refinebio API Python Client Requirements


kurtwheeler:
### Data Model Classes

All data models should have a corresponding class.
* However these models won't be necessary for users to use.
* I don't think it's worth implementing yet, but I think it should be easy to use a cache for objects retrieved from the API.
* All Model classes should have `get` and `list` class method.
     * get should default to accession code (for samples/experiments), but optionally take refinebio IDs instead.
          * We'll probably need some logic to detect the accession type and use them appropriately (accession vs alternate_accession)
    * list should accept all the filter parameters the corresponding endpoint does.
* Experiments will need some additional functionality:
    * search (there's a chance this should just be called list but use the search endpoint)
    * `get_uploader_contact()` will return the contact information for the uploader
* Samples:
    * `get_uploader_contact()` should call the corresponding experiment's `get_uploader_contact()` method.
* Datasets will need a number of special methods:
    * process() will send the dataset to refine.bio for processing
    * check() will see if the dataset has finished processing or not.
    * download(path) will download the dataset to disk if it has finished processing

### High Level Functions

We'll need a number of high level functions:
* `help(entity=none)`
    * if called with no entity, gives high level overview of how to use. If not, prints docstring for method.
* `download_dataset(dataset_dict=None, experiments=None, samples=None, aggregation="EXPERIMENT", transformation="NONE", skip_quantile_normalization=False)`
 * `download_compendium(organism, path, quant_sf_only=False)`
 * `download_quantfile_compendium(organism, path)`
    * will just call `download_compendium(organism, path, quant_sf_only=True)`

### API Tokens

We'll need to have good support for the tokens. The following functions will be needed:
* `create_token(email_address)`
* `agree_to_terms_and_conditions(api_token)`
* `save_token(api_token, file_path=os.getenv("CONFIG_FILE", "~/.refinebio.yaml")`
* `load_token(file_path=os.getenv("CONFIG_FILE", "~/.refinebio.yaml"))`

I'd like to do the config file as a YAML file so we can add additional keys to it in the future if that ever becomes necessary. (Off the top of my head I could think of maybe allowing someone to turn on the cache with them if we ever built that.)
    

### Documentation/Error messages

These will be really important.
We'll want to thoroughly validate all the parameters to all functions so if any are invalid we can explain it nicely.
All classes, methods, and functions should have a docstring that explains all parameters, along with an easy-to-understand description of what it can do and how to use it.

[...]

PR #2540, opened 2020-09-23
WIP Python Client

samuel-tate:
## Issue Number

#2539 

## Methods

## Types of changes

## Functional tests

## Checklist

## Screenshots


python_client/pyrefinebio/sample.py
```diff
@@ -0,0 +1,197 @@
...
+            "compound": compound,
+            "time": time,
+            "is_processed": is_processed,
+            "is_public": is_public,
+            "limit": limit,
+            "offset": offset,
+            "dataset_id": dataset_id,
+            "experiment_accession_code": experiment_accession_code,
+            "accession_codes": accession_codes,
+        }
```
kurtwheeler, line 182: For this, I don't think we should have the string fields default to `""`. I think we should only post the ones that are passed in. There's an easy way to do this with `def search(**kwargs):` (see https://realpython.com/python-kwargs-and-args/). Then you can just do `get("samples", params=kwargs)`, after you verify that they're passing in valid fields/types.

python_client/pyrefinebio/http.py
```diff
@@ -0,0 +1,65 @@
...
+        logging.error("GET %s failed", endpoint)
+        return None
+
+    return results
+
+
+def post(endpoint, payload=None):
+    try:
+        response = requests.post(
+            "http://localhost:8000/v1/" + endpoint + "/",
```
kurtwheeler, line 33: We should probably make this a constant in one place, and then for testing you can switch it back and forth as necessary.

cgreene: Can you load from an environment variable if available? Makes testing even easier. And I agree with a constant in one place for the default :)

python_client/pyrefinebio/http.py
```diff
@@ -0,0 +1,65 @@
...
+logger = logging.getLogger(__name__)
+
+
+def get(endpoint, params=None):
+    url = ""
+
+    if endpoint.startswith("http"):
+        url = endpoint
+    else:
+        url = "https://api.refine.bio/v1/" + endpoint + "/"
```
kurtwheeler, line 14: I like the idea that we force HTTPS, but I don't think this is quite the right way to do it.

I also think that if you want to allow the user to pass a URL instead an endpoint, we should just have two functions for that. Maybe `get()` and `get_by_endpoint()`. Then you can just have `get_by_endpoint()` call `get()` with the right URL.

python_client/pyrefinebio/http.py
```diff
@@ -0,0 +1,65 @@
...
+
+    try:
+        response = requests.get(url, params=params)
+
+        response.raise_for_status()
+        results = response.json()
+        # cache results?
```
kurtwheeler, line 21: Not yet, maybe ever.

python_client/pyrefinebio/sample.py
```diff
@@ -0,0 +1,204 @@
...
+        self.race = race
+        self.subject = subject
+        self.compound = compound
+        self.time = time
+        self.is_processed = is_processed
+        self.created_at = created_at
+        self.last_modified = last_modified
+        self.original_files = original_files
+        self.computed_files = computed_files
+        self.experiment_accession_codes = computed_files
```
kurtwheeler, line 125: This seems wrong. Also, I think samples will need an `experiments` field as well. This should use `@property` (see https://www.programiz.com/python-programming/property) to first check if the field is populated, and if it's not use `experiment_accession_codes` to populate the field. This way we won't have to populate the field until the user tries to use it.

python_client/pyrefinebio/sample.py
```diff
@@ -0,0 +1,204 @@
...
+        ex:
+        >>> import pyrefinebio
+        >>> accession_code = "GSM000000"
+        >>> sample = pyrefinebio.Samples.get(accession_code)
+
+    search for sampes based on filters
+
+        ex:
+        >>> import pyrefinebio
+        >>> samples = pyrefinebio.Samples.search(is_processed=True, specimen_part="soft-tissue sarcoma")
```
kurtwheeler, line 23:
```suggestion
        >>> samples = pyrefinebio.Sample.search(is_processed=True, specimen_part="soft-tissue sarcoma")
```

python_client/pyrefinebio/sample.py
```diff
@@ -0,0 +1,204 @@
...
+
+class Sample:
+    """Samples.
+
+    get a sample based on accession code
+
+        ex:
+        >>> import pyrefinebio
+        >>> accession_code = "GSM000000"
+        >>> sample = pyrefinebio.Samples.get(accession_code)
```
kurtwheeler, line 17:
```suggestion
        >>> sample = pyrefinebio.Sample.get(accession_code)
```

python_client/pyrefinebio/sample.py
```diff
@@ -0,0 +1,204 @@
+from pyrefinebio.common.annotation import Annotation
+from pyrefinebio.common.organism_index import OrganismIndex
+from pyrefinebio.http import get
+from pyrefinebio.organism import Organism
+from pyrefinebio.processor import Processor
+from pyrefinebio.util import generator_from_pagination
+
+
+class Sample:
+    """Samples.
```
kurtwheeler, line 10:
```suggestion
    """Sample.
```

python_client/pyrefinebio/sample.py
```diff
@@ -0,0 +1,204 @@
...
+
+    get a sample based on accession code
+
+        ex:
+        >>> import pyrefinebio
+        >>> accession_code = "GSM000000"
+        >>> sample = pyrefinebio.Samples.get(accession_code)
+
+    search for sampes based on filters
```
kurtwheeler, line 19:
```suggestion
    search for samples based on filters
```

kurtwheeler (changes requested):
I've left some comments, but this looks great so far!

python_client/pyrefinebio/http.py
samuel-tate: This was just a temporary thing to deal with calling the `next` urls when iterating through the pagination because they had the full url and I originally wrote this to use just the endpoints.
I'll split it into the two functions during the next commit though!

kurtwheeler (approved):
This looks great! Nice work!

PR #2549, opened 2020-09-29
Fix the way the user instance data script was prepared for the nomad nodes.

kurtwheeler:
## Issue Number

N/A, I was trying to run some stuff and tracked this down

The node instance startup script wasn't really running because it was `mv`ing the base script instead of `cp`ing it. I checked the instance's user data script and its first line was `exit 0`.

## Types of changes

- Devops


davidsmejia (approved):
LGTM!

Issue #2363, opened 2020-06-29
Clean up the logic around EBS creation and mounting


wvauclain:
### Context

https://github.com/AlexsLemonade/refinebio/pull/2360#pullrequestreview-439453100

### Problem or idea

We have a lot of things commented out related to creating and mounting EBS volumes.

### Solution or next step

Once we know that working without EBS works, we can remove all of the commented-out code relating to EBS.

kurtwheeler:
I cleared this out in the Batch PR.

Issue #2234, opened 2020-04-13
Find a better way to mount the volumes in the instances


arielsvn:
### Context

https://github.com/AlexsLemonade/refinebio/pull/2227

### Problem or idea

Processing instances get the correct volume attached but it doesn't get mounted correctly sometimes.

I was able to get the volume mounted manually by running the following commands from [client-instance-user-data.tpl.sh](https://github.com/AlexsLemonade/refinebio/blob/dev/infrastructure/nomad-configuration/client-instance-user-data.tpl.sh#L59-L73) on the instance:

```bash
ATTACHED_AS=`lsblk -n --sort SIZE | tail -1 | cut -d' ' -f1`
mount /dev/$ATTACHED_AS /var/ebs/
chown ubuntu:ubuntu /var/ebs/
```

#2227 was an attempt at solving that by increasing the time we wait before calculating what volume needs to be mounted.

### Solution or next step

That didn't work, so we'll need to either increase the timeout or find a more sophisticated way to detect when the volume is attached and can be mounted.

kurtwheeler:
This is also made obsolete by Batch!

Issue #2121, opened 2020-02-03
Create a mechanism to change number of volumes without a deploy


kurtwheeler:
### Context

When we want to spin up capacity we have to deploy so we can create volumes. Often times we'll deploy just to do this.

### Problem or idea

Creation of volumes isn't that big of an operation. We should just create a script on the deploy box we can run to change the number of volumes we have in prod without a deploy.

### Solution or next step

A script on the deploy box would be a way to accomplish this, but not necessarily the only one. Open to other ideas.

kurtwheeler:
This is no longer necessary because of Batch!

PR #2552, opened 2020-09-30
Larger smasher volumes

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/1733

Also relevant:
* https://github.com/AlexsLemonade/refinebio/issues/2121 (because at this point, the client instances can run without EBS volumes, and only the smasher needs it, so we don't need to make a number of EBS volumes anymore.)
* https://github.com/AlexsLemonade/refinebio/issues/2363 (still true)
* https://github.com/AlexsLemonade/refinebio/issues/2234

This PR mounts a large EBS volume onto the smasher instance and fixes the needed permissions.

It also tweaks the sizes of things for some processing I want to do.

## Types of changes

- Devops

### DIFF
--- a/README.md
+++ b/README.md
@@ -894,7 +894,7 @@ This can take a long time (>30 minutes)!
 
 ### Tearing Down
 
-A stack that has been spun up via `deploy.sh -u myusername -e dev` can be taken down with `destroy_terraform.sh  -u myusername -e dev`.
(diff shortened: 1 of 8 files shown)


davidsmejia (approved):
LGTM

PR #2559, opened 2020-10-01
Enable full stack on staging to do some debugging, make it cheaper too

kurtwheeler:
## Issue Number

#2552 

The smasher volume in prod didn't get it's volume mounted correctly. I fixed it manually by just running the steps out of the startup script, so I'm not sure why it didn't work on its own, but I can't tell in prod because the cloud init logs get cleared.

This sets full_stack to true while also allowing me to make a smaller smasher volume so it's not as expensive.

## Types of changes

- Devops


davidsmejia (approved):
smol

cgreene (approved):

Issue #2560, opened 2020-10-05
Validate request filter fields


kurtwheeler:
### Context

Sam's been working on the API client. It's come up that we don't return error messages to users when they pass us bad filter fields.

### Problem or idea

We should make sure that filter fields that are specified in all requests have the correct data types and are valid fields to filter on.

### Solution or next step

The best way I can think to do this would be to do it in the [get_queryset](https://github.com/AlexsLemonade/refinebio/blob/dev/api/data_refinery_api/views/compendium_result.py#L105) method of every view. You can check that `self.request.query_params` only contain fields either on the model or listed in `filterset_fields`. Additionally, I think that method also handles special cases like `latest_version` like in the one I linked you to. Let me know if there's any questions.

PR #2562, opened 2020-10-06
Validate Filters

samuel-tate:
## Issue Number

#2560 

Validate filters and return a nice error message if invalid filters have been provided.

## Methods

## Types of changes

## Functional tests

## Checklist

## Screenshots

### DIFF
--- a/api/data_refinery_api/exceptions.py
+++ b/api/data_refinery_api/exceptions.py
@@ -5,11 +5,9 @@ from rest_framework.views import exception_handler
 def custom_exception_handler(exc, context):
     response = exception_handler(exc, context)
 
-    if response:
-        if isinstance(exc, APIException):
-            response.data["error_code"] = exc.detail.code
-
+    if response is not None:
         if isinstance(exc, InvalidFilters):
+            response.data["error_type"] = exc.default_code
             response.data["invalid_filters"] = exc.invalid_filters
 
     return response
--- a/api/data_refinery_api/exceptions.py
+++ b/api/data_refinery_api/exceptions.py
@@ -1,0 +1,7 @@
+from rest_framework.exceptions import APIException
+
+
+class InvalidFilters(APIException):
+    status_code = 400
+    default_detail = "You have supplied invalid filters"
+    default_code = "invalid_query_parameters"
--- a/api/data_refinery_api/exceptions.py
+++ b/api/data_refinery_api/exceptions.py
@@ -1,7 +1,25 @@
 from rest_framework.exceptions import APIException
+from rest_framework.views import exception_handler
+
+
+def custom_exception_handler(exc, context):
+    response = exception_handler(exc, context)
+
+    if response:
+        if isinstance(exc, APIException):
+            response.data["error_code"] = exc.detail.code
+
+        if isinstance(exc, InvalidFilters):
+            response.data["invalid_filters"] = exc.invalid_filters
+
+    return response
 
 
 class InvalidFilters(APIException):
     status_code = 400
-    default_detail = "You have supplied invalid filters"
-    default_code = "invalid_query_parameters"
+    default_detail = "You have supplied invalid filters. A list of the filters that were invalid are provided in the invalid_filters key."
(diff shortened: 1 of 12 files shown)


api/data_refinery_api/test/test_api_general.py
```diff
@@ -286,7 +286,7 @@ def test_all_endpoints(self):
 
         response = self.client.get(
             reverse("transcriptome_indices", kwargs={"version": API_VERSION})
-            + "?organism__name=DANIO_RERIO"
+            + "?organism_name=DANIO_RERIO"
```
kurtwheeler, line 289: I think that extra underscore was intentional, it denotes that `name` is a subfield of `organism`.

api/data_refinery_api/utils.py
```diff
@@ -0,0 +1,24 @@
+def check_filters(view, special_filters=None):
```
kurtwheeler, line 1: This function is defined both here and in common. I think this is the more appropriate place though since I don't think it'd be used outside the API.

kurtwheeler (changes requested):
This looks great! The one thing I think is an easy way to parse the invalid fields. I think having them tucked away in a string in the response is less ideal than having them be their own field in the response.

I think a JSON response like this would be nice:
```json
{
  "detail": "You have supplied invalid filters. A list of the filters that were invalid are provided in the invalid_filters key.",
  "invalid_filters": ["foo", "bar", "baz"]
}
```

I think this can be accomplished by adding a `invalid_filters` field to the custom exception you added, and then implementing a [custom exception handler](https://www.django-rest-framework.org/api-guide/exceptions/#custom-exception-handling) to check if the exception is your custom type and add that key to the response data.

api/data_refinery_api/test/test_api_general.py
samuel-tate: ok it looks like it'll need to be changed in `trancriptome_indices` then

kurtwheeler: I think that's correct. There's no `organism_name` field on the OrganismIndex model.

api/data_refinery_api/exceptions.py
```diff
@@ -0,0 +1,25 @@
+from rest_framework.exceptions import APIException
+from rest_framework.views import exception_handler
+
+
+def custom_exception_handler(exc, context):
+    response = exception_handler(exc, context)
+
+    if response:
+        if isinstance(exc, APIException):
+            response.data["error_code"] = exc.detail.code
```
kurtwheeler, line 10: What's this doing? Is this an HTTP response code?

api/data_refinery_api/test/test_api_general.py
```diff
@@ -306,6 +306,9 @@ def test_all_endpoints(self):
         response = self.client.get(reverse("create_dataset", kwargs={"version": API_VERSION}))
         self.assertEqual(response.status_code, status.HTTP_405_METHOD_NOT_ALLOWED)
 
+        response = self.client.get(reverse("samples", kwargs={"version": API_VERSION}) + "?foo=bar")
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
```
kurtwheeler, line 310: Could you add an assertion to make sure `foo` made it into response's `invalid_filters`?

api/data_refinery_api/exceptions.py
samuel-tate: It's different than the response code number like `400`/`403`. Maybe `type` would be a better field name. 
I think it would be good to return some kind of identifier for the type of error so that, if we ever add more custom 400 exceptions, the client could differentiate between them.

kurtwheeler: Ah ok. Yeah I'm cool with adding an `error_type` field as long as its human readable.

kurtwheeler (approved):
LGTM, although `error_type` would be preferred over `error_code` so savvy users can make use of it without the client's help.

PR #2564, opened 2020-10-07
Reset quantpendia RAM, up compendia RAM, increase smasher size.

kurtwheeler:
## Issue Number

Hmm, not sure which is most appropriate. We want to rerun compendia though.

This makes the smasher and compendia job settings large enough to run human/mouse compendia jobs.

## Types of changes

- Devops


davidsmejia (approved):
LGTM

PR #2566, opened 2020-10-08
Add small sleep to allow EBS volume to be attached.

kurtwheeler:
## Issue Number

#2565 

I think this will fix auto mounting of the EBS volume.

## Types of changes

- Devops


davidsmejia (approved):
👍

PR #2568, opened 2020-10-09
Set full_stack=False now that I no longer need to debug in the staging stack.

kurtwheeler:
## Issue Number

#2345 

Undo the full stack now that debugging is finished


davidsmejia (approved):
LGTM

PR #2563, opened 2020-10-06
Change setting to disable ARP for our instances.

kurtwheeler:
## Issue Number

N/A API went incommunicado again Friday

## Purpose/Implementation Notes

Apparently this happens if you don't have the setting. See https://medium.com/spaceapetech/what-the-arp-is-going-on-b4bc0e73e4d4

I don't know why that isn't set by default...

## Types of changes

- Bugfix (non-breaking change which fixes an issue)

## Functional tests

I can't really, although I can confirm it doesn't fail if I change the setting the other way the tool recommended.

PR #2572, opened 2020-10-14
Remove jobs from original_files list endpoint for performance.

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio/pull/2563 doesn't seem to have done it

Our API has still been freezing up and not even responding to SSH. It seems like we can now reproduce this by hitting the original files endpoint.

I'm not sure this'll fix it, but this does seem to be the only list endpoint that links to jobs, and it's the only one we know of that reliably has this issue, so it seems like it might work. I also don't think the list endpoint needs to link to the jobs, since the detail one does anyway and I think the jobs are less generally useful.

I'll also note that this doesn't happen to staging so I think that these job tables are probably too big in prod because of how many jobs we've run over the years.

## Types of changes

## Functional tests

I'm gonna let the unit tests run.

## Checklist


davidsmejia (approved):
LGTM

PR #2573, opened 2020-10-14
Prevent DRF's browsable API from crashing our server

kurtwheeler:
## Issue Number

#2572 

See http://masnun.com/2016/04/20/django-rest-framework-remember-to-disable-web-browsable-api-in-production.html for context

This should disable the browsable API, and then we already have the API base route set to direct to the redoc page so as long as that's where someone lands for documentation it should Just Work :tm: 

## Types of changes

## Functional tests

I got the redoc page to work locally at the base URL!

## Checklist


davidsmejia (approved):
🤖

PR #2574, opened 2020-10-14
Readd jobs to original file endpoint since we know what the actual issue was now.

kurtwheeler:
#2351  Issue Number

#2573 worked!!

Now that we know what was breaking the original_files endpoint, we don't need this shot in the dark! I'd rather leave it as it was in case it messes with any of the job reporting stuff Ariel did.

## Types of changes


davidsmejia (approved):
🚀

PR #2575, opened 2020-10-14
Prevent original_files endpoint from crashing our server

kurtwheeler:
## Issue Number

#2573 

I tested it on staging. http://54.236.120.207/v1/original_files/ works but certbot hit its limit so the HTTPS route doesn't.


davidsmejia (approved):
🦏

PR #2578, opened 2020-10-19
500 GB is the smallest for ST1s.

kurtwheeler:
## Issue Number

#2552 

This gets overwritten in staging, but needed to be changed for prod as well.


davidsmejia (approved):
👍

PR #2581, opened 2020-10-20
Update the CRAN GPG key we use to install R.

kurtwheeler:
## Issue Number

N/A, noticed tests were failing in #2580 

I'm not sure why these are mostly the same but also different, but things seem to work with this new one.


kurtwheeler:
This was actually handled in #2439

PR #2588, opened 2020-10-29
Finishes upgrading to terraform version 12.

kurtwheeler:
## Issue Number

#2587 accidentally went past terraform version 11.14

## Purpose/Implementation Notes

I was working off that branch to test the jump to 12 so that I could have it ready once the other PR deployed to staging. I must have added some of the changes from that change.

I think that rather than trying to undo the parts of that PR that were related to 12, it will be less error prone to just finish upgrading to 12 properly. I already noted in that PR that I thought that small step could be skipped anyway.

## Types of changes

- Bugfix (non-breaking change which fixes an issue)

## Functional tests

I've tested this successfully on a dev stack.

## Checklist

- [x] Lint and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)
- [x] Any dependent changes have been merged and published in downstream modules

PR #2591, opened 2020-10-30
Upgrade to latest version of terraform: 0.13.5

kurtwheeler:
## Issue Number

#2588 

Also silence an error that is ok.

I want to also upgrade to the latest version of the AWS provider in a subsequent PR.

## Types of changes

- Upgrade

## Functional tests

I was able to deploy successfully with these changes.

### DIFF
--- a/infrastructure/deploy.sh
+++ b/infrastructure/deploy.sh
@@ -402,7 +402,7 @@ container_running=$(ssh -o StrictHostKeyChecking=no \
                         -o ServerAliveInterval=15 \
                         -o ConnectTimeout=5 \
                         -i data-refinery-key.pem \
-                        "ubuntu@$API_IP_ADDRESS"  "docker ps -a" | grep dr_api || echo "")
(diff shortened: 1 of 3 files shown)


infrastructure/deploy.sh
```diff
@@ -402,7 +402,7 @@ container_running=$(ssh -o StrictHostKeyChecking=no \
                         -o ServerAliveInterval=15 \
                         -o ConnectTimeout=5 \
                         -i data-refinery-key.pem \
-                        "ubuntu@$API_IP_ADDRESS"  "docker ps -a" | grep dr_api || echo "")
+                        "ubuntu@$API_IP_ADDRESS"  "docker ps -a" 2> /dev/null | grep dr_api || echo "")
```
cgreene, line 405: confirm that you definitely want this change included

cgreene (approved):
Looks appropriate for new TF version

infrastructure/deploy.sh
kurtwheeler: Definitely! When the API cylces (which is often) this will output an error about not being able to connect right at the end of the deploy that makes a successful deploy seem less obviously successful.

PR #2595, opened 2020-11-04
Fix version of GEOquery for affy image.

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio/pull/2588

Affy didn't fail until we tried to deploy to staging because we don't build it for tests since it's so huge. This fixes that image, which should fix our deploys.

## Types of changes

## Functional tests

I built the image locally!! It took a while though.


samuel-tate (approved):
👍

PR #2596, opened 2020-11-05
Bump up the github actions timeout for deploys to 24 hours.

kurtwheeler:
## Issue Number

#2595 

I think the affy image should be able to build, but it did take like 10 hours or something on my local machine. This should be more than enough. I can lower it once we have a better idea of a good upper bound.

## Types of changes

- CD options.


cgreene (approved):
Looks good 👍

PR #2614, opened 2020-11-12
Make install_with_url fail if any of its packages fail to install.

kurtwheeler:
## Issue Number

#2612 

SCAN.UPC failed to install, but we didn't know because no error was raised. This fixes the `install_with_url` function to fail if any of the packages it is installing fail to install.

## Types of changes

## Functional tests

I put an invalid version and it failed to install it.

## Checklist

### DIFF
--- a/workers/affymetrix_dependencies.R
+++ b/workers/affymetrix_dependencies.R
@@ -7,7 +7,12 @@ options(Ncpus=parallel::detectCores())
 
 # Helper function that installs a list of packages based on input URL
 install_with_url <- function(main_url, packages) {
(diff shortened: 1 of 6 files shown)


jashapiro (approved):
Looks good to me. I might add a comment to note that `install_url` is expected to return a list of package names, because while this is the behavior (and I found where it is in the code (nested a couple function calls deep), it is not clearly documented in the `devtools` (or `remotes`) package, so I do have some fear it could break unexpectedly.

jashapiro:
I also am assuming here that the failed tests are expected!

PR #2615, opened 2020-11-19
Bump these packages to the latest version.

kurtwheeler:
## Issue Number

#2612 

## Purpose/Implementation Notes

 I wish bioc would just chill for a minute so we could get a working image. I swear I've already bumped all of these packages at least once in the last couple weeks trying to get these images to build. They just keep updating them.

## Types of changes

- Bugfix (non-breaking change which fixes an issue)

## Functional tests

I'm building some images locally.

## Checklist

- [x] Lint and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)
- [x] Any dependent changes have been merged and published in downstream modules

PR #2618, opened 2020-11-20
Fix Github Actions set-env change that broke our deploys

kurtwheeler:
## Issue Number

#2615 

This got deprecated: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

I think this should do the same thing: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable

## Types of changes

## Functional tests

The deploys are only tested on deploys.


davidsmejia (approved):
LGTM!

Issue #2620, opened 2020-12-01
Bioconductor packages keep breaking Docker image builds


kurtwheeler:
### Context

https://github.com/AlexsLemonade/refinebio/pull/2612
https://github.com/AlexsLemonade/refinebio/pull/2596
https://github.com/AlexsLemonade/refinebio/pull/2595
https://github.com/AlexsLemonade/refinebio/pull/2589
https://github.com/AlexsLemonade/refinebio/pull/2615

### Problem or idea

I keep bumping bioconductor versions, then the package gets replaced with a later version before our deployer can build it.

### Solution or next step

I can't rebuild our images. We gotta figure something out. Maybe https://github.com/AlexsLemonade/refinebio/issues/2474 will help? Not sure if it'll be helpful with bioconductor specifically though.

PR #2625, opened 2020-12-03
Pin bioconductor versions.

kurtwheeler:
## Issue Number

#2620

I'm hoping that pinning these URLs like this is more reliable and less likely to break by bioconductor changing what versions of things are in the release directory. (If that's what was happening?)

## Types of changes

## Functional tests

The tests have built all the other images (or will have before I merge it) and I tested that all the packages were able to be wgot from their URLs. I didn't build the affy image because I'd rather just do it on the deploy box so if it works it's already done. It's already broken so it can't get worse.


davidsmejia (approved):
🤞

PR #2626, opened 2020-12-08
Fix docker images by pinning to Ubuntu 18/R v3/Bioconductor v3.11

kurtwheeler:
## Issue Number

#2625 

This pins Ubuntu to version 18.04, R to version 3, and bioconductor to version 3.11 because R v4 isn't supported by a lot of bioconductor packages so they weren't available in v3.12. Bioconductor 3.11 doesn't support r v4, so we couldn't use it and it is installed on Ubuntu 20.04 by default so rather than figuring out how to uninstall v4 and reinstall v3 of R, I just bumped us back down to Ubuntu 18.04.

## Types of changes

## Functional tests

The images have built (although affy is still building, but I'd rather wait for it in staging than locally). I'm currently trying to figure out if the unit tests are failing because of ensembl or because of something with our images.

## Checklist


davidsmejia (approved):
LGTM!

PR #2634, opened 2020-12-16
Switch to a zebrafish experiment because the mouse one was getting OOM killed.

kurtwheeler:
## Issue Number

#2631 

There's one affy test that's failing, and it turns out it's because it started getting OOM killed. My guess is that a new version of the mouse probe mappings had more probes which required more RAM than used to.

This experiment had a couple `.CHP` files in it that were getting downloader jobs so I expanded our filter to take them out.

## Methods

I don't really think this changes too much scientifically other than that we're now testing zebrafish instead of mice.

## Types of changes

## Functional tests

Unit tests pass locally!

## Checklist


davidsmejia (approved):
🔥 🚀

Issue #2582, opened 2020-10-20
Upgrade Postgres to version 12 or later


kurtwheeler:
### Context

We received an email from AWS:
```
Hello,

You are receiving this message because you have one or more Amazon RDS for PostgreSQL version 9.5 instances in the us-east-1 Region.

Impacted PostgreSQL Databases:
cognoma


Amazon RDS is starting the end of life process for PostgreSQL major version 9.5. We are doing this because the PostgreSQL community is planning to discontinue support for PostgreSQL 9.5 on February 11, 2021 [1].

Amazon RDS for PostgreSQL 9.5 will reach end of life on February 16, 2021 00:00:01 AM UTC. Amazon RDS for PostgreSQL 9.6 will reach end of life in late 2021. Hence, we strongly encourage you to upgrade your databases running PostgreSQL major version 9.5 to 12 or greater at your convenience before February 16, 2021. PostgreSQL 12 contains major innovations including JSON path queries per SQL/JSON specifications and pluggable table storage interface [2]. PostgreSQL 13 introduced parallel processing of indexes with the VACUUM command and improved duplicate data handling by B-tree indexes [3]. Both versions contain numerous fixes to various software defects in earlier versions of the database.

If you do not upgrade your databases before February 16, 2021, RDS will upgrade your PostgreSQL 9.5 databases to PostgreSQL 12 during a scheduled maintenance window between February 16, 2021 00:00:01 UTC and March 16, 2021 00:00:01 UTC. On March 16, 2021 00:00:01 AM UTC, any PostgreSQL 9.5 databases that remain will be upgraded to version 12 regardless of whether the instances are in a maintenance window or not.

You can initiate an upgrade of your database instance — either immediately or during your next maintenance window — to a newer major version of PostgreSQL using the AWS Management Console or the AWS Command Line Interface. The upgrade process will shut down the database instance, perform the upgrade, and restart the database instance. The database instance may be restarted multiple times during the upgrade process. While major version upgrades typically complete within the standard maintenance window, the duration of the upgrade depends on the number of objects within the database. To avoid any unplanned unavailability outside your maintenance window, we recommend that you first take a snapshot of your database and test the upgrade to get an estimate of the duration. To learn more about upgrading PostgreSQL major versions in RDS, review the Upgrading Database Versions page [4].

We want to make you aware of the following additional milestones associated with upgrading these databases that are reaching end of life.
[...]

PR #2638, opened 2021-01-08
Upgrade to Postgres 11

kurtwheeler:
## Issue Number

#2582 

I know that issue says version 12, but the Preferred Upgrade Target for our current version is 11.1 as per https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_UpgradeDBInstance.PostgreSQL.html#USER_UpgradeDBInstance.PostgreSQL.MajorVersion.Process. If we want to go up to 12 we can do so in a subsequent PR.

This ought to do the trick, for starters. It leaves some things that will have to be cleaned up though. Upgrading the version requires a new parameter group and they cannot be destroyed and then created because apparently there's a super ultra secret backup copy of the RDS instance that's made that still uses for some number of hours. I read 30 minutes, but I waited hours and it was still failing. Then I tried it the next morning and it worked.

What that means is that we'll need another PR to remove the old group after a day or so. Although we could just create an even newer one to go all the way up to version 12 while we're at it.

## Types of changes

- Version change

## Functional tests

I spun up a stack, then ran this, then waited a day to remove the old group. It worked.

## Checklist

### DIFF
--- a/infrastructure/instances.tf
+++ b/infrastructure/instances.tf
@@ -372,10 +372,27 @@ resource "aws_spot_fleet_request" "cheap_ram" {
 # Database
 ##
 
+# Temporary. This will go away once we can delete things.
 resource "aws_db_parameter_group" "postgres_parameters" {
   name = "postgres-parameters-${var.user}-${var.stage}"
   description = "Postgres Parameters ${var.user} ${var.stage}"
   family = "postgres9.6"
+
+  parameter {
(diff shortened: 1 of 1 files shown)


davidsmejia (approved):
LGTM!

kurtwheeler:
I've finished testing the new version in a dev stack and things still work! Onto staging!!!

PR #2643, opened 2021-01-15
Switch to Github Container Repository

kurtwheeler:
## Issue Number

N/A our tests started breaking

We were getting an `unknown blob` error while pushing docker images to `docker.pkg.github.com`. Apparently `ghcr.io` is a new service that does the same thing, and using causes those errors to stop.

I had to add a personal github token to the repo for this because it's still in Beta, so they normal github token didn't cut it.

## Types of changes

## Functional tests

Tests are now passing in CI.

## Checklist

### DIFF
--- a/.github/scripts/pull_docker_images.sh
+++ b/.github/scripts/pull_docker_images.sh
@@ -2,7 +2,7 @@
 
 set -e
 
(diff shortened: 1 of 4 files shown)


davidsmejia (approved):
🏎️ Exciting!

PR #2654, opened 2021-01-28
Gets surveyor/downloader jobs working in AWS Batch

kurtwheeler:
## Issue Number

#2584 

I was going to try to separate this into two PRs, but I don't know where the line is.

This has a lot of infrastructure configuration changes to make AWS Batch work properly. So far it only introduces two jobs specs, surveyor and downloader. The rest of the job specs will have to wait until the next PR. Then after that the foreman will need to be rewritten, then after that we can rip out Nomad.

## Types of changes

## Functional tests

I successfully ran some downloader jobs! I made some changes while opening this to clean things up so I'm currently retesting it!

## Checklist


common/data_refinery_common/message_queue.py
```diff
@@ -85,9 +76,19 @@ def send_job(job_type: Enum, job, is_dispatch=False) -> bool:
...
-        "Queuing %s nomad job to run job %s with id %d.", nomad_job, job_type.value, job.id
-    )
+
+def is_job_processor(job_type):
+    if job_type in list(Downloaders):
+        return False
+    elif job_type in list(SurveyJobTypes):
+        return False
+
+    return True
```
davidsmejia, line 86:
```suggestion
def is_job_processor(job_type):
    return job_type not in list(Downloaders) and job_type not in list(SurveyJobTypes)
```

is this clearer to you?

infrastructure/batch/launch_templates.tf
```diff
@@ -11,13 +11,15 @@ resource "aws_launch_template" "data_refinery_lt_standard" {
...
+  user_data = base64encode(var.data_refinery_worker_user_data)
 }
 
 resource "aws_launch_template" "data_refinery_lt_bigdisk" {
-  name = "data-refinery-launchtemplate-bigdisk"
+  name = "data-refinery-launchtemplate-bigdisk-${var.user}-${var.stage}"
   tags = var.default_tags
   # ccdl-data-refinery-base-v2.0 image
-  image_id = "ami-01c08d4de548477df"
+  image_id = "ami-0f9b369e4ad339b15"
```
davidsmejia, line 22: maybe these should be config variables?

infrastructure/nomad-configuration/client-instance-user-data.tpl.sh
```diff
@@ -1,4 +1,8 @@
-#!/bin/bash -x
+MIME-Version: 1.0
+Content-Type: multipart/mixed; boundary="==MYBOUNDARY=="
```
davidsmejia, line 2: does this boundary get replaced?

davidsmejia (approved):
Well that was quite the read. LGTM 🚀

infrastructure/nomad-configuration/client-instance-user-data.tpl.sh
kurtwheeler: Nope. I guess I could change that but it seems to work as is. Can you think of anything that's significantly better?

infrastructure/batch/launch_templates.tf
kurtwheeler: mmm that's a good call.

infrastructure/nomad-configuration/client-instance-user-data.tpl.sh
davidsmejia: Yeah I just thought it looked strange since I'm used to seeing browser generated strings for that value. There is nothing wrong with it as long as there are no collisions. I think it's fine as is, just checking.

infrastructure/batch/launch_templates.tf
kurtwheeler: I've made this change locally. I'll include it in a later PR. (easier to test that way)

PR #2660, opened 2021-02-08
improve serialization speed

davidsmejia:
## Issue Number

N/A

According to [this blog post](https://hakibenita.com/django-rest-framework-slow) not defining read_only_fields can have up to a 40% slow down in model to json serialization. I'm not confident about the methods used but I am willing to give this a shot.

Adds `read_only_fields = fields` to any serializer that should only be used to convert data from the database to JSON.
I left `dataset` alone because I'm not sure if that uses it and this in a PR mainly concerned with the endpoints that return large responses.

In the blog post they further improved perf by writing their own serializer but the lift for that with the models in question would be significantly higher. If this does allow us to query the api for 1k samples per page we should weight the effort required and which endpoints we would want to apply that technique to.

## Methods

Just about every ModelSerializer.Meta

## Types of changes

## Functional tests

## Checklist

## Screenshots

n/a


kurtwheeler (approved):
This makes sense to me, we don't really allow POSTing to any endpoints other than dataset or token so why not set them all to read-only, especially if it can help that much with performance!

PR #2666, opened 2021-02-25
Fix dataset download numbers 

kurtwheeler:
## Issue Number

N/A, found some discrepancies whilst digging up some stats

Filter out internal email addresses in one place so when we report on it in different places it's consistent.

## Types of changes

## Functional tests

Added unit tests! As far as I can tell things look about the same but I'm not exactly sure how to test actually sending things to slack (and I don't have the local data to do it).

## Checklist


davidsmejia (approved):
Very Nice!

Issue #2663, opened 2021-02-17
There's a lot of 500/400 errors being reported by sentry


kurtwheeler:
### Context

https://sentry.io/organizations/greenelab/issues/?project=1236818

### Problem or idea

These seem to be fairly consistent. We should investigate what's causing them.

### Solution or next step

Figure out why and fix them if possible.

kurtwheeler:
While you're in there, add `filter_by` to special filters in the samples endpoint.

kurtwheeler:
https://github.com/AlexsLemonade/refinebio/pull/2669 did not address this.

https://github.com/AlexsLemonade/refinebio/pull/2711 hopefully will, but we may need to go even larger.

davidsmejia:
This is currently being tested with a larger elasticsearch box

Issue #2659, opened 2021-02-05
50x errors become 404s because we're missing 50x.html for nginx


kurtwheeler:
### Context

Trying to debug why the database/API isn't working right, noticed:
```
data-refinery-log-group-circleci-staging log-stream-api-nginx-error-circleci-staging 2021/02/05 19:55:08 [error] 4520#4520: *4334349 open() "/usr/share/nginx/html/50x.html" failed (2: No such file or directory), client: 68.80.230.239, server: api.staging.refine.bio, request: "GET /v1/experiments/ HTTP/1.1", upstream: "http://127.0.0.1:8081/v1/experiments/", host: "api.staging.refine.bio"
```

### Problem or idea

We just need to populate that path with an HTML page that says there was an internal error that we're looking into.

### Solution or next step

Add that page to the API startup script: https://github.com/AlexsLemonade/refinebio/blob/dev/infrastructure/api-configuration/api-server-instance-user-data.tpl.sh

kurtwheeler:
I haven't seen any of these 404s since we deployed 2669!

PR #2669, opened 2021-03-01
API stability improvements

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/2659
https://github.com/AlexsLemonade/refinebio/issues/2663

I haven't been able to reproduce that 404 issue but it seems like if we don't have the html setup for nginx that we should just remove the custom config, especially since we don't even have a custom page.

The throttling doesn't seem to work but I can't figure out why not. It seems like it should be set up correctly according to https://www.django-rest-framework.org/api-guide/throttling/#throttling and we already have a cache table setup in our DB so I'm not sure what's going on...

## Types of changes

## Functional tests

In addition to unit tests I used the following script to test this:
```python

import multiprocessing
import requests

url = "http://54.175.96.117/v1/samples/?limit=1000&offset={offset}&ordering=id"

def worker(num):
    """thread worker function"""
    this_url = url.format(offset=num * 1000)
    print(this_url)
    response = requests.get(this_url)
    print(response.status_code)
    return

jobs = []
for i in range(100):
    p = multiprocessing.Process(target=worker, args=(i,))
    jobs.append(p)
    p.start()
```

## Checklist

### DIFF
--- a/infrastructure/api-configuration/nginx_config.conf
+++ b/infrastructure/api-configuration/nginx_config.conf
@@ -62,50 +62,6 @@ http {
             proxy_pass http://app;
             proxy_read_timeout 30s;
             proxy_send_timeout 30s;
-
-            # Uncomment this to add CORS at the nginx level.
-            # This Django app uses `corsheaders` now, so duplicating here causes problems.
-            # if ($request_method = 'OPTIONS') {
-            #    add_header 'Access-Control-Allow-Origin' '*';
-            #    add_header 'Access-Control-Allow-Methods' 'GET, POST, PUT, OPTIONS';
-            #    #
-            #    # Custom headers and headers various browsers *should* be OK with but aren't
-            #    #
-            #    add_header 'Access-Control-Allow-Headers' 'DNT,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Range';
-            #    #
-            #    # Tell client that this pre-flight info is valid for 20 days
-            #    #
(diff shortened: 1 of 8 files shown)


infrastructure/api-configuration/nginx_config.conf
```diff
@@ -62,50 +62,6 @@ http {
...
-            # }
-            # if ($request_method = 'PUT') {
-            #     add_header 'Access-Control-Allow-Origin' '*';
-            #     add_header 'Access-Control-Allow-Methods' 'GET, POST, PUT, OPTIONS';
-            #     add_header 'Access-Control-Allow-Headers' 'DNT,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Range';
-            #     add_header 'Access-Control-Expose-Headers' 'Content-Length,Content-Range';
-            # }
-        }
-
-        #error_page  404              /404.html;
```
davidsmejia, line 103: did you want to add a custom nginx error page to handle the various errors?

davidsmejia (approved):
LGTM but maybe we should consider having a fallback nginx error page? I believe we can have it return json as well.

infrastructure/api-configuration/nginx_config.conf
kurtwheeler: We don't have a custom page we want to show right? As far as I can tell this behaves well when there's a 500 or 404.

PR #2677, opened 2021-03-15
Add remaining Batch job templates and formatting code for it.

kurtwheeler:
## Issue Number

#2584 

I should have included `format_batch_with_env.sh` a while ago, but better late than never!!

Besides that this adds a few tweaks to how job names are templated (need to make them stack specific) and adds the remaining job templates.

## Types of changes

## Functional tests

I've tested most of these job types. Not quite all yet, but tweaks can be made if necessary.


davidsmejia (approved):
LGTM!

workers/batch-job-templates/tximport.tpl.json
```diff
@@ -0,0 +1,72 @@
...
+    "jobDefinitionName": "${{USER}}_${{STAGE}}_TXIMPORT",
+    "type": "container",
+    "parameters": {
+        "job_name": "",
+        "job_id": ""
+    },
+    "containerProperties": {
+        "image": "${{DOCKERHUB_REPO}}/${{SALMON_DOCKER_IMAGE}}",
+        "vcpus": 1,
+        "memory": 32768,
```
davidsmejia, line 11: are these supposed to be hardcoded ram values?

kurtwheeler: For some of them yes, for others we template in a variety of RAM tiers. I just double checked all the values that are hardcoded and they seem to match the values we used for the Nomad jobs of the same types.

Issue #2607, opened 2020-11-10
Remove Nomad from refinebio


kurtwheeler:
### Context

#2584 

### Problem or idea

Once we have refinebio using Batch, we won't need Nomad.

### Solution or next step

Remove Nomad from refinebio.

PR #2689, opened 2021-03-22
Remove nomad

kurtwheeler:
## Issue Number

#2607 

After this PR, the word "nomad" shows up in 35 lines of our code! The remaining instances of it are either in documentation, migrations, or end to end tests. I can't take it out of the migrations and I plan on having separate PRs for both documentation and tests!

## Types of changes

## Functional tests

This mostly works! I made one or two changes in a branch based on this one to get it working smoothly, but I think it's just easier to add them in the next one if that's cool.


foreman/data_refinery_foreman/foreman/management/commands/run_tximport.py
```diff
@@ -55,9 +53,9 @@ def run_tximport():
                 processor_job.pipeline_applied = tximport_pipeline.value
                 processor_job.ram_amount = 8192
                 # This job doesn't need to run on a specific volume
-                # but it uses the same Nomad job as Salmon jobs which
+                # but it uses the same Batch job as Salmon jobs which
                 # do require the volume index.
-                processor_job.volume_index = random.choice(list(get_active_volumes()))
+                processor_job.volume_index = choose_job_queue
```
davidsmejia, line 58: does this need to be execute?

davidsmejia (approved):
LGTM! Only concern was mentioned in the call about how we could distinguish nomad vs batch jobs for future debugging - but I think we decided on not retrying jobs older than the change over date or instead adding a new batch_id column.

foreman/data_refinery_foreman/foreman/management/commands/run_tximport.py
kurtwheeler: That's a good catch. I've already made the change in my latest branch that's based off this one. I'll just add it in that one so I don't have to deal with conflicts trying to add it to this one.

PR #2691, opened 2021-03-22
Fix tests failing in Github Actions due to space constraints

kurtwheeler:
## Issue Number

N/A tests started failing

I'm not sure why, but for some reason pulling the docker cache only filled up the disk without actually making the docker builds faster. There may be a better fix that doesn't, but I want to get tests working and they may change a bit with the Batch cutover anyway.

## Types of changes

## Functional tests

They pass!

## Checklist

### DIFF
--- a/.github/workflows/config.yml
+++ b/.github/workflows/config.yml
@@ -22,11 +22,11 @@ jobs:
       - name: Disk Space Check
         run: df -h
 
-      - name: Pull cached docker images
-        run: ./.github/scripts/pull_docker_images.sh
+      # - name: Pull cached docker images
+      #   run: ./.github/scripts/pull_docker_images.sh
 
-      - name: Disk Space Check
-        run: df -h
+      # - name: Disk Space Check
+      #   run: df -h
 
       - name: Start Postgres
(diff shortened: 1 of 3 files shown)


.github/workflows/config.yml
```diff
@@ -83,8 +68,46 @@ jobs:
           ./foreman/run_tests.sh --exclude-tag=salmon --exclude-tag=transcriptome --exclude-tag=affymetrix
           .github/scripts/upload_test_coverage.sh foreman
 
-      - name: Push built docker images
```
samuel-tate, line 86: do we still need to push docker images after main tests and after downloader/no-op tests?

samuel-tate (approved):
LGTM I just had 1 question

.github/workflows/config.yml
kurtwheeler: This is a good question, although main tests are gone now. I'm not actually 100% sure whether we do or not. Not sure why this came up, or why pulling the cache seems to fill almost all of our disk space.

I've made an issue to investigate that: https://github.com/AlexsLemonade/refinebio/issues/2692

PR #2697, opened 2021-03-29
[WIP] Make jobs prioritize processing raw data

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio-examples/issues/455

We found an experiment GSE99039 that has raw data but was processed using NO_OP. This is my best guess at what the issue is, but I have neither reproduced this yet nor tested this fix.

## Methods

Right now we just use the first original file passed to this function. This instead only uses that if there's no original file with raw data.

## Types of changes

## Functional tests

TBD

## Checklist

## Screenshots


davidsmejia (approved):
LGTM

PR #2703, opened 2021-04-05
Don't force surveyor jobs to be run on smasher instances.

kurtwheeler:
## Issue Number

#2697 

I wanted to test #2697 but when I went to queue up the surveyor job for it it wouldn't get queued because we require surveyor jobs to run on the smasher instance, and we don't have a smasher instance for the staging stack unless we deploy a full stack.

This used to make sense when we were running at scale but right now it's just getting in the way of testing things on the staging stack and we won't be running at scale again until the Batch cutover happens, at which point this won't even be relevant any more.

## Types of changes

## Functional tests

This is just operational, and I haven't tested it yet. I will test on staging.

## Checklist

### DIFF
--- a/foreman/nomad-job-specs/surveyor.nomad.tpl
+++ b/foreman/nomad-job-specs/surveyor.nomad.tpl
@@ -72,9 +72,6 @@ job "SURVEYOR_${{RAM}}" {
         max_file_size = 1
       }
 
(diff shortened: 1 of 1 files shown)


davidsmejia (approved):
LGTM!

PR #2711, opened 2021-04-12
Bigger ES instance because we were hitting rate limits.

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio/pull/2669
https://github.com/AlexsLemonade/refinebio/issues/2663

https://github.com/AlexsLemonade/refinebio/issues/2663 only rate limited individual users of our API, but multiple users could hit us simultaneously and cause us to hit our ES rate limit. This bumps up the size of the ES instance so that it can handle more requests. I'm going to see how this change affects our 500 rate once it lands.

## Types of changes

- Scaling

## Functional tests

I haven't, this should be a simple change.


davidsmejia (approved):
LGTM

PR #2716, opened 2021-04-13
Our throttle limits were so low we throttled our own frontend.

kurtwheeler:
## Issue Number

N/A I fixed it rather than open an issue

These overly restrictive limits were an attempt to reduce load on ES, but that was the wrong way to do it because there's nothing stopping multiple users from making requests simultaneously. 

## Types of changes

## Functional tests

N/A just a config change.


davidsmejia (approved):
LGTM

PR #2720, opened 2021-04-16
Rotate creds and remove git-crypt

kurtwheeler:
## Issue Number

N/A came up because of codecov

git-crypt is annoying to reinit. We already had most of secrets in Github Actions secrets, but we didn't have env-specific secrets set up there yet. This adds them there, so that we can have no secrets in the repo, which means that I could remove git-crypt entirely.

I've regenerated and added all the secrets to Github Actions, with the exception of the sentry URLs because this won't land until next week and I'd like sentry to work over the weekend. I'll regenerate them and add them when deploying this.

## Types of changes

- Secrets management

## Functional tests

I'll need to test this on staging.

## Checklist


davidsmejia (approved):
🔒 🔑 LGTM

PR #2734, opened 2021-04-27
Remove the rest of git-crypt references.

kurtwheeler:
## Issue Number

#2720 

## Purpose/Implementation Notes

Copy the ssh key over to the deploy box

PR #2735, opened 2021-04-28
Set full stack to true to make sure staging stack is lookin good.

kurtwheeler:
## Issue Number

#2734 

I want to make sure the smasher smashes, I will reset this after.


davidsmejia (approved):
🛥️

PR #2736, opened 2021-04-28
Revert staging instance back to full_stack=false

kurtwheeler:
## Issue Number

#2735 

Undoes full stack for staging.


davidsmejia (approved):
🔥

Issue #2698, opened 2021-03-30
Get end-to-end tests working for Batch


kurtwheeler:
### Context

Batch can't be run locally, so we can't have end-to-end tests run locally. This is probably for the best though because testing end to end tests on an actual stack is a better test.

### Problem or idea

After staging deploys run we should have end-to-end tests run that:
1. Unsurvey a set of experiments that cover most of our codebase
2. Re-survey those experiments
3. Check to make sure they completed successfully.

This will allow us to check to make sure the end-to-end tests still work before deploying to prod.

### Solution or next step

Write these end to end tests and test that they work on the staging stack.

PR #2738, opened 2021-04-29
Batch end to end tests for downloading and processing

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/2698

This adds end to end tests that test all the processor jobs except for smasher, QN targets, compendia, and quantpendia. This tests only run on staging deploys. These will be a final check before deploying to prod. In addition they can be run with `./foreman/run_end_to_end_tests.sh`.

I added an unqueued job manager for processor jobs as well in this PR.

## Types of changes

## Functional tests

These are passing on my dev stack!

I still need to figure out a good way to test the github actions use of these. I think the best I've got for now is just debugging it in staging... but I want to add another PR for QN targets, compendia, and quantpendia so if I think of something else I'll do it in that one.

## Checklist


foreman/data_refinery_foreman/surveyor/management/commands/surveyor_dispatcher.py
```diff
@@ -73,6 +73,9 @@ def queue_surveyor_for_accession(accession: str) -> None:
 
     # We don't actually send the job here, we just create it.
     # The foreman will pick it up and dispatch it when the time is appropriate.
+    return survey_job
+
+    return survey_job
```
davidsmejia, line 78: how did the linter not catch this?

kurtwheeler: Hmm, I guess that's fine apparently? Not ideal....

davidsmejia (approved):
LGTM!

Issue #2699, opened 2021-03-30
Use multiple AWS Batch job queues/compute environments


kurtwheeler:
### Context

The current plan is to have each AWS Batch Compute Environment run a single instance so that getting downloader jobs and processor jobs running on the same instance is as simple as running them in the same job queue. The AWS Batch branch currently only uses a single compute environment. This simplified some things during development.

### Problem or idea

We should have probably like 5 job queues for the majority of jobs, 1 for smasher jobs, and 1 for compendia.

### Solution or next step

Create these additional job queues/compute environments and implement the logic to assign jobs to them.

kurtwheeler:
Addressed by #2745

PR #2745, opened 2021-05-12
Add multiple compute environments and use them!

kurtwheeler:
## Issue Number

#2699 

Ended up being a good sized change, but I'm happy with how this is setup.

There's one env var set in one place (variables.tf) that controls how many job queues there are, then everything else just iterates over them. There's also job queues for smasher jobs and compendia jobs.

I've changed the message queue to decide which job queue a job goes into, so that logic doesn't get dispersed across the project again.

## Types of changes

## Functional tests

With the current settings, the end to end tests use multiple compute environments. It seems to be working pretty well and I'm tweaking it as I continue to test  it.

## Checklist


kurtwheeler:
I think I finally tweaked the queue depth tracking correctly, this is the kinda distribution I was hoping for:
image

common/data_refinery_common/message_queue.py
```diff
@@ -33,6 +35,196 @@
...
+JOB_QUEUE_DEPTHS = {}
+for queue_name in settings.AWS_BATCH_QUEUE_ALL_NAMES:
+    JOB_QUEUE_DEPTHS[queue_name] = 0
+
+DOWNLOADER_JOB_QUEUE_DEPTHS = {}
+for queue_name in settings.AWS_BATCH_QUEUE_WORKERS_NAMES:
+    DOWNLOADER_JOB_QUEUE_DEPTHS[queue_name] = 0
+
+
+def count_jobs_in_queue(batch_job_queue) -> int:
```
davidsmejia, line 53: would it make more sense to have the name reflect the return value(s)

scripts/kill_all_jobs.sh
```diff
@@ -10,7 +10,7 @@ cd "$script_directory" || exit
 while read -r row; do
     # Exporting an expansion rather than a variable, which is exactly what we want to do.
```
davidsmejia, line 11: is this comment still relevant?

davidsmejia (approved):
just had some small comments but LGTM

common/data_refinery_common/message_queue.py
wvauclain: The type declaration is also wrong, because this function returns a pair of ints

PR #2748, opened 2021-06-01
Small Fixes for Test Bugs

wvauclain:
## Issue Number

N/A

Fixing some small bugs that I ran into while trying to get the tests running on my machine. It looks like there was a bug in the transcriptome tests that wasn't caught because we have them disabled, so I fixed that. Also, on my machine
```
$ readlink -f /usr/bin/python
/usr/bin/python2.7
```
so I had a couple issues running tests until I replaced `python ...` with `python3 ...`.

## Types of changes

## Functional tests

The transcriptome tests pass locally and the other tests were able to start running.

## Checklist

## Screenshots

### DIFF
--- a/.github/workflows/config.yml
+++ b/.github/workflows/config.yml
@@ -341,6 +341,12 @@ jobs:
       - name: Rebuild the Elasticsearch index
         run: ./scripts/rebuild_es_index.sh
 
+      - name: Run transcriptome tests
+        run: |
+          mkdir -p workers/test_volume
+          sudo chmod -R a+rw workers/test_volume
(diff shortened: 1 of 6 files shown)


wvauclain:
I added some fixes to the SRA tests to make them pass even during the latest ENA outage. This time, instead of 550-ing, the server would refuse to connect, which caused an Exception to be raised.

One of the ArrayExpress tests was also failing due to the outage but the outage ended before I could finish testing the fix for that.

kurtwheeler (approved):
LGTM!

PR #2750, opened 2021-06-03
Add tests that make sure we can import our test files

### LINKED ISSUE #2749: Import issues cause our tests to not run
### Context

See https://github.com/AlexsLemonade/refinebio/issues/1348 for more context on the issue the last time we faced it. The compendia tests were not running because there was some issue that I am currently tracking down in the `IterativeSVD` package.

### Problem or idea

It looks like there are some classes of issues which cause tests not to run, including syntax errors. I found a new error that also causes tests to get skipped, but this one is not statically analyzable, so we need to find a new solution to this problem.

### Solution or next step

I think we should be able to make a new test in each directory that dynamically imports all of the other test files. As long as the imports are dynamic and happen inside the body of the test, this should avoid the issue from above.

wvauclain:
## Issue Number

Fixes #2749 and #1348

When we run our tests, we sometimes filter based on specific tags. Django has a bug/feature that means that if it cannot load a specific test file, it just skips it. This turns out to be a necessary part of our tests, because, for example, only the compendia docker image contains all the packages needed to successfully import the compendia processor.

Unfortunately, a side effect of this is that even when we are on the right docker image, tests can sometimes fail silently. This fixes that by adding a test corresponding to each project/tag which imports each relevant test file dynamically, to make sure that no exceptions are thrown.

## Types of changes

## Functional tests

All of the tests pass, except the compendia tests which we expect to fail because of the bug.

## Checklist

## Screenshots

### DIFF
--- a/api/data_refinery_api/test_imports.py
+++ b/api/data_refinery_api/test_imports.py
@@ -1,0 +1,16 @@
+from django.test import TestCase, tag
+
+
+class ImportTestCase(TestCase):
+    def test_imports(self):
+        # Make sure we can import the api tests
+        import data_refinery_api.management.commands.test_post_downloads_summary
+        import data_refinery_api.test.test_api_general
+        import data_refinery_api.test.test_compendia
+        import data_refinery_api.test.test_dataset
+        import data_refinery_api.test.test_dataset_stats
+        import data_refinery_api.test.test_processor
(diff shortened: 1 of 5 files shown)


kurtwheeler (approved):
Oh this is quite clever. I like this.

PR #2761, opened 2021-06-08
Change version scheme for local versions

### LINKED ISSUE #2760: Fix the warning that appears when installing the data-refinery-common wheel
### Context

Saw this issue while running some tests

### Problem or idea

Whenever we build docker images, we now get this warning, some of it highlighted in red text:
```
Building wheels for collected packages: data-refinery-common
  Building wheel for data-refinery-common (setup.py): started
  Building wheel for data-refinery-common (setup.py): finished with status 'done'
  Created wheel for data-refinery-common: filename=data_refinery_common-local1623085273-py3-none-any.whl size=127887 sha256=34d44dc490189212bbbc6afdbbd18fa958c927125e4806a920b81ca9d78e591f
  Stored in directory: /root/.cache/pip/wheels/6b/35/ed/dfd04a03593c647f7b6cb2f98e2d3530e3cf7bd61dbb87e54d
  WARNING: Built wheel for data-refinery-common is invalid: Metadata 1.2 mandates PEP 440 version, but 'local1623085273' is not
Failed to build data-refinery-common
Installing collected packages: data-refinery-common
    Running setup.py install for data-refinery-common: started
    Running setup.py install for data-refinery-common: finished with status 'done'
  DEPRECATION: data-refinery-common was installed using the legacy 'setup.py install' method, because a wheel could not be built for it. A possible replacement is to fix the wheel build issue reported above. You can find discussion regarding this at https://github.com/pypa/pip/issues/8368.
Successfully installed data-refinery-common-local1623085273
```

### Solution or next step

First, we need to make sure that this is a warning, and that we are not going to fail right away because of this. Because all of the tests pass, I lean towards this being a warning for a deprecated feature. We should also do more research on what this warning means and the new way that python expects us to handle installing common.

wvauclain:
## Issue Number

Fix #2760

The old version scheme was not PEP 440 compatible, so there was a deprecation warning every time we installed the data_refinery_common package that might eventually turn into an error in future pip versions.

From https://www.python.org/dev/peps/pep-0440/#version-scheme, `0.0.0.devN` is a valid PEP 440 version, so we'll use that instead.

## Types of changes

## Functional tests

I built some docker images locally and verified that the deprecation warning no longer appears.

## Checklist

## Screenshots

### DIFF
--- a/scripts/update_models.sh
+++ b/scripts/update_models.sh
@@ -20,9 +20,9 @@ if ! docker ps | tail -n +2 | awk '{ print $NF }' | grep drdb > /dev/null; then
     exit 1
 fi
 
-# Default to "local" for system version if we're not running in the cloud.
+# Default to "0.0.0.dev" for system version if we're not running in the cloud.
 if [ -z "$SYSTEM_VERSION" ]; then
-    SYSTEM_VERSION="local$(date +%s)"
+    SYSTEM_VERSION="0.0.0.dev$(date +%s)"
     export SYSTEM_VERSION
 fi
 


kurtwheeler (commented):
Works for me! Thanks for cleaning this up!

kurtwheeler (approved):
Works for me! Thanks for cleaning this up!

PR #2779, opened 2021-06-21
Revert "Add terraform module for AWS Batch. "

kurtwheeler:
Reverts AlexsLemonade/refinebio#2601

Turns out you can't upgrade terraform from version 12 to 13 and delete resources at the same time. I need to revert all infra changes, upgrade terraform to version 13, then reapply the rest of the PR.

This has illustrated an important test I forgot to run: setting up a dev stack with the old code and then changing branches and applying the newer infrastructure and making sure it goes smoothly. This would have caught this issue.


wvauclain (approved):
Oof

PR #2790, opened 2021-06-22
Fix a couple small issues with the new send_job function.

kurtwheeler:
## Issue Number

#2601 

If the job queues were full we were logging exceptions because `batch_job_queue` was `None`. We also were dispatching Janitor tests incorrectly because we were appending ram_amount onto the job name when it's not necessary and therefore incorrect.

## Types of changes

## Functional tests

I'm running end to end tests with this change now to make sure I didn't break the message queue.

## Checklist

### DIFF
--- a/DEBUGGING.md
+++ b/DEBUGGING.md
@@ -88,7 +88,7 @@ from data_refinery_common.models.jobs.downloader_job import DownloaderJob
 from django.utils import timezone
 from datetime import datetime
 
-JOB_CREATED_AT_CUTOFF = datetime(2019, 9, 19, tzinfo=timezone.utc)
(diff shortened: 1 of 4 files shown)


wvauclain (approved):

wvauclain (changes requested):
It looks like these changes might have broken the foreman tests

kurtwheeler:
Well I set the cutoff date for today and the tests ran yesterday, so that makes a lot of sense. I'm rerunning them, let's see if it's a bigger issuel

kurtwheeler:
Hah it did!

wvauclain (approved):

Issue #2650, opened 2021-01-21
Cleanup Smasher Tests


samuel-tate:
### Context

While documenting the Smasher tests I realized that some tests had some confusing code.
To me it seemed like this code was unnecessary and made the test a little bit more confusing to understand.

### Problem or idea

The extra code makes the tests more confusing than they have to be

### Solution or next step

We should go through the tests and clean up/refactor a bit to make the tests easier to understand.
I first started noticing the extra setup code in `test_no_smash_dupe` and continued noticing it in future tests.
Additionally, `test_qn_targets_only` is named incorrectly and should be renamed.

PR #2795, opened 2021-06-23
Processor Test Improvements

wvauclain:
## Issue Number

Refactors a lot of the tests to add suggestions from https://github.com/AlexsLemonade/refinebio-processor-documentation/
Fixes https://github.com/AlexsLemonade/refinebio/issues/2650

Makes all of the recommended changes to the following processors:
- Affymetrix
- Agilent 2-color
- NO_OP
- QN reference
- Smasher
- Transcriptome Index
  - Except for the end-to-end test. I know what changes to make because I refactored the old one, but the Batch PR changed the end-to-end tests so now I need to rewrite my patch. Once we decide on how to compare the output file against the reference file I can go ahead and make the changes.

The following tests still need to be finalized, and will be included in separate PRs:
- Illumina
  - The whole processor needs to be debugged first, there are weird things going on and it has never actually succeeded on prod
- Compendia / quantpendia
  - Not blocking us from processing
- Cron job test
  - Not blocking us from processing

### Affy and NO_OP

The following reference files were downloaded from refine.bio to compare against the most recent processor output:
- Affymetrix:
  - [GSM1426071_CD_colon_active_1.PCL](https://data-refinery-test-assets.s3.amazonaws.com/GSM1426071_CD_colon_active_1.PCL)
  - [GSM45588.PCL](https://data-refinery-test-assets.s3.amazonaws.com/GSM45588.PCL)
  - [GSM1364667_U_110208_7-02-10_S2.PCL](https://data-refinery-test-assets.s3.amazonaws.com/GSM1364667_U_110208_7-02-10_S2.PCL)
- NO_OP
  - [gene_converted_GSM557500-tbl-1.txt](https://data-refinery-test-assets.s3.amazonaws.com/gene_converted_GSM557500-tbl-1.txt)
  - [GSM269747.PCL](https://data-refinery-test-assets.s3.amazonaws.com/GSM269747.PCL)
  - [gene_converted_GSM1234847-tbl-1.txt](https://data-refinery-test-assets.s3.amazonaws.com/gene_converted_GSM1234847-tbl-1.txt)
  - [gene_converted_GSM1089291-tbl-1.txt](https://data-refinery-test-assets.s3.amazonaws.com/gene_converted_GSM1089291-tbl-1.txt)

For the NO_OP and Affymetrix tests (and eventually Illumina and transcriptome), we use the following function to check the reference output against the actual output, based on Jackie's feedback:
https://github.com/AlexsLemonade/refinebio/blob/fa0b7d859a6db18936a8bdb10b1f19c662470610/workers/data_refinery_workers/processors/testing_utils.py#L9,L29
where `squish_duplicates` was factored out of the smasher code and is defined here:
https://github.com/AlexsLemonade/refinebio/blob/fa0b7d859a6db18936a8bdb10b1f19c662470610/workers/data_refinery_workers/processors/utils.py#L744,L750

### Transcriptome Index
~We use the following reference file: [ERR1562482_quant.sf](https://data-refinery-test-assets.s3.amazonaws.com/ERR1562482_quant.sf)~ **Edit:** I will have to pick a new reference file because the new end-to-end tests use a different organism.

The transcriptome index comparison is a little bit more complicated, because of examples like the following: in our reference file we have a reading for `3R5.1a` and in the output we have a (nearly identical) reading for `3R5.1a.1` instead. It turns out that the transcript IDs aren't really stable, so we can have files that are almost identical in scientific content but they don't compare against each other using the method described above.

I had a couple ideas of how to fix this. The first was to strip version suffixes wherever possible being somewhat clever:
https://github.com/AlexsLemonade/refinebio/blob/f253bc48bdc7b5c1a3d4b41a4bb84ac6e2e992ac/foreman/data_refinery_foreman/surveyor/test_end_to_end.py#L764,L817
This only improved correlation to ~85% though, because of observations like this:
| | Reference | Actual |
| --- | --- | --- |
| B0024.9a.1 | 189 | 0 |
| B0024.9a.2 | 0 | 189 |

The other thing I tried, which did give me ~99.9% correlation but is a bit sketchy, was converting transcript IDs to gene IDs using the files generated as a part of the transcriptome index:
https://github.com/AlexsLemonade/refinebio/blob/3854383f0f389b2ee9368803152fd4e2f056200e/foreman/data_refinery_foreman/surveyor/test_end_to_end.py#L769,L792
This works because Ensembl Gene IDs are much more stable than the transcript IDs, but we do lose some information by doing this. I did make sure, however, that the resulting files still contain all of the right observations, so the ~99.9% correlation is legit and would be smaller if we were comparing different samples.

In conclusion, I'm not really sure what to do because the only method that works in comparing files is also the most iffy, so I am open to suggestions.

### QN reference
I added a test to the QN reference which compares the output against an expected output, but I calculated the expected output using a spreadsheet and manually walking through the steps of the algorithm.

See: https://github.com/AlexsLemonade/refinebio/blob/fa0b7d859a6db18936a8bdb10b1f19c662470610/workers/data_refinery_workers/processors/test_qn_reference.py#L68,L100 and https://github.com/AlexsLemonade/refinebio/blob/fa0b7d859a6db18936a8bdb10b1f19c662470610/workers/data_refinery_workers/processors/test_qn_reference.py#L194,L195

## Types of changes

## Functional tests

I ran all the updated tests locally to make sure that they worked.

## Checklist


jaclyn-taroni:
@wvauclain What organism is used for the new end-to-end tests that impact the transcriptome index? Percolating on some things...

wvauclain:
@jaclyn-taroni right now the end-to-end tests use Saccharomyces cerevisiae

kurtwheeler:
I chose Saccharomyces cerevisiae because it's small so it's quick.

workers/data_refinery_workers/processors/test_no_op.py
```diff
@@ -14,6 +17,62 @@
...
+
+    sample = Sample()
+    sample.accession_code = job_info["accession_code"]
+    sample.title = job_info["accession_code"]
+    sample.platform_accession_code = job_info["platform_accession_code"]
+    man = job_info.get("manufacturer", None)
+    if man is not None:
+        sample.manufacturer = man
+    # The illumina samples need the human organism
+    if job_info.get("manufacturer", None) == "ILLUMINA":
```
kurtwheeler, line 43:
```suggestion
    if man == "ILLUMINA":
```
I don't love the variable name `man` lol

workers/data_refinery_workers/processors/test_no_op.py
```diff
@@ -24,181 +83,127 @@ def setUp(self):
...
+            "/home/user/data_store/TEST/NO_OP/EXPECTED/gene_converted_GSM1089291-tbl-1.txt",
+            sep="\t",
+            names=["", "VALUE"],
+            index_col=0,
+        )["VALUE"]
+        actual_data = pd.read_csv(
+            final_context["output_file_path"], sep="\t", names=["", "VALUE"], index_col=0
+        )["VALUE"]
+
+        assertMostlyAgrees(self, expected_data, actual_data)
```
kurtwheeler, line 206: I like how succinct all these tests are now that you've refactored all the setup away :+1:

workers/run_tests.sh
```diff
@@ -190,55 +223,69 @@ if [ -z "$tag" ] || [ "$tag" = "agilent" ]; then
...
+        echo "Downloading NOOP expected file3."
+        wget -q -O "$no_test_exp_dir/$no_exp_file3" \
+             "$test_data_repo/$no_exp_file3"
+    fi
+    no_exp_file4="gene_converted_GSM1089291-tbl-1.txt"
+    if [ ! -e "$no_test_exp_dir/$no_exp_file4" ]; then
+        mkdir -p "$no_test_exp_dir"
+        echo "Downloading NOOP expected file4."
+        wget -q -O "$no_test_exp_dir/$no_exp_file4" \
+             "$test_data_repo/$no_exp_file4"
```
kurtwheeler, line 288: Could this just be done as a loop over `"gene_converted_GSM557500-tbl-1.txt GSM269747.PCL gene_converted_GSM1234847-tbl-1.txt gene_converted_GSM1089291-tbl-1.txt"`?

kurtwheeler (approved):
This is fantastic. I went over it pretty slowly but only had a few minor comments. Well done!

jaclyn-taroni:
Hi @wvauclain, I am excited to see this. 

The Affymetrix and NO_OP changes seem very reasonable. The selection of Affymetrix samples means we've covered platforms with and without BrainArray and we need a specific check for HuEx because of the BrainArray situation. Similarly, good selection of different platforms for NO_OP.

I had a question about the QN reference changes to check my understanding. 

I also had the most comments about the transcriptome index test, which I think makes sense given your initial comment. 

I've divided those up under different headers below.

#### QN reference

To summarize my understanding, we've "manually" calculated a QN reference and then we'll use the same samples to generate the QN reference we're testing.

This is going to be distinct from the cron job that essentially asks: How much did the QN reference change? ([ref](https://github.com/AlexsLemonade/refinebio-processor-documentation/pull/15/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R10))

Is that all correct?


#### Transcriptome index

* We're currently testing that gene to transcript mapping with a new tx2gene file is giving us gene-level similar results. This is good, but is perhaps more of a happy side effect. It may not want to be the only thing we do.
* We may want to check transcript ID agreement between reference and test data (e.g., maybe something like a Jaccard index calculation, but the 95% of the smaller set check we have is probably fine, too).
* My suspicion is that the transcript IDs are likely to be more stable in say human or mouse, but I'm not sure that's true. I'm also not sure how feasible it is to run one of those species experiments in an end-to-end test. (I suppose we could find an experiment that is pretty shallow.) It may also not be true when the _genome build_ changes, as we're expecting it will/did for mouse ([now on GRCm39](http://www.ensembl.org/Mus_musculus/Info/Index)). If genome builds change, we **definitely want to know about that**! Is there a way to include that in our tests?
  * Is the problem we encountered with the _C. elegans_ data a release issue or a genome build issue?

wvauclain:
> #### QN reference
> 
> To summarize my understanding, we've "manually" calculated a QN reference and then we'll use the same samples to generate the QN reference we're testing.
>
> This is going to be distinct from the cron job that essentially asks: How much did the QN reference change? (ref)
>
> Is that all correct?

Yup, that's all correct. Just so we're on the same page, the "samples" that we are using in this test are, from what I can tell, dummy files that only contain ~30 "genes", rather than actual samples.

jaclyn-taroni:
> Just so we're on the same page, the "samples" that we are using in this test are, from what I can tell, dummy files that only contain ~30 "genes", rather than actual samples.

Thanks for clarifying – that makes sense given the number of values. While we're on the subject, I just want to make sure we're testing to make sure that if we encounter problematic samples in the platform that we're using to generate the reference that we drop them. (For context: https://github.com/AlexsLemonade/refinebio-processor-documentation/pull/12#discussion_r539612668 😄 )

wvauclain:
> Thanks for clarifying – that makes sense given the number of values. While we're on the subject, I just want to make sure we're testing to make sure that if we encounter problematic samples in the platform that we're using to generate the reference that we drop them. (For context: [AlexsLemonade/refinebio-processor-documentation#12 (comment)](https://github.com/AlexsLemonade/refinebio-processor-documentation/pull/12#discussion_r539612668) smile )

`7.tsv` has some NA rows and `6.tsv` has an extra row, and we make sure that they both get filtered when comparing against the expected QN reference.

wvauclain:
> Is the problem we encountered with the C. elegans data a release issue or a genome build issue?

@jaclyn-taroni I checked the API, and it looks like on prod we ran [ERR1562482](https://api.refine.bio/v1/samples/ERR1562482/) with assembly `WBcel235` and Ensembl release version 96. The version in our tests with incompatible transcript IDs is also `WBcel235` with Ensembl release version 99, so it looks like the problem is a release issue.

Is that more of a problem or less of a problem?

jaclyn-taroni:
> Is that more of a problem or less of a problem?

I'd say probably _less of a problem_ overall. I wonder how stable transcript IDs are for _Saccharomyces cerevisiae_.

wvauclain:
I just added the changes to the end-to-end tests, which should be the last additions before this gets merged

foreman/data_refinery_foreman/foreman/test_end_to_end.py
```diff
@@ -302,3 +305,58 @@ def test_all_the_things(self):
...
+            .organism_index.assembly_name,
+            "R64-1-1",
+        )
+        # Note that the reference file was processed using version 96
+        self.assertEqual(
+            sample.results.all()
+            .filter(processor__name=ProcessorEnum.SALMON_QUANT.value["name"])
+            .first()
+            .organism_index.release_version,
+            "104",
```
kurtwheeler, line 330: This number changes over time as new assemblies are released. We get this number by querying ensembl's API because we used to have it hardcoded and it would break every now and then when a new version was released. It seems this test will now do the same.

wvauclain: My thinking here was that this falls under the category of "things we want to know about", but I can remove this check so we are just notified when the genome build itself changes.

@jaclyn-taroni which would you prefer?

jaclyn-taroni: Genome only build sounds good to me 👍🏻  ideally we'd process everything with the same release, but I don't know that that is realistic.

kurtwheeler: I don't think it is. The API for the current and past releases are not consistent. We potentially could change things to use the API for a pinned past release, but that would mean we couldn't use the latest if we wanted to...

jaclyn-taroni (approved):
LGTM

PR #2796, opened 2021-06-23
Allow users to opt out of dataset notification emails

### LINKED ISSUE #2794: Allow consumers of our API to opt out of email notifications
### Context

https://github.com/AlexsLemonade/refinebio-py/issues/58

### Problem or idea

API consumers might not want e-mail notifications when their datasets are ready.

### Solution or next step

Here we should probably make email notifications opt-out, but then we can make the client opt-in.

wvauclain:
## Issue Number

Fixes #2794

Adds a `notify_me` field to the Dataset model that API users can use to opt out of email notifications. `notify_me` defaults to `True` so that we don't break the existing behavior with the frontend, but note that the API client will default to explicitly setting it to False.

## Types of changes

## Functional tests

## Checklist

## Screenshots


kurtwheeler (approved):
LGTM! Nice work and nice tests!

Issue #2775, opened 2021-06-17
Create mechanism to run pipelines locally


kurtwheeler:
### Context

#2584 

We can't run Batch locally like we could with Nomad. This means that there's not currently a way to run a sample all the way through locally without using the `tester.sh` script which hasn't gotten a lot of use in the last couple years.

### Problem or idea

I think we need a script that will monitor the database and will start containers for any jobs that are waiting to be run. This is necessary because no one docker image has all the dependencies for all jobs.

### Solution or next step

Create that script or a similar mechanism to be able to run all jobs locally, along with running downstream jobs.

kurtwheeler:
Closed by #2799

PR #2799, opened 2021-06-24
Add script to run full pipeline locally

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/2775

This adds a python script that invokes bash scripts to run surveyor, downloader, and processor jobs for a given experiment accession (or organism/Ensembl division in the case of transcriptome indices).

I also ended up doing some refactoring because the transcriptome jobs were failing because of an issue in `needs_processing`. Fixing that issue ended up creating a cyclical import issue. The right way to fix it seemed to split the `job_lookup` namespace because it was really doing two things.

## Types of changes

## Functional tests

I've run transcriptome and no_op smoothly and I'm running others still.

## Checklist

### DIFF
--- a/foreman/data_refinery_foreman/foreman/management/commands/get_job_to_be_run.py
+++ b/foreman/data_refinery_foreman/foreman/management/commands/get_job_to_be_run.py
@@ -1,0 +1,61 @@
+import json
+
+from django.core.management.base import BaseCommand
+
+from data_refinery_common.models import DownloaderJob, ProcessorJob
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **options):
+        """Outputs a JSON dict of a job that need to be run.
+
+        The most recently created unstarted ProcessorJob will be prioritized.
+
+        If there are no unstarted ProcessorJobs, then the most
+        recently created unstarted DownloaderJob will be prioritized.
+
+        The dict will have two top level keys: downloader_jobs and processor_jobs.
+        The JSON output will be a dict containing the following keys:
+        * job_name
+        * job_id
+        * image_name
+        * job_type
+        """
+        processor_jobs_to_run = [
+            "AFFY_TO_PCL",
+            "SALMON",
+            "TXIMPORT",
+            "ILLUMINA_TO_PCL",
+            "TRANSCRIPTOME_INDEX_LONG",
+            "TRANSCRIPTOME_INDEX_SHORT",
+            "NO_OP",
+        ]
+        processor_job = (
+            ProcessorJob.objects.filter(
+                start_time=None, success=None, pipeline_applied__in=processor_jobs_to_run
+            )
+            .order_by("-id")
+            .first()
+        )
+        if processor_job:
+            job_dict = {
+                "job_id": processor_job.id,
+                "job_type": "ProcessorJob",
+                "job_name": processor_job.pipeline_applied,
+            }
+            print(json.dumps(job_dict))
+            return
+
+        downloader_job = (
+            DownloaderJob.objects.filter(start_time=None, success=None).order_by("-id").first()
+        )
+        if downloader_job:
+            job_dict = {
+                "job_id": downloader_job.id,
+                "job_type": "DownloaderJob",
+                "job_name": downloader_job.downloader_task,
+            }
+            print(json.dumps(job_dict))
+            return
+
+        print("{}")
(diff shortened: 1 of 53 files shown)


scripts/reinit_database.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+# Reintializes the database so there's no data or migrations run against it.
```
wvauclain, line 3: Fancy, I like it

foreman/data_refinery_foreman/foreman/management/commands/get_job_to_be_run.py
```diff
@@ -0,0 +1,61 @@
...
+class Command(BaseCommand):
+    def handle(self, *args, **options):
+        """Outputs a JSON dict of a job that need to be run.
+
+        The most recently created unstarted ProcessorJob will be prioritized.
+
+        If there are no unstarted ProcessorJobs, then the most
+        recently created unstarted DownloaderJob will be prioritized.
+
+        The dict will have two top level keys: downloader_jobs and processor_jobs.
```
wvauclain, line 17: I think this line is outdated

scripts/run_full_pipeline.py
```diff
@@ -0,0 +1,99 @@
+import argparse
+import json
+import subprocess
+
+
+def parse_args():
+    description = """This script can be used to run the full pipeline.
```
wvauclain, line 7: :fire:

scripts/run_full_pipeline.py
```diff
@@ -0,0 +1,99 @@
...
+        elif job_name == "SALMON":
+            image_name = "salmon"
+        elif job_name == "ILLUMINA_TO_PCL":
+            image_name == "illumina"
+        elif job_name in ["TRANSCRIPTOME_INDEX_LONG", "TRANSCRIPTOME_INDEX_SHORT"]:
+            image_name = "transcriptome"
+        elif job_name == "NO_OP":
+            image_name = "no_op"
+
+    completed_command = subprocess.check_call(
```
wvauclain, line 59: From [here](https://docs.python.org/3/library/subprocess.html#subprocess.check_call), it looks like `check_call` throws if the return code is nonzero, rather than returning the nonzero error code.

Also, does the output of this get sent to the terminal? That would be preferable, or at least somewhere that the output can be inspected.

scripts/run_full_pipeline.py
```diff
@@ -0,0 +1,99 @@
...
+        [
+            "./workers/run_job.sh",
+            "-i",
+            image_name,
+            subcommand,
+            f"--job-name={job_name}",
+            # job_name,
```
wvauclain, line 66: I think you can get rid of these commented-out elements

scripts/run_full_pipeline.py
```diff
@@ -0,0 +1,99 @@
...
+            # str(job_id),
+        ]
+    )
+
+    if completed_command != 0:
+        print(f"{job_name} Job with id {job_id} failed!")
+
+
+def survey_accession(accession_code):
+    completed_command = subprocess.check_call(
```
wvauclain, line 77: Same comment as above

scripts/run_full_pipeline.py
```diff
@@ -0,0 +1,99 @@
...
+    if completed_command != 0:
+        print(f"The Survey Job failed!")
+
+
+def run_full_pipeline(accession_code):
+    survey_accession(accession_code)
+
+    job_to_run = get_job_to_run()
+
+    while job_to_run:
```
wvauclain, line 90: You might want to make this `while job_to_run is not None:` and then explicitly return `None` in `get_job_to_run()` if we fall through all the if statements

wvauclain (changes requested):
Few small things

scripts/run_full_pipeline.py
kurtwheeler: There's only one if statement in `get_job_to_run`. The `get_job_to_be_run` management command has several but it has to return JSON, and `None` isn't valid JSON. Instead `get_job_to_be_run` returns a dict representing a processor job, a downloader job, or an empty dict. An empty dict `{}` is both valid JSON and falsey in python so it works here I think.

Is there any advantage to using an explicit `None` here?

scripts/run_full_pipeline.py
kurtwheeler: Oh good catch. I guess I don't gotta check the return codes.

The output does get sent to the terminal. It's a little bit chatty at the moment because the `run_job.sh` and `run_surveyor.sh` scripts rebuild the docker images every time but it is nice having them take care of that for you as you iterate.

I think that maybe in the future we could have the pipeline script look at what docker images would need to be built and build them once before kicking off work, but that's a larger refactor to the way the scripts work. In this PR I just wanted to get the flow working with the existing scripts and it already ended up being nontrivial.

scripts/run_full_pipeline.py
wvauclain: Good point, I didn't think of that. It's good as is then.

wvauclain (approved):

PR #2805, opened 2021-06-25
Fix deregistering Batch job definitions

kurtwheeler:
## Issue Number

#2601
Came up from LGTM complaint

The boto docs lied to me. The list parameter doesn't work with names, just ARNs. See https://github.com/boto/boto3/issues/2908

This basically just reverts to an older method.

## Types of changes

## Functional tests

I've tested this against my dev stack and am continuing to do so.

## Checklist


infrastructure/deregister_batch_job_definitions.py
```diff
@@ -15,19 +15,12 @@
...
-    for job_definition in job_definition_list:
-        job_definitions = batch.describe_job_definitions(
-            jobDefinitionName=job_definition, status="ACTIVE"
-        )
-        # There can be multiple revisions per job deifinition. We want them all gone.
-        for job_definition_revision in job_definitions["jobDefinitions"]:
-            batch.deregister_job_definition(
-                jobDefinition=job_definition_revision["jobDefinitionArn"]
-            )
+    # There can be multiple revisions per job deifinition. We want them all gone.
```
wvauclain, line 24:
```suggestion
    # There can be multiple revisions per job definition. We want them all gone.
```

scripts/format_batch_with_env.sh
```diff
@@ -185,7 +187,8 @@ if [ "$project" = "workers" ]; then
                 echo "Made $FILEPATH"
             done
             echo "Made $output_dir/$OUTPUT_FILE"
-        elif [ "$OUTPUT_FILE" = "smasher.json" ] || [ "$OUTPUT_FILE" = "create_qn_target.json" ] || [ "$OUTPUT_FILE" = "create_compendia.json" ] || [ "$OUTPUT_FILE" = "create_quantpendia.json" ] || [ "$OUTPUT_FILE" = "tximport.json" ]; then
+        # From https://unix.stackexchange.com/a/111517
+        elif (echo "$NO_RAM_JOB_FILES" | grep -Fqw "$OUTPUT_FILE"); then
```
wvauclain, line 191: It's moments like this when I most wish for switching to python like in resources-portal

wvauclain (approved):

scripts/format_batch_with_env.sh
kurtwheeler: Absolutely. I've started putting new scripts in python. Eventually maybe we'll even take the time to rewrite some of the old bash ones.

PR #2807, opened 2021-06-28
Prefer 'title' over 'sample name' for the title field in harmonizer

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/2619

Surveyor jobs for GSE6532 are failing because accessing of:
```
  File "/home/user/data_refinery_foreman/surveyor/external_source.py", line 171, in survey
    experiment, samples = self.discover_experiment_and_samples()
  File "/home/user/data_refinery_foreman/surveyor/geo.py", line 501, in discover_experiment_and_samples
    experiment, samples = self.create_experiment_and_samples_from_api(experiment_accession_code)
  File "/home/user/data_refinery_foreman/surveyor/geo.py", line 319, in create_experiment_and_samples_from_api
    GeoSurveyor._apply_harmonized_metadata_to_sample(
KeyError: 'OXFU_23 (HG-U133A)'
```

This was because we were using the sample name instead of title as the title field. 

I think that I doubly fixed this because just using `harmony.extract_title()` in both places should guarantee that there's no mismatch. 

Also Fix ILLUMINA job name because I saw some error messages in the logs complaining that job didn't exist.

## Methods

## Types of changes

## Functional tests

## Checklist

## Screenshots

### DIFF
--- a/foreman/data_refinery_foreman/surveyor/harmony.py
+++ b/foreman/data_refinery_foreman/surveyor/harmony.py
@@ -27,11 +27,10 @@ def extract_title(sample: Dict) -> str:
         "extract name",
     ]
     title_fields = add_variants(title_fields)
-    for key, value in sorted(sample.items(), key=lambda x: x[0].lower()):
-        lower_key = key.lower().strip()
 
-        if lower_key in title_fields:
-            return value
+    for title_field in title_fields:
+        if title_field in sample:
+            return sample[title_field]
 
     # If we can't even find a unique title for this sample
     # something has gone horribly wrong.
@@ -724,6 +723,6 @@ def preprocess_geo(items: List) -> List:
 
             # Probably won't be a list with length greater than one,
             # but maybe?
-            new_sample[key] = " ".join(value)
+            new_sample[key.lower()] = " ".join(value)
         preprocessed_samples.append(new_sample)
     return preprocessed_samples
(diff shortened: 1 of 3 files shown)


foreman/data_refinery_foreman/surveyor/harmony.py
```diff
@@ -27,11 +27,10 @@ def extract_title(sample: Dict) -> str:
...
     ]
     title_fields = add_variants(title_fields)
-    for key, value in sorted(sample.items(), key=lambda x: x[0].lower()):
-        lower_key = key.lower().strip()
 
-        if lower_key in title_fields:
-            return value
+    for title_field in title_fields:
+        if title_field in sample:
+            return sample[title_field]
```
kurtwheeler, line 33: I'm not actually sure if I should leave this change in or not. I think just using `harmony.extract_title()` in both places will be enough to fix it. I do think that preferring `title` over `sample name` would be a beneficial change, but this also is less consistent with how we've done it previously.

I'm leaning with changing it rather than staying consistent, but I'd love to hear other opinions if they exist.

wvauclain: I think that the code you have now is better, regardless of whether or not we want to keep preferring `title` over `sample name`. Looking at the old code, I couldn't tell why `sample name` had higher priority than `title` until I realized that it would sort earlier. No matter what we choose, I think you should add a comment above `title_fields` that says something like "sorted in decreasing order of priority".

I am hesitant to change the default behavior though, because a lot of our tests are written using the implicit assumption that `sample name` is prioritized over `title` since they take metadata from prod rather than surveying every time.

wvauclain: We also use `title` in places like column detection in the Illumina processor so changing the behavior has a nontrivial effect on our processing.

kurtwheeler:
https://github.com/AlexsLemonade/refinebio/pull/2850 does this much better

Issue #2812, opened 2021-06-29
Add `is_queued` parameter to job endpoints


kurtwheeler:
### Context

I removed the `nomad` parameter: https://github.com/AlexsLemonade/refinebio/commit/73dc56d84567f076d79b577af2ea242e91e12563#diff-a7ce68dcae10038cb21e2da530be6a6f05e1cd96155be7dce54fc237070ef3a4L53

### Problem or idea

That was used by the frontend, which was broken by this change.

### Solution or next step

Readd the parameter, except call it `is_queued` instead.

wvauclain:
Closed by https://github.com/AlexsLemonade/refinebio/pull/2813

Issue #2791, opened 2021-06-22
API tests sometimes fail because we don't trigger a 429


kurtwheeler:
### Context

The failing test looks like:

```
FAIL: test_all_endpoints (data_refinery_api.test.test_api_general.APITestCases)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/user/data_refinery_api/test/test_api_general.py", line 339, in test_all_endpoints
    self.assertEqual(response.status_code, status.HTTP_429_TOO_MANY_REQUESTS)
AssertionError: 200 != 429
```

### Problem or idea

It only happens every now and then. I suspect we're relying too carefully on timing and it sometimes doesn't work out. It makes sense since the test is about rate limiting requests.

### Solution or next step

Make the test more reliable.

wvauclain:
https://github.com/AlexsLemonade/refinebio/pull/2813 tried to address this issue, and the situation is definitely a lot better now, but we still have some runs where this check fails: https://github.com/AlexsLemonade/refinebio/runs/3088940490.

PR #2813, opened 2021-06-30
Add isQueued field to all three job endpoints.

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio-frontend/issues/950
https://github.com/AlexsLemonade/refinebio/issues/2812
https://github.com/AlexsLemonade/refinebio/issues/2791

This removes `nomad` as a parameter to the job endpoints and instead just adds a `is_queued` to all job objects. This field isn't filterable because it isn't added until after the job has been retrieved.

I've also tried to address https://github.com/AlexsLemonade/refinebio/issues/2791 by just making a few more API calls to make sure a bad timing window won't prevent the 429 from being triggered.

## Types of changes

## Functional tests

I've updated unit tests to test this. I've also run the unit tests a bunch of times to see if I can still trigger the 429. I saw it a couple times before I added my change to that test but not since.

## Checklist

### DIFF
--- a/api/data_refinery_api/test/test_api_general.py
+++ b/api/data_refinery_api/test/test_api_general.py
@@ -256,19 +256,19 @@ class APITestCases(APITestCase):
 
         response = self.client.get(reverse("survey_jobs", kwargs={"version": API_VERSION}))
         self.assertEqual(response.status_code, status.HTTP_200_OK)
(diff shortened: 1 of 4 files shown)


davidsmejia (approved):
LGTM

kurtwheeler:
Ok I improved some things as I tested this against my dev stack. Seems to finally be working correctly though!

Issue #2701, opened 2021-03-30
Fix API stats endpoint to query Batch instead of Nomad


kurtwheeler:
### Context

This endpoint is only used by our dashboards, so fixing it isn't going to break anything for our users.

I've punted on this issue by just removing the code that queries nomad.

### Problem or idea

The stats endpoint will need to query Batch.

### Solution or next step

Reimplement the stats endpoint by querying Batch.

PR #2814, opened 2021-06-30
Add Batch info to the `/stats` endpoint

wvauclain:
## Issue Number

Fixes https://github.com/AlexsLemonade/refinebio/issues/2701

Query batch for the stats endpoint.

I feel like this is really slow, but from the comments on @kurtwheeler's other PR it looks like what I am doing is the fastest way to get all of the information that we need from Batch, so we might just have to deal with it.

## Types of changes

## Functional tests

I ran the API with my changes locally pointed at my devstack and made sure that everything worked before spinning the devstack down.

## Checklist

## Screenshots

### DIFF
--- a/api/data_refinery_api/views/stats.py
+++ b/api/data_refinery_api/views/stats.py
@@ -200,7 +200,7 @@ class Stats(APIView):
             )
         ]
     )
-    # @method_decorator(cache_page(CACHE_TIME_SECONDS))
+    @method_decorator(cache_page(CACHE_TIME_SECONDS))
     def get(self, request, version, format=None):
         range_param = request.query_params.dict().pop("range", None)
         is_dashboard = request.query_params.dict().pop("dashboard", False)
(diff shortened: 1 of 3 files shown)


kurtwheeler:
Have you tried pointing the frontend at your dev stack API? Does the dashboard work with these stats? Will it need another PR to integrate with them? I think the queue breakdown is new. Is there a frontend issue to deal with that?

api/data_refinery_api/views/stats.py
```diff
@@ -58,8 +71,123 @@ def paginate_queryset_response(queryset, request):
...
+    data["pending_jobs"] = count_occurrences(is_pending, all_jobs)
+    data["running_jobs"] = count_occurrences(is_running, all_jobs)
+
+    def get_job_type(job):
+        return job["type"]
+
+    # groupby must be executed on a sorted iterable
+    # ref: https://docs.python.org/3.8/library/itertools.html#itertools.groupby
+    # XXX: in the old code this was filtered by if the job had the key
+    # "ParameterizedJob". Does Batch have an equivalent?
```
kurtwheeler, line 162: No, Batch does not because Batch didn't have to create a weird job-that's-not-a-job because Batch's primary purpose is to be a job queue unlike an orchestration tool that comports itself to be a job queue that we are too familiar with :roll_eyes:

kurtwheeler (approved):
This LGTM! If needed we should create a frontend issue to work with these changes.

wvauclain:
It looks like https://github.com/AlexsLemonade/refinebio-frontend/issues/953 deals with these changes

PR #2815, opened 2021-06-30
Add a script for calculating RAM-hours for a given job.

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/AdminItems/issues/358

This is currently blocked by https://github.com/AlexsLemonade/refinebio-py/pull/61 which is blocked by https://github.com/AlexsLemonade/refinebio/pull/2813

However if I only attempt to calculate it for processor jobs it works!!

```
Job took 0.023554304999999998 hours to run.
Job took 2.0 GB to run.
Sample GSM65317 took 0.047108609999999995 RAM-hours to process.
```

(With a couple extra prints to verify the calculation)

## Types of changes

## Functional tests

I've run this with `python3 scripts/calculate_ram_hours.py "GSM65317" --start-date 2018-10-01`

## Checklist


scripts/calculate_ram_hours.py
```diff
@@ -0,0 +1,89 @@
...
+
+    ram_hours = 0
+    for original_file_id in sample.original_files:
+        original_file = pyrefinebio.OriginalFile.get(original_file_id)
+
+        for downloader_job in original_file.downloader_jobs:
+            if downloader_job.start_time and downloader_job.start_time > utc_start_date:
+                ram_hours += calculate_ram_hours_for_job(downloader_job)
+
+        for processor_job in original_file.processor_jobs:
```
wvauclain, line 77: I think you are overcounting the ram hours here for some sample types.

Salmon files with unmated reads, for example, have multiple original files but only a single downloader job:
https://github.com/AlexsLemonade/refinebio/blob/dev/workers/data_refinery_workers/downloaders/sra.py#L259,L287

so you count the same downloader job for each read file.

scripts/calculate_ram_hours.py
```diff
@@ -0,0 +1,89 @@
...
+import datetime
+
+import pyrefinebio
+import pytz
+
+
+def parse_args():
+    description = """This script calculates how many RAM-hours a sample has consumed.
+
+    A RAM-hour shall be defined as using 1GB of random access memory for 1 hour.
```
wvauclain, line 25: I think you have to un-indent these lines, because the leading spaces on each line are stored as part of the string, so when the description is printed it will look like this:
```
This script calculates how many RAM-hours a sample has consumed.

    A RAM-hour...

    All downloader...
```

kurtwheeler: It actually strips out all the whitespace so it looks like:
```
$ python3 ./scripts/calculate_ram_hours.py -h
usage: calculate_ram_hours.py [-h] [--start-date START_DATE] accession_code

This script calculates how many RAM-hours a sample has consumed. A RAM-hour
shall be defined as using 1GB of random access memory for 1 hour. All
downloader and processor jobs that were run will be considered, regardless of
whether they were successful or not. The goal is to determine how many
resources were spent to download and process a sample, regardless of success.

positional arguments:
  accession_code        Specify the accession_code of the sample to calculate
                        RAM-hours for.

optional arguments:
  -h, --help            show this help message and exit
  --start-date START_DATE
                        Do not include jobs started previous to start_date in
                        the calculations. The format is expected to be YYYY-
                        MM-DD.
```

kurtwheeler: But I guess I don't need that extra newline there, although I think it makes it a bit easier to read from within the code.

kurtwheeler: I should have just put that in the PR description to start with! Next time!

scripts/calculate_ram_hours.py
kurtwheeler: Oh that's a good catch!

kurtwheeler:
Okay! https://github.com/AlexsLemonade/refinebio-py/pull/61 has been released so this now works for both downloader and processor jobs!

I've also used a couple sets to keep track of job ids that have been counted!

scripts/calculate_ram_hours.py
```diff
@@ -70,13 +70,25 @@ def calculate_ram_hours(accession_code, start_date):
     for original_file_id in sample.original_files:
         original_file = pyrefinebio.OriginalFile.get(original_file_id)
 
+        seen_downloader_job_ids = set()
```
wvauclain, line 73: You need to lift these sets outside the above `for` loop, because the overcounting happens across original files instead of within a single original file

kurtwheeler: Oh good catch again, that was sloppy.

wvauclain (approved):
LGTM

Issue #2803, opened 2021-06-25
Fix weekly digests


kurtwheeler:
### Context

Weekly digests stopped working

#2215 

### Problem or idea

I tried to fix this, it didn't work. Not sure what's going wrong.

### Solution or next step

Get it working. While we're changing things with it, also filter out the end-to-end test email so that doesn't get counted or even make its way into #ccdl-refinebio-emails

kurtwheeler:
It worked!

PR #2818, opened 2021-07-01
Change ENGAGEMENT_WEBHOOK->ENGAGEMENTBOT_WEBHOOK

kurtwheeler:
## Issue Number

#2803 

I didn't noticed the difference between these two. I've regenerated and added ENGAGEMENTBOT_WEBHOOK to the secrets for the repo.

## Types of changes

## Functional tests

This will need to be tested by deploying to prod as staging doesn't even use it.


wvauclain (approved):
Sneaky

PR #2739, opened 2021-04-29
Tag AWS resources with team and project

kurtwheeler:
## Issue Number

#2704 

## Purpose/Implementation Notes

This adds a bunch of tags with `team=engineering` and `project=refine.bio`.

## Types of changes

- New feature (non-breaking change which adds functionality)

## Functional tests

I've deployed this to a dev stack and it wroked second try!

## Checklist

- [x] Lint and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)
- [x] Any dependent changes have been merged and published in downstream modules

PR #2829, opened 2021-07-06
Tag database instance and parameter group.

kurtwheeler:
## Issue Number

#2739 

I missed these before. Noticed a large percentage of our untagged costs were coming from RDS and found these!

## Types of changes

- Infrastructure tagging

## Functional tests

I deployed a dev stack with these to make sure it was a valid configuration.


davidsmejia (approved):
LGTM

PR #2830, opened 2021-07-06
Rotate SSH keys

kurtwheeler:
## Issue Number

N/A freshening up some creds

Rotates SSH key

Taints a key created by terraform to force it to rotate

Also removes the defunct deployer key.

## Types of changes

- Security

## Functional tests

I've deployed a dev stack with these changes. I will test that I can ssh onto the API server with the new key.


davidsmejia (approved):
LGTM

PR #2832, opened 2021-07-07
Cycle instances to change SSH keys

kurtwheeler:
## Issue Number

#2830 

Taint instances because updating the keypair won't cylce instances, despite not updating authorized_keys.


davidsmejia (approved):
lgtm

PR #2835, opened 2021-07-07
Apply db password change immediately

kurtwheeler:
## Issue Number

#2832 

While testing staging I noticed that I need this flag. I think this one is probably good as a permanent change because we already have downtime during deploys so we may as well change the DB then.

## Types of changes

## Functional tests

I've spun up a dev stack with the changes to make sure they're valid.


davidsmejia (approved):
LGTM

PR #2836, opened 2021-07-08
Remove SSH key from AMI template

kurtwheeler:
## Issue Number

#2832 

After cycling the instances, the old key was still there because it was baked into the AMI. This removes it. 

## Types of changes

## Functional tests

I've already updated the AMIs and tested it on staging. (Once the AMIs were updated, I just reran the deploy since they're outside the deploy process.)


davidsmejia (approved):
LGTM

Issue #2793, opened 2021-06-23
Samples that make it to `illumina.R` don't process successfully


wvauclain:
### Context

Still trying to find a reference file for the Illumina tests

### Problem or idea

I found an experiment [GSE56080](https://www.refine.bio/experiments/GSE56080/orphan-nuclear-receptor-err-is-required-for-pancreatic-islet-beta-cell-maturation) that processed a sample successfully, but if you look at the output file these are the entire contents:
```tsv
GeneID	Ad-EGFP-1.AVG_Signal
```

The actual input file is not empty though.

### Solution or next step

It looks like the process of converting `GSE56080_non-normalized.txt.sanitized` into `GSE56080_non-normalized.PCL` filters out all of the rows; maybe for a good reason but the processor job probably shouldn't succeed.

wvauclain:
I figured out the cause of this issue; it turns out that
https://github.com/AlexsLemonade/refinebio/blob/65b724cf8dc47c211cb37fd2a252e5423df7032a/workers/data_refinery_workers/processors/illumina.R#L428
is not selecting the right column, so R _helpfully_ gives us an empty vector instead of throwing an error :disappointed: 

This causes a new error, however, because it turns out that probe IDs do not correspond to a unique gene ID!

For example, [ILMN_1651268](https://useast.ensembl.org/Homo_sapiens/Location/Genome?fdb=funcgen;ftype=ProbeFeature;array=HumanRef-8_V3;vendor=ILLUMINA;id=ILMN_1651268) maps to both [ENSG00000165714](https://useast.ensembl.org/Homo_sapiens/Gene/Summary?db=core;g=ENSG00000165714;r=12:12357078-12471233) and [ENSG00000280689](https://useast.ensembl.org/Homo_sapiens/Gene/Summary?db=core;g=ENSG00000280689;r=12:12357078-12471233). @jaclyn-taroni hypothesized that this could be an issue with poor probe quality, but `ILMN_1651268` has perfect probe quality so filtering based on quality didn't help.

[...]

Issue #2792, opened 2021-06-23
Lots of Illumina jobs are failing before reaching `illumina.R`


wvauclain:
### Context

Trying to find a reference file on prod for the Illumina tests

### Problem or idea

A suspicious number of samples processed using the Illumina processor fail on prod, including [the experiment we currently use in our tests](https://www.refine.bio/experiments/GSE22427/effect-of-t350-phosphorylation-on-gene-repression-activity-of-ezh2-in-bj-cells).

### Solution or next step

~It looks like this might be a RAM issue; I think we need to make Illumina jobs have multiple RAM tiers like some of the other processors.~ **Edit:** The Illumina jobs do now have RAM tiers, they were just added after most of the samples were processed.

wvauclain:
For some of the samples it might be a RAM issue, but this is really suspicious:
```
>>> Sample.objects.filter(results__processor__name="Illumina SCAN").count()
10
```

wvauclain:
It looks like these are the 10 most common failure reasons:
```python3
[
    (None, 2800),
    ("Could not detect PValue column!", 889),
    ("Missing file for processor job but unable to recreate downloader jobs!", 411),
    ("'NoneType' object is not subscriptable", 195),
    (
        "Unhandled exception caught while running processor function _prepare_files in pipeline: 'ascii' codec can't decode byte 0x8b in position 1: ordinal not in range(128)",
        145,
    ),
    ("We can not process the data because it is not on the disk", 81),
    (
        "Unhandled exception caught while running processor function _prepare_files in pipeline: 'ascii' codec can't decode byte 0xfe in position 10: ordinal not in range(128)",
        71,
    ),
    (
        "Unhandled exception caught while running processor function _prepare_files in pipeline: 'ascii' codec can't decode byte 0xff in position 0: ordinal not in range(128)",
        36,
    ),
    (
        "Unhandled exception caught while running processor function _detect_platform in pipeline: 'MACACA_MULATTA'",
        35,
    ),
    (
        "Unhandled exception caught while running processor function _prepare_files in pipeline: 'ascii' codec can't decode byte 0x81 in position 9: ordinal not in range(128)",
        27,
    ),
]
```

The most common (`None`) usually happens when a job gets OOM-killed; not much we can do about that. Then there are a couple other ones related to our infrastructure like data missing from the disk. There are three interesting ones though:
- `Could not detect PValue column`
[...]

PR #2844, opened 2021-07-13
Fix the Illumina processor

wvauclain:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/2792, https://github.com/AlexsLemonade/refinebio/issues/2793, https://github.com/AlexsLemonade/refinebio/issues/2846, https://github.com/alexslemonade/refinebio/issues/1560

I overhauled the early parts of the Illumina processor and added lots of tests to account for the different ways that I saw Illumina samples fail on prod.

This also implements the suggestions that the science team had for how to fix the Illumina processor.

## Methods

For probes with multiple associated genes, we use the mappings generated using https://github.com/AlexsLemonade/illumina-refinery/pull/1 to determine which probe ID to use. (I will migrate this repo to the Alexslemonade org, but I didn't want to do that without input on which license I should use and if I need to add anything to the README) The readme for that repo contains the information on the methods used to generate the mappings.

## Types of changes

## Functional tests

I added lots of new tests and made sure that the experiments in those tests don't fail for the same reasons they used to.

## Checklist


wvauclain:
The only thing left to implement (besides a few more checks in some of the tests) is this feedback:
> ...I think also maybe useful to add in a warning that there was a duplicate gene name and that the main assembly gene name was used.
How would users of refine.bio expect to see warnings like this? Should it appear in the metadata in the smasher output? Tagging @dvenprasad @jaclyn-taroni

jaclyn-taroni:
I will say that my expectation would be that this is added as one of the FAQs and we could link to it in the front end – looking forward to @dvenprasad's input though

jaclyn-taroni (commented):
Wow! 👏🏻  Between the different file extensions out there and the fact that some of these files are encoded in latin1, this almost surprised me as much as the time that I learned about all the custom printed chicken microarrays out there. Great work!

At this point, I had a couple questions about the R processing code to check my understanding but between that and the README from `wvauclain/illumina-refinery` this approach seems very reasonable 👍🏻 

> For probes with multiple associated genes, we use the mappings generated using https://github.com/wvauclain/illumina-refinery to determine which probe ID to use. (I will migrate this repo to the Alexslemonade org, but I didn't want to do that without input on which license I should use and if I need to add anything to the README)

As far as the Illumina refinery code goes, I think it's a good idea to have that go through code review – what do you think the best way to facilitate that is after transfer (e.g., does it need a new default branch before it gets moved over)?

Regarding the licensing question, did you happen to take a look at the BrainArray and Illumina Bioconductor package licenses? This seems like a good starting point.

workers/data_refinery_workers/processors/illumina.R
```diff
@@ -377,8 +378,11 @@ if (length(exprColumns) == 0)
 exprData <- as.matrix(data[,exprColumns,drop=FALSE])
 rownames(exprData) <- probeIDs
 
+# Throw out NA rows
+exprData <- na.omit(exprData)
```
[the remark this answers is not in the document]
wvauclain: I should take this out, I don't think I intended to commit this change, because in my experience, usually only a handful of columns have missing values.

I initially added this as a temporary fix for some samples that were failing with the following issue:
```
Error in if (max(x) > 100) x <- log2(x) :
  missing value where TRUE/FALSE needed
Calls: scanNorm -> doLog2
```
I will take another look at the code and see if there is an easy way to ignore NAs that occur only in single columns.

workers/data_refinery_workers/processors/illumina.R
```diff
@@ -424,8 +428,27 @@ probesToKeep <- intersect(probesToKeep, rownames(normData))
...
+probe_map <- read.csv(paste0("/home/user/probe_maps/", platform, ".tsv"), sep="\t", na.strings=c(""))
+rownames(probe_map) <- probe_map$probe_id
+
+choose_gene = function(probe_id, ensembl_ids) {
+  probe_id <- probe_id[1]
+  if (length(ensembl_ids) == 1) {
+    return(ensembl_ids[1])
+  }
+
+  # TODO: emit some kind of warning. Also, we need to decide what to do about NAs
```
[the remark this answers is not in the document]
wvauclain: Right now, this dataframe only has probes that map to multiple gene IDs, but
> If none of the associated gene IDs appear in any Brainarray platform, an NA is emitted

Instead, should I use the same tiebreaker and pick the one with the lowest gene ID?

workers/data_refinery_workers/processors/illumina.R
```diff
@@ -424,8 +428,27 @@ probesToKeep <- intersect(probesToKeep, rownames(normData))
...
+  probe_map[probe_id,"ensembl_id"][1]
```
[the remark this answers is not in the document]
wvauclain: I agree that this is a little bit hacky. My Illumina refinery only returns the first preference, so there should only be one element, but R always returns a vector when you index this way so I am basically doing `c(1)[1]`. Is there a different way I can index so that I don't have to extract the one value in the vector like this?

workers/illumina_probe_maps/illuminaRatv1.tsv
```diff
@@ -0,0 +1,169 @@
+probe_id	ensembl_id
```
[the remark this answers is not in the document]
wvauclain: It looks like the `ensembl_id` column is blank here because of a bug in the Illumina refinery; I will re-generate these.

workers/data_refinery_workers/processors/illumina.R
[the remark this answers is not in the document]
wvauclain: It turns out that the indexing was unnecessary, so I removed it to avoid confusion.

dvenprasad:
I agree with @jaclyn-taroni that it is add a note and link out situation. 

I think this information should be presented to the users when they are downloading it. So on the `Download Now` modal and My Dataset page

For the `Download Now` modal, it is easy enough, I would add it in the modal, just before we present the transformation options.
Screen Shot 2021-07-23 at 2 48 15 PM

For the My Dataset page it is a bit trickier. My question is:

- Should this warning be limited on an experiment level or on a dataset level? If they are aggregating by species, no problem, but if they are aggregating my experiment?
- What do we usually use, if not main assembly gene name?

jaclyn-taroni:
> * Should this warning be limited on an experiment level or on a dataset level? If they are aggregating by species, no problem, but if they are aggregating by experiment?

I'm not sure where exactly I come down on this so I'm going to "talk it out." 

If folks are aggregating by species where the other datasets come from Affymetrix, we're favoring IDs that are on BrainArray (and since Illumina is limited to human, mouse, and rat – most of the Affymetrix chips for those species are BrainArray-supported) so it is definitely _less_ of an issue. The thing is – every single processor includes some kind of scientific decision-making and if folks are very concerned with individual genes, for example, rather than assessing broad signal they should read the docs.

> * What do we usually use, if not main assembly gene name?

BrainArray is usually doing the heavy-lifting for us in this regard. It also may be the case that these problems are somewhat mitigated by the Affymetrix probe design, but I'm not sure to be honest!

dvenprasad:
>If folks are aggregating by species where the other datasets come from Affymetrix, we're favoring IDs that are on BrainArray (and since Illumina is limited to human, mouse, and rat – most of the Affymetrix chips for those species are BrainArray-supported) so it is definitely less of an issue. The thing is – every single processor includes some kind of scientific decision-making and if folks are very concerned with individual genes, for example, rather than assessing broad signal they should read the docs.

So it would be okay if we put a blanket statement on the My dataset page saying that some gene names maybe main assembly gene names?

jaclyn-taroni:
> So it would be okay if we put a blanket statement on the My dataset page saying that some gene names maybe main assembly gene names?

Do you mean _any time_ one of the affected platforms is in the dataset? Then, yes. I wouldn't put that warning there if there were no Illumina BeadArray samples in the dataset.

It's not a simple as the main assembly genes, though, because we're favoring gene identifiers represented on Brainarray.  (I would also not use the term gene names – I would make sure to say Ensembl gene identifiers.)

All of this to say, the message should probably say something like: Illumina probe IDs map to multiple Ensembl gene identifiers, learn about how we choose which Ensembl gene identifier to use -> link to docs.

Side note: @wvauclain can you file an issue on [`refinebio-docs`](https://github.com/AlexsLemonade/refinebio-docs) about the Illumina refinery?

dvenprasad:
>Do you mean any time one of the affected platforms is in the dataset? Then, yes. I wouldn't put that warning there if there were no Illumina BeadArray samples in the dataset. 

Yep, only when affected platforms are present. (we'll be able to tell via the API right?)

scripts/run_full_pipeline.py
```diff
@@ -49,7 +51,7 @@ def run_job(job):
         elif job_name == "SALMON":
             image_name = "salmon"
         elif job_name == "ILLUMINA_TO_PCL":
-            image_name == "illumina"
+            image_name = "illumina"
```
kurtwheeler, line 54: oof

workers/data_refinery_workers/processors/illumina.py
```diff
@@ -43,31 +44,134 @@ def _prepare_files(job_context: Dict) -> Dict:
...
-                    and line[0] != "<"
-                    and line[0] != "\t"
-                ):
+                HEADER_CHARS = ["#", "!", "^"]
+
+                # Sometimes we have a quoted header, so we need to check both
+                is_header = line[0] in HEADER_CHARS or (
+                    line[0] in ["'", '"'] and line[1] in HEADER_CHARS
+                )
+                is_empty = line.strip() == ""
```
kurtwheeler, line 119: Would it not make sense to strip the line before any operations? That way a leading space doesn't make us misinterpret header lines?

workers/data_refinery_workers/processors/illumina.py
```diff
@@ -455,13 +606,23 @@ def _create_result_objects(job_context: Dict) -> Dict:
...
 
-def illumina_to_pcl(job_id: int) -> None:
+def illumina_to_pcl(job_id: int, cleanup=None) -> None:
     pipeline = Pipeline(name=PipelineEnum.ILLUMINA.value)
+    initial_job_context = {"job_id": job_id, "pipeline": pipeline}
+
+    # When running the tests, don't clean up original files so we don't have to
+    # keep downloading them.
+    if cleanup is not None:
+        initial_job_context["cleanup"] = cleanup
```
kurtwheeler, line 616: We don't seem to pass this anywhere from the tests?

workers/data_refinery_workers/processors/test_illumina.py
```diff
@@ -67,25 +69,108 @@ def prepare_illumina_job(organism):
...
+        # find its detection column using the description given as an annotation
+        sample = Sample.objects.get(title="LV-T350A&si-EZH2-3")
+        sample.title = "ignoreme_for_description"
+        sample.accession_code = "ignoreme_for_description"
+        sample.save()
+
         final_context = illumina.illumina_to_pcl(job.pk)
-        self.assertTrue(final_context["success"])
+        self.assertSucceeded(job)
+        # XXX: remove this but because the job failed the rest of this won't succeed
```
kurtwheeler, line 171: I'm confused by this comment.

workers/data_refinery_workers/processors/test_illumina.py
```diff
@@ -113,57 +198,361 @@ def test_bad_illumina_detection(self):
...
+            self.assertTrue(os.path.exists(smashme.absolute_file_path))
+
         # Cleanup after the job since it won't since we aren't running in cloud.
         shutil.rmtree(final_context["work_dir"], ignore_errors=True)
+
+    @tag("illumina")
+    def test_illumina_latin1_input(self):
+        """Test a latin1-encoded Illumina file.
+
+        GSE106321 is encoded in latin1 and uses μ in the title of some
```
kurtwheeler, line 238: Whaaaaaaat?

kurtwheeler: Sorry, I don't need any further clarification here, this is just crazy.

workers/data_refinery_workers/processors/test_illumina.py
```diff
@@ -113,57 +198,361 @@ def test_bad_illumina_detection(self):
...
+                "organism": organism,
+                "samples": [("GSM836222", "IMGUS_32"), ("GSM836223", "IMGUS_33"),],
+            }
+        )
+
+        final_context = illumina.illumina_to_pcl(pj.pk)
+        self.assertSucceeded(pj)
+
+        # Make sure that the row names are no longer quoted after sanitizing the file
+        def assertNotQuoted(string: str):
```
kurtwheeler, line 336: It seems like stripping `string` would make this a little more robust.

workers/data_refinery_workers/processors/test_illumina.py
wvauclain: Agh, I meant to remove these before marking as ready for review. Will do.

kurtwheeler (approved):
This looks really great to me! Super comprehensive, nice work!

I think it'd maybe be good to do a little extra `strip()`ping and it'd be good to use the `cleanup` flag by default in the tests, but this is other G2G!

Well, assuming tests pass that is.

workers/data_refinery_workers/processors/illumina.R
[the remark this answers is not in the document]
wvauclain: Marking as resolved for now because we decided to open https://github.com/AlexsLemonade/refinebio/issues/2870 to track this

PR #2847, opened 2021-07-13
Add a separate `downloadable_organism_names` field to the experiment doc

### LINKED ISSUE #2840: `/search/` responses contain incorrect information
### Context

I found what was causing https://github.com/AlexsLemonade/refinebio-frontend/issues/955

### Problem or idea

Sometimes results returned from `/search/` contain incorrect information, which causes incorrect information like `No Species` to be displayed on the search page.

An example is GSE112517:

Hitting https://api.refine.bio/v1/search/?search=accession_code:GSE112517 gives
```
...
            "organism_names": [],
...
```
but hitting https://api.refine.bio/v1/experiments/GSE112517/ gives
```
...
            "organism_names": ["HOMO_SAPIENS"],
...
```

### Solution or next step

My first instinct is that the Elasticsearch database is out of sync with our database. One easy way to test this would be to check whether or not manually refreshing Elasticsearch solves the problem for this experiment.

I also think that this _could_ be happening because `organism_names` is a computed property, which might cause some weirdness. Surveying the experiment locally and seeing what a local instance of Elasticsearch returns would be a good way to test if this is true.

wvauclain:
## Issue Number

Fixes #2840

This PR moves the old behavior to a field called `downloadable_organism_names` field and makes `organism_names` return all of the organism names. I wired up the filters to point to `downloadable_organism_names`, but the frontend takes the data from `organism_names` to populate the UI, so this makes the new behavior consistent with the experiment detail view.

I need to file a tiny frontend PR to go along with this one that just changes the facet name.

## Types of changes

## Functional tests

I ran the API locally alongside the frontend changes and made sure everything looks right.

## Checklist

## Screenshots

image

### DIFF
--- a/api/data_refinery_api/views/experiment_document.py
+++ b/api/data_refinery_api/views/experiment_document.py
@@ -282,7 +282,8 @@ class ExperimentDocumentView(DocumentViewSet):
         "accession_code": "accession_code",
         "alternate_accession_code": "alternate_accession_code",
         "platform": "platform_accession_codes",
-        "organism": "downloadable_organism_names.raw",
+        "organism": "organism_names.raw",
+        "downloadable_organism": "downloadable_organism_names.raw",
         "num_processed_samples": {
             "field": "num_processed_samples",
             "lookups": [LOOKUP_FILTER_RANGE, LOOKUP_QUERY_IN, LOOKUP_QUERY_GT],
--- a/api/data_refinery_api/views/experiment_document.py
+++ b/api/data_refinery_api/views/experiment_document.py
@@ -141,6 +141,7 @@ class ExperimentDocumentSerializer(DocumentSerializer):
             "platform_names",
             "platform_accession_codes",
             "organism_names",
+            "downloadable_organism_names",
             "pubmed_id",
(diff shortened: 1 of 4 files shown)


api/data_refinery_api/views/experiment_document.py
```diff
@@ -281,7 +282,7 @@ class ExperimentDocumentView(DocumentViewSet):
         "accession_code": "accession_code",
         "alternate_accession_code": "alternate_accession_code",
         "platform": "platform_accession_codes",
-        "organism": "organism_names.raw",
+        "organism": "downloadable_organism_names.raw",
```
kurtwheeler, line 285: This change looks like it will force everyone to search by downloadable_organism_names, just because that's the behavior we want in the frontend.

Could we instead a `downloadable_organism` filter field and then have the frontend use that? Or would the frontend then just show both filters and confuse users?

kurtwheeler (approved):
This LGTM. I am curious if we can allow users to choose whether they want to search by organism_names or downloadable_organism_names without cluttering the frontend though.

Issue #2828, opened 2021-07-06
Refactor harmonizer


kurtwheeler:
### Context

It currently harmonizes all the samples for an experiment, then tries to go through them again and match up samples with their metadata.

### Problem or idea

This matching process is error prone. Sometimes it doesn't match up correctly.

### Solution or next step

We should just harmonize samples one at a time.

kurtwheeler:
Closed by https://github.com/AlexsLemonade/refinebio/pull/2850

PR #2850, opened 2021-07-14
Refactor harmonizer

kurtwheeler:
## Issue Number

#2828 

Try to make the harmonizer more reliable, readable, and maintainable. Also try to address the issue https://github.com/AlexsLemonade/refinebio/pull/2807 was trying to fix.

I was able to change the GEO and SRA surveyors to harmonize the metadata sample by sample, which should eliminate the kind of mismatch errors that could exist before. Array express can't be done that way because we're combining two different responses, metadata from one and file data from the other.

## Types of changes

- refactor, bug fixes

## Functional tests

Unit tests are passing! I added one for the experiment that wouldn't survey before 

## Checklist

### DIFF
--- a/foreman/data_refinery_foreman/surveyor/harmony.py
+++ b/foreman/data_refinery_foreman/surveyor/harmony.py
@@ -288,7 +288,7 @@ def parse_sdrf(sdrf_url: str) -> List:
         sample = {}
         for col, value in enumerate(sample_values):
             key = keys[col]
-            sample[key] = value
+            sample[key.strip().lower()] = value
         samples.append(sample)
 
     return samples
@@ -310,12 +310,12 @@ def preprocess_geo(items: List) -> List:
                     # a malformed response from the server.
                     if ":" in pair:
                         split = pair.split(":", 1)
-                        new_sample[split[0].strip()] = split[1].strip()
+                        new_sample[split[0].strip().lower()] = split[1].strip()
                 continue
 
             # Probably won't be a list with length greater than one,
             # but maybe?
-            new_sample[key.lower()] = " ".join(value)
+            new_sample[key.strip().lower()] = " ".join(value)
         preprocessed_samples.append(new_sample)
     return preprocessed_samples
 
@@ -333,8 +333,8 @@ def extract_title(sample: Dict) -> str:
         "sample title",
         "title",
         "subject number",
-        "labeled extract name",
         "extract name",
+        "labeled extract name",
     ]
     title_fields = create_variants(title_fields)
 
@@ -562,7 +562,7 @@ class Harmonizer:
         ]
         self.time_fields = create_variants(self.time_fields)
 
-    def harmonize_value(field_name, value):
+    def harmonize_value(self, field_name: str, value):
         if field_name == "age":
             try:
                 return float(value)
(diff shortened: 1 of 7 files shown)


foreman/data_refinery_foreman/surveyor/harmony.py
```diff
@@ -697,33 +357,303 @@ def parse_sdrf(sdrf_url: str) -> List:
...
+            "strain/genotype",
+        ]
+        self.genetic_information_fields = create_variants(genetic_information_fields)
+
+        disease_fields = [
+            "disease",
+            "disease state",
+            "disease status",
+            "diagnosis",
+            "disease",
```
wvauclain, line 475: Is this redundant?

foreman/data_refinery_foreman/surveyor/test_harmony.py
```diff
@@ -271,48 +280,67 @@ def test_geo_harmony(self):
         """
         Thoroughly tests a specific GEO harmonization
         """
-
         # Weird ones caused bugs
         gse = GEOparse.get_GEO("GSE94532", destdir="/tmp/GSE94532/", silent=True)
-        preprocessed_samples = preprocess_geo(gse.gsms.items())
+        for _, sample in gse.gsms.items():
```
wvauclain, line 285: What are we checking for here?

wvauclain (commented):
Almost all of the PR looks really good, I just have one question about the sample title changes

foreman/data_refinery_foreman/surveyor/test_harmony.py
kurtwheeler: From the comment I assume that calling preproces_geo() used to cause errors on this experiment until it got fixed?

It'd be nice if the comment was a little bit more descriptive... I'll add a couple checks of some sort here.

foreman/data_refinery_foreman/surveyor/harmony.py
```diff
@@ -10,657 +254,73 @@
 logger = get_and_configure_logger(__name__)
 
 
-def extract_title(sample: Dict) -> str:
+TITLE_FIELDS = [
```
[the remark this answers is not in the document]
kurtwheeler: Hmm, I'm not sure that I needed to but I think this order makes the most sense... While I'm not sure this change is necessary, I think it's probably worth doing. I can't help but think that if a sample has both a `sample name` and a `labeled extract name` that the `sample name` field is more likely to cause us to succeed at processing if we're using it as a column name.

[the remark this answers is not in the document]
kurtwheeler: That's certainly an assumption so I'll run the surveyor for some or all of the experiments in our processor tests and see if there's any differences.

kurtwheeler:
@wvauclain I've rerun the surveyor for the two experiments in `test_illumina`. Here are the sample titles, which match exactly what's already in the tests:
```
data_refinery=# select title from samples order by created_at desc;
       title        
--------------------
 CB CD34+ normoxia
 CB CD34+ hypoxia
 LV-T350A&si-EZH2-3
 LV-T350A&si-EZH2-2
 LV-T350A&si-EZH2-1
 LV-EZH2&si-EZH2-3
 LV-EZH2&si-EZH2-2
 LV-EZH2&si-EZH2-1
 LV-C&si-EZH2-3
 LV-C&si-EZH2-2
 LV-C&si-EZH2-1
 LV-C&si-Control-3
 LV-C&si-Control-2
 LV-C&si-Control-1
```

I didn't see any other processors that used the `title` field. Were there any others that you were aware of?

wvauclain (approved):
None that I'm aware of, so we should be good to go!

PR #2852, opened 2021-07-16
Poll the smasher end to end test so it doesn't fail when the API isn't up yet.

kurtwheeler:
## Issue Number

N/A Will and I noticed this in the end to end tests

The end to end tests run right after the deploy, so sometimes the API isn't yet up by the time we're trying to use it to download a dataset.

## Types of changes

## Functional tests

I ran an end-to-end test against a dev stack, it worked! I tried to have the API down, but I couldn't get it quick enough. I guess staging will be the final test.

## Checklist

### DIFF
--- a/foreman/data_refinery_foreman/foreman/test_end_to_end.py
+++ b/foreman/data_refinery_foreman/foreman/test_end_to_end.py
@@ -171,15 +171,14 @@ class SmasherEndToEndTestCase(TestCase):
         prepare_computed_files()
 
         # The API sometimes takes a bit to come back up.
-        stop = False
         start_time = timezone.now()
-        while not stop:
+        while True:
             try:
                 pyrefinebio.create_token(agree_to_terms=True, save_token=False)
-                stop = True
+                break
             except pyrefinebio.ServerError:
                 if timezone.now() - start_time > timedelta(minutes=15):
-                    stop = True
+                    break
                 else:
                     sleep(30)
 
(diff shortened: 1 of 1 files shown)


foreman/data_refinery_foreman/foreman/test_end_to_end.py
```diff
@@ -168,7 +169,19 @@ def test_smasher_job(self):
...
+        # The API sometimes takes a bit to come back up.
+        stop = False
+        start_time = timezone.now()
+        while not stop:
+            try:
+                pyrefinebio.create_token(agree_to_terms=True, save_token=False)
+                stop = True
+            except pyrefinebio.ServerError:
+                if timezone.now() - start_time > timedelta(minutes=15):
+                    stop = True
```
wvauclain, line 182: Should we fail the test here instead of breaking out of the loop? It seems like this would cause the test to fail later for a less-clear reason than if we raised an `AssertionError("Server not up after 15 minutes")`

foreman/data_refinery_foreman/foreman/test_end_to_end.py
```diff
@@ -168,7 +169,19 @@ def test_smasher_job(self):
             purge_experiment(accession_code)
 
         prepare_computed_files()
-        pyrefinebio.create_token(agree_to_terms=True, save_token=False)
+
+        # The API sometimes takes a bit to come back up.
+        stop = False
+        start_time = timezone.now()
+        while not stop:
```
wvauclain, line 176: Can this be rewritten to use `break` instead?

foreman/data_refinery_foreman/foreman/test_end_to_end.py
kurtwheeler: That's a good point!

wvauclain (approved):

PR #2853, opened 2021-07-16
E2E tests: Download the quant file using the smasher

wvauclain:
## Issue Number

Came up while deploying

Because we locked down the S3 buckets, `.sync_from_s3()` no longer works inside the end-to-end tests. Because we were only trying to grab a quant file, the easiest solution is to use the smasher to download the quant file.

## Types of changes

## Functional tests

I spun up a devstack and verified that the tests pass.

## Checklist

## Screenshots


kurtwheeler (approved):
Sweet, this LGTM! Nice work!

Issue #2854, opened 2021-07-19
A lot of computed files have null s3_bucket/s3_key


kurtwheeler:
### Context

```
data_refinery=> select count(*) from computed_files where s3_bucket is NULL or s3_key is NULL;
 count 
-------
 40542
```

This is probably the cause of https://sentry.io/share/issue/5523b06e30aa4eb8bb3bec7363c1aead/

### Problem or idea

Why are we making computed files that we aren't storing? If we don't have the file, it's probably worth cleaning up. We also should check to see if there's samples we think we've processed but really we don't even have the output anywhere.

### Solution or next step

See if we can stop creating computed files that aren't uploaded. Do something about the 40.5k that are already in that state, probably some combination of deleting/reprocessing.

kurtwheeler:
The command added in #2932 removed all the computed files!

```
>>> ComputedFile.objects.filter(s3_key=None, s3_bucket=None).count()
0
```

Issue #2657, opened 2021-02-03
Notifications for failed smash jobs stopped working


kurtwheeler:
### Context

They should go to slack: https://github.com/AlexsLemonade/refinebio/blob/dev/workers/data_refinery_workers/processors/smasher.py#L392

### Problem or idea

They haven't been working for a week and no one realized.

### Solution or next step

Figure out why that notification isn't working and fix it.

kurtwheeler:
This is working again!

PR #2858, opened 2021-07-19
If we don't have it in s3, we can't smash a file.

kurtwheeler:
## Issue Number

#2854 

This check is there for quant files, seems like it'd be good here too.

## Types of changes

## Functional tests

Unit tests should cover this.

## Checklist


kurtwheeler:
I've also added a one-line fix for #2657

wvauclain (changes requested):
I think this breaks the tests, because we don't upload to s3 so `is_processed` is now false. That might be why the smasher tests are failing.

wvauclain (approved):
Oops, meant to approve this sooner. LGTM!

Issue #2811, opened 2021-06-29
End to end tests fail frequently because of ENA's FTP server


kurtwheeler:
### Context

#2738 

### Problem or idea

Often when running the end to end tests they will fail because a downloader job wasn't able to reach ENA's FTP server.

### Solution or next step

This seems like an issue that we're likely to run into prod. It's probably worth seeing if it can be mitigated before running lots of salmon jobs.

One thought I had was to move this logic into the surveyor. It's the check to see if there's unmated reads so the surveyor could determine this for all the downloader jobs. I think it might be an improvement because it won't be multiple simultaneous connections to the FTP server, but I think we've also seen some issues while running a single salmon job. However it also might be nice to have the surveyor job succeed or fail so it can be rerun until it passes, rather than having to retry hundreds of downloader jobs until they all pass.

wvauclain:
From our discussion over text, the real problem is this function right here:
https://github.com/AlexsLemonade/refinebio/blob/fab031300fcadd670e27cd61ef2829f150082f04/workers/data_refinery_workers/downloaders/sra.py#L215,L256

it is worth looking into if there is a different way to determine if a sample has unmated reads that doesn't rely on ENA's FTP server.

Issue #2593, opened 2020-10-30
Our method for building ENA URLs seems to have broken


kurtwheeler:
### Context

I was trying to run the experiment for #2592 but the downloader jobs all failed.

### Problem or idea

It seems to be caused by us creating the wrong ENA URL here:

We create `ftp://ftp.sra.ebi.ac.uk/vol1/fastq/SRR115/008/SRR11523548/SRR11523548.fastq.gz`, but [ENA](https://www.ebi.ac.uk/ena/browser/view/SRR11523548) shows it as being `ftp://ftp.sra.ebi.ac.uk/vol1/fastq/SRR115/048/SRR11523548/SRR11523548_1.fastq.gz`.

### Solution or next step

I believe that function works based off of patterns that we found. It seems like things may have shifted or went an increment that caused it to seem different. I'm not exactly sure, but perhaps it can be fixed or perhaps there's a less hacky/brittle way of determining this FTP URL.

jashapiro:
There is! Use the ENA API: Here is some R code I used to get file urls a project at a time. It should be adaptable...

```
get_sample_df <- function(SRA_project){
  fields <- paste("run_accession",
                  "sample_accession",
                  "sample_alias",
                  "sample_title",
                  "experiment_title",
                  "fastq_ftp",
                  "fastq_md5",
                  "fastq_bytes",
                  sep = ",")
  ena_url <- stringr::str_interp("http://www.ebi.ac.uk/ena/data/warehouse/filereport?accession=${SRA_project}&result=read_run&fields=${fields}")
  ena_table <- readr::read_tsv(ena_url) %>%
    tidyr::separate_rows(fastq_ftp, fastq_md5, fastq_bytes, sep = ";")
}

**Comment by jashapiro** [comment]
The docs have improved somewhat from when I was working that out: https://www.ebi.ac.uk/ena/portal/api/#/Portal%20API/fileReportUsingGET

**Comment by jashapiro** [comment]
For the specific query listed above, the problem seems to be there are 2 files rather than the 1 expected?

I used the following query to the API:
`https://www.ebi.ac.uk/ena/portal/api/filereport?accession=SRR11523548&format=JSON&result=read_run`

And received the following response:
```
[
  {
    "run_accession": "SRR11523548",
[...]

PR #2865, opened 2021-07-22
Use file-report endpoint to determine unmated reads and check file properties

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/2811
https://github.com/AlexsLemonade/refinebio/issues/2593

This fixes both of the above issues by using ENA's file-report endpoint to determine the file names and check for the unmated reads. This eliminates our need to hit the FTP server altogether. This endpoint also has md5 and size_in_bytes properties for the files so I am checking those in the downloader when they're set.

## Types of changes

## Functional tests

The unit tests are going to test this pretty well. I'll run an end-to-end test as well.

## Checklist

## Screenshots


common/data_refinery_common/utils.py
```diff
@@ -262,6 +262,13 @@ def calculate_sha1(absolute_file_path):
         for buf in iter(partial(open_file.read, io.DEFAULT_BUFFER_SIZE), b""):
             hash_object.update(buf)
```
wvauclain, line 263: It looks like you stole the line `return hash_object.hex_digest()` from the end of this function

kurtwheeler:
Ok! I think the tests should pass! The foreman tests finally passed... the downloader tests only failed because of building an image, so I've retried them but they worked in the previous commit and I only updated foreman code!

foreman/data_refinery_foreman/surveyor/test_sra.py
```diff
@@ -112,6 +112,34 @@ def test_srp_survey(self, mock_send_task):
...
+        sra_surveyor = SraSurveyor(survey_job)
+        experiment, samples = sra_surveyor.discover_experiment_and_samples()
+
+        self.assertEqual(experiment.accession_code, "SRP048683")
+        self.assertEqual(len(samples), 12)
+
+        # Just check one file for one sample's expected file size/md5
+        for sample in samples:
+            if sample.accession_code == "SRR1603661":
+                for original_file in sample.original_files.all():
```
wvauclain, line 136: I think we should also assert that `set(o.source_filename for o in sample.original_files.all())` is equal to what we expect, because if there isn't an original file with source filename `SRR1603661_1.fastq.gz` then the check succeeds when it should not.

foreman/data_refinery_foreman/surveyor/test_sra.py
```diff
@@ -112,6 +112,34 @@ def test_srp_survey(self, mock_send_task):
         self.assertEqual(experiment.alternate_accession_code, None)
         self.assertEqual(len(samples), 9)
 
+    @vcr.use_cassette("/home/user/data_store/cassettes/surveyor.sra.survey_file_report.yaml")
+    @patch("data_refinery_foreman.surveyor.external_source.send_job")
+    def test_survey_file_report(self, mock_send_job):
```
wvauclain, line 117: Do we have a test that makes sure that unmated reads are detected properly?

common/data_refinery_common/models/sample_annotation.py
```diff
@@ -19,6 +19,7 @@ class Meta:
     sample = models.ForeignKey("Sample", blank=False, null=False, on_delete=models.CASCADE)
 
     # Properties
+    name = models.TextField(null=True)
```
wvauclain, line 22: Do we use this field anywhere?

workers/data_refinery_workers/downloaders/sra.py
```diff
@@ -30,19 +28,22 @@
...
+    download_url = original_file.source_url
     # SRA files have Apsera downloads.
-    if "ftp.sra.ebi.ac.uk" in download_url and not force_ftp:
-        # From: ftp://ftp.sra.ebi.ac.uk/vol1/fastq/ERR036/ERR036000/ERR036000_1.fastq.gz
-        # To: era-fasp@fasp.sra.ebi.ac.uk:/vol1/fastq/ERR036/ERR036000/ERR036000_1.fastq.gz
-        download_url = download_url.replace("ftp://", "era-fasp@")
-        download_url = download_url.replace("ftp", "fasp")
+    if "ftp.sra.ebi.ac.uk" in download_url:
+        # From: ftp.sra.ebi.ac.uk/vol1/fastq/SRR735/005/SRR7353755/SRR7353755_1.fastq.gz
+        # To: era-fasp@fasp.sra.ebi.ac.uk:vol1/fastq//SRR735/005/SRR7353755/SRR7353755_1.fastq.gz
```
wvauclain, line 39:
```suggestion
        # To: era-fasp@fasp.sra.ebi.ac.uk:/vol1/fastq/SRR735/005/SRR7353755/SRR7353755_1.fastq.gz
```

wvauclain (approved):
Few questions, but other than that LGTM!

common/data_refinery_common/models/sample_annotation.py
kurtwheeler: Oh good catch, I had added this because I was going to record the file report as an annotation but then decided to instead just stick it directly on that model.

foreman/data_refinery_foreman/surveyor/test_sra.py
kurtwheeler: Yes, that's this one. I've renamed it to make that more clear.

kurtwheeler:
The only error was a 429. I'm gonna merge this!

PR #2871, opened 2021-07-26
Adds a management command for building a chart to show dataset downloads

kurtwheeler:
## Issue Number

Yum yums

This will be useful in the future, although I'll need to modify it a bit for next quarter.


wvauclain (approved):
Exciting!

Issue #2849, opened 2021-07-13
Improve the compendia tests


wvauclain:
### Context

Making tracking issues for the outstanding processor test improvements.

### Problem or idea

> The are two main parts of the compendia pipeline that I think could be tested more thouroughly
> 
> Firstly, _prepare_frames could be tested pretty easily I think. This function uses smashing_utils.process_frames_for_key to set up the job context before the next steps. We could call this function and assert that the resulting job context is set up correctly, or it might make more sense to just test smashing_utils.process_frames_for_key in the util test file.
> 
> Secondly, and most importantly we should definitely test the massive function _perform_imputation. Right now our tests just verify that the pipeline ran successfully and that the compendium_result exists. We should make sure that the intermediate tests work as well:
> 
> - We should test that imputation is done correctly:
>   - Remove some values from the matrix.
>   - Run imputation.
>   - Compare the actual and imputed values.
> - We filter both rows and columns during creation of compendia, we should test that the expected rows and columns are dropped from the matrices. These filters are explained in the comments of the create_compendia.py processor file, but repeated here for convenience:
>   - Drop all rows in rnaseq_expression_matrix with a row sum < 10th percentile of rnaseq_row_sums; this is now filtered_rnaseq_matrix
>   - Remove samples (columns) with >50% missing values in combined_matrix
> - We should also test that quantile normalization is skipped for quantpendia.

### Solution or next step

I started working on these, but I put them aside so I could get the tests finished that were stopping us from processing. I need to re-evaluate what I have done so far, and continue from there.
- [x] We should test that imputation is done correctly:
  - Remove some values from the matrix.
  - Run imputation.
  - Compare the actual and imputed values.
- We filter both rows and columns during creation of compendia, we should test that the expected rows and columns are dropped from the matrices. These filters are explained in the comments of the create_compendia.py processor file, but repeated here for convenience:
  - [x] Drop all rows in rnaseq_expression_matrix with a row sum < 10th percentile of rnaseq_row_sums; this is now filtered_rnaseq_matrix
  - [x] Remove samples (columns) with >50% missing values in combined_matrix
- [x] We should also test that quantile normalization is skipped for quantpendia.

PR #2879, opened 2021-07-30
Compendia Test Improvements

wvauclain:
## Issue Number

Fixes https://github.com/AlexsLemonade/refinebio/issues/2849

Add some more tests to the compendia:
- Make sure that we can smash with microarray data only
  - The comments in `create_compendia.py` indicate that it was designed to do this, but we never verified that it still worked
- Check that we drop rows correctly when their row sums are in the < 10th percentile
- Check that we drop samples correctly when there is >50% missing observations
- Try removing some values and performing imputation, and verify that the results are close to expected
- Check that QN is skipped for quantpendia by checking the output metadata

## Methods

The one thing I would like some science input on is the imputation testing. I picked 3 entries to set to NA, which I do here:
https://github.com/AlexsLemonade/refinebio/blob/wvauclain/compendia-test-improvements/workers/data_refinery_workers/processors/test_compendia.py#L594,L604

and then I check to make sure that after imputation they are close to the measured values here:
https://github.com/AlexsLemonade/refinebio/blob/wvauclain/compendia-test-improvements/workers/data_refinery_workers/processors/test_compendia.py#L638-L661

@jaclyn-taroni is this what you had in mind? Or is there a different way you want me to remove observations / see how good our imputation was?

## Types of changes

## Functional tests

The tests pass locally

## Checklist

### DIFF
--- a/workers/data_refinery_workers/processors/create_compendia.py
+++ b/workers/data_refinery_workers/processors/create_compendia.py
@@ -85,6 +85,14 @@ def _prepare_input(job_context: Dict) -> Dict:
 
 
 def _prepare_frames(job_context: Dict) -> Dict:
+    """
+    Takes the inputs and places them into matrices
+     - Combine all microarray samples with a full join to form a
+       microarray_expression_matrix (this may end up being a DataFrame).
+     - Combine all RNA-seq samples (lengthScaledTPM) with a full outer join
+       to form a rnaseq_expression_matrix.
+     - Adds unsmashable files to `job_context["unsmashable_files"]`
+    """
     start_prepare_frames = log_state("start _prepare_frames", job_context["job"].id)
 
     job_context["unsmashable_files"] = []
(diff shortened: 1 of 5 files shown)


workers/data_refinery_workers/processors/create_compendia.py
```diff
@@ -85,6 +85,14 @@ def _prepare_input(job_context: Dict) -> Dict:
 
 
 def _prepare_frames(job_context: Dict) -> Dict:
+    """
+    Takes the inputs and places them into matrices
+     - Combine all microarray samples with a full join to form a
+       microarray_expression_matrix (this may end up being a DataFrame).
```
kurtwheeler, line 91: I know you only copied this, but does it end up being a DataFrame?

workers/data_refinery_workers/processors/create_compendia.py
```diff
@@ -119,132 +127,164 @@ def _prepare_frames(job_context: Dict) -> Dict:
...
 
 
-def _perform_imputation(job_context: Dict) -> Dict:
+def _filter_rnaseq_matrix(job_context: Dict) -> Dict:
+    """
+    - Calculate the sum of the lengthScaledTPM values for each row (gene) of
+      the rnaseq_expression_matrix (rnaseq_row_sums).
+    - Calculate the 10th percentile of rnaseq_row_sums
+    - Drop all rows in rnaseq_expression_matrix with a row sum < 10th percentile of
+      rnaseq_row_sums; this is now filtered_rnaseq_matrix
```
kurtwheeler, line 136: This also drops genes that are NA, probably worth mentioning here.

workers/data_refinery_workers/processors/create_compendia.py
```diff
@@ -119,132 +127,164 @@ def _prepare_frames(job_context: Dict) -> Dict:
...
+    - Perform a full outer join of microarray_expression_matrix and
+      log2_rnaseq_matrix; combined_matrix
+    """
 
-        log_state("end caching zeroes", job_context["job"].id, cache_start)
+    # We potentially can have a microarray-only compendia but not a RNASeq-only
+    # compendia, so we need to provide a default
+    log2_rnaseq_matrix = job_context.pop("log2_rnaseq_matrix", None)
 
     outer_merge_start = log_state("start outer merge", job_context["job"].id)
```
kurtwheeler, line 250: Can we actually move this into the `if` statement so we only log about it if we actually do it?

workers/data_refinery_workers/processors/test_compendia.py
```diff
@@ -71,81 +63,74 @@ def test_create_compendia(self):
...
+                "data_dir": DATA_DIR,
+            },
+            experiment,
+        )
 
         dset = Dataset()
-        dset.data = {"GSE1487313": ["GSM1487313", "GSM1487222"], "SRX332914": ["SRS332914"]}
+        dset.data = {
+            "GSE1487313": ["GSM1487313", "GSM1487222"],
+            "SRX332914": ["SRS332914"],
```
kurtwheeler, line 128: Is this supposed to be an SRX? I know it was before... but that's kinda confusing.

workers/data_refinery_workers/processors/test_compendia.py
```diff
@@ -325,16 +269,393 @@ def test_create_compendia_danio(self):
...
+        job_context["microarray_matrix"].loc[
+            "ENSDARG00000000151", "02c3v1urzlq9o499ye6315v3_GSM396530.PCL"
+        ] = np.nan
+
+        final_context = utils.run_pipeline(
+            job_context, [create_compendia.COMPENDIA_PIPELINE[imputation_index]]
+        )
+        self.assertDidNotFail(job)
+
+        print(
```
kurtwheeler, line 611: Was this a debugging print?

kurtwheeler (approved):
Only a couple minor comments but other than that this looks fantastic! Nice work!

jaclyn-taroni:
> The one thing I would like some science input on is the imputation testing. I picked 3 entries to set to NA, which I do here:
> https://github.com/AlexsLemonade/refinebio/blob/wvauclain/compendia-test-improvements/workers/data_refinery_workers/processors/test_compendia.py#L594,L604
> 
> and then I check to make sure that after imputation they are close to the measured values here:
> https://github.com/AlexsLemonade/refinebio/blob/wvauclain/compendia-test-improvements/workers/data_refinery_workers/processors/test_compendia.py#L638-L661
> 
> @jaclyn-taroni is this what you had in mind? Or is there a different way you want me to remove observations / see how good our imputation was?

I'm going to start with summarizing my understanding of what the test compendium looks like and then how the imputation testing might be improved, provided my understanding is correct. 

The test compendium (getting this from [here](https://github.com/AlexsLemonade/refinebio/blob/9ab4da9266ccd5599dd4fd1b99266c81702681b6/workers/data_refinery_workers/processors/test_compendia.py#L310-L312)):

- Comes from zebrafish
- Is comprised of 420 microarray and 420 RNA-seq samples
- 1 sample has a missing file that we make sure gets dropped
- 9 samples have 50% or more missing values

A common scenario we anticipate in compendia is that we will have Ensembl gene identifiers where a substantial (up to 30%) of the entries are missing and sometimes that will be because a gene is not assayed on a microarray platform (or platforms) that makes up a decent share of the samples that get included.

So I think about masking (setting to NA) values in a few different ways: 

- Selecting entries to mask completely at random – basically what you did here but on a larger scale!
- Selecting rows randomly and masking up to 30% of the values

_I'll note that for RNA-seq we expect that genes with certain properties (namely shorter genes) to be more likely to be "missing," but I don't think we necessarily need to get into that **here**._ 

And then as far as assessing the performance of imputation: 

- I think we should be calculating something like RMSE. 
- We can calculate the RMSE initially, e.g., under these conditions, and then we'd want to set some tolerance for test whether things have changed/deviated too much (`test RMSE < initial RMSE + some amount`)


I don't feel comfortable commenting on the feasibility / scope of implementing what I've outlined, so let's call what I've outlined a starting point for "negotiations" :smile:

wvauclain:
@jaclyn-taroni I was able to implement all of your suggestions, and it should pretty much be ready to go now.

There are two last things that need to be hammered down in my implementation:
- I used a fix seed for the random generator so that the tests would behave predictably. Is this okay, or would you prefer it if we don't use a set seed?
- Right now the RMSE is around `0.28`, and I allowed for a tolerance of `0.05`. This might be a little bit generous, especially if we don't have a set seed, so I might reduce it to something like `0.01` if we choose not to randomize.

jaclyn-taroni:
> * I used a fix seed for the random generator so that the tests would behave predictably. Is this okay, or would you prefer it if we don't use a set seed?
> 
> * Right now the RMSE is around `0.28`, and I allowed for a tolerance of `0.05`. This might be a little bit generous, especially if we don't have a set seed, so I might reduce it to something like `0.01` if we choose not to randomize.

Both of these items – setting a seed and reducing the tolerance to `0.01` – seem good to me (at least for now)! I think our _main objective_ with these are to make sure the behavior of the imputation has not changed without us knowing about it, which I expect setting a seed would still allow us to accomplish.

jaclyn-taroni:
I'm going to let @kurtwheeler review the _implementation_, but I am onboard with the concepts! So, I'd say that – provided the implementation is aligned with the concepts laid out – you should consider this a 👍🏻  from me!

workers/data_refinery_workers/processors/test_compendia.py
```diff
@@ -591,44 +633,68 @@ def test_imputation(self):
...
 
-        job_context["microarray_matrix"].loc[
-            "ENSDARG00000000151", "02c3v1urzlq9o499ye6315v3_GSM396530.PCL"
-        ] = np.nan
+        # Now pick some rows to remove according to the instructions from
+        # https://github.com/AlexsLemonade/refinebio/pull/2879#issuecomment-895143336
+
+        random.seed(42)
+
+        # Select some rows randomly and mask ~30% of the values
```
kurtwheeler, line 646: Where does the 30% come in? Is that because there's like 366 columns?

workers/data_refinery_workers/processors/test_compendia.py
```diff
@@ -325,16 +273,428 @@ def test_create_compendia_danio(self):
...
+        rare_rows = random.sample(list(job_context["microarray_matrix"].index), k=25)
+        rare_genes = {}
+        for row in rare_rows:
+            cols = random.sample(list(job_context["microarray_matrix"].columns), k=110)
+            rare_genes[row] = cols
+            for col in cols:
+                job_context["microarray_matrix"].loc[row, col] = np.nan
+
+        # Now some entries from the other rows to mask completely at random
```
kurtwheeler, line 655:
```suggestion
        # Now some entries from the other rows to completely mask at random
```
I originally thought this meant "completely at random" until I reread Jackie's message.

workers/data_refinery_workers/processors/test_compendia.py
```diff
@@ -325,16 +273,428 @@ def test_create_compendia_danio(self):
...
+
+        index = set(final_context["merged_no_qn"].index) & set(
+            expected_context["merged_no_qn"].index
+        )
+        columns = set(final_context["merged_no_qn"].columns) & set(
+            expected_context["merged_no_qn"].columns
+        )
+
+        N = 0
+        se = 0
```
kurtwheeler, line 681: What does `se` mean? Is there a more descriptive name we can use? Maybe `error_squared`?

kurtwheeler (approved):
These changes LGTM! It seems you're doing the things Jackie recommended! Just a couple comments around readability to address.

workers/data_refinery_workers/processors/test_compendia.py
wvauclain: Good catch, turns out that when I rewrote the tests we ended up with a lot more columns than I expected.

I updated this code to be more explicit and fixed this to be the correct number.

Issue #2861, opened 2021-07-20
Run some never-before-run experiments just for cost estimation


kurtwheeler:
### Context

https://github.com/AlexsLemonade/refinebio/issues/2859 mentions that

> We'll continue to hone our estimates with the dataset request backlog.

However, a lot of the backlog are experiments we've attempted before, so the costs for them are convoluted because some samples have been run already, and the ones that haven't are generally more problematic.

### Problem or idea

The best cost estimation of "happy path" or "best case" processing will be on fresh, successful experiments.

### Solution or next step

We should run some experiments we've never tried before just to get a sense for how much they cost if they go well.

PR #2886, opened 2021-08-04
Rework get_quant_results_for_experiment to run in a few smaller queries.

kurtwheeler:
## Issue Number

#2861 

I was trying to run tximport for one of the experiments that should have been eligible and the command errored out because of a database timeout. I think this should be more manageable for the database.

## Types of changes

- Optimization

## Functional tests

Unit tests are passing. I tried this in a shell on the foreman but I got
```
django.core.exceptions.FieldError: Invalid field name(s) given in select_related: 'computedfile'. Choices are: processor, organism_index
```
which is weird because it works in the code that's already there and in the tests.

## Checklist


common/data_refinery_common/rna_seq.py
```diff
@@ -78,40 +77,37 @@ def get_quant_results_for_experiment(experiment: Experiment, filter_old_versions
...
-    newest_computational_results = eligible_results.filter(
-        samples=OuterRef("id"), processor__name=ProcessorEnum.SALMON_QUANT.value["name"],
-    ).order_by("-created_at")
-
-    # Annotate each sample in the experiment with the id of the most recent computational result
-    computational_results_ids = (
-        experiment.samples.all()
-        .annotate(
-            latest_computational_result_id=Subquery(newest_computational_results.values("id")[:1])
+        all_results.select_related("computedfile").filter(
```
wvauclain, line 91: Should this line be outside the `if` statement?

common/data_refinery_common/rna_seq.py
```diff
@@ -78,40 +77,37 @@ def get_quant_results_for_experiment(experiment: Experiment, filter_old_versions
...
-        .values_list("latest_computational_result_id", flat=True)
-    )
 
-    # return the computational results that match those ids
-    return ComputationalResult.objects.all().filter(id__in=computational_results_ids)
+    def get_sample_id_set(result):
+        return {sample.id for sample in result.samples.all()}
+
+    latest_results = set()
+    for k, group in groupby(all_results, get_sample_id_set):
```
wvauclain, line 99: Do you have to sort before you `groupby` here? The `itertools` docs say that `groupby` doesn't automatically sort beforehand: https://docs.python.org/3/library/itertools.html#itertools.groupby

> The operation of groupby() is similar to the uniq filter in Unix. It generates a break or new group every time the value of the key function changes (which is why it is usually necessary to have sorted the data using the same key function).

foreman/data_refinery_foreman/foreman/management/commands/rerun_salmon_old_samples.py
```diff
@@ -18,15 +18,21 @@
...
+            if quant_result.created_at > last_created.created_at:
+                last_created = quant_result
+
+    latest_salmon_version = last_created.organism_index.salmon_version
 
     total_samples_queued = 0
-    latest_salmon_version = None
     for quant_result in quant_results:
         if not latest_salmon_version:
             # we can safely ignore the latest salmon version, that will be the first
```
wvauclain, line 38: I don't think the comment here still applies, because we are no longer ordering by `-organism_index__created_at`. This branch also might be unreachable because `latest_salmon_version` is already set before this loop.

common/data_refinery_common/rna_seq.py
```diff
@@ -36,17 +37,15 @@ def should_run_tximport(experiment: Experiment, results, is_tximport_job: bool):
 
     results is a queryset of ComputationalResults for the samples that had salmon quant run on them.
     """
-    num_quantified = results.count()
+    num_quantified = len(results)
```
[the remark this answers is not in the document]
kurtwheeler: Well the primary reason for switching here wasn't because of an optimization between count/len, it was because it's not a quesryset any more. I just read the docstring right above this that is out of date now. I'll update that.

common/data_refinery_common/rna_seq.py
kurtwheeler: Eww WTF? Why does python's groupby suck so bad? I miss clojure

Good catch though, I didn't read the docs carefully enough and made assumptions about how easy it should be to use.

foreman/data_refinery_foreman/foreman/management/commands/rerun_salmon_old_samples.py
kurtwheeler: Good catch, this is dead code that should have come up when I added the `last_created` loop.

common/data_refinery_common/rna_seq.py
```diff
@@ -36,17 +37,15 @@ def should_run_tximport(experiment: Experiment, results, is_tximport_job: bool):
...
-    num_salmon_versions = (
-        results.filter(organism_index__salmon_version__isnull=False)
-        .values_list("organism_index__salmon_version")
-        .distinct()
-        .count()
-    )
-    if num_salmon_versions > 1:
+    salmon_versions = set()
+    for result in results:
+        salmon_versions.add(result.organism_index.salmon_version)
```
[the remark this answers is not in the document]
kurtwheeler: I think that shouldn't be None, but it's a good thing to check!

wvauclain (approved):
LGTM! It looks like there's one small thing that needs to be fixed for the tests though

common/data_refinery_common/rna_seq.py
```diff
@@ -75,43 +72,40 @@ def should_run_tximport(experiment: Experiment, results, is_tximport_job: bool):
...
-            latest_computational_result_id=Subquery(newest_computational_results.values("id")[:1])
-        )
-        .filter(latest_computational_result_id__isnull=False)
-        .values_list("latest_computational_result_id", flat=True)
-    )
+    def get_sample_id_set(result):
+        return {sample.id for sample in result.samples.all()}
+
+    latest_results = set()
+    for k, group in groupby(sorted(list(all_results), get_sample_id_set), get_sample_id_set):
```
wvauclain, line 97: `sorted` accepts any iterable, so I don't think you need `list(all_results)`. Also the tests are failing because `sorted` needs `key` to be passed as a kwarg

kurtwheeler: Yeah I figured that out despite the red herring error message. It's weird to me that calling sorted on a set is fine but I guess python just assumes that if you want something sorted it's gotta be in a list.

Issue #2848, opened 2021-07-13
Add cron job tests


wvauclain:
### Context

Making tracking issues for the outstanding processor test improvements

### Problem or idea

> In addition to improving the existing tests as listed in the links below, we also would like to run a few tests on a weekly basis to make sure things aren't drifting without us knowing. These can be configured to run as cron jobs on the foreman server. These should check:
> 
> - that brainarray has not drastically changed genes since the last time it was run
> - that a newly generated QN reference file is similar enough to the latest version currently being used.
> 

From a comment on the PR:
> I think we'd probably want to compare the distributions instead (Kolmogorov-Smirnov test is what immediately comes to mind). I'd say that this is fine for now and we can nail down specifics later.

### Solution or next step

Figure out how we want the tests to run, and how we want to be notified if they fail. Probably a slack message and a GitHub issue.

wvauclain:
Comment on the processor jobs PR:
> Similar to what checking about genes for BrainArray does for our Affymetrix microarray samples, we could also consider checking the genes (or transcripts) that show up in different transcriptome indices that affect our RNA-seq samples.

Issue #2887, opened 2021-08-09
QN reference generation is broken


wvauclain:
### Context

Trying to test my last cron job test.

### Problem or idea

```
2021-08-09 21:10:36,468 i-0c7b928a370a5cc81 data_refinery_common.message_queue WARNING [job_name: QN_REFERENCE] [reason: An error occurred (ClientException) when calling the SubmitJob operation: JobDefinition name will_dev_QN_REFERENCE does not exist or is not in ACTIVE status] [job_id: 5757]: Unable to Dispatch Batch Job.
Traceback (most recent call last):
  File "manage.py", line 22, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.6/dist-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.6/dist-packages/django/core/management/__init__.py", line 413, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python3.6/dist-packages/django/core/management/base.py", line 354, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python3.6/dist-packages/django/core/management/base.py", line 398, in execute
    output = self.handle(*args, **options)
  File "/home/user/data_refinery_foreman/surveyor/management/commands/dispatch_qn_jobs.py", line 107, in handle
    dispatch_qn_job_if_eligible(organism)
  File "/home/user/data_refinery_foreman/surveyor/management/commands/dispatch_qn_jobs.py", line 87, in dispatch_qn_job_if_eligible
    send_job(ProcessorPipeline.QN_REFERENCE, job)
  File "/usr/local/lib/python3.6/dist-packages/data_refinery_common/message_queue.py", line 322, in send_job
    parameters={"job_name": job_type.value, "job_id": str(job.id)},
  File "/usr/local/lib/python3.6/dist-packages/botocore/client.py", line 386, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/usr/local/lib/python3.6/dist-packages/botocore/client.py", line 705, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.errorfactory.ClientException: An error occurred (ClientException) when calling the SubmitJob operation: JobDefinition name will_dev_QN_REFERENCE does not exist or is not in ACTIVE status
```

### Solution or next step

[...]

PR #2888, opened 2021-08-10
Cron job tests

wvauclain:
## Issue Number

Fixes https://github.com/AlexsLemonade/refinebio/issues/2848

Adds three new tests that will run every week as cron jobs.

Something that we still need to figure out, possibly after fixing https://github.com/AlexsLemonade/refinebio/issues/2887, is how we want to run the QN reference cron job tests. I don't think that the foreman has enough RAM to generate a new QN reference for some of the large species, so for now this test is not enabled.

## Methods

For Brainarray gene IDs and transcriptome index transcript IDs, we use a Jaccard Index test, and signal failure if the Jaccard Index calculated between the IDs from last week and from this week is less than 0.95.

For the QN references, we use a 2-sample KS test to check how similar a newly-generated QN reference is to the current one we are using and signal failure if the p-value is less than 0.05.

## Types of changes

## Functional tests

I spun up a devstack and made sure that all of the tests behaved as expected and had the proper S3 creds.

## Checklist


wvauclain:
I do not expect the Affy tests to fail until we do a staging deploy, because I updated the version of `rpy2`. The Affy tests pull the current staging Affy docker image, which contains the old version of `rpy2`, but there were some backwards-incompatible changes in the newer version.

jaclyn-taroni:
Methods sound reasonable 👍🏻  Nice work!

workers/data_refinery_workers/processors/test_qn_reference.py
```diff
@@ -251,3 +252,32 @@ def test_qn_management_command(self):
         # There's not enough samples available in this scenario so we
         # shouldn't have even made a processor job.
         self.assertEqual(ProcessorJob.objects.count(), 0)
+
+    @tag("qn")
+    def test_qn_target_without_updating_organism(self):
+        """In the cron job tests, we want to create a new QN target without
+        actually setting it as the new default for the experiment."""
```
kurtwheeler, line 259:
```suggestion
        """Test that create_results=False will not update the experiment.
        
        This is because in the cron job tests, we want to create a new QN target without
        actually setting it as the new default for the experiment."""
```

workers/data_refinery_workers/processors/qn_reference.py
```diff
@@ -120,40 +120,41 @@ def _build_qn_target(job_context: Dict) -> Dict:
...
+#             "Encountered error in R code while running qn_reference.R"
+#             " pipeline during processing of {0}: {1}"
+#         )
+#         error_message = error_template.format(job_context["smashed_file"], str(e))
+#         logger.warn(error_message, processor_job=job_context["job_id"])
+#         job_context["job"].failure_reason = error_message
+#         job_context["success"] = False
 
-    return job_context
+#     return job_context
```
kurtwheeler, line 152: Is this unused? Any reason to keep it around still? It looks like it's been replaced by _build_qn_target?

workers/data_refinery_workers/processors/management/commands/check_qn_reference_agreement.py
```diff
@@ -0,0 +1,117 @@
...
+    )
+
+
+def check_qn_reference_agreement():
+    bad_organisms = []
+    for organism in Organism.objects.filter(qn_target__isnull=False):
+        final_context = create_qn_target.create_qn_target(
+            organism,
+            platform=create_qn_target.get_biggest_platform(organism),
+            create_results=False,
```
kurtwheeler, line 61: `create_results` will only prevent the database objects from being created right? It doesn't prevent the files from being written to disk? Will these accumulate?

kurtwheeler (changes requested):
This looks pretty good to me but I had some questions/comments.

workers/data_refinery_workers/processors/management/commands/check_qn_reference_agreement.py
wvauclain: Nice catch! I originally was using the files in the cron job test. Because I use the raw dataframe in the cron job test now, I added a check for `create_results` before writing files to the disk.

workers/data_refinery_workers/processors/management/commands/check_qn_reference_agreement.py
```diff
@@ -0,0 +1,117 @@
...
+
+        existing_target_computedfile = organism.qn_target.computedfile_set.latest()
+        existing_target_filename = existing_target_computedfile.sync_from_s3()
+        existing_target = pd.read_csv(
+            existing_target_filename, header=None, sep="\t", encoding="utf-8"
+        )[0]
+        # We don't want this file hanging around on the foreman for too long
+        existing_target_computedfile.delete_local_file()
+
+        _, pvalue = scipy.stats.ks_2samp(new_target, existing_target)
```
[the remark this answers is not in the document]
wvauclain: @jaclyn-taroni can confirm, but I think checking the pvalue makes more sense than trying to interpret the raw test statistic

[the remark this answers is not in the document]
jaclyn-taroni: Agree that using the arbitrary but commonly used threshold with a p-value makes sense. Something that this brought up for me, though, is that we probably want to be able to easily surface the values themselves (KS test p-value or Jaccard index) even if the conditions are passing…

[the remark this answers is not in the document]
wvauclain: All of the test statistics are logged, along with whether or not that test passed, so if we want to check the values for each organism/platform/etc. we can look at the log file on the foreman instance.

kurtwheeler:
I fixed the syntax but there's a weird error coming out of the `qn_test` that is fixed in #2894 so I'm just going to wait for that to get merged to fix that test.

davidsmejia (approved):
LGTM

kurtwheeler (approved):

PR #2891, opened 2021-08-11
Trying only switching to fancySVD.

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/fancySVD/pull/1

The tests broke because of a new fancyimpute version makes the delicate dependency balancing act we were doing fall apart. Therefore I've created a fancySVD package that doesn't need many dependencies. This is much easier to install and I think is maybe even working.

## Types of changes

## Functional tests

`./workers/run_tests.sh data_refinery_workers.processors.test_imports.ImportTestCase.test_compendia_imports` worked locally!

## Checklist

### DIFF
--- a/workers/data_refinery_workers/processors/compendia.yml
+++ b/workers/data_refinery_workers/processors/compendia.yml
@@ -14,4 +14,4 @@ python:
   - data-refinery-common
   - pandas
   - numpy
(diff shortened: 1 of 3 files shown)


davidsmejia (approved):
🕯️

Issue #2878, opened 2021-07-30
Rewrite KS test for smasher/compendia jobs without using R


kurtwheeler:
### Context

We don't seem to be able to run the KS test for human/mouse samples because it OOM kills us.

### Problem or idea

A cursory glance of the `_test_qn`  function shows that this is likely because we're passing the entire matrix over to R to run the KS test: https://github.com/AlexsLemonade/refinebio/blob/dev/workers/data_refinery_workers/processors/smashing_utils.py#L512

### Solution or next step

scipy has a [KS test](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.kstest.html). If we just use this, then we can run it on the matrix that we already have loaded into memory. This should prevent us from being OOM-killed (unless the KS test is like really really inefficient).

PR #2894, opened 2021-08-12
Rewrite KS test in python

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/2878

This uses python to run a KS test rather than passing the whole matrix to R. We still use rpy2 to quantile normalize but we're already doing it in chunks so it shouldn't be too bad of a RAM-suck.

I've also reenabled the qn_test for compendia and set them back to public when they're created.

## Methods

@jaclyn-taroni it's probably appropriate for you to take a look to make sure I didn't miss any subtleties of the R code, but I think it was pretty straightforward.

## Types of changes

## Functional tests

Compendia and smasher tests are still passing! Or they were, they may have stopped, still investigating

## Checklist

## Screenshots


workers/data_refinery_workers/processors/smashing_utils.py
```diff
@@ -509,52 +511,33 @@ def _test_qn(merged_matrix):
...
+    if num_columns < 2:
         return None
 
-    # This wont work with larger matricies
-    # https://github.com/AlexsLemonade/refinebio/issues/1860
-    ncolumns = ncol(merged_R_matrix)
-
-    if ncolumns[0] <= 200:
+    if num_columns <= 200:
         # Convert to NP, Shuffle, Return to R
```
jaclyn-taroni, line 522: This comment needs updating

workers/data_refinery_workers/processors/smashing_utils.py
```diff
@@ -509,52 +511,33 @@ def _test_qn(merged_matrix):
...
+    random.seed(123)
+    num_columns = merged_matrix.shape[1]
     # Not enough columns to perform KS test - either bad smash or single sample smash.
-    if n < m:
+    if num_columns < 2:
         return None
 
-    # This wont work with larger matricies
-    # https://github.com/AlexsLemonade/refinebio/issues/1860
-    ncolumns = ncol(merged_R_matrix)
```
jaclyn-taroni, line 533: Do we still have this problem, such that we need the logic below, now that we're not passing things to R? We probably don't know _yet_ do we?

jaclyn-taroni (commented):
Doesn't look to me like you missed any subtleties or R code!

davidsmejia (approved):
This looks much cleaner. nice job!

workers/data_refinery_workers/processors/smashing_utils.py
kurtwheeler: I don't think we do, but I would be surprised if we did.

PR #2896, opened 2021-08-13
Install certbot via snap instead of apt.

kurtwheeler:
## Issue Number

N/A Certbot failed to install last night so we had no HTTPS

This is the new recommended way to install certbot. I did this on the prod server last night, so I think it'll work but I can't really test without at least hitting staging.


davidsmejia (approved):
🔒

Issue #2884, opened 2021-08-03
Add MetaSRA metadata to compendia/smasher metadata files


kurtwheeler:
### Context

Once we run https://github.com/AlexsLemonade/refinebio/issues/2868 we'll have metaSRA metadata for a lot of samples.

### Problem or idea

We should return this to users in their downloads!

### Solution or next step

Add these metadata values to the metadata files for smasher and compendia jobs.

Issue #2872, opened 2021-07-27
How should the predicted metadata be included in the downloadable metadata files?


dvenprasad:
### Context

Discussing how MetaSRA and contributor predicted metadata should be provided to users on download.

### Problem or idea

We have the MetaSRA predicted metadata and will likely have predicted metadata from other contributors in the future. Along with the predicted values, there is also confidence information available.

- How should these be included in the download file?
    - Separate file?
    - Same metadata file, with fields suffixed?
    - something else???
    
- How should the confidence information be provided? (if it should be provided)

### Solution or next step

We can start by dogfooding this one. Tagging our resident scientists  for their opinions @cbethell @jashapiro @allyhawkins @jaclyn-taroni

jaclyn-taroni:
Updated the title to be a little more informative. Here's the section of the docs on downloadable files for reference: http://docs.refine.bio/en/latest/main_text.html#downloadable-files

jaclyn-taroni:
These answers should be considered a starting point.

> * How should these be included in the download file?
>   
>   * Separate file?
>   * Same metadata file, with fields suffixed?
>   * something else???

Of these, same metadata with fields suffixes strikes me as the way to go, i.e., use `predicted_` the same way we use `refinebio_`. I would also like to mention that we may want to include the source in the prefix, too, as that may help to orient users when they go to look at docs or a publication for example. So for MetaSRA, perhaps we could use `predicted_metasra_` or maybe the prefix relies _solely_ on source: `metasra_`. 

However, I expect _a lot_ of ontology terms to be included in MetaSRA data. So a question I have, because I do not remember off of the top of my head, is: Do we expect every ontology term supported by MetaSRA to be included for every sample?

_If so_, I might change my answer to separate file?

> 
> * How should the confidence information be provided? (if it should be provided)

I think that this is probably a question for the TSV file because that seems trickier to me than the JSON, where including it would likely be pretty straightforward (but tell me if I'm wrong!). What about something like `predicted_metasra_<term>.confidence`?

wvauclain:
<details>
<summary>Here's an excerpt from the JSON file that we ingest from</summary>
```json
[
    {
        "sample_accession": "ERR2080620",
        "attributes": [
            {
                "EFO:0000721": {
                    "value": 6.0
                }
            }
        ]
    },
    {
        "sample_accession": "ERR3348231",
        "attributes": [
            {
                "EFO:0000246": {
                    "value": 3.0
                }
            }
[...]

PR #2899, opened 2021-08-13
Add MetaSRA metadata to the smasher's output

wvauclain:
## Issue Number

Closes https://github.com/AlexsLemonade/refinebio/issues/2884
Closes https://github.com/AlexsLemonade/refinebio/issues/2872

This PR (as of now) contains some WIP code to add MetaSRA metadata to the smasher's output. I tried to follow the guidelines in https://github.com/AlexsLemonade/refinebio/issues/2872 as closely as I could, but before merging this PR it is worth going back through and figuring out if we want to change anything.

## Methods

In the TSV files, metadata is added according to these rules for each attribute:

A column is created with the name `{source name}_{ontology term's human readable name}`. If the attribute has a unit, a second column is added with the name `{source name}_{ontology term's human readable name}.unit`, where the value is the unit's ontology term's human readable name. If the attribute has a confidence probability, we add a column with the name `{source name}_{ontology term's human readable name}.unit`.

One thing that still needs to be determined is what happens when the value is an ontology term, like if the attribute was `biological sex`.

In the JSON files, we serialize attributes according to the logic here: https://github.com/AlexsLemonade/refinebio/blob/5a219e644a396213edb1f6546ec0fe9744d399de/common/data_refinery_common/models/attributes.py#L34-L49

where the ontology term `to_dict()` method is
https://github.com/AlexsLemonade/refinebio/blob/5a219e644a396213edb1f6546ec0fe9744d399de/common/data_refinery_common/models/ontology_term.py#L86-L90

We also return the methods URLs of each contribution as a dict under the `methods` key in the metadata JSON file.

## Types of changes

## Functional tests

I modified the tests to check for some new attributes I added.

## Checklist


davidsmejia (approved):
LGTM

Issue #2092, opened 2020-01-22
Re-enable QN Test step in compendia generation


arielsvn:
### Context

Looks like we generate compendias with `is_public=False` https://github.com/AlexsLemonade/refinebio/pull/1891

### Problem or idea

We disabled QN Test step in the compendia and made them private while we restore that. I was testing the compendia generation and spent some time trying to figure out why they didn't appear in the API.

Created this issue to keep track of this.

### Solution or next step

- re-enable QN test step
- Make the generated `ComputationalResult` public https://github.com/AlexsLemonade/refinebio/blob/dev/workers/data_refinery_workers/processors/create_compendia.py#L396-L397

kurtwheeler:
Closed by #2905

PR #2905, opened 2021-09-02
Set compendia results back to public, remove ks-check as an option

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/2092

Pretty straightforward. We'll see how this works on smaller, then larger compendia.

This option shouldn't be needed any more, so I've just removed it entirely.

## Types of changes

## Functional tests

Unit tests will cover it some, but then I'll test in prod for scale.

## Checklist


davidsmejia (approved):
🕯️

Issue #2784, opened 2021-06-22
Add end to end tests for QN targets, compendia, quantpendia


kurtwheeler:
### Context

The end to end tests test almost everything, but QN targets, compendia, and quantpendia are still missing.

### Problem or idea

Add end to end tests for them.

kurtwheeler:
Closed by https://github.com/AlexsLemonade/refinebio/pull/2908

Issue #2772, opened 2021-06-16
Create separate RAM tiers for Compendia jobs and dispatch based on Organism


kurtwheeler:
### Context

Right now Compendia jobs only have one set amount of ram. However compendia jobs for Human and Mouse require way more RAM than the rest of the organisms.

### Problem or idea

In the past we've dealt with this by changing that value and redeploying. That is burdensome, requires dev time, and is slower.

### Solution or next step

We should have two ram tiers for Compendia/Quantpendia jobs and put Human/Mouse jobs on the big one and the rest on the small one.

wvauclain:
Once we do this we should be able to fix these lines:
https://github.com/AlexsLemonade/refinebio/blob/dev/workers/data_refinery_workers/processors/create_compendia.py#L396,L397

wvauclain:
Related to https://github.com/AlexsLemonade/refinebio/issues/2744

kurtwheeler:
Closed by #2908

Issue #2744, opened 2021-05-12
Queue small compendia jobs into smasher job queue


kurtwheeler:
### Context

AWS Batch has enabled us to have separate job queues for different types of jobs. Since human and mouse compendia are so large, I've created a separate queue for them.

### Problem or idea

At the moment all compendia jobs are being queued on the compendia queue, despite not needing to be so large.

### Solution or next step

We should only queue human and mouse into the compendia job queue. The other compendia jobs should just be run on the smasher instance since it's large enough.

kurtwheeler:
Datasets and processors do not have a relationship to organism though, so that will need to be established as part of this.

PR #2908, opened 2021-09-07
Get the *pendia jobs running

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/2784
https://github.com/AlexsLemonade/refinebio/issues/2772
https://github.com/AlexsLemonade/refinebio/issues/2744

This fixes queuing for compendia jobs. It splits where the jobs are queued based on organism, along with the RAM size of those jobs. This required adding RAM tiers for those job types.

## Types of changes

## Functional tests

End to end tests pass with a dev stack!!!

## Checklist


davidsmejia (approved):
LGTM

two comments though...
In `foreman/data_refinery_foreman/foreman/job_requeuing.py` 
* could we use `old_ram_amount` and set the `new_ram_amount` it is hard to read when we use `new_ram_amount` everywhere
* Should store the ram tiers in an enum or some other mechanism so we can tell which tier they were and what the next ram tier should be?

PR #2913, opened 2021-09-09
Fix the ACM certs and also remove a debugging log.

kurtwheeler:
## Issue Number

We should probably have an issue, but we've just been working off an email AWS sent about our cert being unable to renew

This hopefully will allow our cert to renew.

## Functional tests

Can't really test without at least staging.


davidsmejia (approved):
LGTM!

PR #2923, opened 2021-09-14
Remove ACM cert, cloudfront distribution, and related s3 buckets.

kurtwheeler:
## Issue Number

#2922 

We can't get ACM to work, but it turns out we don't really need it! This removes that cert, the cloudfront distribution that used it, the static bucket the cloudfront distribution used to server, and a policy related to it.

## Types of changes

- Removing dead code

## Functional tests

I've run a dev deploy just to make sure terraform's happy but these resources didn't get setup for dev anyway.

## Checklist


davidsmejia (approved):
LGTM!

Issue #2534, opened 2020-09-03
Include compendium_version in compendium metadata files


kurtwheeler:
### Context

Steven asked me about compendium versions. We're tracking them, but not making them easy to use.

### Problem or idea

If you download a compendium, you might not keep track of what version it is. Even if you don't, it should be discoverable later by looking at the metadata files.

### Solution or next step

Write the compendium version to the metadta files for compendia and quantpendia.

PR #2926, opened 2021-09-14
Add compendium versions in metadata

kurtwheeler:
## Issue Number

#2534 

I ran into an inconsistency between `ComputationalResult.compendium_version` and `ComputedFile.compendia_version` so I resolved that as well. This will be backwards incompatible though...

I have a branch for refinebio-py ready. @davidsmejia will this impact the frontend at all?

## Types of changes

## Functional tests

I got the unit tests to pass!

## Checklist


davidsmejia (approved):
This should be good on the FE side of things. I think we only a special query param on the compendia endpoint which in turn uses `compendium_version` to get the correct/latest per organism


LGTM!

PR #2929, opened 2021-09-15
Make the compendia queue use on demand instances.

kurtwheeler:
## Issue Number

N/A I just thought of this. I should have thought about it sooner but at least I thought about it before instance cycling interrupted a job.

The big compendia jobs that get put into the compendia job queue take like 5 days, which means they'll probably get cycled at some point if they're spot instances. This puts them onto on demand instances.

## Types of changes

## Functional tests

I'll test this on staging because it has data for it.

### DIFF
--- a/infrastructure/batch/compute.tf
+++ b/infrastructure/batch/compute.tf
@@ -2,8 +2,10 @@
 # The default environment is a 2 vCPU spot cluster
 
 locals {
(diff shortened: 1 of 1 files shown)


davidsmejia (approved):
LGTM

kurtwheeler:
I'll note that I've run a dev deploy just to make sure terraform is cool with it. It hasn't finished, but it's cranking along happily.

PR #2930, opened 2021-09-16
Tweaks params for compendia compute env

kurtwheeler:
## Issue Number

#2929 

I think that one instance of that size was more vCPUs then our max, so it never came up and the jobs got stuck at runnable.

Also fixes RAM amounts for *pendia jobs.

## Types of changes

## Functional tests

I'm going to test in staging because there's enough human/mouse data there already.

## Checklist

- [x] Any dependent changes have been merged and published in downstream module


davidsmejia (approved):
LGTM

PR #2932, opened 2021-09-16
Remove computed files that were not uploaded to s3

kurtwheeler:
## Issue Number

#2854 

This seeks to prevent computed files from being left around if they fail to be uploaded to s3.

In addition, it cleans removes computed files that weren't uploaded to S3, along with cleaning up computational results that are relying on these computed files. It also marks samples that are processed as unprocessed if they don't have another live computed file or result, and then requeues a downloader job for them to be processed and have their files uploaded to s3.

I also renamed foreman/run_surveyor.sh to foreman/run_management_command.sh because I did most of that before, including in the README before. It just needed to renamed and changed in one script.

## Types of changes

## Functional tests

I ran it locally, but I don't have anything for it to do. I will dry run it on staging, run it for real on staging, dry run it on prod, then run it for real on prod, assuming that everything looks good along the way.

## Checklist


davidsmejia (approved):
LGTM

PR #2931, opened 2021-09-16
Deploy compendia infra tweaks

kurtwheeler:
## Issue Number

#2930 
#2929 

The compendia job is cranking along in staging! Let's go to prod!

PR #2933, opened 2021-09-17
Bump the timeouts for compendia to 10 days.

kurtwheeler:
## Issue Number

#2931 

The timeout was set for 3 hours. The compendia take more like 5 days I think. This sets it to 10 days, just so that if it takes longer than I thought it won't be wasted, but if it goes past 5 days I'll be checking on it daily to make sure it's not hung.

## Types of changes

## Functional tests

N/A


davidsmejia (approved):
🕯️

PR #2936, opened 2021-09-22
Make remove dead computed files less noisy

kurtwheeler:
## Issue Number

#2935 

I saw this exception a whole lot, which was to be expected since the whole reason we're deleting these files is that these fields are null.

It worked well and cleaned up the files though. There were no samples that needed be requeued, but that's not surprising to me since we've tested `retry_samples --computed-files-not-uploaded` in staging before.

## Types of changes

## Functional tests

N/A, I'll test again on staging but there's nothing left to be done so we won't even be able to hit this branch point. I think this is low risk though.

## Checklist


davidsmejia (approved):

davidsmejia (approved):
LGTM

PR #2943, opened 2021-09-29
Combine dep bumps and do 'em right.

kurtwheeler:
## Issue Number

#2937
#2938
#2939 
#2940 
#2941

Bump the deps based on dependabot's suggestions.


davidsmejia (approved):
LGTM

PR #2944, opened 2021-09-29
Add stop jobs command so I don't have to go into the database as much.

kurtwheeler:
## Issue Number

N/A This came up while I was running compendia

## Purpose/Implementation Notes

Stopping a compendia job I don't want to run any more is annoying because it gets retried. This is a better way of doing it.

## Types of changes

- New feature (non-breaking change which adds functionality)

## Functional tests

I've tested this on a dev stack

## Checklist

- [x] Lint and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)
- [x] Any dependent changes have been merged and published in downstream modules

PR #2945, opened 2021-09-29
Deploy compendia tweaks and dependecies to prod

kurtwheeler:
## Issue Number

#2944 
https://github.com/AlexsLemonade/refinebio/pull/2936
#2943 

I don't think any of these need to be tested on staging


davidsmejia (approved):
🚀

PR #2965, opened 2021-11-22
Fix transcriptome index file downloading/deletion

kurtwheeler:
## Issue Number

N/A trying to get staging end to end tests to pass

Most of the short original file's properties should be the same, but the absolute file path should not.

## Types of changes

## Functional tests

It's a bit late to start spinning up a dev stack, but I'll test this on one tomorrow.

## Checklist


davidsmejia (approved):
LGTM

PR #2971, opened 2021-11-30
Fix aws events log group ARN.

kurtwheeler:
## Issue Number

N/A discovered while pulling some stats

We discovered that a CloudTrail event was failing because of a ResourceNotFoundException. This event was trying to create a log stream inside of `"logGroupName": "/aws/events/data-refinery-compendia-log-group-circleci-pr",`, which seems to be missing the last two characters.

I think this may have been necessary in an older terraform/AWS provider version. The last log event that worked was around June 20th which is around the time we upgraded.

## Types of changes

## Functional tests

I've deployed a dev stack to make sure it's not broken, but this will really get tested in staging.

## Checklist


davidsmejia (approved):
🕯️

PR #2972, opened 2021-12-02
Removes checks from deploy.sh for unused env var

kurtwheeler:
## Issue Number

N/A came up during the deploy tour

We have determined these checks don't do anything because the GITHUB_ACTIONS env var isn't passed through to the deploy box. I did leave the check for it in `scripts/prepare_image.sh` because that is invoked withing github actions during the tests.

## Types of changes

- Remove dead code

## Functional tests

This will be checked partially by unit tests, but when this goes into a staging deploy I will double check its outputs to make sure there's nothing that shouldn't be there.

## Checklist


davidsmejia:
Can we add a note, probably in remote_deploy that reminds people that the output from this could be sensitive so point it to the log file.

davidsmejia (approved):

PR #2973, opened 2021-12-03
Remove CI ingress for production environment.

kurtwheeler:
## Issue Number

N/A came up while going through deploy processes

We haven't been doing this since we switched over to AWS Batch, but we should. However staging and dev still need access to the database to test end to end tests.

## Types of changes

## Functional tests

I am testing this in a dev stack, then will test with a staging deploy.

## Checklist


davidsmejia (approved):
LGTM

PR #2976, opened 2021-12-08
Add some changes to API startup for debugging certbot challenge

kurtwheeler:
## Issue Number

#2975 

See if curling ourselves can be a better test for the API server coming up. If the output changes before the target group attachment comes up, then I can use that output to poll until it does. In some cases it might be faster, but it'll also prevent us from ever running into that timing issue again.

## Types of changes

- Temporary debugging


davidsmejia (approved):
LGTM

davidsmejia (approved):

Issue #2883, opened 2021-08-03
Make run_tximport ignore samples that failed due to fragment mapping errors


kurtwheeler:
### Context

If someone uploads bad samples with a small experiment we won't be able to run tximport on it as things are now.

### Problem or idea

TBD

### Solution or next step

TBD

PR #2978, opened 2021-12-10
Ignore bad rna samples in tximport eligibility

kurtwheeler:
## Issue Number

#2883 

This changes the way tximport completion percentages are calculated. We now exclude samples that had fragment mapping errors entirely.

In order to make this more efficient, we are adding some fields to the samples:

* is_unable_to_be_processed
* last_processor_job
* last_downloader_job
* most_recent_smashable_file
* most_recent_quant_file

There is also an associated command to populate these fields. It may or may not be efficient enough as it is.

## Methods

This PR changes the calculation for when an experiment is eligible for tximport. The general formula was `number of samples quantified` divided by `number of samples in the experiment`. After this change the denominator is instead `number of samples in the experiment which are able to be processed` as denoted by having `is_unable_to_be_processed` as false, essentially ignoring samples which had fragment mapping errors.

I also included some common microarray errors that we can't recover. This won't really affect anything, as that field is only checked for rnaseq samples during tximport, but it seemed like a potentially useful field to have populated wherever appropriate.

## Types of changes

## Functional tests

I'm mostly relying on the unit tests and end to end tests, which I'm still running now. I suspect they might fail because they expect the one rna-seq experiment to need a manual tximport run because of its bad sample, but that might not be the case any more.

## Checklist


jaclyn-taroni:
Methods section looks reasonable to me 👍🏻

davidsmejia:
LGTM!

davidsmejia (approved):
👍 Well done

PR #2982, opened 2021-12-16
Handle the cases where a sample doesn't have any jobs.

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/refinebio/pull/2978

My functional tests didn't include cases where there were no jobs for a sample. This happens in staging/prod.

## Types of changes

## Functional tests

I've tested this before, just not without jobs. Once this command runs it can be removed so I don't want to add unit tests.

## Checklist


davidsmejia (approved):
LGTM

Issue #3019, opened 2022-05-05
Improve high-level code style/readability


arkid15r:
### Context

High level overview of code base current state

### Problem or idea

Fix obvious easy to do stuff.

- remove wildcard imports
- created_at/last_updated_at model fields
- str concatenation
- redundant  queryset calls
- class Meta placement
-

PR #3023, opened 2022-05-06
Add Timestamped base model (created_at, last_updated_at auto fields).

arkid15r:
## Issue Number

#3019 
Introduce Timestamped model base with created_at, last_modified_at auto fields


davidsmejia:
I'd much prefer we kept `created_at` and `updated_at` conventions across models.

davidsmejia (changes requested):
Let's update the field name and make a ticket to verify that the new fields don't get auto inserted, I am concerned about there being a timeout, might make sense to add the field and then paginate and insert the values.

PR #3125, opened 2022-10-07
Update AMI terraform config.

### LINKED ISSUE #3111: Update foreman AMI
### Context
The current foreman image is quite out of date with a lot of pending updates 

### Problem or idea
The image has to be updated with the minor updates (note that major version upgrade is also available)

### Solution or next step
Generate a new AMI with the most recent updates.

arkid15r:
## Issue Number

Resolves #3111 

Update terraform config for foreman AMI generation.

## Types of changes


ami/instance.tf
```diff
@@ -26,10 +25,14 @@ resource "aws_instance" "ubuntu-ami-template-instance" {
...
   # instance. For more information see the definition of this resource.
-  user_data = data.template_file.ubuntu_instance_user_data.rendered
+  user_data = templatefile("ubuntu-instance-user-data.tpl.sh", {
+      docker_apt_key = data.local_file.docker_apt_key.content
+    }
+  )
 
-  subnet_id = data.aws_subnet.ccdl_dev_subnet.id
   associate_public_ip_address = true
+  key_name = "data-refinery-key-circleci-prod"
```
davidsmejia, line 34:
```suggestion
  key_name = "data-refinery-key-${var.user}-${var.stage}"
```

davidsmejia: Use templated values so there isn't collision for dev/stage/prod

davidsmejia (approved):
Can we change the target of this pr to `feature/automatic-accessions`

ami/instance.tf
arkid15r: Not sure if these variables are applicable in `ami` context. Keeping as is for now.

PR #3131, opened 2022-10-25
Update `validate_dataset` to allow for experiments with duplicate samples

### LINKED ISSUE #3055: Duplicate Sample Accessions Breaks Dataset State
### Context

@josh was adding some samples to a dataset recently when he noticed on the client side he was no longer able to modify his dataset. After some investigation I noticed that the issue occurred when samples with the same accessions codes are present in multiple experiments. 

```python
  josh = Dataset.objects.get(id='5f3dd9da-e1fd-4ee6-8c07-bd97a83a0dd1')
  for key, value in josh.data.items():
    accessions.extend(value)
    
  len(accessions)
  # 1715

  samples = Sample.public_objects.filter(accession_code__in=accessions)
 
  samples.count()
  # 1709

  accession_set = Set(accessions)
  len(accession_set)
  # 1709

```

```python
for key, value in josh.data.items():
   items = set(value) & set(duplicates))
   if len(items) > 0:
     data[key] = items
 
   print(data)
   #  {
   #    "GSE98279": {
   #      "GSM2590437",
   #      "GSM2590436",
   #      "GSM2590433",
   #      "GSM2590434",
   #      "GSM2590432",
   #      "GSM2590435",
   #    },
   #    "GSE98384": {
   #      "GSM2590437",
   #      "GSM2590436",
   #      "GSM2590433",
   #      "GSM2590434",
   #      "GSM2590432",
   #      "GSM2590435",
   #    },
   # }
```
### Problem or idea

We should update this [line](https://github.com/AlexsLemonade/refinebio/blob/dev/api/data_refinery_api/views/dataset.py#L118) to either coerce the `accessions` list to a set or update the querset that defines `samples` to allow for duplicates.

We should also determine where in the logic that allows users to initially add the samples without failing this validation since the code should run before save. It only seems to be an issue after the duplicate sample accession codes are saved on the dataset instance and then another update is attempted.

### Solution or next step

We should update the code that checks if the dataset is in a valid state.

arkid15r:
## Issue Number

Resolves #3055 

Change some checks to more readable versions.
Use sets instead of lists for entity IDs tracking (accessions, experiments).
Get rid of str + str.

## Types of changes

## Checklist

### DIFF
--- a/api/data_refinery_api/views/dataset.py
+++ b/api/data_refinery_api/views/dataset.py
@@ -54,7 +54,7 @@ def experiment_has_downloadable_samples(experiment, quant_sf_only=False):
             results__computedfile__s3_bucket__isnull=False,
         )
 
-        if samples.count() == 0:
+        if not samples.exists():
             return False
 
     else:
@@ -128,7 +128,7 @@ def validate_dataset(data):
             results__computedfile__s3_key__isnull=False,
             results__computedfile__s3_bucket__isnull=False,
         )
-        if samples_without_quant_sf.count():
+        if samples_without_quant_sf.exists():
             raise InvalidData(
                 message="Sample(s) in dataset are missing quant.sf files. See `details` for a full list",
                 details=[s.accession_code for s in samples_without_quant_sf],
@@ -136,7 +136,7 @@ def validate_dataset(data):
 
     else:
         unprocessed_samples = samples.exclude(is_processed=True)
-        if unprocessed_samples.count():
+        if unprocessed_samples.exists():
             raise InvalidData(
                 message="Non-downloadable sample(s) in dataset. See `details` for a full list",
                 details=[s.accession_code for s in unprocessed_samples],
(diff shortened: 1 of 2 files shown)


api/data_refinery_api/views/dataset.py
```diff
@@ -75,68 +75,68 @@ def validate_dataset(data):
...
-    if data.get("quant_sf_only", False):
+    if data.get("quant_sf_only"):
         samples_without_quant_sf = samples.exclude(
             # Exclude samples that have at least one uploaded quant.sf file associated with them
             results__computedfile__filename="quant.sf",
             results__computedfile__s3_key__isnull=False,
             results__computedfile__s3_bucket__isnull=False,
         )
-        if samples_without_quant_sf.count() > 0:
+        if samples_without_quant_sf.count():
```
davidsmejia, line 131:
```suggestion
        if samples_without_quant_sf.exists():
```

davidsmejia: We should do this for the rest of the `count` checks in this file

davidsmejia (approved):
LGTM - as a general rule we will try to use more semantic code

api/data_refinery_api/views/dataset.py
arkid15r: Done.

PR #3156, opened 2022-11-29
Update common docker images

arkid15r:
## Issue Number

N/A

Common Docker image files clean up. 
  - Pin markupsafe==2.0.1
  - Inherit images from dr_base image.
  - Reorder and clean up the instructions.

## Methods

N/A

## Types of changes

N/A

## Functional tests

N/A

## Checklist

## Screenshots

N/A


davidsmejia (approved):
LGTM 

Though looking at the migration dockerfile... I don't see how this could be true. I also don't think anything bad will happen as it is now so we can leave it alone.
https://github.com/AlexsLemonade/refinebio/pull/3156/files#diff-a363c7b8565f52109cb4d35ced8895ca5bc85fb18326803d09cd7036bdb541b0R15

Issue #3129, opened 2022-10-24
Fix M1 Dev Deploys


davidsmejia:
### Context

Currently it is not possible to perform a dev deploy from a M1 based mac.

### Solution or next step

Determine what the problems are and either document local changes required or file a PR that fixes any related bugs.

PR #3158, opened 2022-12-02
Refactor docker image building scripts.

arkid15r:
## Issue Number

#3129 

Update scripts for better docker cache leveraging, improve readability:
  - Add remote builder support option.
  - Add existing image based cache.
  - Set `amd64` as a default architecture.
  - Add optional `arm64` build support for `update_my_docker_images.sh`.
  - Consolidate api/common/workers images build logic.

## Methods
NA

## Types of changes

NA

## Functional tests
NA

## Checklist

## Screenshots
NA

### DIFF
--- a/scripts/update_my_docker_images.sh
+++ b/scripts/update_my_docker_images.sh
@@ -1,9 +1,6 @@
 #!/bin/sh
 
-# This script is very similar to .circleci/update_docker_images.sh but it has less
-# production/cloud related checks.
-
-# Exit on failure
+# Exit on failure.
 set -e
 
 # This script should always run as if it were being called from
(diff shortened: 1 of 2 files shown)


scripts/update_my_docker_images.sh
```diff
@@ -13,39 +10,48 @@ script_directory="$(perl -e 'use File::Basename;
...
     cat << EOF
 There are two required arguments for this script:
--d specifies the Dockerhub repo you would like to deploy to.
+-d specifies the Dockerhub repository you would like to deploy to.
 -v specifies the version you would like to build. This version will passed into
     the Docker image as the environment variable SYSTEM_VERSION.
     It also will be used as the tag for the Docker images built.
 
-There is also one optional argument:
+There is also optional arguments:
```
davidsmejia, line 29:
```suggestion
There are also optional arguments:
```

scripts/update_my_docker_images.sh
```diff
@@ -13,39 +10,48 @@ script_directory="$(perl -e 'use File::Basename;
...
 -a also build the affymetrix image
     (we normally don't because it is so intense to build)
+-b sets a remote Docker builder for amd64 image builds.
```
davidsmejia, line 32: We should simplify this, default to `adm64` and have optarg `-x` to target `arm64`. Though I also think we probably shouldn't enable this until all images can be built on `arm64`. Would like to hear your thoughts on this.

davidsmejia (changes requested):
Default to amd64 and then have an optional flag to target arm

scripts/update_my_docker_images.sh
davidsmejia: Actually I am unsure what this option does functionally.

arkid15r:
Suppressed by https://github.com/AlexsLemonade/refinebio/pull/3285

PR #2751, opened 2021-06-03
Fix the compendia tests

wvauclain:
## Issue Number

N/A came up while working on revamping the tests.

## Purpose/Implementation Notes

I cherrypicked this commit from my changes so we can get it merged and fix the tests.

Tensorflow updated, and the new version was incompatible with fancyimpute, so I had to freeze tensorflow at 2.4.1.

## Types of changes

What types of changes does your code introduce?

- Bugfix (non-breaking change which fixes an issue)

## Functional tests

All the tests pass now for the compendia.

## Checklist

_Put an `x` in the boxes that apply._

- [x] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

## Screenshots

Please attach any screenshots that illustrate these changes.

PR #3240, opened 2023-03-09
Update compendia dependencies

arkid15r:
## Issue Number

#2751 

Update compendia Docker.
Update image and R dependencies.

## Types of changes

## Checklist


davidsmejia:
Do you think we should extract the compendia specific python requirements into another file and install them via pip as well?

arkid15r:
> Do you think we should extract the compendia specific python requirements into another file and install them via pip as well?

Extracted into a separate file. PTAL.

davidsmejia (approved):
LGTM

Issue #3283, opened 2023-04-24
Fix RNA-Seq end to end tests


arkid15r:
### Context

In order to complete restoring staging deploy process the CI/CD errors need to be fixed.

### Problem or idea

```
Polling SurveyJobs. Currently waiting for job id: 129117
SurveyJob 129117 failed!
F
======================================================================
FAIL: test_all_the_things (tests.foreman.test_end_to_end.FullFlowEndToEndTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python3.8/unittest/case.py", line 60, in testPartExecutor
    yield
  File "/usr/lib/python3.8/unittest/case.py", line 676, in run
    self._callTestMethod(testMethod)
  File "/usr/lib/python3.8/unittest/case.py", line 633, in _callTestMethod
    method()
  File "/home/user/tests/foreman/test_end_to_end.py", line 235, in test_all_the_things
    self.process_experiments()
  File "/home/user/tests/foreman/test_end_to_end.py", line 279, in process_experiments
    self.assertTrue(wait_for_job(survey_job))
  File "/usr/lib/python3.8/unittest/case.py", line 765, in assertTrue
    raise self.failureException(msg)
AssertionError: False is not true

----------------------------------------------------------------------
Ran 2 tests in 3422.129s

FAILED (failures=1)
Error: Process completed with exit code 1.
```

### Solution or next step
Troubleshoot and fix the end2end tests errors.

arkid15r:
After a quick investigation it seems the problem is with an out of date SRA surveyor:

```
2023-04-24 19:34:49,854 i-03ccf74ae760b7e5e data_refinery_foreman.surveyor.external_source ERROR [survey_job: 129117]: Exception caught while discovering samples. Terminating survey job.
Traceback (most recent call last):
  File "/home/user/data_refinery_foreman/surveyor/external_source.py", line 170, in survey
    experiment, samples = self.discover_experiment_and_samples()
  File "/home/user/data_refinery_foreman/surveyor/sra.py", line 661, in discover_experiment_and_samples
    returned_experiment, samples = self._generate_experiment_and_samples(
  File "/home/user/data_refinery_foreman/surveyor/sra.py", line 446, in _generate_experiment_and_samples
    files_urls = [SraSurveyor._build_ncbi_file_url(run_accession)]
  File "/home/user/data_refinery_foreman/surveyor/sra.py", line 346, in _build_ncbi_file_url
[...]

PR #3292, opened 2023-05-17
3283 sra metadata download

davidsmejia:
## Issue Number

#3283 

- Updates out of date vcr cassettes for sra surveyor metadata requests
- Removes url creation for ncbi via ftp and uses s3 location
- since the response changed from ENA I have updated the test values and removed keys that are no longer returned
  - this is honestly kind of weird, the dates for first published and last update are now further in the past...
  - some attributes (spot count, base count) are no longer on sample and only on run

ENA changes information here: https://docs.google.com/document/d/1RPHmK8Pvm9UxSa21Ej3MkGoGYO9baSxwxk_dOuWWyNE

## Methods

There aren't any methods, that I can tell, that should be called out here.
There isn't much we can do about the `sample_ena_base_count` and `sample_ena_spot_count` being removed. For newly surveyed samples we will only be able to capture what is presented to us. The test validates that the method in which we capture the data works, and this test still captures that truthiness.

## Types of changes

## Functional tests

- n/a
- sra surveyor tests are passing
- sra downloader tests are passing

## Checklist

## Screenshots

n/a

### DIFF
--- a/foreman/data_refinery_foreman/surveyor/sra.py
+++ b/foreman/data_refinery_foreman/surveyor/sra.py
@@ -1,4 +1,3 @@
-import random
 import re
 import xml.etree.ElementTree as ET
 from typing import Dict, List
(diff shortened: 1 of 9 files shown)


arkid15r (approved):
LGTM

PR #3295, opened 2023-05-19
codecov informational

davidsmejia:
## Issue Number

- N/A

- Makes code coverage reports non-blocking
- Ignores tests folders for coverage reports
- Also ignores shell scripts though we may want to revisit this in the future / make them python

## Methods

- n/a

## Types of changes

n/a

## Functional tests

n/a ran on a test PR with bogus changes
ran `cat codecov.yml | curl --data-binary @- https://codecov.io/validate` to validate the config.

## Checklist

## Screenshots

n/a


codecov.yml
```diff
@@ -1,2 +1,15 @@
 github_checks:
     annotations: false
+
+ignore:
```
arkid15r, line 4: Would it make sense to add `manage.py`, `setup.py`, and `wsgi.py` as well?

arkid15r (approved):
LGTM

codecov.yml
davidsmejia: Ok I will remove this and we can omit it in `setup.cfg` instead.

PR #3299, opened 2023-06-13
Implement Docker image cache cross-environment support

### LINKED ISSUE #3288: Cache Docker Images by Branch
### Context

We want to be able to cache docker image layers that are created locally as testing artfacts locally to be used by github actions.
The current prepare_images.sh does this but there was an issue with the definition for branch_name.
We also don't want to remove support non-ccdl members developing locally.

image

### Solution or next step

- After #3285 is merged, we should set sensible defaults that can be overridden for external contributors.
- Get current branch name or tag to be set when pushing images to ccdl(staging) repo.

Determine:
- If they don't have access to the docker repo should we just build locally and not push?
- How long can docker tags be / are they compatible with our longer branch names.

arkid15r:
## Issue Number

Closes #3288 

This PR consolidate the recent efforts towards refine.bio Docker images build process optimization. 

Every image name is now based on the branch hash value and cache layers are uploaded into a registry for further sharing purpose. This allowed retiring `Docker.affymetrix_local` image and related `prepare_image.sh` logic.

The API image has been split into base and local/production images. 
The affymetrix/agilent tests are 2 separate jobs now and use larger runners (-m suffix).

These changes resulted in >50% (~1 hour vs ~30 minutes) tests execution time improvement: [run a](https://github.com/AlexsLemonade/refinebio/actions/runs/5027859460) or [run b](https://github.com/AlexsLemonade/refinebio/actions/runs/5027486100) VS [this PR run](https://github.com/AlexsLemonade/refinebio/actions/runs/5257956290).

Some points to keep in mind during review:
  - Name Docker images based on the branch hash value (shasum)
  - Add setup.py PEP-440 workaround for common package
  - Reorganize GHA jobs
  - Utilize Docker registry cache for image cache layers sharing
  - Use separate Docker images (_cache) for caching
  - Optimize Docker images size (foreman, salmon)
  - Add remote builder support (used by default in GHA jobs)
  - Set umask explicitly to avoid cache miss due to permissions diff
  - Introduce API base image, derive api local/production from it
  - Reformat affected files for better readability
  - Use better names for scripts/variables
  - Consolidate renv related scripts location (use common/ everywhere)
  - Retire -d option of `prepare_image.sh`
  - Delete affymetrix_local image as no longer used (affymetrix cache boost satisfies our tests execution time requirements)
  - Sort .dockerignore entries
  - Update .pre-commit-config.yaml
  - Update README

`pre-commit-terraform` has been disabled temporarily due to unrelated failures.

The PR contains all changes required for further staging/production deploy process optimization identified at this moment.

## Checklist


.github/workflows/config.yml
```diff
@@ -1,513 +1,552 @@
-name: test-and-deploy
+name: refine.bio CI/CD
+
+env:
```
davidsmejia, line 3: This is minor but I think the order of `env` and `on` should be switched for clarity.

scripts/update_docker_images.sh
```diff
@@ -71,91 +65,72 @@ while getopts ":r:v:abx:h" opt; do
...
+    set_up_docker_builder
 
-    DOCKER_BUILDKIT=1 docker buildx build \
-        --build-arg BUILDKIT_INLINE_CACHE=1 \
+    docker buildx build \
         --build-arg DOCKERHUB_REPO="$DOCKERHUB_REPO" \
         --build-arg SYSTEM_VERSION="$SYSTEM_VERSION" \
-        --cache-from "$IMAGE_NAME:$SYSTEM_VERSION" \
-        --cache-from "$IMAGE_NAME:latest" \
+        --"$CACHE_FROM_LATEST" \
```
davidsmejia, line 126: Do you think this looks more consistent with how we set the other values?
```suggestion
        cache-from="$CACHE_FROM_LATEST" \
```

davidsmejia (approved):
LGTM just some minor suggestions you can change at your discretion.

.github/workflows/config.yml
arkid15r: Done.

scripts/update_docker_images.sh
arkid15r: That's right, initially `cache-from/cache-to` options had more extras around them. It's no longer needed. I updated `scripts/update_docker_images.sh` too.

Issue #3320, opened 2023-06-28
Harmonizer should support new ENA surveyor changes


davidsmejia:
### Context

When updating the SRA surveyor we had to do a science review. This ticket is for tracking some of the work that went into creating the changes for the harmonizer piece of that work.

### Problem or idea

The harmonizer behavior will need to be modified in order to support the upcoming changes to the SRA surveyor. Additionally there is some behavior that makes it diverge from what we outline in the documentation. For that we will prioritize updating the harmonizer to conform to the documentation, as a fallback we should update the documentation for anything we can no longer support.

### Solution or next step

Features that should be added to support new changes
- Fix genetic information parsing
- Support multiple values
- support new flattened metadata dict (flattening implementation will be in another pr)
- support Regex key values

PR #3321, opened 2023-06-28
Support new ENA surveyor changes in Harmonizer

davidsmejia:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/3320

TODO: explain changes below, at least update the test to use an example with multiple values

- genotype / genetic informaton
- regular expressions
- multiple values match
- 

## Methods
- 

## Types of changes

## Functional tests

Unit Tests

## Checklist

## Screenshots

N/A

### DIFF
--- a/foreman/data_refinery_foreman/surveyor/harmony.py
+++ b/foreman/data_refinery_foreman/surveyor/harmony.py
@@ -5,7 +5,7 @@ Given a list of samples and their metadata, extract these common properties:
   `sex`,
   `age`,
   `specimen_part`,
-  `genetic_information`,
+  `genotype`,
   `disease`,
   `disease_stage`,
   `cell_line`,
@@ -245,6 +245,7 @@ GEO:
 """
 
 import csv
+import re
 from io import StringIO
 from typing import Dict, List
 
@@ -252,6 +253,9 @@ from data_refinery_common.logging import get_and_configure_logger
 from data_refinery_foreman.surveyor.utils import requests_retry_session
 
 logger = get_and_configure_logger(__name__)
+
+# re.Pattern / re._pattern_type were not defined in 3.8?
+RE_TYPE = re.compile("").__class__
 
 
 TITLE_FIELDS = [
@@ -292,7 +296,7 @@ def create_variants(fields_list: List):
 
     Ex, given 'cell line', add the ability to hit on 'characteristic [cell_line]' as well.
     """
-    variants = []
+    variants = set()
 
     for field in fields_list:
         space_variants = [field]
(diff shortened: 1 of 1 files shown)


foreman/data_refinery_foreman/surveyor/harmony.py
```diff
@@ -253,6 +424,9 @@
 
 logger = get_and_configure_logger(__name__)
 
+# re.Pattern / re._pattern_type were not defined in 3.8?
+RE_TYPE = re.compile("").__class__
+
```
arkid15r, line 429: It seems that at least `re.Pattern` was.

foreman/data_refinery_foreman/surveyor/harmony.py
```diff
@@ -303,24 +477,27 @@ def create_variants(fields_list: List):
...
-            variants.append("factor value[" + space_variant + "]")
-            variants.append("factorvalue [" + space_variant + "]")
-            variants.append("factor value [" + space_variant + "]")
-            variants.append("sample_" + space_variant)
-            variants.append("sample_host" + space_variant)
-            variants.append("sample_sample_" + space_variant)  # Yes, seriously.
-
-    return variants
+            variants.add(space_variant)
+            variants.add("characteristic [" + space_variant + "]")
```
arkid15r, line 481: It should be less expensive with
```suggestion
            variants.add(f"characteristic [{space_variant}]")
```

or join (here and below).

foreman/data_refinery_foreman/surveyor/harmony.py
```diff
@@ -585,25 +761,42 @@ def harmonize_value(self, field_name: str, value):
...
 
-    def harmonize_field(self, sample_metadata: Dict, harmonized_sample: Dict, field: str):
-        field_name = field.split("_fields")[0]
+    def harmonize_field(
+        self, sample_metadata: Dict, harmonized_sample: Dict, variants_attribute: str
+    ):
+        sample_attribute = variants_attribute.split("_fields")[0]
+        harmonized_values = set()
+        key_variants = getattr(self, variants_attribute)
+        key_regexes = [r for r in key_variants if isinstance(r, RE_TYPE)]
```
arkid15r, line 770: Did you try `re.Pattern` here?

foreman/data_refinery_foreman/surveyor/harmony.py
```diff
@@ -585,25 +761,42 @@ def harmonize_value(self, field_name: str, value):
...
+        key_strings = [s for s in key_variants if isinstance(s, str)]
 
         for key, value in sample_metadata.items():
             lower_key = key.lower().strip()
+            if lower_key in key_strings:
```
arkid15r, line 775: Use a `set` for `key_strings` (or its copy) for ~O(1) `in` operation.

arkid15r (commented):
A couple of performance related suggestions:

foreman/data_refinery_foreman/surveyor/harmony.py
davidsmejia: Could you elaborate?

foreman/data_refinery_foreman/surveyor/harmony.py
davidsmejia: I'm going to leave this as is and will revisit when I do array express.

foreman/data_refinery_foreman/surveyor/harmony.py
davidsmejia: Ah, well when I had re.Pattern in the tests it failed as undefined so for now I will leave it as is. I am open to creating an issue to tackle this later on.

foreman/data_refinery_foreman/surveyor/harmony.py
davidsmejia: Yeah it is in the docs, when I run the tests it is undefined. I am going to leave this as is for now and when the functionality is proven then we can come back and revisit this. I think it might be that our python version in the tests is different or something.

foreman/data_refinery_foreman/surveyor/harmony.py
```diff
@@ -585,25 +761,42 @@ def harmonize_value(self, field_name: str, value):
...
-    def harmonize_field(self, sample_metadata: Dict, harmonized_sample: Dict, field: str):
-        field_name = field.split("_fields")[0]
+    def harmonize_field(
+        self, sample_metadata: Dict, harmonized_sample: Dict, variants_attribute: str
+    ):
+        sample_attribute = variants_attribute.split("_fields")[0]
+        harmonized_values = set()
+        key_variants = getattr(self, variants_attribute)
+        key_regexes = [r for r in key_variants if isinstance(r, RE_TYPE)]
+        key_strings = set([s for s in key_variants if isinstance(s, str)])
```
arkid15r, line 771:
```suggestion
        key_strings = set((s for s in key_variants if isinstance(s, str)))
```

or 

`key_strings = {s for s in key_variants if isinstance(s, str)}`

arkid15r (approved):

jashapiro (commented):
Everything here looks as I would expect it.

PR #3281, opened 2023-04-24
Add newer and additional RNA-seq supported platforms

jaclyn-taroni:
## Issue Number

Closes #3280 

## Purpose/Implementation Notes

Here I'm adding newer (and some older) Illumina instrument models to our list of supported RNA-seq platforms.

## Methods

I am requesting both David and Josh to review due to these methods.

My approach was to snag a TSV of human or mouse RNA-seq data with raw reads that were generated on an Illumina platform from European Nucleotide Archive:

```sh
curl -X POST -H "Content-Type: application/x-www-form-urlencoded" -d 'result=read_run&query=(tax_eq(9606)%20OR%20tax_eq(10090))%20AND%20instrument_platform%3D%22ILLUMINA%22%20AND%20library_strategy%3D%22RNA-Seq%22&fields=study_accession%2Cinstrument_model%2Ctax_id&format=tsv' "https://www.ebi.ac.uk/ena/portal/api/search"
```

Then over to R:

```r
# Currently supported platforms
refinebio_supported_df <- read_csv("https://raw.githubusercontent.com/AlexsLemonade/refinebio/ac3170a8cfd14a754282334204c048182963dac9/config/supported_rnaseq_platforms.txt", col_names = FALSE)

# RNA-seq mouse or human data generated on Illumina platform
instrument_models_df <- read_tsv("Downloads/results_read_run_tsv.txt")

# Get a list of unique instrument models for RNA-seq of mouse and human
unique_illumina_models <- instrument_models_df %>% pull(instrument_model) %>% unique()
```

Then we can look at what platforms are in the public data but not yet in our supported list of platforms with:

```r
setdiff(unique_illumina_models, refinebio_supported_df[[1]])
 [1] "Illumina MiSeq"        "HiSeq X Ten"           "Illumina MiniSeq"     
 [4] "Illumina NovaSeq 6000" "NextSeq 1000"          "Illumina NovaSeq X"   
 [7] "Illumina iSeq 100"     "NextSeq 2000"          "unspecified"          
[10] "HiSeq X Five"          "Illumina HiSeq X"  
```

I included the platforms that are currently positioned for transcriptome sequencing (by Illumina): Illumina NovaSeq 6000, NextSeq 1000, Illumina NovaSeq X, NextSeq 2000. 

Somewhat obviously, I will not be including “unspecified.” There are also a few benchtop sequencers that I didn’t include at this point (MiSeq, iSeq 100, MiniSeq). 

[...]

PR #3326, opened 2023-07-14
ENA Surveyor Metadata Changes

davidsmejia:
## Issue Number

#3320

This is based on #3321 and includes those changes, they should be merged in first.

The focus of this PR is to support migrating from the XML API to JSON.

- Changes how we pull in metadata from ENA to use the new JSON API
- Completely reworks the Surveyor for SRA to handle either a sample or an experiment accession code.
- Adds recursive dict flattening utility function for handling deeply nested JSON responses.
- Adds tests with updated values from the API because some of them have changed since the migration.

## Methods

Science: 
- Most of the attributes look very similar and are mapped according to common sense. I think it would be nice to have a one over specifically in `_apply_metadata_to_experiment` in `surveyor/sra.py`.

## Types of changes

## Functional tests

n/a local tests

## Checklist

## Screenshots

n/a


foreman/data_refinery_foreman/surveyor/sra.py
```diff
@@ -56,275 +54,137 @@ def source_type(self):
...
-        formatted_metadata_URL = ENA_METADATA_URL_TEMPLATE.format(metadata["submission_accession"])
-        response = utils.requests_retry_session().get(formatted_metadata_URL)
-        submission_xml = ET.fromstring(response.text)[0]
-        submission_metadata = submission_xml.attrib
-
-        # We already have these
-        submission_metadata.pop("accession", "")
-        submission_metadata.pop("alias", "")
+    def get_ena_json_api(endpoint: str, accession, params: dict = {}):
+        query_params = "&".join([f"{k}={params[k]}" for k in params])
```
arkid15r, line 58: I believe you want `urllib.parse.urlencode` here.

foreman/data_refinery_foreman/surveyor/utils.py
```diff
@@ -43,6 +44,29 @@ def flatten(d, parent_key="", sep="_"):
...
+# TODO: Merge this with above method after updating array express
+#       to support recursive flattening.
+def flatten_dict(data: Union[Dict, List], prefix: str = "", sep="_") -> dict:
+    """Flattens a nested dictionary or list"""
+    flattened = {}
+    iterator = enumerate(data) if isinstance(data, list) else data.items()
+
+    for k, v in iterator:
+        key = (k if not prefix else f"{prefix}{sep}{k}").replace(" ", sep)
+        if isinstance(v, dict) or isinstance(v, list):
```
arkid15r, line 56:
```suggestion
        if isinstance(v, (dict, list)):
```

foreman/data_refinery_foreman/surveyor/utils.py
```diff
@@ -43,6 +44,29 @@ def flatten(d, parent_key="", sep="_"):
...
+        else:
+            flattened[key] = v
+
+    return flattened
+
+
+def get_nonempty(hash: dict, key: str, fallback):
+    if key in hash and hash[key]:
+        return hash[key]
+    return fallback
```
arkid15r, line 67:
```suggestion
    return hash.get(key) or fallback
```

arkid15r (approved):
Nicely done!

foreman/data_refinery_foreman/surveyor/sra.py
```diff
@@ -345,220 +205,186 @@ def _apply_metadata_to_experiment(experiment: Experiment, metadata: dict):
...
-            ):
-                sample_object.manufacturer = "ILLUMINA"
-            elif "ION TORRENT" in sample_object.platform_name.upper():
-                sample_object.manufacturer = "ION_TORRENT"
+            sample.platform_accession_code = sample.platform_name.replace(" ", "")
+            sample.technology = "RNA-SEQ"
+
+            platform_name = sample.platform_name.upper()
+            if platform_name in ["ILLUMINA", "NEXTSEQ"]:
+                sample.manufacturer = "ILLUMINA"
```
jashapiro, line 350: Noting the reversal of logic here: I was assuming the previous configuration (`"ILLUMINA" in sample_object.platform_name.upper()`)was being used as a simple regex to catch things like `ILLUMINA MISEQ` which would be missed by the changes here.

jashapiro: Related: do we want to add the other Illumina platforms we are planning to support here (i.e. MiSeq)? I'm not sure how well this will catch all of them: I would have expected `ILLUMINA` to be sufficient, but the inclusion of `NEXTSEQ` says otherwise!

jashapiro (commented):
I think this is generally doing what we want, but I had some questions, notably about the change in logic around checking platforms, which has gotten more "exact" with the current code, which may not be what we want.

foreman/data_refinery_foreman/surveyor/sra.py
jaclyn-taroni: @jashapiro - I expect this list is guiding the logic here: https://github.com/AlexsLemonade/refinebio/blob/7d87133d820973692fa6465ff09ede9395422521/config/supported_rnaseq_platforms.txt 

Illumina MiSeq is listed as `Illumina MiSeq`, whereas the NextSeq platforms are listed as `NextSeq <number>`. That list was compiled using the platforms returned by ENA (see: https://github.com/AlexsLemonade/refinebio/pull/3281 for methods).

jashapiro: Based on that list we will also need to add `HISEQ` and reverting to the previous structure (`"ILLUMINA" in sample.platform_name.upper()`, etc.) or a version that does preserve partial matches will be required.

foreman/data_refinery_foreman/surveyor/utils.py
```diff
@@ -43,6 +44,29 @@ def flatten(d, parent_key="", sep="_"):
...
 
+# TODO: Merge this with above method after updating array express
+#       to support recursive flattening.
+def flatten_dict(data: Union[Dict, List], prefix: str = "", sep="_") -> dict:
+    """Flattens a nested dictionary or list"""
+    flattened = {}
+    iterator = enumerate(data) if isinstance(data, list) else data.items()
+
+    for k, v in iterator:
+        key = (k if not prefix else f"{prefix}{sep}{k}").replace(" ", sep)
```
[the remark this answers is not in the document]
davidsmejia: Yeah thats fair, I think this is safe because of how we are using it. I don't think we actually need the separator argument but I kept it because I wanted more overlap of functionality when we go to merge it with the other non-recursive flatting utility method. The collision here wouldn't really matter with how the harmonizer works as it checks spaces, dashes, and underscores regardless. It is harder to work out which spaces were nested keys, but in all honesty we don't really care we just need keys to be unique. When I implement array express and then merge these functions I will probably remove it in favor of a more opinionated approach.

foreman/data_refinery_foreman/surveyor/sra.py
```diff
@@ -56,275 +54,137 @@ def source_type(self):
...
+            experiment_accession = accession
+
+        experiment_metadata = SraSurveyor.gather_experiment_metadata(experiment_accession)
+
+        if "accession" in experiment_metadata:
+            pubmed_metadata = SraSurveyor.gather_experiment_pubmed_metadata(
+                experiment_metadata["accession"]
+            )
+            if len(pubmed_metadata) > 0:
+                experiment_metadata["pubmed_id"] = pubmed_metadata["Source Primary Accession"]
```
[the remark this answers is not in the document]
davidsmejia: This is a limitation of our current db structure, which I would like to avoid changing at this time. I do think you are right that we would want to handle more than one.

foreman/data_refinery_foreman/surveyor/sra.py
```diff
@@ -56,275 +54,137 @@ def source_type(self):
...
-        return metadata
+        if response.status_code != 200:
+            logger.error(
+                "Could not discover Pubmed ID.",
+                accession=accession,
+                status_code=response.status_code,
+                url=response.url,
+            )
+            return []
+        return response.json()
```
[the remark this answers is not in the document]
davidsmejia: Yeah this is always an unpaginated list in the response, and `[]` when there are no matches.

foreman/data_refinery_foreman/surveyor/sra.py
davidsmejia: I looked through RNA-Seq samples in the database, that we have previously processed and the only manufacturers that are NEXTSEQ are for GEO accessions. I am unsure if that is a problem. I think that check was unnecessary as we should only be surveying geo accessions in the geo surveyor. All of our SRA accessions have unknown, illumina, or ion_torrent.

davidsmejia: >>> sra = Sample.objects.filter(source_database="SRA")
>>> sra.distinct('manufacturer').values_list('manufacturer', flat=True)
<QuerySet ['ILLUMINA', 'ION_TORRENT', 'UNKNOWN']>

foreman/data_refinery_foreman/surveyor/sra.py
```diff
@@ -345,220 +205,184 @@ def _apply_metadata_to_experiment(experiment: Experiment, metadata: dict):
...
+            sample.platform_name = sample_metadata.get("instrument_model", "UNKNOWN")
+            sample.platform_accession_code = sample.platform_name.replace(" ", "")
+            sample.technology = "RNA-SEQ"
 
+            # In the rare case that we get something we don't expect
+            # we don't want to process this sample so we can set it to UNKNOWN.
+            if sample.manufacturer not in ["ILLUMINA", "ION_TORRENT"]:
+                sample.manufacturer = "UNKNOWN"
+                sample.platform_name = "UNKNOWN"
+                sample.platform_accession_code = "UNKNOWN"
```
davidsmejia, line 351: @jashapiro the value we want for manufacturer matches exactly what we are getting in `instrument_platform` ie `ILLUMINA` or `ION_TORRENT`. `platform_model` will be the whole name so `Illumina HiSeq 2000` etc. I added back in a check to prevent processing if we do get a weird value back. I checked NextSeq 500 which will return ILLUMINA for platform_name.

Issue #3333, opened 2023-07-21
e2e test throws "value must be a decimal number" error


arkid15r:
### Context

As a part of post deployment process we run end to end tests. 

### Problem or idea
The most recent run failed w/ the following error:

```
SurveyJob 129220 failed!
F
======================================================================
FAIL: test_all_the_things (tests.foreman.test_end_to_end.FullFlowEndToEndTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python3.8/unittest/case.py", line 60, in testPartExecutor
    yield
  File "/usr/lib/python3.8/unittest/case.py", line 676, in run
    self._callTestMethod(testMethod)
  File "/usr/lib/python3.8/unittest/case.py", line 633, in _callTestMethod
    method()
  File "/home/user/tests/foreman/test_end_to_end.py", line 235, in test_all_the_things
    self.process_experiments()
  File "/home/user/tests/foreman/test_end_to_end.py", line 279, in process_experiments
    self.assertTrue(wait_for_job(survey_job))
  File "/usr/lib/python3.8/unittest/case.py", line 765, in assertTrue
    raise self.failureException(msg)
AssertionError: False is not true

----------------------------------------------------------------------
Ran 2 tests in 921.220s
```
```
Traceback (most recent call last):
--
File "/home/user/data_refinery_foreman/surveyor/geo.py", line 275, in create_experiment_and_samples_from_api
sample_object = Sample.objects.get(accession_code=sample_accession_code)
File "/usr/local/lib/python3.8/dist-packages/django/db/models/manager.py", line 85, in manager_method
return getattr(self.get_queryset(), name)(*args, **kwargs)
File "/usr/local/lib/python3.8/dist-packages/django/db/models/query.py", line 435, in get
raise self.model.DoesNotExist(
data_refinery_common.models.sample.Sample.DoesNotExist: Sample matching query does not exist.
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
File "/usr/local/lib/python3.8/dist-packages/django/db/models/fields/__init__.py", line 1552, in to_python
return decimal.Decimal(value)
decimal.InvalidOperation: [<class 'decimal.ConversionSyntax'>]
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
File "/home/user/data_refinery_foreman/surveyor/external_source.py", line 170, in survey
experiment, samples = self.discover_experiment_and_samples()
[...]

PR #3334, opened 2023-07-24
prevent age from being set as string, adds test

davidsmejia:
## Issue Number

#3333 

There was a flaw in how we assigned hamonized attributes. For every attribute but `age` we can set an empty string, but age must be omitted because you cannot coerce an empty string into a decimal type. This PR also adds a test for the experiment in the e2e tests that was throwing the error  so we can catch similar issues earlier. 

## Types of changes

## Functional tests

Surveyed the failing experiment in the e2e tests.

## Checklist

## Screenshots
n/a


arkid15r (approved):
LGTM

PR #3362, opened 2023-08-25
TXImport Eligibility - Check if SHORT and LONG indices exist

davidsmejia:
## Issue Number

N/A

When determining if `TXImport` can be run on a particular experiment either as part of the `Salmon` job or run early via the management command `run_tximport` we verify that the latest trascriptome index is what that salmon job was run with.

The problem encountered prior to this PR is that if the order of transcriptome indices is different, ie `SHORT` was created after `LONG`, this function would return `SHORT` even though we are interested in checking if the latest `LONG` index matches the index used to process the quant file. 

My understanding is that this is to prevent different versions of salmon from being used throughout processing, which this change would still preserve. 

This PR makes the following changes:

- Rename `get_latest_organism_index` to `get_latest_organism_indices`.
- Renamed method now tries to return a list of the latest short and long index for the organism.
- Additionally, instead of checking equality when confirming the index exists, we now check if the index is contained in the list of the latest for each type.

## Methods

n/a or see implementation notes

## Types of changes

## Functional tests

tested on staging

## Checklist

## Screenshots

n/a

### DIFF
--- a/common/data_refinery_common/rna_seq.py
+++ b/common/data_refinery_common/rna_seq.py
@@ -28,13 +28,16 @@ EARLY_TXIMPORT_MIN_SIZE = 25
 EARLY_TXIMPORT_MIN_PERCENT = 0.80
 
 
-def get_latest_organism_index(organism):
+def get_latest_organism_indices(organism):
     # Salmon version gets saved as what salmon outputs, which includes this prefix.
     current_salmon_version = "salmon " + get_env_variable("SALMON_VERSION", "0.13.1")
+    indices = OrganismIndex.objects.filter(
+        salmon_version=current_salmon_version, organism=organism
+    ).order_by("-created_at")
+
     return (
-        OrganismIndex.objects.filter(salmon_version=current_salmon_version, organism=organism)
-        .order_by("-created_at")
-        .first()
+        indices.filter(index_type="TRANSCRIPTOME_LONG").first(),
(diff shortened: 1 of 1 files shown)


common/data_refinery_common/rna_seq.py
```diff
@@ -28,13 +28,16 @@
 EARLY_TXIMPORT_MIN_PERCENT = 0.80
 
 
-def get_latest_organism_index(organism):
+def get_latest_organism_indices(organism):
     # Salmon version gets saved as what salmon outputs, which includes this prefix.
     current_salmon_version = "salmon " + get_env_variable("SALMON_VERSION", "0.13.1")
+    indices = OrganismIndex.objects.filter(
+        salmon_version=current_salmon_version, organism=organism
+    ).order_by("-created_at")
```
arkid15r, line 36: Not sure whether the `order_by()` needs to be (re-)moved.

jaclyn-taroni:
Purpose/implementation section seems reasonable 👍🏻  -- I agree with this assessment:

> My understanding is that this is to prevent different versions of salmon from being used throughout processing, which this change would still preserve.

arkid15r (approved):
LGTM

common/data_refinery_common/rna_seq.py
davidsmejia: I just tested that the order_by is preserved on staging. We don't need to move it to after the filter.

PR #3388, opened 2023-09-15
Update common/foreman scripts

arkid15r:

Update common/foreman scripts:
  - Export SYSTEM_VERSION when unset
  - Set platform for `run_management_command.sh`

## Types of changes

## Checklist


davidsmejia (approved):
We should test running the `run_management_command` from the foreman instance on staging after this change to see how it is affected.

Issue #3390, opened 2023-09-19
Move github secrets to 1Password


davidsmejia:
### Context

We are now keeping the project secrets in 1Password. This issue is to track adding in that functionality and addressing side effects of the transition.

### Solution or next step

- Update actions to include loading secrets from 1Password
- Use a terraform variable for SSH public key in `infrastructure/`
- Add [environment](https://docs.sentry.io/platforms/python/legacy-sdk/advanced/) to sentry sdk initializations to [RAVEN_CONFIG](https://github.com/search?q=repo%3AAlexsLemonade%2Frefinebio+RAVEN_CONFIG&type=code)
- Update scripts to reference updated variable names (the OP_ name but without the OP for consistency across projects)

arkid15r:
Resolved in https://github.com/AlexsLemonade/refinebio/pull/3391

PR #3391, opened 2023-09-22
Use 1Password as a secrets storage

arkid15r:
## Issue Number

https://github.com/AlexsLemonade/refinebio/issues/3390

Use 1Password as a secrets storage:
  - Unify secret names
  - Move SSH_PUBLIC_KEY to vars
  - Set GH env secrets in necessary scopes only
  - Add environment and release version into Sentry context

The SSH public key migration to vars required some extra `IFS` manipulations in `deploy.sh`

## Types of changes

## Checklist


davidsmejia (changes requested):
This is looking good to me.

We do need a bit more specificity in the environment value for sentry. Lets add another static setting in each of the projects with the name `app_name`. Then concat that with environment before sending the configuration. So we get something like `prod-refinebio-api`. It would be nice to break up the workers into their own environment scope but for now we can group all the workers together.

Edit: updated the example config value to match what we do elsewhere

arkid15r (commented):
I updated the PR based on the new requirements. PTAL.

davidsmejia (approved):
LGTM

PR #3415, opened 2023-10-12
Rename Sample::genotype to Sample::genetic_information

### LINKED ISSUE #3408: Change Sample property genotype to genetic_information
### Context

We currently use `genotype` as the attribute on the `Sample` model. This should have been changed to the more appropriately named `genetic_information`. 

https://github.com/AlexsLemonade/refinebio-admin/pull/116#issuecomment-1605154840
> The attribute on the sample is genotype, but the way that the harmonizer was written it assigns the value to sample.gentetic_information which does match the docs. We should either change the docs to say genotype instead and update the harmonizer (which I will do for now) or make a migration on the DB to change the name of the column to genetic_information.

### Problem or idea

Updating this should be pretty straightforward. 

One caveat is that the harmonizer. For the accompanying PR for this issue lets not rewrite the harmonizer and instead rename the `genotype_fields`, `self.genotype_fields`, and the string that is passed in here: https://github.com/AlexsLemonade/refinebio/blob/dev/foreman/data_refinery_foreman/surveyor/harmony.py#L800 to matching versions of `genetic_information`.

### Solution or next step

- [ ] Update documentation before deploy
- [ ] Create migration to change Sample property `genotype` to `genetic_information`
- [ ] Update harmonizer attribute from `genotype_fields` to `genetic_information_fields`
- [ ] This probably warrants a new test to ensure the `Sample` attribute names and output of `to_metadata_dict` match what we describe in the docs.

arkid15r:
## Issue Number

Resolves #3408 

Rename sample field to better reflect its contents.
Fix `make_migrations.sh` to run it in a fresh branch.

The downloaders/no_op tests job is failing due to the remote server rate limiting (sort of) with `421 There are too many connected users, please try later` error. It needs to addressed separately.
## Types of changes

## Checklist


davidsmejia (approved):
LGTM

Issue #3436, opened 2023-10-23
Test no-op job is failing: Illumina file won't process


arkid15r:
### Context

After [separating downloaders and no-op test jobs](https://github.com/AlexsLemonade/refinebio/pull/3431) the latter one [keeps failing](https://github.com/AlexsLemonade/refinebio/actions/runs/6615577970/job/17968120770) 

### Problem or idea

The test no-op job is failing with `We found an Illumina file to NO_OP that we're not set up to process yet` error message. 

### Solution or next step

Find out how to set up the file processing and implement it.

arkid15r:
Presumably related e2e test job (circleci_staging_NO_OP_2048_290063) error:

```
2023-10-24 19:52:00,243 i-0fb9189c45a799243 data_refinery_workers.processors.no_op ERROR [job_id: 290063] [platform: yeast2]: Missing gene index file for platform!
--
2023-10-24 19:52:00,526 i-0fb9189c45a799243 data_refinery_workers.processors.utils ERROR [failure_reason: Missing gene index for yeast2] [processor_job: 290063]: Processor function _convert_genes failed. Terminating pipeline.


```

PR #3444, opened 2023-10-27
Follow redirect when downloading gene_indexes

davidsmejia:
## Issue Number

#3436 

NOTE: This PR encountered an error with isort that I could not reproduce locally. I saw that the repo we were using has recently changed so I am now targeting the branch that contains that fix. https://github.com/AlexsLemonade/refinebio/tree/davidsmejia/isort-upgrade-5.11.5

Running `docker run -it ccdlstaging/dr_no_op ls -l /home/user/gene_indexes`. Would give the correct output but rebuilding the container failed.

While testing out the code locally I noticed that curling the `ID_REFINERY_URL` was not downloading the indexes and the subsequent unzip command exit code 9.

This change tells curl to follow the redirect. Which then allows the no_op to build correctly.

This change invalidates the cache layer and makes the files available that were missing during the no_op tests.

## Methods

n/a

## Types of changes

## Functional tests

## Checklist

## Screenshots

n/a

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,10 +15,11 @@ repos:
       # - id: check-docstring-first # We can enable this rule once models.py is simplified
       - id: check-yaml
 
-  - repo: https://github.com/PyCQA/isort
-    rev: 4.3.21
+  - repo: https://github.com/pycqa/isort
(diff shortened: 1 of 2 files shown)


davidsmejia:
closing in favor of #3446

Issue #3452, opened 2023-11-03
Move logs and static files from /tmp to /var


davidsmejia:
### Context

Currently, when we launch the API we write Nginx access and error logs to `/tmp` as well as serving generated static files. This seem to be an issue after deploy. Trying to tail logs from `/tmp/access.log` does nothing which indicates that the container cannot write to that file. Similarly, static files are unable to be served from `/tmp`.

### Problem or idea

We should move nginx logs and static content to their appropriate location in `/var`.
Nginx logs should be located at `/var/log/nginx/access.log` and `/var/log/nginx/error.log`
Static content should be served in nginx from `/var/www/volumes_static`

### Solution or next step

- update nginx.conf to write logs to `/var/log/nginx``
- update nginx.conf to serve static files from `/var/www/volumes_static`
- update user data script template to rotate logs from `/var/log/nginx`
- update user data script to map static directory to `/var/www/volumes_static`

PR #3455, opened 2023-11-07
3452 Ensure /var/www/volumes_static exists on container restart

davidsmejia:
## Issue Number

#3452 

This PR adds the same creation and permission settings to the static dir when the API container is restart. We have already moved the static dir out of `tmp`. This change could safely be removed in the future.

Thoughts for the future: We probably should consider pulling out the starting of the container to it's own script so that we can just call that remotely on the API instead of duplicating the setup in user data and deploy scripts. That change is more involved than what we need at this moment. If we encounter issues with the API standing up or docs not being available again I strongly recommend addressing that then.

## Methods

n/a

## Types of changes

## Functional tests

n/a

## Checklist

## Screenshots

n/a

### DIFF
--- a/infrastructure/deploy.sh
+++ b/infrastructure/deploy.sh
@@ -388,6 +388,19 @@ if [[ -n $container_running ]]; then
         -i data-refinery-key.pem \
         api-configuration/environment "ubuntu@$API_IP_ADDRESS:/home/ubuntu/environment"
 
+    # Ensure the API's static file dir exists and is accessible"
+    ssh -o StrictHostKeyChecking=no \
+        -o ServerAliveInterval=15 \
+        -o ConnectTimeout=5 \
+        -i data-refinery-key.pem \
+        "ubuntu@$API_IP_ADDRESS" "mkdir -p /var/www/volumes_static"
+
+    ssh -o StrictHostKeyChecking=no \
+        -o ServerAliveInterval=15 \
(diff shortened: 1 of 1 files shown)


infrastructure/deploy.sh
```diff
@@ -388,6 +388,19 @@ if [[ -n $container_running ]]; then
...
+        -o ServerAliveInterval=15 \
+        -o ConnectTimeout=5 \
+        -i data-refinery-key.pem \
+        "ubuntu@$API_IP_ADDRESS" "mkdir -p /var/www/volumes_static"
+
+    ssh -o StrictHostKeyChecking=no \
+        -o ServerAliveInterval=15 \
+        -o ConnectTimeout=5 \
+        -i data-refinery-key.pem \
+        "ubuntu@$API_IP_ADDRESS" "chmod a+rwx /var/www/volumes_static"
```
arkid15r, line 402:
```suggestion
    # Ensure the API's static file dir exists and is accessible.
    ssh -o StrictHostKeyChecking=no \
        -o ServerAliveInterval=15 \
        -o ConnectTimeout=5 \
        -i data-refinery-key.pem \
        "ubuntu@$API_IP_ADDRESS" "mkdir -m a+rwx -p /var/www/volumes_static"
```

arkid15r (approved):
LGTM

Issue #3438, opened 2023-10-24
Implement a backfill script for Sample::developmental_stage field


arkid15r:
### Problem or idea

The Sample::developmental_stage field has been added in #3425. This value needs to be populated for all existing samples when possible.

### Solution or next step

Implement `refresh_metadata` management command to run periodically for Sample::developmental_stage values population. It should be similar to surveyor's logic (metadata update part).

PR #3461, opened 2023-12-11
Add a prototype of Sample::developmental_stage backfill script

arkid15r:
## Issue Number

#3438 

This is a draft/prototype of a Foreman command to use for `Sample::developmental_stage` backfill process. The code is untested and supports SRA source DB only.

## Types of changes

## Checklist


common/data_refinery_common/models/sample.py
```diff
@@ -94,6 +94,10 @@ def __str__(self):
     created_at = models.DateTimeField(editable=False, default=timezone.now)
     last_modified = models.DateTimeField(default=timezone.now)
 
+    # Auxiliary field for tracking latest metadata update time.
+    # Originally added to support Sample::developmental_stage values backfilling.
+    last_refreshed = models.DateTimeField(auto_now=True, null=True)
```
davidsmejia, line 99: - We will probably want `last_refreshed` on Experiment as well, since a sample could belong to more than one experiment.
- We probably want to add `last_refresh_failure` as a timestamp on both as well to help with re-running

foreman/data_refinery_foreman/foreman/management/commands/refresh_sample_metadata.py
```diff
@@ -0,0 +1,43 @@
...
+    def handle(self, *args, **options):
+        for sample in Sample.objects.filter(
+            developmental_stage__isnull=True,
+            last_refreshed__isnull=True,
+            source_database=options["source"],
+        ).order_by("id")[: options["limit"]]:
+            logger.info(f"Refreshing metadata for a sample {sample.accession_code}")
+            try:
+                _, sample_metadata = SraSurveyor.gather_all_metadata(sample.accession_code)
+                SraSurveyor._apply_harmonized_metadata_to_sample(sample_metadata)
```
davidsmejia, line 37: This takes sample as the first argument

davidsmejia: One more consideration here is that after updating the sample we will want to update the cached values on experiment. 

ie: https://github.com/AlexsLemonade/refinebio/blob/07d375941ec45e6e692b467c58142ec8eb835d1f/foreman/data_refinery_foreman/surveyor/external_source.py#L207-L211

davidsmejia (commented):
Looking good, a couple comments about approach.

Looking forward, I think that we will want to update the `_apply_harmonized_metadata_to_sample` to specifically handle updates vs new samples. This depends on what the science team says is appropriate.

Issue #3554, opened 2026-04-28
Upgrade Postgres to 17.x or 18.x


davidsmejia:
# Context

The current legacy DB is no longer a viable option to have on AWS. This PR is just to track upgrading (and deploying) the postrgres db on staging and production once we can at lease run the full test suite for the API and smashing. 

# Next Steps

- [x] Review and merge issues related to #3538 
- [ ] Upgrade Postgres
- [ ] Tag deploy to staging
- [ ] Tag deploy to prod

davidsmejia:
Tested with 17.6, but want to test with 18.3 before committed as it should give us more runway

```bash
--------------------------
|DescribeDBEngineVersions|
+------------------------+
|  11.22-rds.20250814    |
|  11.22-rds.20251114    |
|  12.22-rds.20250508    |
|  12.22-rds.20250814    |
|  12.22-rds.20251114    |
|  13.21                 |
|  13.22                 |
|  13.23                 |
|  14.18                 |
|  14.19                 |
|  14.20                 |
|  14.21                 |
|  14.22                 |
|  15.13                 |
|  15.14                 |
|  15.15                 |
|  15.16                 |
|  15.17                 |
|  16.9                  |
|  16.10                 |
|  16.11                 |
|  16.12                 |
|  16.13                 |
|  17.5                  |
|  17.6                  |
|  17.7                  |
|  17.8                  |
|  17.9                  |
|  18.1                  |
|  18.2                  |
|  18.3                  |
+------------------------+
```

davidsmejia:
Closed as this is now live on prod

PR #3568, opened 2026-05-06
fix: require SSL for pgbouncer server connections to RDS

davidsmejia:
## Issue Number

#3554 

Previously we thought that pg bouncer just needed more time to boot up before we could try to connect and then run migrations and finish the deploy process. However in subsequent PRs once the migration took place pg bouncer fails to connect because RDS requires now requires connections to be over ssl. This PR sets that in the ini.

- Adds `server_tls_sslmode = require` to `pg-bouncer-instance-user-data.tpl.sh`

## Methods

N/A

## Types of changes

## Functional tests

SSHed onto the staging pgbouncer ec2 instance and manually edited the ini file before restarting the service. This resolved the connection issue on staging.

## Checklist

## Screenshots

N/A


nozomione (approved):
👍

Issue #3572, opened 2026-05-12
Migrate remaining adhoc scripts to rbio


davidsmejia:
# Migrate everything in `scripts/` to `bin/rbio`

## Goal

Eliminate `scripts/` entirely. Every script that lives there today should either move into a `bin/lib/` namespace, be deleted as obsolete, or have its helper functions absorbed into `bin/lib/` internals. End state: `rbio --help` is the single discoverable surface for every dev, deploy, and ops workflow.

## Background

`bin/rbio` already owns these namespaces: `build`, `common`, `db`, `dev`, `es`, `test`. Several legacy `scripts/` entries have already migrated or been shimmed (`run_postgres`, `run_es`, `install_db_docker`, `rebuild_es_index`, `run_tests`, db helpers, etc.). The remaining scripts span dev convenience, host bootstrapping, deploy-time formatting, and ad hoc ops. Folding them all in removes the "is it in `scripts/` or `bin/`?" guess game.

### Dev convenience

| Script | Proposed `rbio` command | Notes |
|---|---|---|
| `scripts/run_manage.sh` | `dev:manage <image> <service> -- <args>` | Map `-i` / `-s` directly. |
| `scripts/run_shell.sh` | `dev:shell [-i image] [-s service]` | Defaults to foreman. |
| `scripts/run_full_pipeline.py` | `dev:pipeline` | Wraps a chain of management commands; arg passing TBD. |
| `scripts/run_all_tests.sh` | (already a shim) | Delete after callers updated. |

### Ops / Batch

Introduce a new `ops` namespace following `bin/lib/README.md`.

| Script | Proposed `rbio` command | Notes |
|---|---|---|
| `scripts/kill_all_jobs.py` + `kill_all_jobs.sh` | `ops:kill-jobs` | Reads `AWS_REGION` + `REFINEBIO_JOB_QUEUE_WORKERS_NAMES`; iterates queues, terminates jobs. |
| `scripts/calculate_ram_hours.py` | `ops:ram-hours` | Argparse-driven; uses `pyrefinebio`. |

### Deploy

Introduce a new `deploy` namespace (or fold into `ops` if we'd rather keep namespace count low).

| Script | Proposed `rbio` command | Notes |
|---|---|---|
| `scripts/format_batch_with_env.sh` | `deploy:format-batch -p <project> -e <env> -o <out>` | Called from `infrastructure/deploy.sh`; update the deploy script in the same PR. |

### Host bootstrapping

[...]

PR #3583, opened 2026-05-26
rbio: add @requires decorator + mtime-gated common sdist rebuild

davidsmejia:
## Issue Number

#3572 

- bin/lib/_requirements.py (new) — @requires(env=, env_optional=, tools=)
    decorator + epilog_from() helper. Commands declare their env vars and
    external tool deps; future debug:env / debug:deps will read this
    registry to render per-command and aggregate views.
- bin/lib/common.py — factor the inline sdist rebuild out of cmd_common_update
  into rebuild_common_sdist(), which now:
    * uses an mtime gate to skip the rebuild when no common source file
      is newer than the existing sdist (preserves docker COPY layer cache
      for downstream worker/foreman/api images)
    * prefers dr_env/bin/python3 when the venv is present (system python3
      lacks setuptools on PEP-668 distros)
  Surfaces the helper as a standalone command, rbio common:build-sdist.
- bin/lib/build.py — invoke rebuild_common_sdist() before bake so
  downstream targets pick up the latest common package. No-op when source
  is unchanged (gate above).

## Methods

N/A

## Types of changes

## Functional tests

rbio functional tests only

## Checklist

## Screenshots

N/A

### DIFF
--- a/bin/lib/_requirements.py
+++ b/bin/lib/_requirements.py
@@ -1,0 +1,45 @@
+"""@requires decorator and epilog helper for rbio commands.
+
+Commands declare env vars + external tools they need via @requires. Future
+debug:env and debug:deps commands will aggregate this registry to render
+per-command and cross-command views. Per-command help text pulls the same
+metadata via epilog_from().
+"""
+
+
+def requires(*, env=None, env_optional=None, tools=None):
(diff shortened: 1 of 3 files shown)


nozomione (approved):
👍

avrohomgottlieb (approved):
👍

PR #3587, opened 2026-05-26
relocate format_batch_with_env.sh into infrastructure/

davidsmejia:
## Issue Number

#3572

The script is invoked exclusively from infrastructure/deploy.sh (six call sites, all relative to deploy.sh's CWD). Moving it adjacent to its only caller eliminates one of the two reasons the deploy flow had to traverse `../scripts/`.

  - git mv scripts/format_batch_with_env.sh infrastructure/
  - infrastructure/deploy.sh: ../scripts/format_batch_with_env.sh
                            → ./format_batch_with_env.sh (×6)

The script itself is unchanged. It still sources ../scripts/common.sh from inside its own cd'd CWD (api/, foreman/, workers/) - that path is relative to runtime CWD, not the script's file location, so the relocation doesn't affect it. scripts/common.sh remains in place until we drop it later.

## Methods

N/A

## Types of changes

- housekeeping

## Functional tests

N/A

## Checklist

## Screenshots

N/A


nozomione (approved):
👍

PR #3589, opened 2026-05-26
infra/deploy.sh: drop scripts/common.sh source, inline its consumers

davidsmejia:
## Issue Number

#3572

- infrastructure/deploy.sh only used two things from scripts/common.sh:
  - the $ALL_IMAGES list (consumed by the for-loop that builds per-image TF_VAR_*_docker_image vars) and get_deploy_repo (a two-line case mapping branch → Dockerhub namespace). Both are inlined now so this file no longer depends on scripts/.
- ALL_IMAGES is now a local string literal, matching the expansion of common.sh's `base api_base api foreman $WORKER_IMAGES`.
- get_deploy_repo is replaced with the two literal repo names (ccdlstaging / ccdl) at the only call site, with a comment naming what each is for.

## Methods

N/A

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A

### DIFF
--- a/infrastructure/deploy.sh
+++ b/infrastructure/deploy.sh
@@ -93,15 +93,14 @@ if [[ -z $SYSTEM_VERSION ]]; then
     exit 1
 fi
 
-# Load $ALL_IMAGES and helper functions.
-# shellcheck disable=SC1091
-. ../scripts/common.sh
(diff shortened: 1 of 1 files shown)


nozomione (approved):
👍

PR #3601, opened 2026-06-29
fix list compendia timeout

davidsmejia:
## Issue Number

N/A

This pull request improves the performance and reliability of querying and accessing related data for `CompendiumResult` objects. The main changes focus on optimizing database queries in the API and making the retrieval of computed files more robust.

`CompendiumResult.get_computed_file used .first()`, which appends ORDER BY id LIMIT 1. On the ~5.5M-row computed_files table Postgres walked the primary-key index filtering by result_id, discarding millions of rows (~1.6s per call). The compendia listing serializes one computed file per row, so all 41 rows together ran ~67s and the endpoint returned a 504.

**Query performance improvements:**

* In `compendium_result.py` API view, the queryset for public results now uses `select_related` for `primary_organism` and `result`, and `prefetch_related` for `organisms` and `result__computedfile_set`, reducing the number of database queries when accessing related objects.

**Code reliability and maintainability:**

* The `get_computed_file` method in the `CompendiumResult` model now reliably returns the computed file with the lowest ID (or `None` if there are none), instead of just the first returned by the queryset, ensuring deterministic behavior.
* The unused import of `ComputedFile` was removed from `compendium_result.py`, cleaning up the code.

## Methods

N/A

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/data_refinery_api/views/compendium_result.py
+++ b/api/data_refinery_api/views/compendium_result.py
@@ -110,7 +110,11 @@ class CompendiumResultListView(generics.ListAPIView):
         if invalid_filters:
             raise InvalidFilters(invalid_filters=invalid_filters)
 
-        public_result_queryset = CompendiumResult.objects.filter(result__is_public=True)
+        public_result_queryset = (
+            CompendiumResult.objects.filter(result__is_public=True)
+            .select_related("primary_organism", "result")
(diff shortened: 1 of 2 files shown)


nozomione (approved):
👍

